PR #92, opened 2019-06-02
Give users the option to set a passphrase

kaloudis:


decentropy:
This is implemented?

How does it work... I cant find any lock or password options.

Issue #118, opened 2019-07-26
Support LNURL withdraws


fiatjaf:
![demo](https://user-images.githubusercontent.com/1653275/61960649-3d512d80-af9c-11e9-97d4-b0ba00d3043d.gif)

https://github.com/btcontract/lnurl-rfc/blob/master/spec.md#3-withdrawing-funds-from-a-service

It's specially useful in these days post 0-sat invoice exploit.

It requires only that when a QR code starting with `lnurl` is found you 
1. decode it to get an `https://...` URL;
2. `GET` that URL;
3. Use the response parameters to show an invoice prompt to the user like https://lightning-wallet.com/ does -- or just use the defaults and make the invoice automatically without user interaction as t.me/lntxbot does;
4. Send the invoice back to the URL given also as a parameters.

That will generally trigger the service to pay the invoice automatically.

ZeusLN:
Added in v0.2.0 by @fiatjaf

PR #119, opened 2019-07-27
Add support for LNURL withdraws

fiatjaf:
See #118.

---

This pull request

* Adds a new scanning match for `AddressQRScanner`, `lnurl...` strings.
* When found, shows the `Receive` view prefilled with values taken from the URL (decoded as bech32).
* After confirmation by the user, the invoice is generated as normal, but automatically sent to the callback URL provided by the first lnurl call -- so the third-party service can proceed to pay the invoice.

---

Now I'm hitting a ton of bugs when trying to build it locally so I'm not sure it works. I'll wait for feedback before trying to build it locally again.

### DIFF
--- a/utils/AddressUtils.ts
+++ b/utils/AddressUtils.ts
@@ -2,6 +2,7 @@ const btcNonBech = /^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$/;
 const btcBech = /^(bc1|[13])[a-zA-HJ-NP-Z0-9]{25,39}$/;
 
 const lnInvoice = /^(lnbcrt|lntb|lnbc|LNBCRT|LNTB|LNBC)([0-9]{1,}[a-zA-Z0-9]+){1}$/;
+const lnUrl = /^(lnurl|LNURL)([0-9]{1,}[a-zA-Z0-9]+){1}$/;
 
 /* testnet */
 const btcNonBechTestnet = /^[2][a-km-zA-HJ-NP-Z1-9]{25,34}$/;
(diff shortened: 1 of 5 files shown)


fiatjaf:
I've finally manage to compile my branch. It seems to work on the surface, however the QR code scanner can't read anything, not even normal Lightning invoices (why that may be?), so I was unable to test the actual lnurl-withdraw functionality.

kaloudis:
This looks good. Let's return to it once we get in the the better QR support for Android. About to leave a comment on that PR.

utils/AddressUtils.ts
```diff
@@ -2,6 +2,7 @@ const btcNonBech = /^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$/;
 const btcBech = /^(bc1|[13])[a-zA-HJ-NP-Z0-9]{25,39}$/;
 
 const lnInvoice = /^(lnbcrt|lntb|lnbc|LNBCRT|LNTB|LNBC)([0-9]{1,}[a-zA-Z0-9]+){1}$/;
+const lnUrl = /^(lnurl|LNURL)([0-9]{1,}[a-zA-Z0-9]+){1}$/;
```
kaloudis: can we add some tests for this in `AddressUtils.test.ts`?

fiatjaf: ok. will do it.

fiatjaf:
This PR needs a refactor as I want to implement lnurl-withdraw in a way that leaves room open for lnurl-pay later.

Maybe create an lnurl library with the basic types, decoders and calls and import it here (I'm thinking it could be also used in BlueWallet and Joule later).

PR #120, opened 2019-07-29
Invoke an external QR code reader.

fiatjaf:
I'm hoping this code will make the app degrade nicely to the internal QR code scanner if no zxing-whatever-capable reader apps are present, but didn't test that yet. Will wait for feedback.

Also I don't know how this will work on iPhone, but I expect it to be fallback as well.

### DIFF
--- a/android/app/src/main/java/com/zeus/MainApplication.java
+++ b/android/app/src/main/java/com/zeus/MainApplication.java
@@ -14,6 +14,7 @@ import com.facebook.react.ReactPackage;
 import com.facebook.react.shell.MainReactPackage;
 import com.facebook.soloader.SoLoader;
 import com.zeusln.zeus.extensions.CustomClientFactory;
+import com.microsoft.ActivityResultPackage;
 
 import java.util.Arrays;
 import java.util.List;
@@ -34,7 +35,8 @@ public class MainApplication extends Application implements ReactApplication {
             new KeychainPackage(),
             new RNCameraKitPackage(),
             new RNGestureHandlerPackage(),
-            new VectorIconsPackage()
+            new VectorIconsPackage(),
+            new ActivityResultPackage()
       );
     }
 
--- a/components/QRCodeScanner.tsx
+++ b/components/QRCodeScanner.tsx
@@ -26,7 +26,7 @@ interface QRState {
 export default class QRCodeScanner extends React.Component<QRProps, QRState> {
   state = {
     hasCameraPermission: null,
-    useInternalScanner: false
+    useInternalScanner: Platform.OS !== 'android' // only try to use the external scanner on android
   }
 
   async startExternalQRScanner() {
--- a/components/QRCodeScanner.tsx
+++ b/components/QRCodeScanner.tsx
@@ -1,86 +1,157 @@
-import * as React from 'react';
-import { Platform, StyleSheet, Text, View } from 'react-native';
-import { Header } from 'react-native-elements';
-import { CameraKitCamera, CameraKitCameraScreen } from 'react-native-camera-kit';
-import Permissions from 'react-native-permissions';
+/** @format */
+
+import * as React from 'react'
+import {Platform, StyleSheet, Text, View} from 'react-native'
+import {Header, Icon} from 'react-native-elements'
+import ActivityResult from 'react-native-activity-result-fork'
+import {CameraKitCamera, CameraKitCameraScreen} from 'react-native-camera-kit'
+import Permissions from 'react-native-permissions'
+
+const QRINTENT = 'com.google.zxing.client.android.SCAN'
+
+const badQRscanners = {'com.motorola.camera': true}
 
 interface QRProps {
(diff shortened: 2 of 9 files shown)


android/app/src/main/java/com/zeus/MainApplication.java
```diff
@@ -14,6 +14,7 @@
 import com.facebook.react.shell.MainReactPackage;
 import com.facebook.soloader.SoLoader;
 import com.zeusln.zeus.extensions.CustomClientFactory;
+import com.microsoft.ActivityResultPackage;
```
kaloudis: I wonder if the fdroid compiler will throw a fit about this.

package.json
```diff
@@ -20,6 +20,7 @@
     "object-hash": "1.3.1",
     "react": "16.6.3",
     "react-native": "0.59.8",
+    "react-native-activity-result-fork": "^0.1.3",
```
kaloudis: please lock in the dependency version. remove the `^`

views/AddressQRScanner.tsx
```diff
@@ -49,9 +58,47 @@ export default class AddressQRScanner extends React.Component<AddressQRProps, {}
             navigation.navigate('Send');
         }
     }
+
+    async startExternalQRScanner () {
```
kaloudis: Let's break this function out into its own utility function and file, perhaps `utils/QRUtils.ts`

It needs to be used in `AddressQRScanner` as well as `BTCPayConfigQRScanner`, `LNDConnectQRScanner`, and `NodeQRScanner` in `views`

kaloudis: Also let's make sure this function doesn't hinder the iOS QR Scanner. If necessary you can wrap the whole function with an if checking for `Platform.OS === 'android'` to make sure it only executes on Android.

More details on the Platform module here: https://facebook.github.io/react-native/docs/platform-specific-code

android/app/src/main/java/com/zeus/MainApplication.java
fiatjaf: It's open-source!
If the name "microsoft" triggers them we can fork it and change, I guess.

package.json
fiatjaf: No one trusts semver anymore.

fiatjaf:
1. Moved all the logic to inside the `components/QRCodeScanner.tsx` (I couldn't see how it could be an "util" function as it has to be tangled with the React component `CameraKitCameraScreen` that is used as fallback);
2. Improved code that checks for the presence of capable third-party apps and launches them;
3. Removed the `BackButton` property from `QRCodeScanner` (since they were passed always exactly the same from the parent component and I needed a `goBack` function to use inside the external app handler flow, now parent components pass a `goBack` function directly and the `BackButton` is built inside `QRCodeScanner`);
4. Added a blacklist for bad apps that advertise themselves as QR scanners, but aren't, like my own vendor `com.motorola.camera`. Hopefully there aren't many others that do this. If someone has one of these apps and not other capable ones they won't be presented with a choice and will be stuck with the bad app. To solve that they can just install "Barcode Scanner", "Binary Eye", "SecScanQR" or similar apps so they can be presented with a choice and never again see the bad app -- and/or report the issue with the package name and we hardcode it in the blacklist.
5. If the user has no app at all that advertises itself as capable of reading QR codes (or if it has only a blacklisted app) the internal `CameraKitCameraScreen` will be used.
6. All this will only run on `Platform.OS === 'android'`.

kaloudis:
Excellent work. Code looks good, my only quip is that the styling is different but we will make it uniform with https://github.com/ZeusLN/zeus/issues/128.

I'll test it on both platforms and let you know how it's looking.

fiatjaf:
Oh, seriously? I remember taking much care to keep the style the first time I touched the code. Now yesterday my editor suddenly started reformatting things automatically so I thought: "hey, they have some Prettier rules here, I shouldn't have worried about styling at all".

But now that you're saying I think that was just my own global Prettier working on its own.

I'm sorry about that, and glad that #128 exists to save me.

components/QRCodeScanner.tsx
```diff
@@ -1,86 +1,157 @@
...
+import * as React from 'react'
+import {Platform, StyleSheet, Text, View} from 'react-native'
+import {Header, Icon} from 'react-native-elements'
+import ActivityResult from 'react-native-activity-result-fork'
+import {CameraKitCamera, CameraKitCameraScreen} from 'react-native-camera-kit'
+import Permissions from 'react-native-permissions'
+
+const QRINTENT = 'com.google.zxing.client.android.SCAN'
+
+const badQRscanners = {'com.motorola.camera': true}
```
kaloudis: shouldn't this be an array?

fiatjaf: I don't know. This object construction is more efficient, I think, and here it doesn't hurt as it's hardcoded. It allows us to do the `something in object` thing.

PR #137, opened 2019-10-16
add scan button to MainPane.

fiatjaf:
The idea is to make `AddressQRCodeScanner` a scanner for both invoices and lnurl things, so even if you're in the Pay screen and click scan you'll be able to scan an lnurl. So why not have a scanner button in the main app pane?

This is useful because after lnurl-withdraw we can introduce lnurl-pay, lnurl-auth and lnurl-channel reusing the same basic lnurl interface and scanning everything from the same place, without having to make a custom scanner for each and add buttons whoknowswhere for all these kinds of lnurls.

Even if you're not interested in adding all the bizarre lnurl types, I still think having the scanner in the front of the app is a good idea.

I also like the idea of swiping to the right to scan, instead of clicking. Is that an addition I should investigate?

### DIFF
--- a/views/Wallet/MainPane.tsx
+++ b/views/Wallet/MainPane.tsx
@@ -235,12 +235,36 @@ export default class MainPane extends React.Component<
                                     borderRadius: 30
                                 }}
                                 containerStyle={{
-                                    marginLeft: 10
+                                    marginLeft: 10,
+                                    marginRight: 10
                                 }}
                                 titleStyle={{
                                     color: theme === 'dark' ? 'white' : 'black'
                                 }}
                                 onPress={() => navigation.navigate('Receive')}
+                                raised={theme !== 'dark'}
+                            />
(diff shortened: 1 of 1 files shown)


fiatjaf:
Let me know if this is a desired change. If yes, let me know if I should pile the other lnurl commits here or if this will be merged separatedly and I open a new PR for them.

kaloudis:
I sorta like this idea. Can you post some screenshots to show how it looks?

views/Wallet/MainPane.tsx
```diff
@@ -236,13 +236,37 @@ export default class MainPane extends React.Component<
                                 }}
                                 containerStyle={{
                                     marginLeft: 10
```
kaloudis: you need to have a comma here

fiatjaf: oops

Issue #154, opened 2019-10-22
Sending screen: catch routing failure


kaloudis:
Often times the screen will go white when trying to send an LN payment. We need to review our logic to catch this routing failure

ZeusLN:
this is fixed with https://github.com/ZeusLN/zeus/pull/160 and will be released in v0.1.2

PR #160, opened 2019-10-27
ZEUS-154: Sending screen: catch routing failure

kaloudis:
Often times the screen would go white when trying to send an LN payment. Also added the 'Go to Wallet' button for every scenario


kaloudis:
https://github.com/ZeusLN/zeus/issues/154

PR #177, opened 2019-12-11
unify pasted/scanned content.

fiatjaf:
This PR moves logic from `Send.tsx` and `AddressQRScanner.tsx` to `utils/handleAnything.tsx`. There lives a function that takes a Bitcoin address, a Lightning invoice or any kind of LNURL and returns the route to which the user must be directed with parameters.

Also included in this PR is the usage of react-native's `Clipboard` at `Send.tsx`. If anything is copied at the user clipboard when that component is mounted it is automatically checked with `handleAnything()`, no need to click in the field, then click "paste", then click "Lookup payment request".


fiatjaf:
Superseced by https://github.com/ZeusLN/zeus/pull/178 because I don't know how to play with these PRs.

PR #178, opened 2019-12-12
Finishing lnurl-pay stuff.

fiatjaf:
Viewing successAction, browsing past, reusing, viewing past successActions.
Other fixes.


fiatjaf:
This should be ready to merge.

PR #221, opened 2020-01-20
lnd: Option to set sat per byte fee when closing channels

kaloudis:
Currently unsupported by c-lightning-REST


views/Channel.tsx
```diff
@@ -484,23 +491,52 @@ export default class ChannelView extends React.Component<
...
+                            {implementation === 'lnd' && <React.Fragment>
+                                <Text
+                                    style={{
+                                        color: theme === 'dark' ? 'white' : 'black'
+                                    }}
+                                >
+                                    (Optional) Sat per byte closing fee
+                                </Text>
+                                <TextInput
+                                    placeholder={'2'}
```
benthecarman: Should the following so the user can only enter numbers 
```
numeric value   // This prop makes the input to get numeric only 
keyboardType={'numeric'} // This prop help to open numeric keyboard
```

PR #267, opened 2020-05-12
Set iOS dev code signing to auto mode

kaloudis:


ZeusLN:
covered in https://github.com/ZeusLN/zeus/pull/270

PR #269, opened 2020-05-15
lnurlpay: replace realm with async-storage.

fiatjaf:
also:

 - [x] stop checking pending transactions, insted delete every lnurlpay thing older than 30 days (a check is done in 10% of the times the wallet is turned on).
 - [x] show lnurlpay images when available.


fiatjaf:
This is a work in progress, but I forgot to mark it as a Draft PR.

stores/LnurlPayStore.ts
```diff
@@ -30,187 +25,88 @@ interface LnurlPaySuccessAction {
...
+            if (key.slice(0, 9) === 'lnurlpay:') {
+                let item = JSON.parse(await AsyncStorage.getItem(key));
+                if (
+                    (item.last_stored && item.last_stored < daysago30) ||
+                    (item.time && item.time < daysago30)
+                ) {
+                    toRemove.push(key);
                 }
             }
         }
```
fiatjaf, line 62: @kaloudis what do you think of this? Deleting entries related to lnurlpay older than 30 days on startup (only every 10% of times, see [above](https://github.com/ZeusLN/zeus/pull/269/files#diff-1ca457fae175aff8e2ccf7b02b225555R38-R43)). Too dumb to be true?

stores/LnurlPayStore.ts
```diff
@@ -30,187 +25,88 @@ interface LnurlPaySuccessAction {
...
-                        this.clear(pendingHash);
-                        return;
-                    default:
-                        // leave it as is
-                        return;
+    deleteOld = () => {
+        // delete all lnurlpay keys older than 30 days
+        const daysago30 = new Date().getTime() + 1000 * 60 * 60 * 24 * 30;
+        const allKeys = await AsyncStorage.getAllKeys();
+        var toRemove = [];
```
ZeusLN, line 50: this can be a `const`

stores/LnurlPayStore.ts
ZeusLN: I generally hate using `Math.random` at all but the approach may be reasonable. Gonna sleep on this one.

kaloudis: I'm actually ok with this

fiatjaf:
I think this is almost ready, but I didn't try a real payment with it, only tried to pay on my regtest lnd and these failed, of course.

Will try to make a Spark PR first so I can use my node and do a real test.

fiatjaf:
The Spark PR includes these changes so maybe let's just use that.

PR #314, opened 2020-08-02
Multi-path payments with lnd

kaloudis:
Simulator Screen Shot - iPhone 11 - 2020-08-01 at 20 22 21
Simulator Screen Shot - iPhone 11 - 2020-08-11 at 00 18 18
Simulator Screen Shot - iPhone 11 - 2020-08-11 at 00 16 41


kaloudis:
hit an issue with the new v2 endpoints. reached out to the lnd team

kaloudis:
Need to tweak the request format and also test when doing keysends

kaloudis:
update: found that the /v2/ routes use websockets and configured RESTUtils to be able to leverage them, however we cannot authorize as you cannot send headers from websockets in the browser/React. Looking for solutions.

PR #323, opened 2020-08-14
Eclair support

fiatjaf:


views/Send.tsx
```diff
@@ -155,7 +157,8 @@ export default class Send extends React.Component<SendProps, SendState> {
             destination,
             amount,
             fee,
-            error_msg
+confirmationTarget,
```
kaloudis, line 160: indent is messed up here, be sure to run `npm run linter`

views/Send.tsx
```diff
@@ -67,6 +67,7 @@ export default class Send extends React.Component<SendProps, SendState> {
             destination: destination || '',
             amount: amount || '',
             fee: '2'
```
kaloudis, line 69: missing comma here

kaloudis:
Awesome job refactoring `RESTUTils`

fiatjaf:
This is not ready yet. I must include it in the node selection UI and then test it on my phone which I'm trying to bring back from the grave, that will certainly reveal a ton of bugs.

views/Send.tsx
```diff
@@ -124,12 +125,13 @@ export default class Send extends React.Component<SendProps, SendState> {
...
         const { TransactionsStore, navigation } = this.props;
-        const { destination, fee } = this.state;
+        const { destination, fee, confirmationTarget } = this.state;
 
         TransactionsStore.sendCoins({
             addr: destination,
             sat_per_byte: fee,
-            amount: satAmount.toString()
+            amount: satAmount.toString(),
+            target_conf: Number(confirmationTarget // eclair-only
```
kaloudis, line 134: missing closing parenthesis

fiatjaf:
Wow, that's a lot of conflicts.

views/Transaction.tsx
```diff
@@ -141,7 +141,7 @@ export default class TransactionView extends React.Component<TransactionProps> {
                                 </Text>
                             </TouchableOpacity>
                         </View>
-                    )}
+                    ) : null}
```
kaloudis, line 144: weird that this change was needed, it's effectively the same

fiatjaf: It was crashing without this because it was trying to output a value outside `<Text>`.

The issue was that since my onchain transactions were all coinbases (because I was on regtest and Eclair uses the Bitcoin Core wallet directly, not because I'm secretly a solo miner) `total_fees` was 0. And `0 && something` returns `0`, which gets translated to a text but then everything breaks because it is outside `<Text>`.

kaloudis: makes sense. alternatively you could do `!!total_fees &&` but this is acceptable too

fiatjaf: Wouldn't that output "false"?

kaloudis: yes and have the same effect, it wouldn't render the `<View>` component that followed

fiatjaf: But wouldn't it try to render the string `false` and cause an error?
I thought the `something && <Component>` trick only worked with `something == null`.

kaloudis: nope it works (doesn't render the component that follows) with falsy values too

fiatjaf:
Untested yet.

utils/RESTUtils.ts
```diff
@@ -1,658 +1,38 @@
...
+import Spark from '../backends/Spark';
+import LND from '../backends/LND';
+import LndHub from '../backends/LndHub';
+import Eclair from '../backends/Eclair';
+
+import LND from '../backends/LND';
+import CLightningREST from '../backends/CLightningREST';
+import LndHub from '../backends/LndHub';
+import Spark from '../backends/Spark';
+import Eclair from '../backends/Eclair';
```
ZeusLN, line 12: you duplicated the backend imports here

ZeusLN (approved):

fiatjaf:
Wait, wait, I think it's still broken.

fiatjaf:
Actually, no, it is fine.

I had made that change and tested it some days ago and gotten bugs. I now realize I was testing not Eclair but lndhub integration against a buggy backend of mine, but the bugs had discouraged me.

kaloudis (approved):

PR #365, opened 2020-09-22
fix eclair getInvoices "from" filter.

fiatjaf:
# Description

Invoices were not showing up on Eclair because the `from` filter was formatted as a float but needed to be an int.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Other

## Checklist

## Testing

- [x] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [ ] iOS

- [ ] LND
- [ ] c-lightning-REST
- [ ] spark
- [x] Eclair
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/backends/Eclair.ts
+++ b/backends/Eclair.ts
@@ -148,7 +148,7 @@ export default class Eclair {
         );
     getInvoices = () => {
         // 90 days ago
-        const since = parseInt(new Date().getTime() / 1000 - 60 * 60 * 24 * 90);
+        const since = Math.round(new Date().getTime() / 1000 - 60 * 60 * 24 * 90);
 
         return Promise.all([
             this.api('listinvoices', { from: since }),
--- a/backends/Eclair.ts
+++ b/backends/Eclair.ts
@@ -147,7 +147,9 @@ export default class Eclair {
             })
         );
     getInvoices = () => {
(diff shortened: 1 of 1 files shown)


kaloudis (approved):

backends/Eclair.ts
```diff
@@ -147,7 +147,9 @@ export default class Eclair {
             })
         );
     getInvoices = () => {
-        const since = new Date().getTime() / 1000 - 60 * 60 * 24 * 90; // 90 days ago
+        // 90 days ago
+        const since = parseInt(new Date().getTime() / 1000 - 60 * 60 * 24 * 90);
```
kaloudis, line 151: For the sake of the typechecker can you change it to:
```const since = parseInt((new Date().getTime() / 1000 - 60 * 60 * 24 * 90).toString());```

param of `parseInt` must be a string. I'm not sure if you'd prefer `Math.round` here instead as `parseInt` always rounds down

fiatjaf: Doesn't make much difference, but let's go with `Math.round` as it is prettier.

kaloudis (approved):

Issue #382, opened 2020-10-18
Open channel for Eclair node crashes.


dscotese:
**Describe the bug**
I have to unlock my bitcoin core wallet for Eclair to open a new channel, so I unlocked it (on my laptop).  Then I went to my Eclair node in Zeus beta (on my Galaxy 7 running Android 8.0.0) and clicked "Open channel" even though it still said "0 on chain funds."  It crashed.

**To Reproduce**
1. Run Bitcoin core on a computer.
1. Ensure that you have a wallet with funds in it but no outputs that are p2pk (legacy) outputs because they stop Eclair from running properly.
1. Make sure your wallet is password protected.
2. Run Eclair on the computer too.
3. Connect Zeus beta to that Eclair node.
2. Verify that Eclair cannot open a new channel because you have not unlocked your bitcoin core wallet.
4. Open your Bitcoin core console and use the walletpassphrase command to unlock your wallet for several seconds.
4. Before all those seconds pass, try again to open a new channel.

Note: When I tried to establish that this series of steps worked, I discovered a much simpler way to crash it:
1. Click on the "Channels" tab.
1. Click the "Open channel" button.
1. CRASH!

**Desktop running Eclair:**
 - OS: Windows 10 (but my Eclair node wasn't involved)
 - NA
 - Version 1909

**Smartphone running Zeus beta:**
 - Device: Galaxy 7
 - OS: Android 8.0.0
 - Zeus Version 0.4.0 beta-3

Issue #381, opened 2020-10-18
Zeus Beta: Fee Report shows same (wrong) amount for today, this week, and this month.


dscotese:
For my c-lightning node using Zeus on my Android Galaxy 7 Andriod version 8.0.0,

Non-Beta Zeus tells me my fees are 354.565 sats. Yay!
Beta Zeus tells me: 
Earned Today: 95.363 sats
Earned this week: 95.363 sats
Earned this month: 95.363 sats

It would be nice if it also indicated:
Earned since ____ (If I could pick a number of days, __THAT__ would be awesome! If not, maybe 3 months?  Or all time?): ____ sats

fiatjaf:
This is related to c-lightning-REST, not Spark. c-lightning-REST doesn't implement `day_fee_sum`, `week_fee_sum`, `month_fee_sum`, only the deprecated `total_fee_sum`.

fiatjaf:
But actually this has led me to discover issues on Spark and Eclair too, so that was something.

PR #385, opened 2020-10-22
A handful of bugfixes in Éclair and Spark

fiatjaf:
# Description

Fixes https://github.com/ZeusLN/zeus/issues/382
Partially fixes https://github.com/ZeusLN/zeus/issues/381 (except c-lightning-REST)
Supersedes https://github.com/ZeusLN/zeus/pull/380 (because I had to fix it in my own way -- basically changing the button render condition slightly -- it would be impossible to test).

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] LND
- [ ] c-lightning-REST
- [x] spark
- [x] Eclair
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/backends/Eclair.ts
+++ b/backends/Eclair.ts
@@ -349,7 +349,6 @@ export default class Eclair {
                       1000000
                     : null
             })),
-            total_fee_sum: allTimes / 1000,
             day_fee_sum: lastDay / 1000,
             week_fee_sum: lastWeek / 1000,
             month_fee_sum: lastMonth / 1000
--- a/backends/Eclair.ts
+++ b/backends/Eclair.ts
@@ -437,6 +437,7 @@ export default class Eclair {
     supportsChannelManagement = () => true;
     supportsCustomHostProtocol = () => false;
     supportsMPP = () => false;
+    supportsCoinControl = () => false;
 }
 
 const mapInvoice = (isPending: any) => ({
--- a/backends/Eclair.ts
+++ b/backends/Eclair.ts
@@ -51,7 +51,7 @@ export default class Eclair {
             });
 
         setTimeout(
-            (id) => {
+            id => {
                 delete calls[id];
             },
             5000,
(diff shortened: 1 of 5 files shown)


backends/Eclair.ts
```diff
@@ -50,6 +50,14 @@ export default class Eclair {
...
 
+        setTimeout(
+            id => {
+                delete calls[id];
+            },
+            5000,
+            id
+        );
+
         return calls[id];
```
ZeusLN, line 61: won't this always return null if you're deleting it?

kaloudis:
Couple of things to take from #380 I believe:

-make sure block height is passed in: https://github.com/ZeusLN/zeus/pull/380/files#diff-0e600bf73bc7165dc56072761a578b076f9f6fc4b8c126aff34413e01b43149bR157
-don't display implemetation version if not defined: https://github.com/ZeusLN/zeus/pull/380/files#diff-e35361e9913d3c91a24c350489110554e0a7c068ede78828a70db5e7b9812d86R226

We can add them later in a later PR too - just want to comment so we don't forget them

backends/Eclair.ts
fiatjaf: I'm deleting only after 5 seconds.

This may be the most dumb solution ever, but I was hitting a bug in which this call failed because a problem in my desktop node and then it kept failing again and again until I restarted Zeus.

fiatjaf:
Maybe merge #380 and I rebase this.

backends/Eclair.ts
kaloudis: I'm ok with this, please just leave a code comment

kaloudis:
I merged in #380. I like the way you changed the `MainPane` logic, much cleaner. you can overwrite with your changes

fiatjaf:
Rebased finally.

kaloudis (approved):

Issue #339, opened 2020-08-27
Fix "Fee Rate mili msat" so that its meaning is clearer.


dscotese:
**Describe the bug**
A user who wishes to update the fee rate in a channel (not the base fee, but the fee rate) needs to know the current fee and the label does not make this clear.  The label is "Fee Rate mili msat".  For a node I own (03e1b9e6ca5d912254060927794efd3b96a8149f9d360bbec68e693fe7c47a8235), the fee rate is 10 "fee per millionth" which means that 10 out of every million transferred units will be charged as a fee, regardless of the unit.  Zeus displays this as "1,000,000 mili msat."  My rate is ten out of a million, and that's the same as 10,000 out of a billion, or 1,000,000 out of a hundred billion.  Thus, what Zeus is displaying is "fee_per_hundred-billionth".  [Bolt 11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md) uses "fee_proportional_millionths."  I assume some LN implementations allow the operator to set this to a number with up to five decimal places (which would be "fee_proportional_hundred-billionths") which is why Zeus displays it that way. I believe the label should be clearer.  I propose "hundred-billionths" as that is clear and concise.

**To Reproduce**
1. Go to "channels"
1. Click on any channel.
2. Click on 'Set new fees'
3. Notice that under "Base Fee msat" and its current value is "Fee Rate millil msat" along with the number of hundred-billionths of a payment that will be charged.

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy 7
 - OS: Android 8.0.0
 - Zeus Version 0.2.2

dscotese:
I wanted to tackle this issue, but I searched the repository](https://github.com/ZeusLN/zeus/search?q=mili&unscoped_q=mili) for "mili" and only found it as part of a variable name.  If changing that variable name from `newFeeRateMiliMsat` to `newFeeHundredBillionths` would work, let me know and I'll do it.  If so, however, I'd like to learn why the display capitalizes "Fee" and "Rate" but not "mili" or "msat."

Jestopher-BTC:
Please just make it parts per million (ppm). The fee rate is sats fee per sats routed, which is unitless. So, using milli msat is not only confusing, but it's also inaccurate.
Part sats fee per million sats routed is far more intuitive and a common convention in sciences.

kaloudis:
[...]

PR #387, opened 2020-10-25
Patch 1

dscotese:
# Description

This is simply a change to the label for the fee rate so that its meaning is clearer.  Issue #339 provides more detail.


kaloudis:
I think it'd be clearer to label it PPM (Parts Per Million) and multiply the value we send over currently by 1,000,000

kaloudis:
5ah390rq

ZeusLN:
Hey @dscotese - really appreciate the PR but closing in favor of https://github.com/ZeusLN/zeus/pull/388

dscotese:
That's good.  I favored #388 too.

PR #509, opened 2021-08-08
Amount component

futurepaul:
# Description

A `<Value />` component for displaying values throughout the app.

This needs some cleanup and would love to chat if there's anything I'm doing here that could be simpler / more straightforward.

Also it should be pretty straighforward to rewrite the existing `getAmount` method based on the new logic of `getUnformattedAmount` method but since I believe we can deprecate `getAmount` eventually, so not sure if that's necessary.

Another thing this component could be set up to support is displaying bytes / vbytes. Didn't bother with that for now but should be easy addition.

Also, in addition to various rules about left / right and spacing for monies, there are also rules about commas:

https://fastspring.com/blog/how-to-format-30-currencies-from-countries-all-over-the-world/

For instance here's Argentina:
`$ 1.234,56`
And here's USA:
`$1,234.56`

Shouldn't be too hard to add but haven't done that here.

I'd love some recommendations on how to kick the tires of this to make sure it's truly robust.

Finally, this won't every look like the mockups in Figma until we get Lato installed :)

image
image
image
image

- [x] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Other

## Checklist

## Testing

- [x] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [ ] iOS

- [ ] LND
- [ ] c-lightning-REST
- [ ] spark
- [ ] Eclair
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/LayerBalances/index.tsx
+++ b/components/LayerBalances/index.tsx
@@ -15,6 +15,8 @@ import { inject, observer } from 'mobx-react';
 
 import OnChain from './../../images/SVG/OnChain.svg';
 import Lightning from './../../images/SVG/Lightning Circle.svg';
+import { Value } from '../Value';
+import { Spacer } from '../layout/Spacer';
 
 interface LayerBalancesProps {
     BalanceStore: BalanceStore;
@@ -27,7 +29,7 @@ I18nManager.allowRTL(false);
 
 type DataRow = {
     layer: string;
-    balance: string;
+    balance: string | number;
 };
 
 const Row = ({ item }: { item: DataRow }) => (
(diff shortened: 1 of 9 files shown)


futurepaul:
We talked and we're going to change this component name to `Amount`.

stores/UnitsStore.ts
```diff
@@ -47,6 +57,65 @@ export default class UnitsStore {
...
 
+    @action getUnformattedAmount = (
+        value: string | number = 0,
+        fixedUnits?: string
+    ): ValueDisplayProps => {
+        const { settings } = this.settingsStore;
+        const { fiat } = settings;
+        const units = fixedUnits || this.units;
+
+        const wholeSats = value.toString().split('.')[0];
```
kaloudis, line 68: `const [wholeSats] = value.toString().split('.');` maybe

views/Wallet/MainPane.tsx
```diff
@@ -53,71 +54,23 @@ export default class MainPane extends React.PureComponent<MainPaneProps, {}> {
...
-                >
-                    {units &&
-                        PrivacyUtils.sensitiveValue(
-                            getAmount(lightningBalance),
-                            8,
-                            true
-                        )}{' '}
-                    ⚡
-                </Text>
+                <Value jumbo={true} sats={lightningBalance} sensitive />
```
futurepaul, line 57: pass the jumbo prop like the sensitive prop

futurepaul:
Okay this should be good to go sorry for the delay.

I tried rebasing this to master and there were no conflicts which is amazing to me. Let me know if I should push that version up and point this PR at master.

kaloudis:
@futurepaul awesome, please do

components/LayerBalances/index.tsx
```diff
@@ -80,29 +75,22 @@ const SwipeableRow = ({
     );
 };
 
-@inject('UnitsStore')
+@inject()
```
kaloudis, line 78: do you still need this `inject` here if it's empty?

kaloudis:
Can we add a color prop to amount? Some balances will be red, green, or the highlight themeColor

ZeusLN:
Merging this in as is so we can start integrating it other places.

We can iterate on it in future PRs.

PR #558, opened 2021-08-30
Add color prop to Amount and use in Activity view

futurepaul:
# Description

Relates to issue: Follow up to #509 

`Amount` component can now take a "color" prop in addition to the debit / credit options. Testing this out by putting it in the `Activity` view.

I changed up the underlying `AmountDisplay` component to just take a color and removed the debit / credit options from `Body` and just added a color prop to keep it more logic agnostic.

I'm also using the semantic "warning" / "success" color names instead of "red" / "green" because that seems to offer the most flexibility to future theme authors.

Speaking of colors, we don't have all of these colors for the non-Dark themes. What's a good solution to that? I also had an idea to create an enum for all the keys in the theme which we could to typecheck anyone using this `Amount` or `Body` component to make sure they're passing a valid key and not some random string.

Next obvious step would be to use `KeyValue` component to refactor this `Activity` view but wanted to keep this pr focused.

image

- [x] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Other

## Checklist
- [x] I’ve run `npm run tsc` and make sure my code didn’t produce errors 
- [x] I’ve run `npm run prettier` and formatted my code correctly

## Testing

- [x] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [ ] iOS

### DIFF
--- a/components/Amount.tsx
+++ b/components/Amount.tsx
@@ -20,8 +20,7 @@ interface AmountDisplayProps {
     rtl?: boolean;
     space?: boolean;
     jumboText?: boolean;
-    credit?: boolean;
-    debit?: boolean;
+    color?: 'text' | 'success' | 'warning' | 'highlight' | 'secondaryText';
 }
 
 function AmountDisplay({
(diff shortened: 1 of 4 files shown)


kaloudis:
If a color doesn't exist on a theme it falls back to the value on the default dark theme. Feel free to add new definitions for debit/credit etc if you see fit.

futurepaul:
Ah I wasn't aware that's good to know. I'm using semantic names now because I'm using the "success" / "warning" colors from the figma.

kaloudis (approved):

PR #748, opened 2022-02-01
Keysend troubleshooting

kaloudis:
Troubleshooting keysends for LND v0.14+

So far:

- Removed old `v1` endoint, always run through `/v2/router/send`
- Fixed timeout to 60s (pertinent to `react-native-tor` work we'll do later this week or next)
- Removed timeout field from view

### DIFF
--- a/backends/CLightningREST.ts
+++ b/backends/CLightningREST.ts
@@ -117,6 +117,11 @@ export default class CLightningREST extends LND {
             invoice: data.payment_request,
             amount: Number(data.amt && data.amt * 1000)
         });
+    sendKeysend = (data: any) =>
+        this.postRequest('/v1/pay/keysend', {
+            pubkey: data.pubkey,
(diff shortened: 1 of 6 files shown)


kaloudis:
Regular payments work, regular payments with amp work, just getting `FAILURE_REASON_INCORRECT_PAYMENT_DETAILS` on keysends

kaloudis:
Welp, looks like my regtest counterparties just didn't have ``--accept-amp`` enabled. All works now.

deregs:
The logic as of now will try to send an amp keysend if the connected node supports amp. I think if amp keysends are to be supported, it should really be checking if the node on the receiving end of the keysend supports amp as well. Not sure if it's even possible to check that though.

kaloudis:
@deregs created an issue to track here: https://github.com/ZeusLN/zeus/issues/750

PR #749, opened 2022-02-01
Keysend Messaging

deregs:
Implemented so far:
- send keysend messages through lnd
- send messages through c-lightning

To do:
- look into displaying received keysend messages in a new keysend page? Maybe leave this for a separate pr.


backends/CLightningREST.ts
```diff
@@ -120,7 +120,8 @@ export default class CLightningREST extends LND {
     sendKeysend = (data: any) =>
         this.postRequest('/v1/pay/keysend', {
             pubkey: data.pubkey,
-            amount: Number(data.amt && data.amt * 1000)
+            amount: Number(data.amt && data.amt * 1000),
+            label: data.label
```
kaloudis, line 124: remove the `label` field please. this is used for internal labeling, not keysend

kaloudis:
Don't modify any of the locale files other than `en.json` please

backends/CLightningREST.ts
deregs: done

views/Send.tsx
```diff
@@ -224,12 +226,14 @@ export default class Send extends React.Component<SendProps, SendState> {
...
         const { TransactionsStore, navigation } = this.props;
-        const { destination, maxParts, maxShardAmt, feeLimitSat } = this.state;
+        const { destination, maxParts, maxShardAmt, feeLimitSat, message } =
+            this.state;
 
         if (RESTUtils.supportsAMP()) {
             TransactionsStore.sendPayment({
                 amount: satAmount.toString(),
                 pubkey: destination,
+                message: message,
```
kaloudis, line 236: if a property and the variable you are assigning to it have the same name you don't have to repeat it. See https://alligator.io/js/object-property-shorthand-es6/

we'll make sure we catch this with our linter by added this rule https://eslint.org/docs/rules/object-shorthand

views/Send.tsx
```diff
@@ -238,7 +242,8 @@ export default class Send extends React.Component<SendProps, SendState> {
         } else {
             TransactionsStore.sendPayment({
                 amount: satAmount.toString(),
-                pubkey: destination
+                pubkey: destination,
+                message: message
```
kaloudis, line 246: same here

utils/Base64Utils.ts
```diff
@@ -84,6 +84,9 @@ class Base64Utils {
             (memo: any, i: number) => memo + ('0' + i.toString(16)).slice(-2),
             ''
         );
+
+    utf8ToHexString = (hexString: string) =>
+        Buffer.from(hexString, 'utf8').toString('hex');
```
kaloudis, line 89: would be good to add some tests for this

deregs: added a few

kaloudis (approved):
LFG!

PR #776, opened 2022-02-07
Nodes: refresh list after adding/delete node

kaloudis:


deregs:
Noticed this issue when adding the copy node button. I tested this change on iphone 11 Pro with iOS 13.7 and it's still not refreshing the list of nodes after adding a new one. The new node only appears if I press back and then click on the nodes list again.

kaloudis:
@deregs thanks for testing! needed to do some refactoring. please give it a go when you have a chance.

deregs:
works great now!

PR #834, opened 2022-02-16
fixed keyboard collapse on set fees form

deregs:
Initial fix to collapse keyboard on set fees form.

### DIFF
--- a/components/SetFeesForm.tsx
+++ b/components/SetFeesForm.tsx
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { StyleSheet, Text, View } from 'react-native';
+import {
+    StyleSheet,
+    Text,
+    View,
+    Keyboard,
+    TouchableWithoutFeedback
+} from 'react-native';
 import { inject, observer } from 'mobx-react';
 
 import Button from './../components/Button';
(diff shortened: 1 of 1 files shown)


deregs:
can add this in other areas if we think this is the right approach

kaloudis:
Did changing the input type to `number-pad` not offer the same functionality?

deregs:
> Did changing the input type to `number-pad` not offer the same functionality?

It looks like the only difference between `numeric` and `number-pad` keyboards is that `number-pad` doesn't have a decimal point. [Here](https://lefkowitz.me/visual-guide-to-react-native-textinput-keyboardtype-options/) are the visual differences in keyboards.

From reading [this thread](https://stackoverflow.com/questions/29685421/hide-keyboard-in-react-native), it seems that using `keyboard.dismiss` is the proper way to fix this.

However, I could be wrong because I only tested `number-pad` on an emulator since I can't run the app on my phone without a developer account. And the keyboard collapse issue is only happening on iOS, not android.

deregs:
@kaloudis I checked everywhere else that uses `numeric` keyboard, and the collapse issue is not a problem because the pages are longer so they're using `ScrollView`. In `ScrollView`, tapping outside of the text boxes always collapses the keyboard.

So we can either go through with this change for the set fees form, or we can use `ScrollView` for this form. Do you prefer one over the other?

kaloudis:
Good find. Let's go with `ScrollView` then - it can't hurt and is cleaner

kaloudis:
We have `keyboardType="numeric"` several places in the app. Do we need to apply this fix elsewhere or are they all already covered by `ScrollView`s?

deregs:
> We have `keyboardType="numeric"` several places in the app. Do we need to apply this fix elsewhere or are they all already covered by `ScrollView`s?

All covered by ScrollView. It's used in `EditFee`, `OpenChannel`, `PaymentRequest`, `Receive`, `Send`, `Channel`, and `AddEditNode`, always within a scrollview from what I see.

kaloudis (approved):

Issue #832, opened 2022-02-16
Add attempt amp toggle for keysend payments


deregs:

deregs:
working on this one

Bosch-0:
Fixed in https://github.com/ZeusLN/zeus/pull/840 yeah?

deregs:
> Fixed in #840 yeah?

yup

PR #840, opened 2022-02-17
Adding amp toggle for keysends

deregs:
# Description

Relates to issue: **[ZEUS-832](https://github.com/ZeusLN/zeus/issues/832)**

Previously, for LND 0.13 and 0.14 nodes, the keysend page would default to sending an amp payment. This could potentially cause issues, because amp payments will fail unless they're being sent to another LND 0.13 / 0.14 node with accept-amp turned on.

This change introduces an amp toggle for keysends. The toggle will only appear if the node supports amp (meaning it's an LND 0.13 / 0.14 node).

Screenshot when toggle is disabling amp payment:

Screen Shot 2022-02-16 at 9 07 33 PM

Screenshot when toggle is enabling amp payment:

Screen Shot 2022-02-16 at 9 08 13 PM

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -327,6 +327,7 @@
     "views.Send.editNetworkFees": "Edit network fees",
     "views.Send.keysendNotSupported": "does not support sending keysend payments at the moment.",
     "views.Send.lookup": "Look Up Payment Request",
+    "views.Send.amp": "Attempt atomic multi-path payment",
     "views.SendingLightning.sending": "Sending Transaction",
     "views.SendingLightning.success": "Transaction successfully sent",
     "views.SendingLightning.paymentHash": "Payment Hash",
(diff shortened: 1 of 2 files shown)


locales/en.json
```diff
@@ -327,6 +327,7 @@
     "views.Send.editNetworkFees": "Edit network fees",
     "views.Send.keysendNotSupported": "does not support sending keysend payments at the moment.",
     "views.Send.lookup": "Look Up Payment Request",
+    "views.Send.amp": "Attempt atomic multi-path payment",
```
kaloudis, line 330: honestly, we can just reuse `views.PaymentRequest.amp` so we can spare the translators having to translate the field twice

deregs: sure, changed that

kaloudis:
I think the toggle should stay above the Max parts field. Thoughts @Bosch-0?

At the very least it should be consistent with the `PaymentRequest` view

deregs:
> I think the toggle should stay above the Max parts field. Thoughts @Bosch-0?
> 
> At the very least it should be consistent with the `PaymentRequest` view

I agree, to remain consistent with `PaymentRequest`, and also users might toggle it on just to see what it is, and then they'd have to scroll all the way down to toggle it off. I made the change, and here's a screenshot for when toggle is enabling amp payment:

Screen Shot 2022-02-18 at 4 59 34 PM

kaloudis (approved):

Issue #857, opened 2022-02-19
Add PIN/Duress PIN and Duress Password


deregs:
Duress modes only enabled if main password/PIN is set. User can enter duress pin to delete all their node configs. Can be added for v0.6.1.

deregs:
Working on this, couldn't find an existing open issue for it.

deregs:
@kaloudis @Bosch-0 I saw this screen in figma:
![Screen Shot 2022-02-19 at 2 32 48 PM](https://user-images.githubusercontent.com/93104057/154816496-0bb4a2b6-e7c4-433d-8e0e-79632925bbd7.png)

Along with implementing duress pin/password, do we want to change this screen to set a pin instead of password?

<img width="372" alt="Screen Shot 2022-02-19 at 2 44 06 PM" src="https://user-images.githubusercontent.com/93104057/154816625-ed7829e5-248a-4331-a261-1c3f98de1e93.png">

If so, which is higher priority? Should we implement duress password then later convert everything to pin, or switch to pin before implementing duress pin?

Bosch-0:
I think by default every user should have to set a PIN when they first register with Zeus - this is the easier option that most users will most likely use and is in general a good security practice. In the settings, users should have the option to disable this PIN and have an option to use passphrase over a the PIN. Depending on which lock screen method they have set they should be able to set a duress that is the same as the lock screen method they have set (keeps the UI clean that way).

deregs:
Sounds good, will get everything working for passphrase / duress passphrase first, and then we can add pin / duress pin.

PR #858, opened 2022-02-19
PIN, Duress PIN, and Duress Passphrase

deregs:
Relates to issues: #857, #859 
Features added for password/duress password:
- Set/Change Duress Password screen added.
- If a password is set, user now has to enter their password in order to modify password/duress password
- Delete button added to set password and set duress password screens, only appears if the password/duress password is already set.
- If duress password entered while logging in, all node configs are deleted and user is logged in.

Features added for PIN/duress PIN
- Set/Change PIN and Set/Change Duress PIN screens added.
- Set/Change PIN/Duress PIN screens allow PINs of length 4-8 to be set.
- Any time the Pin Pad is displayed, the order of the numbers is shuffled and none of the numbers visually respond to touch events - this is to prevent against attacks.
- If PIN is set, user has to enter PIN in order to modify PIN/Duress PIN.
- Delete options added in Security settings for PIN/Duress PIN, each one appears only when the corresponding PIN/Duress PIN has already been set.
- If duress PIN is entered while logging in, all node configs are deleted and user is logged in.

Other notes:
- PinPad component can be reused on the SendReceive pane with numbers not shuffled.
- PIN and Password can not be set at the same time - must choose one or the other. But user can switch from one to the other by deleting PIN/Password and then setting the other in Security settings.
- User is restricted to 4 failed authentication attempts in a row, and failed attempts are saved even if the app is closed. When the user submits a 5th failed attempt, app takes duress behavior.
- Security Settings will show different options depending on what values are set:
  - If no Password or PIN is set, shows options to set PIN or set Password
  - If Password is set, shows options to set/change Password or set/change Duress Password
  - If PIN is set, shows options to set/change PIN, Delete PIN, or set/change Duress PIN
  - If PIN and Duress PIN are set, shows all the PIN options plus option to Delete Duress Password

pin


Bosch-0:
> Should duress password be cleared when password is changed?

I'd say no as well. This is an awesome feature! ty for this deregs. Will be tidying up the UI frames this week and will make sure this is integrated cleanly

Bosch-0:
Closes https://github.com/ZeusLN/zeus/issues/857 https://github.com/ZeusLN/zeus/issues/859

components/Pin.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export default function Pin({
+    onSubmit,
+    onPinChange = () => void 0,
+    hidePinLength,
+    pinLength = 4,
+    pinConfirm = false
+}: PinProps) {
+    const [pinValue, setPinValue] = useState('');
```
kaloudis, line 21: love the use of hooks. we should use them wherever we can.

components/PinCircles.tsx
```diff
@@ -0,0 +1,47 @@
...
+            margin: 10
+        }
+    });
+
+    let filled = [];
+    for (let i = 0; i < numFilled; i++) {
+        filled.push(<Filled style={styles.circles} />);
+    }
+
+    let hollow = [];
```
kaloudis, line 32: nit: `filled` and `hollow` can be consts.

package.json
```diff
@@ -112,8 +112,8 @@
     "@types/jest": "23.3.13",
     "@types/lodash": "4.14.120",
     "@types/object-hash": "2.2.1",
-    "@types/react": "16.9.19",
-    "@types/react-native": "0.61.4",
+    "@types/react": "17.0.2",
+    "@types/react-native": "0.67.2",
```
kaloudis, line 116: good fix

components/PinCircles.tsx
deregs: done

PR #871, opened 2022-03-01
Update the link path for the RaspiBolt installation guide

VajraOfIndra:
# Description

An updated version of the RaspiBolt Zeus guide is now part of the main guide! (it was part of the bonus section before,). And the overall guide has also now its own domain name.

This PR updates the link path of the RaspiBolt entry in the `README.md`

PS: in the comment just above the list of installation guide, it is stated 

> On Android Zeus has support for connecting to you node entirely over the Tor network.

Is the "On Android" still relevant these days? or can Zeus be used over Tor on iOS as well?

### Other:

- [x] Changes were made that require an update to the README

image


kaloudis:
Yes, you can use Tor on both the Android and iOS versions of Zeus. It's integrated in app/

Issue #876, opened 2022-03-02
Channels page showing incorrect amount of total outbound, inbound, and offline


deregs:
This one is a bit tricky to debug/reproduce because it's due to a race condition, but I'll outline some scenarios where it can happen:

App is closed > open app to main wallet page > go to channels (bottom right) > in total outbound, inbound, and offline fields it shows multiples (I've seen 2x and 3x multiple) of the actual amounts. See screenshot below for an example - my test node has about 509,000 sats outbound and about 233,500 sats outbound, and total outbound and inbound are showing twice those amounts.

<img width="392" alt="Screen Shot 2022-03-02 at 6 26 39 PM" src="https://user-images.githubusercontent.com/93104057/156466865-9b2b13ea-6f6f-4171-8b0b-81739cdd50b2.png">

**What's causing this:**
When the main wallet page is opened, 3 different methods are fired at about the same time that all call `getChannels` in `ChannelsStore`. `getChannels` adds up total inbound, outbound, and offline in a loop, and the loop is executing in parallel multiple times so the addition is getting duplicated. The 3 methods that call `getChannels` on wallet page open are:

https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/views/Wallet/Wallet.tsx#L98-L107
https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/views/Wallet/Wallet.tsx#L87-L92
https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/stores/ChannelsStore.ts#L46-L53

There are other workflows that cause a different duplicate call as well, such as:
wallet > channels > settings > back

This will call `getChannels` 4 times, generally in this order:

https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/views/Wallet/Wallet.tsx#L109-L116
https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/stores/ChannelsStore.ts#L46-L53
https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/views/Wallet/Wallet.tsx#L87-L92
https://github.com/ZeusLN/zeus/blob/982ac06e7917ee52fefedf70fdff51f763276d68/stores/ChannelsStore.ts#L46-L53

**Proposal to fix**
[...]

PR #877, opened 2022-03-03
Speeding up wallet refresh and fixing channel outbound/inbound totals

deregs:
Relates to #876

### DIFF
--- a/stores/ChannelsStore.ts
+++ b/stores/ChannelsStore.ts
@@ -44,15 +44,6 @@ export default class ChannelsStore {
         this.settingsStore = settingsStore;
 
         reaction(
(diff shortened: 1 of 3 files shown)


deregs:
Did some more thorough testing and found that amounts (lightning balance, on-chain balance, and total balance) were not always updating when switching nodes with this change. Fixed that by injecting `BalanceStore` on `MainPane`.

The only reason this bug wasn't appearing before was that `UNSAFE_componentWillMount` was getting called before the render, so the balances were already set by render time.

Anyway, all async methods should be called after render in `componentDidMount`, so it's good we're removing the balance fetching async calls from `componentWillMount`.

kaloudis (approved):

Issue #888, opened 2022-03-09
Zeus: linking prefix


deregs:
Support deep links, request from Tim Edwards in telegram group.

deregs:
Messaged Tim about this and can take a stab at it unless someone else gets to it first.

tim-edwards:
Thanks for looking into this for me. I'll try and provide some extra context, and as much information as I can from my deep dive into your codebase.

At THNDR Games we give out sats as prizes. In order to claim, the player needs to have a lightning wallet installed on their device (iOS or Android). By claiming we generate a LNURL withdraw and open this with a deep link.

Initially we used the url scheme "lightning:", which does indeed open a wallet; however it soon became clear that (especially on iOS), the OS just chooses the last installed wallet with the scheme "lightning". Since I believe all wallets have that scheme, at least in my experience this is no good, especially when you want to send your winning sats to a specific wallet  like Zeus. I've been changing this in our platform to use custom deep link url schemes for each wallet, and so far this has been pretty successful.

I believe you can add in a custom scheme (perhaps "zeusln"?) by just editing these three places:

1. 
https://github.com/ZeusLN/zeus/blob/007c516d0152b32be0f869548b0de2acb3e11d79/ios/zeus/Info.plist

In the Info.plist add in another string for CFBundleURLSchemes
`<string>zeusln</string>`

2.
https://github.com/ZeusLN/zeus/blob/906ae36877d6ba36fff96ec9b08615b85e93c691/android/app/src/main/AndroidManifest.xml

In the AndroidManifest, add another Intent filter

```
<intent-filter>
    <action android:name="android.intent.action.VIEW" />
    <category android:name="android.intent.category.DEFAULT" />
    <category android:name="android.intent.category.BROWSABLE" />
    <data android:scheme="zeusln" />
</intent-filter>
```

3.
https://github.com/ZeusLN/zeus/blob/bbfe141bea3f63a37d6c2fadbd8b1ec8c3072740/utils/AddressUtils.ts

Insert another if/else to check for the prefix

```
} else if (input.includes('zeusln:')) {
    value = input.split('zeusln:')[1];
} else {
    value = input;
}
```

-------------

[...]

PR #889, opened 2022-03-10
Adding zeusln deep link and fixing iOS deep link handling

deregs:
Relates to issue #888 

Adding the following for deep links:
1. Adding 'zeusln' scheme for deep links because iOS opens up the last installed wallet with the scheme 'lightning'.

2. There was a bug on iOS where deep links were not being handled, and it was only happening when Zeus was already running in the background. If Zeus is completely closed, an LNURL deep link works fine - it will open up the app and bring up the Send page. However, if Zeus was running in the background, the deep link would just open up the app to the Wallet page and do nothing. This is happening because iOS Zeus was not configured for deep links as specified [here](https://reactnative.dev/docs/linking). Also, once it was configured correctly, there was a bug where `handleOpenURL` calls `handleDeepLink`, but does not pass in `navigation`.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -94,6 +94,12 @@
             <data android:scheme="lndconnect" />
         </intent-filter>
         <intent-filter>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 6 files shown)


deregs:
Tim was able to test and says the problem is fixed for him:

Bosch-0:
Closes https://github.com/ZeusLN/zeus/issues/888

kaloudis (approved):

PR #897, opened 2022-03-14
Create feature request issue template

Bosch-0:
Quite a few issues made on issue are for feature requests, figured a template would come in handy.


deregs:
+1. Could also clear up confusion because "Issues" kinda sounds like it's only for logging bugs.

PR #962, opened 2022-04-23
add scramble pin toggle

deregs:
# Description

Adds toggle in Security settings to choose whether pin numbers should be scrambled.

Because all current users' pins are scrambled, made it so that default state is `scrambledPin = true`

Also changed the Security Settings view to use a class instead of a function in order to remain consistent with the rest of the views, and also to make it easier to inject `SettingsStore`. Without injecting `SettingsStore` I was running into bugs.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

- [ ] LND
- [ ] c-lightning-REST
- [ ] spark
- [ ] Eclair
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Settings/AddEditNode.tsx
+++ b/views/Settings/AddEditNode.tsx
@@ -285,6 +285,7 @@ export default class AddEditNode extends React.Component<
                           duressPassphrase: settings.duressPassphrase,
                           pin: settings.pin,
                           duressPin: settings.duressPin,
+                          scramblePin: settings.scramblePin,
                           authenticationAttempts:
                               settings.authenticationAttempts,
                           privacy: settings.privacy
@@ -306,7 +307,7 @@ export default class AddEditNode extends React.Component<
 
     copyNodeConfig = () => {
         const { SettingsStore, navigation } = this.props;
-        const { setSettings, settings } = SettingsStore;
+        const { settings } = SettingsStore;
         const {
             nickname,
             host,
@@ -322,7 +323,7 @@ export default class AddEditNode extends React.Component<
             implementation,
             certVerification
         } = this.state;
-        const { nodes, lurkerMode, passphrase, fiat, locale } = settings;
+        const { nodes } = settings;
 
         const node = {
             nickname: `${nickname} copy`,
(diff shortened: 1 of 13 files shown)


views/Settings/AddEditNode.tsx
```diff
@@ -340,24 +341,11 @@ export default class AddEditNode extends React.Component<
             enableTor
         };
 
-        setSettings(
```
kaloudis, line 343: why is `setSettings` ripped out here?

deregs: There's no purpose to this call as far as I can see, we're just getting the settings and setting it to the exact same settings. Which makes sense, when copying the node config nothing in the settings changes. The copied node is only added to settings once save node config is pressed on the copied configuration.

views/Settings/Security.tsx
```diff
@@ -133,35 +154,97 @@ function Security(props: SecurityProps) {
...
+                    }}
+                >
+                    <ListItem.Content>
+                        <ListItem.Title
+                            style={{
+                                color: themeColor('secondaryText'),
+                                fontFamily: 'Lato-Regular'
+                            }}
+                        >
+                            {'Scramble PIN numbers'}
```
kaloudis, line 210: This should be added to `locales/en.json`

Bosch-0:
I'd recommend this be off by default and have it opt-in for a better UX

views/Settings/Security.tsx
deregs: Done

deregs:
> 

I agree, but discussed with @kaloudis and we thought for now it should default to scrambled. Since all current users have scramble PIN on, we want to avoid defaulting to non-scrambled and unscrambling current users' pins. Ideally, we could design out a create pin flow that includes the scramble pin toggle so the user can just decide scrambled/non-scrambled off the bat. It also might make more sense to users if they could toggle scrambled on/off during pin creation and see the numbers on the pin pad get scrambled/unscrambled and decide which they prefer.

kaloudis (approved):

PR #967, opened 2022-04-27
Only show pin toggle when a pin is set

deregs:


views/Settings/Security.tsx
```diff
@@ -63,6 +65,12 @@ export default class Security extends React.Component<
             scramblePin: settings.scramblePin ?? true
         });
 
+        if (settings.pin) {
+            this.setState({
+                pinExists: true
+            });
+        }
```
kaloudis, line 72: Why should we write this to the state instead of just checking for `settings.pin` on line 206?

deregs: I found there are sometimes bugs if we don't fetch the updated `settings` using `getSettings`. When we just get the `settings` object from `SettingsStore`, `settings` is sometimes out of date. Observed this on multiple pages, but especially on this page where the settings are changing a lot.

kaloudis (approved):

Issue #1090, opened 2022-10-11
Remove LND default fee limit of 10 sats


kaloudis:
This is far too low for many users and causes a bunch of payment failures. Let's make this an optional field and let LND defaults take over if not specified.

otech47:
will cut a small PR for this to change the limit to 100 sats just to reduce payment failures

another PR should involve a bigger refactor to use the streaming API instead so that the fee can be dynamically managed by LND

PR #1101, opened 2022-10-13
Change fee limit

otech47:
# Description

Relates to issue: https://github.com/ZeusLN/zeus/issues/1090

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance 
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android ---- emulator Pixel 4 Android 12
- [ ] iOS

- [x] LND
- [ ] c-lightning-REST
- [ ] spark
- [ ] Eclair
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/PaymentRequest.tsx
+++ b/views/PaymentRequest.tsx
@@ -11,7 +11,9 @@ import LoadingIndicator from './../components/LoadingIndicator';
 import TextInput from './../components/TextInput';
 
 import InvoicesStore from './../stores/InvoicesStore';
-import TransactionsStore from './../stores/TransactionsStore';
+import TransactionsStore, {
+    SendPaymentReq
+} from './../stores/TransactionsStore';
 import UnitsStore from './../stores/UnitsStore';
 import ChannelsStore from './../stores/ChannelsStore';
 import SettingsStore from './../stores/SettingsStore';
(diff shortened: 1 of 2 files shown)


otech47:
@kaloudis should this bigger refactor be handled in scope of the same issue or is this mergeable as is?

kaloudis:
We can merge this. Want to create a new issue for implementing the streaming `sendPayment` LND endpoint?

otech47:
I'm not sure I understand the streaming `sendPayment` solution, but if you can clear it up I'll create the issue and add the details

which API are you referring to exactly?

I skimmed the LND API reference and couldn't find an API that provides the ability to do the dynamic fee estimation

is the idea to use the same API endpoint but also make sure to grab the stream of payments updates being returned to figure out if the payment failed and then retry the payment with a higher fee? or is there something built-in to LND that I'm missing?

kaloudis:
@otech47 here's the streaming endpoint https://api.lightning.community/#sendpaymentv2

kaloudis (dismissed):

views/PaymentRequest.tsx
```diff
@@ -556,7 +582,7 @@ export default class PaymentRequest extends React.Component<
                                             color: 'white'
                                         }}
                                         onPress={() => {
-                                            TransactionsStore.sendPayment({
+                                            this.sendPayment({
```
deregs, line 585: What is the purpose of creating a `sendPayment` method here instead of just calling `TransactionsStore.sendPayment` directly? I remember we were talking about a reason to do this before, but since the scope of this PR is reduced to just changing the default fee amount I'm not sure that this method is needed.

otech47: Reasoning is in this commit message

https://github.com/ZeusLN/zeus/pull/1101/commits/b5c62a8ca430b991bc565c13d5c26f10c5276989

But yes I agree with your point and adding the type check is the only other difference to how it was written before. I can revert this commit but if we plan on writing our own fee estimation logic may as well leave it

otech47:
> @otech47 here's the streaming endpoint https://api.lightning.community/#sendpaymentv2

@kaloudis from what I can tell, the REST counterpart to the GRPC call you linked ^ is already implemented [here](https://github.com/setlife-network/zeus/blob/b5c62a8ca430b991bc565c13d5c26f10c5276989/backends/LND.ts#L235)

Screen Shot 2022-10-18 at 9 44 57 PM

so it's not about using a different endpoint because from what I can tell there is no easy parameter we can pass that will do dynamic fee estimation... so to solve the original problem more completely, we will need to expand the functionality of the response handler for `payLightningInvoice` so that at the very least it can respond with a "fee too low" error.. will open an issue and try to settle on some specs for a solution

views/PaymentRequest.tsx
deregs: Ah ok should be fine to leave as is then

kaloudis (approved):

Issue #739, opened 2022-01-28
c-lightning with spark end in spark error LightningError: No method


cryptoteun:
**Describe the bug**
I try to connect to a c-lightning node via spark (via a cloudflare proxy). When I add this config Zeus cant connect. In the spark log I see the following:

```
POST /rpc 500 3.559 ms - 103
    at Socket.emit (node:events:394:28)
LightningError: No method
    at createError (/opt/spark/node_modules/error/typed.js:40:22)
    at LightningClient.<anonymous> (/opt/spark/node_modules/clightning-client/index.js:122:28)
    at Object.onceWrapper (node:events:514:26)
    at LightningClient.emit (node:events:394:28)
    at Interface.<anonymous> (/opt/spark/node_modules/clightning-client/index.js:76:17)
    at Interface.emit (node:events:394:28)
    at Interface._onLine (node:readline:487:10)
    at Interface._normalWrite (node:readline:661:12)
    at Socket.ondata (node:readline:269:10)
    at Socket.emit (node:events:394:28)
```

**To Reproduce**
1. Add url (https://domain)
2. Add key
3. No cert. validation 
4. Save
5. Zeus tries to connect

**Expected behavior**
Zeus can connect

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Zeus Version 0.6 Beta2 /Beta1
 - Clearnet
 - Spark 0.3.1

**Additional context**

kaloudis:
Were you able to connect to this setup in earlier versions?

cryptoteun:
In beta1 I had the same error. Version before I was experiencing the flickering issue so couldn't test.

LightningK0ala:
I have the same problem, it breaks the Zeus app for me. I'm unable to go to settings (app crashes) and every time I restart the app it automatically tries to connect to the node and then I'm stuck on the error again.

sha-265:
Same problem

SimonVrouwe:
Some more detail by running `spark-wallet --verbose`

On a regtest setup with latest c-lightning, Zeus wallet (on Android) connected to spark via local LAN via port forwarding using JuiceSSH.
```
spark-wallet --verbose
  clightning-client Connecting to /home/simon/.lightning2/regtest/lightning-rpc +0ms
  clightning-client #1 --> getinfo [] +4ms
  clightning-client Lightning client connected +4ms
  clightning-client #1 <-- {
[...]

PR #1132, opened 2022-11-25
fix original spark backend

SimonVrouwe:
This is a work in progress and just wanted to share, for feedback.
**warning** I am new to javascript / typescript / node / React and to this repo, so big chance I made some blunders.

The Zeus + [Spark](https://github.com/shesek/spark-wallet) + c-lightning combination appears pretty broken, did it ever work?
Relates to issue: #739

this PR fixes:
- missing "Content-type": "application/json" in the POST header
- /rpc method parameters now passed as array and in the correct order
- a weird network Error, which seems related to ReactNativeBlobUtil.fetch
- accept the latest _msat amount fields returned by c-lightning (but also still accept old ones)

**edit**:  I discovered the [Sparko](https://github.com/fiatjaf/sparko) plugin (not Spark), which is less strict about the POST header's "Content-type" and the body's "params" value (it accepts array or object) and has no issues with `ReactNativeBlobUtil.fetch()`. Perhaps the Node Interface option **"Spark (c-lightning)"** was really meant to be called **"Sparko (c-lightning)"** ? This is confusing to say the least.

Sparko also has some issues:
- [x] duplicate URI in Node info screen **solved**: Sparko also sees the 'dns' address type, Spark only the 'ipv4'
- [x] ~pressing the ^ button in home screen kills the app~ stays alive, but difference in Spark/Sparko, see below

~Some small fixes and 1st and 2nd commits are mostly for debugging (can be dropped).~

**edit**: Basic functionality and buttons in the (Android) App have been tested, but no full coverage yet.
Testing with latest cln `--allow-deprecated-apis=false` will probably reveal more deprecated methods we still rely on.
Here an (incomplete) list:
- [x] `setchannelfee`, is replaced by `setchannel` (since [0.11.0.1](https://github.com/ElementsProject/lightning/blob/master/CHANGELOG.md))
- [ ] incorrect dates in payment list and/or some extra entries (compared to Sparko), see picture
Screenshot_1671104775

Will try to make it work with (old) Spark an Sparko.

- [x] Bug fix

## Checklist

## Testing

- [x] Android
- [ ] iOS

- [ ] LND
- [ ] c-lightning-REST
- [x] spark
- [ ] sparko
- [ ] Eclair
- [ ] LndHub

### Other:

### DIFF
--- a/backends/Spark.ts
+++ b/backends/Spark.ts
@@ -72,7 +72,17 @@ export default class Spark {
 
     getTransactions = () =>
         this.rpc('listfunds').then(({ outputs }: any) => ({
-            transactions: outputs
+            transactions: outputs.map((o: any) => ({
+                // I guess Transaction class expects satoshi's
+                amount:
+                    typeof o.amount_msat === 'number'
+                        ? o.amount_msat / 1000
+                        : Number(o.amount_msat.replace('msat', '')) / 1000,
+                blockheight: o.blockheight,
+                status: o.status,
+                txid: o.txid,
+                address: o.address
+            }))
         }));
     getChannels = () =>
         this.rpc('listpeers').then(({ peers }: any) => ({
--- a/backends/Spark.ts
+++ b/backends/Spark.ts
@@ -156,10 +156,24 @@ export default class Spark {
         this.rpc('listfunds').then(({ channels }: any) => ({
             balance: channels
                 .filter((o: any) => o.state === 'CHANNELD_NORMAL')
-                .reduce((acc: any, o: any) => acc + o.channel_sat, 0),
+                .reduce(
+                    (acc: any, o: any) =>
+                        acc +
+                        (typeof o.our_amount_msat === 'number'
+                            ? o.our_amount_msat / 1000
+                            : o.channel_sat),
+                    0
+                ),
             pending_open_balance: channels
                 .filter((o: any) => o.state === 'CHANNELD_AWAITING_LOCKIN')
-                .reduce((acc: any, o: any) => acc + o.channel_sat, 0)
+                .reduce(
+                    (acc: any, o: any) =>
+                        acc +
+                        (typeof o.our_amount_msat === 'number'
+                            ? o.our_amount_msat / 1000
+                            : o.channel_sat),
+                    0
+                )
         }));
     sendCoins = (data: TransactionRequest) =>
         this.rpc('withdraw', {
--- a/backends/Spark.ts
+++ b/backends/Spark.ts
@@ -141,10 +141,24 @@ export default class Spark {
         this.rpc('listfunds').then(({ outputs }: any) => {
             const unconf = outputs
                 .filter((o: any) => o.status !== 'confirmed')
-                .reduce((acc: any, o: any) => acc + o.value, 0);
+                .reduce(
+                    (acc: any, o: any) =>
+                        acc +
+                        (typeof o.amount_msat === 'number'
+                            ? o.amount_msat / 1000
+                            : o.value),
+                    0
+                );
             const conf = outputs
                 .filter((o: any) => o.status === 'confirmed')
-                .reduce((acc: any, o: any) => acc + o.value, 0);
+                .reduce(
+                    (acc: any, o: any) =>
+                        acc +
+                        (typeof o.amount_msat === 'number'
+                            ? o.amount_msat / 1000
+                            : o.value),
+                    0
+                );
 
             return {
                 total_balance: conf + unconf,
--- a/backends/Spark.ts
+++ b/backends/Spark.ts
@@ -342,10 +342,10 @@ export default class Spark {
         };
     };
     setFees = (data: any) =>
-        this.rpc('setchannelfee', {
+        this.rpc('setchannel', {
             id: data.global ? 'all' : data.channelId,
-            base: data.base_fee_msat,
-            ppm: data.fee_rate * 1000000
+            feebase: data.base_fee_msat,
+            feeppm: data.fee_rate * 1000000
         });
     getRoutes = async (urlParams: Array<string>) => {
         const msatoshi = Number(urlParams[1]) * 1000;
--- a/backends/Spark.ts
+++ b/backends/Spark.ts
@@ -213,16 +213,15 @@ export default class Spark {
                     memo: inv.description,
                     r_preimage: inv.payment_preimage,
                     r_hash: inv.payment_hash,
-                    value: inv.msatoshi / 1000,
-                    value_msat: inv.msatoshi,
+                    amt_paid_sat: (typeof inv.amount_msat === 'number'
+                        ? Number(inv.amount_msat) / 1000
+                        : Number(inv.amount_msat.replace('msat', '')) / 1000
+                    ).toString(),
                     settled: inv.status === 'paid',
                     creation_date: inv.expires_at,
                     settle_date: inv.paid_at,
                     payment_request: inv.bolt11,
                     expiry: inv.expires_at,
-                    amt_paid: inv.msatoshi_received / 1000,
-                    amt_paid_sat: inv.msatoshi_received / 1000,
-                    amt_paid_msat: inv.msatoshi_received
                 }))
             })
         );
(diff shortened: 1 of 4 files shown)


kaloudis:
Great to see this. I'm going to try to add Spark to my CLN node this upcoming week for testing.

views/Send.tsx
```diff
@@ -156,7 +156,9 @@ export default class Send extends React.Component<SendProps, SendState> {
         }
 
         if (Platform.OS === 'android') {
-            await this.enableNfc();
+            await this.enableNfc().catch((e) => {
+                console.log('Send.tsx componentDidMount:', e.message);
+            });
```
kaloudis, line 161: were you having issues with NFC?

index.js
```diff
@@ -9,4 +9,6 @@ import './shim.js'
 import App from './App.tsx';
 import {name as appName} from './app.json';
 
+console.disableYellowBox = true;
```
kaloudis, line 12: please remove all code used for debugging

views/Send.tsx
SimonVrouwe: No just to suppress some warnings "Possible Unhandled Promise Rejection", but they have disappeared. Will drop this commit.

index.js
SimonVrouwe: yes

kaloudis:
@SimonVrouwe  I think you're correct in your assessment. When @fiatjaf implemented this, I believe he was using a Sparko backend.

fiatjaf:
Oops, sorry about this mess. I really gave my best at the time, but I think I was a little confused since I had never done any Android development and my development UX was very bad.

SimonVrouwe:
> I had never done any Android development

@fiatjaf The same here :), if you want I will take a step back and close this PR.

Otherwise maybe it is possible to keep Zeus's spark backend compatible with both (original) Spark and Sparko.
But at the cost of not using Sparko's conveniences like accepting a "params" **object** and not having issues with `ReactNativeBlobUtil.fetch()`.

Any idea why Sparko has no issues with `ReactNativeBlobUtil.fetch()` and Spark does? (showing "Error: unexpected end of stream") It seems Zeus doesn't make use the SSE /stream, maybe it should?

fiatjaf:
No, please keep it. Technically Sparko should support everything Spark supports and some other things, so if this makes Spark work Sparko should too. If there is any problem I'd rather fix it on Sparko side.

backends/Spark.ts
```diff
@@ -19,23 +18,37 @@ export default class Spark {
...
         }
-        const body = JSON.stringify({ method: rpcmethod, params: param });
+
+        /* Spark-wallet '/rpc' API (v0.3.2-rc) does not accept objects and
+         * "params" should be an array of parameter values in the correct order
+         * per the API https://lightning.readthedocs.io. Required but empty
+         * values should be set JSON null. */
+        const body = JSON.stringify({
+            method: rpcmethod,
+            params: Object.values(param)
```
fiatjaf, line 35: I believe this won't work because you'll be sending the parameters in some undefined order which is likely to not match the order expected by CLN.

I don't understand why Spark is now enforcing this crazy policy, do you? Now if a client has to send parameters to Spark it must know the exact order CLN expects them.

kaloudis:
@SimonVrouwe how are you feeling about where things are at right now?

SimonVrouwe:
> @SimonVrouwe how are you feeling about where things are at right now?

Sorry I got a little side tracked.

The rebase removed the debug code and this is now at bare minimum to at least communicate with the Spark/Sparko backend.

There is a small difference between Spark and Sparko w.r.t. the payment list (^ button), Spark shows some extra entries with weird dates (`75).
 
Then I have a bunch of patched to make it compatible with latest c-lightning (returning '_msat' fields as numbers) **and** keep it backward compatible with earlier versions (returning '_msat' fields as strings ending in "msat"), but am not there yet. Maybe that is too ambitious.

To be honest I got a bit lost exploring Zeus code, it seems `msat` to `sat` conversions and "msat" replacements are done at various levels. For example at the backends (interface) level:
https://github.com/ZeusLN/zeus/blob/bf164b40d3d0b465d217e85865c3dad853503354/backends/Spark.ts#L80-L82
but also at the models level:
https://github.com/ZeusLN/zeus/blob/bf164b40d3d0b465d217e85865c3dad853503354/models/Payment.ts#L98-L105

Do you have developer guidelines for that?

Perhaps you want all conversions at a single level (in backends?), but again I have no experience in Reactive programming.

Then finally, is there a (integration) test framework to test all wallet functions (clicking buttons etc.) with a particular backend?

kaloudis:
@SimonVrouwe backends should be preferred for conversion over models. Unfortunately we don't have a UI testing framework in place :(

SimonVrouwe:
>backends should be preferred for conversion over models

Ok but for that there should be an interface (in models class?) that clearly defines what fields (names + types) it expects from a backend. Currently this comes down to guessing, for example the abstraction of a Payment:
https://github.com/ZeusLN/zeus/blob/8b684a9072d99252d78a535084ce03e4da34071c/models/Payment.ts#L8-L14

it has a long list of fields, many optional.
How can I know what fields are required and what their meaning is to Zeus?
It is very painful to first have to understand the entire Payment class (and beyond, see below) to figure this out.

Conversions/filtering seems to happen all over the place.
For example when using an Eclair backend, the `PaymentsStore.getPayments()` method awaits `RESTUtils.getPayments()`:
https://github.com/ZeusLN/zeus/blob/8b684a9072d99252d78a535084ce03e4da34071c/stores/PaymentsStore.ts#L34-L43

which calls Eclair backend, where some conversion/selection (**1**) happens.
https://github.com/ZeusLN/zeus/blob/8b684a9072d99252d78a535084ce03e4da34071c/backends/Eclair.ts#L210-L212
 
Then at L37 in `PaymentsStore.ts` another selection (**2**) is done, which is passed to create a `new Payment`.

And then `Payment` does another bunch of conversions/selection (**3**) in its methods, for example:
https://github.com/ZeusLN/zeus/blob/8b684a9072d99252d78a535084ce03e4da34071c/models/Payment.ts#L86-L91

So the conversions/filtering happens at various levels:
1. inside `Eclair.ts`
2. inside `PaymentsStore.ts`
3. inside `Payment.ts`

Some conversions even appear to reverse the effects of one applied at another level.
Unifying this into a single level would make reviewing/understanding the code much easier.

To continue this PR (the Spark backend), more info is needed.
- what data the backend should return, i.e. which field names + types
- the meaning of that data

Maybe for starters some comments can be added to the models classes (Payment, Transaction etc.) about the minimal set of attributes (fields) they require and their meaning to Zeus.

I will push what I have so far and turn this PR into a draft.

kaloudis:
I've updated the README to specify compatibility with Sparko, not Spark as we work through this: https://github.com/ZeusLN/zeus/pull/1205/files

SimonVrouwe:
@kaloudis I gave up, IMHO this project is not ready for mainnet, at least not for c-lightning (neither Spark nor Sparko) as it is untested. Are any other backends tested?

Although I am still reading books on Javascript and Typescript so who am I to give lectures, but here my intuitions:

- No clear separation between logic and view, for example this last minute switch to show blockheight instead of a date:
https://github.com/ZeusLN/zeus/blob/85cb487c5a5738bb6cec87312fafb370e8e972fb/views/Activity/Activity.tsx#L315-L319
Such decisions should probably not be made in the render section. 

- Inconsistent use of typescript, especially w.r.t. backend responses, for example:
https://github.com/ZeusLN/zeus/blob/85cb487c5a5738bb6cec87312fafb370e8e972fb/models/Payment.ts#L8-L11
suggests `value` is of type string, but then `eclair` returns many fields, which all seem equivalent, as `number`?:
https://github.com/ZeusLN/zeus/blob/85cb487c5a5738bb6cec87312fafb370e8e972fb/backends/Eclair.ts#L225-L227
When was the last time anybody tested with the eclair?

- As mentioned earlier, these `/models` classes becomes an incomprehensible soup of fieldnames (many redundant) returned by the various backends. Either pick one popular backend as a standard or define some generalized field names/types (typescript interface?) with a clearly stated meaning. Stick to it and make all the (other) backends adapt in the backend-side.

End-of-rant,
and good-luck

kaloudis:
@SimonVrouwe no worries. appreciate the feedback. Largely agree with your sentiment and have begun to track efforts to tackle the issues you've listed here: https://github.com/ZeusLN/zeus/issues/1209

PR #1259, opened 2023-01-13
Point of Sale: Square integration

kaloudis:
# Description

This PR adds a Point of Sale interface for Square terminals

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance 
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [x] iOS

- [ ] LND (REST)
- [ ] Core Lightning (Spark)
- [ ] Eclair
- [ ] LndHub

### Locales
- [x] I’m aware that new translations should be made on the Zeus [Transfix page](https://www.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Other:


views/Wallet/PosPane.tsx
```diff
@@ -0,0 +1,219 @@
...
+            filteredOrders: result
+        });
+    };
+
+    renderItem = (order) => {
+        const { navigation, FiatStore } = this.props;
+        const { getRate, loading } = FiatStore;
+        const { item } = order;
+        return (
+            <TouchableHighlight
```
Talej, line 95: - While fiat rate was being fetched the order couldn't be viewed but it wasn't entirely obvious why... maybe touchability needs to be disabled while rates are loaded?

views/Wallet/PosPane.tsx
```diff
@@ -0,0 +1,219 @@
...
+        );
+    };
+
+    render() {
+        const { SettingsStore, PosStore, FiatStore, navigation } = this.props;
+        const { search, filteredOrders } = this.state;
+        const { loading, getOrders } = PosStore;
+        const orders = filteredOrders;
+        const { getRate, getFiatRates } = FiatStore;
+        const fiatLoading = FiatStore.loading;
```
Talej, line 118: Intermittently the fiat rate fetch was failing or taking a long time. Might be worth having a short timeout + cache fallback so it can still be used if the fetch fails

views/Wallet/PosPane.tsx
```diff
@@ -0,0 +1,219 @@
...
+                item.getItemsList.includes(value) ||
+                item.getItemsList.toLowerCase().includes(value)
+        );
+        this.setState({
+            search: value,
+            filteredOrders: result
+        });
+    };
+
+    renderItem = (order) => {
```
Talej, line 90: Not hugely important but RN complaining about no key for items

`Warning: Each child in a list should have a unique "key" prop.`

views/Wallet/PosPane.tsx
```diff
@@ -0,0 +1,219 @@
...
+                    date={item.getDisplayTime}
+                />
+            </TouchableHighlight>
+        );
+    };
+
+    render() {
+        const { SettingsStore, PosStore, FiatStore, navigation } = this.props;
+        const { search, filteredOrders } = this.state;
+        const { loading, getOrders } = PosStore;
```
Talej, line 115: two unusual things happening with getOrders.

1. Occasionally when refreshed the orders listed would disappear even though the square sandbox endpoint seemed to be returning a result
2. After I added my first test order I added 2 more orders in exactly the same way but the square endpoint was only ever returning the 1st order (DM me and I'll provide you my auth token and location ID for this if it helps).

Talej (commented):
A couple of general items

- Should the balance / keypad views still be accessible? I couldn't find any way to get back to them with POS enabled

- Was occasionally getting this error popping up. Not really sure what was causing it - maybe related to fetching fiat rates?

IMG_9197

views/Wallet/PosPane.tsx
kaloudis: were you sure to update your idempotency_key on the POST call?

Talej: That's what it was. All good now

Talej:
Just one more item I have found this morning. I created an invoice and paid the invoice by switching between apps on the same device. It looks like because of that the order didn't get marked as paid even though the invoice was paid. 

This was using LND over clearnet.

Maybe there's a bug or perhaps it needs a way to manually mark an order as paid just as a fallback?

IMG_9221

IMG_9222

views/Wallet/PosPane.tsx
kaloudis: This should be covered by the flashing `Fetching fiat rates` at the top of the screen but will revisit if there's confusion out in the field

kaloudis:
@Talej I think the listener gets killed when you navigate away from it. Will evaluate it.

Will assess this weekend whether to build the mark function into the UI or if it's OK to just have staff do that from the Square terminal. Likely needs an admin menu.

Issue #833, opened 2022-02-16
Add support for QR code imports from gallery


kaloudis:

kaloudis:
I believe BlueWallet is using this library to parse text out of images: https://www.npmjs.com/package/@remobile/react-native-qrcode-local-image

kaloudis:
https://github.com/BlueWallet/BlueWallet/blob/94039fa0049e84d7222053a5320d2e340ce23db0/screen/send/ScanQRCode.js

kaloudis:
Added in v0.7.3

PR #1296, opened 2023-02-03
QRCodeScanner: import QR code images from gallery

shubhamkmr04:
# Description

Relates to issue: **https://github.com/ZeusLN/zeus/issues/833**

The description of the Issue and how did I approach it with the solution is in this [documentation](https://docs.google.com/document/d/1aSSd3oH0Ud_W7hNUZrbk1uBGpWxGRQDKnpb58-hjQBg/edit)

- [X] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance 
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [X] N/A

- [X] Android
- [ ] iOS

- [X] LND (REST)
- [ ] Core Lightning (Spark)
- [ ] Eclair
- [ ] LndHub

### Locales

- [X] Contributors will need to run `yarn` after this PR is merged in

### Other:


backends/LND.ts
```diff
@@ -144,7 +144,7 @@ export default class LND {
         route: string,
         ws?: boolean
     ) => {
-        const hostPath = host.includes('://') ? host : `https://${host}`;
+        const hostPath = host.includes('://') ? host : `http://${host}`;
```
kaloudis, line 147: gotta undo this change as it's only for dev

package.json
```diff
@@ -75,6 +76,7 @@
     "react-native-encrypted-storage": "4.0.2",
     "react-native-gesture-handler": "2.2.0",
     "react-native-hce": "0.1.2",
+    "react-native-image-picker": "^5.0.1",
```
kaloudis, line 79: I like to lock the version numbers here. Please remove the `^` from both new additions and run yarn again, please.

components/QRCodeScanner.tsx
```diff
@@ -67,6 +70,29 @@ export default class QRCodeScanner extends React.Component<QRProps, QRState> {
...
     };
 
+    handleOpenGallery = () => {
+        launchImageLibrary(
+            {
+                mediaType: 'photo'
+            },
+            (response) => {
+                if (response.didCancel) {
+                    console.log('USER CANCELLED IT');
```
kaloudis, line 80: Do we need to take any other action in this instance? perhaps we remove this block and wrap the other action in `if (!response.didCancel)`

backends/LND.ts
shubhamkmr04: got it!

components/QRCodeScanner.tsx
shubhamkmr04: Okay sure. We can do this for now. Though we might use this instance to change a loading variable like `loading == false` in future

android/app/build.gradle
```diff
@@ -202,11 +202,12 @@ dependencies {
     implementation 'com.android.support:multidex:1.0.3'
     implementation project(':react-native-randombytes')
     implementation "androidx.core:core:1.1.0"
-    implementation "com.facebook.react:react-native:+"  // From node_modules
+    implementation "com.facebook.react:react-native:0.70.6!!"  // From node_modules
```
kaloudis, line 205: is it required to specify the version here?

shubhamkmr04: Yes I was getting some BUILD error, so I had to specify the version. Like I read it from [here](https://stackoverflow.com/questions/74325460/a-failure-occurred-while-executing-com-android-build-gradle-internal-tasks-merge)

shubhamkmr04:
The previous commit introduce a patch file to upgrade build.gradle file of `@remobile/react-native-qrcode-local-image`

package.json
```diff
@@ -11,13 +11,14 @@
   },
   "scripts": {
     "start": "react-native start",
-    "postinstall": "rn-nodeify --install --hack; npx jetify",
+    "postinstall": "rn-nodeify --install --hack; npx jetify; git apply patches/rnqli-build.gradle.patch",
```
kaloudis, line 14: up here let's turn the patch func to `yarn run patch` since we have it defined below

shubhamkmr04: Okay! then would I have to shift "postinstall" below "patch" ?

kaloudis: I don't think you have to but please double check before committing the change

shubhamkmr04: tested that. Its working perfectly

kaloudis (approved):
tACK. great job!

PR #1317, opened 2023-02-15
Channels: maintain channelsType state between renders

Talej:
# Description

After viewing a closed channel and then going back to the list of channels the channel list is refreshed and sent back to the open channel view. This change maintains the current channel list view.

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance 
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [X] N/A

- [ ] Android
- [X] iOS

- [ ] LND (REST)
- [X] LND (Lightning Node Connect)
- [X] Core Lightning (c-lightning-REST)
- [ ] Core Lightning (Spark)
- [ ] Eclair
- [ ] LndHub

### Locales
- [X] I’m aware that new translations should be made on the Zeus [Transfix page](https://www.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -193,6 +193,8 @@
     "views.Wallet.Wallet.invoices": "Invoices",
     "views.Wallet.Wallet.onchain": "On-chain",
     "views.Wallet.Wallet.channels": "Channels",
+    "views.Wallet.Wallet.pendingChannels": "Pending Channels",
+    "views.Wallet.Wallet.closedChannels": "Closed Channels",
     "views.Wallet.Wallet.startingUp": "Zeus is starting up.",
     "views.Wallet.Wallet.connecting": "Zeus is connecting to your node.",
     "views.Wallet.restart": "Restart",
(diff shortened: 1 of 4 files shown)


kaloudis:
- No need to maintain state on the page if we're tracking it in the store
- If we're gonna write the state to the store, it should have a proper function in the store
- As is, there will be issues when switching to implementations that don't support pending/closed channels yet when you're on one of those states

Talej:
> * No need to maintain state on the page if we're tracking it in the store
> * If we're gonna write the state to the store, it should have a proper function in the store
> * As is, there will be issues when switching to implementations that don't support pending/closed channels yet when you're on one of those states

All done... only thing we may still want is to flesh out the reading/writing in the store a bit more? Open to your thoughts here...

kaloudis:
Looking much better. Perhaps we should create some enums for channel state though

coreweb:
enum added in. Also added some missing translations

Talej:
enum added in. Also added some missing translations

kaloudis (approved):
Much better now. Nice work!

PR #1486, opened 2023-06-09
Fixed updating selected node after deleting a node

myxmaster:
# Description

When deleting a node which has a lower index than the selected one, the index of the selected node was not updated. So another node became active or the index even was out of bounds.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance 
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [x] LND (REST)
- [ ] Core Lightning (Spark)
- [ ] Eclair
- [ ] LndHub

### Locales
- [ ] I’m aware that new translations should be made on the Zeus [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Other:

### DIFF
--- a/views/Settings/NodeConfiguration.tsx
+++ b/views/Settings/NodeConfiguration.tsx
@@ -37,6 +37,7 @@ import SettingsStore, {
 } from '../../stores/SettingsStore';
 
 import Scan from '../../assets/images/SVG/Scan.svg';
+import { Settings } from '../../stores/SettingsStore';
 
 interface NodeConfigurationProps {
     navigation: any;
@@ -90,7 +91,7 @@ export default class NodeConfiguration extends React.Component<
         port: '',
         macaroonHex: '',
         saved: false,
-        index: null,
+        index: null as number | null,
         active: false,
         newEntry: false,
         implementation: 'lnd',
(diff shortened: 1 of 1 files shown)


kaloudis:
Does this address https://github.com/ZeusLN/zeus/issues/1487?

myxmaster:
No, this here only makes sure that the index of the selectedNode is updated correctly after user deletes a node which has index<current Index. In edge cases it could happen that an index is out of bound, simple example: There are only to node configs (index 0 and 1), user is currently on 1. Now user deletes 0, but selectedNode stays on 1, which is now out of bound becuase there is only index 0 existing after the deletion of 0 (1 becomes 0).

Since it is clear that this is unwanted behavior, I didn't start a discussion about it. #1487 is different, because you might have an opinion about how the app should handle this. But we should discuss that in that issue.

kaloudis:
Left a comment on https://github.com/ZeusLN/zeus/issues/1487

myxmaster:
Ok, but again, this here is not changing the behavior. It only fixes a bug.

kaloudis (approved):
tACK

PR #1518, opened 2023-06-23
Display human readable time until expiration

### LINKED ISSUE #1509: Display expiry time of invoices in minutes, hours or days if useful
### Describe the feature

In activity screen the expiry time of invoices are shown in seconds. It should be converted to minutes, hours or days if useful.

myxmaster:
# Description

Instead of displaying the initially defined time until expiration, the remaining time is now displayed in a human readable form. Also made the layout more consistent by putting labels on the left and values on the right.

grafik

grafik

Before:

grafik

grafik

I changed the keys of LOCALE_KEYS to ISO 639-1 to be able to use this as language for humanize-duration. Might also be needed for similar cases in the future.

This fixes #1509.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance 
- [x] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [x] LND (REST)
- [ ] Core Lightning (Spark)
- [ ] Eclair
- [ ] LndHub

### Locales
- [ ] I’m aware that new translations should be made on the Zeus [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Other:


myxmaster:
Forgot to mention that duration is displayed with 2 units max and won't go lower than seconds, examples:
1 year, 210 days
2 days, 4 hours
3 hours, 24 minutes
16 minutes, 21 seconds
53 seconds

package.json
```diff
@@ -31,6 +31,7 @@
     "@react-navigation/native": "6.0.16",
     "@remobile/react-native-qrcode-local-image": "github:BlueWallet/react-native-qrcode-local-image#b8baa79",
     "@tradle/react-native-http": "2.0.1",
+    "@types/humanize-duration": "^3.27.1",
```
kaloudis, line 34: please lock the versions by removing the `^` for both added dependencies and run `yarn` again

stores/SettingsStore.ts
```diff
@@ -127,36 +127,65 @@ export const LNC_MAILBOX_KEYS = [
...
+    { key: 'el', value: 'Ελληνικά' },
+    { key: 'fa', value: 'زبان فارسي' },
+    { key: 'sl', value: 'Slovenski jezik' },
+    { key: 'ru', value: 'русский язык' },
+    { key: 'fi', value: 'Suomen kieli' },
+    { key: 'it', value: 'Italiano' },
+    { key: 'vi', value: 'Tiếng Việt' },
+    { key: 'jp', value: '日本語' },
+    { key: 'he', value: 'עִבְרִית' },
+    { key: 'hr', value: 'Hrvatski' }
```
kaloudis, line 156: the locale changes here should be a separate PR

package.json
myxmaster: Well, the yarn.lock file already locks the version. Locking the version in the package.json file means that packages cannot be simply updated by running [yarn upgrade](https://classic.yarnpkg.com/lang/en/docs/cli/upgrade/).

I'm aware this might cause problems when adding a package because all the packages are potentially updated (which should normally not be a problem since only major updates should bring breaking changes). But keeping dependencies up-to-date is important for security and thus the work needs to be done regularly.

By the way, this is no problem for reproducible builds since yarn install is run with `--frozen-lockfile`.

stores/SettingsStore.ts
myxmaster: This PR depends on this change since the two-letter language ISO code is needed by humanize-duration for localization.

package.json
kaloudis: We'd rather handle the upgrades manually. Please lock the version. Keeping packages up to dates is important but it can also be poor for security to just blindly accept updates.

We do use `--frozen-lockfile` for repro builds.

stores/SettingsStore.ts
kaloudis: It's a big change. Please make a separate PR for it and then you can base these changes off of that PR.

We also have to test this for regression. Will someone who has set their language in an older release lose their setting - will it break the display here?

myxmaster: Alright, I'll create new PRs.

I tested the migration of the language setting and it works, no settings are lost.

PR #1524, opened 2023-06-29
Bug Fix: External links not re-directing to mempool

shubhamkmr04:


kaloudis (changes requested):
While this fix works, we lose the functionality of the touchable opacity on copy events that indicate to users that something has happened.

I think a better fix would be to flesh out the `KeyValue` component to have some sort of prop that we can use to block copy functionality.

kaloudis (approved):

Issue #1542, opened 2023-07-15
Note is the same for all sent transactions


thebullishbitcoiner:
### Describe the bug

I tried setting a unique note for a few transactions and noticed that the same note was being applied to all of them. 

### Reproduce

1. Set a note for a sent transaction
2. Open another sent transaction 
3. The previous note is loaded 

### Zeus version

v0.7.6

### Node interface

LndHub

### Network

Clearnet

### Device

iPhone 11

### Device operating system

iOS 16.5.1

kaloudis:
@thebullishbitcoiner could you provide any more info on the LNDHub instance you're using? There are quite a few conflicting implementations.

We use the payment hash for an identifier on notes on sent payments on LN. Seems to work fine for me with LN.tips' instance

thebullishbitcoiner:
No problem. The host is https://ln.getalby.com and "Use Tor" is not enabled. Let me know if you need anything else.

thebullishbitcoiner:
I was trying to test the fix in 0.7.7-beta1 but the keyboard won't go away (so I'm unable to click save) Not sure if that's an iOS thing or not.
![image](https://github.com/ZeusLN/zeus/assets/139607837/9964d580-6709-4477-a07c-26630aea56ac)

PR #1551, opened 2023-07-18
Zeus-1542: Fixed the notes being same for invoice and lightning payme…

shubhamkmr04:
Fixes [1542](https://github.com/ZeusLN/zeus/issues/1542)


kaloudis:
Looks OK to me. One case to test that might not be covered is hitting the `Add Note` button from a successful payment screen made with Alby.

kaloudis:
uTACK but looks good too me. think all the cases are covered. let's put it in the new beta

kaloudis (approved):

PR #1663, opened 2023-09-12
deps: React Native 0.72.4

kaloudis:


shubhamkmr04 (commented):
Tested for both android and iOS.

PR #1668, opened 2023-09-19
Contacts: Fixes

shubhamkmr04:
- Fixing the size of Header icons, mainly the **Back** button on the `AddContact,` `Contacts` and `ContactDetails` view.
- Disabling the `SAVE CONTACT` button if an extra field of pubkey is generated and not being filled correctly.
- On `AddContact`Previously we had to tap on the plus icon to open the gallery for adding photos, we changed that to make the whole background circle pressable.
- If the contact has only one address available (either lnAddress, onchainAddress, or pubkey) , we get forwarded to the payments and skip going to `ContactDetails` view only if the user is coming from the `Send` view (we still will be forwarded to the `ContactDetails` view if we coming from `Settings` so the user can edit the contact).

### DIFF
--- a/views/ContactDetails.tsx
+++ b/views/ContactDetails.tsx
@@ -163,6 +163,7 @@ export default class ContactDetails extends React.Component<
                 }}
                 color={themeColor('text')}
                 underlayColor="transparent"
+                size={35}
             />
         );
         const StarButton = () => (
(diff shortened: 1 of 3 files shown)


kaloudis (approved):
tACK

Issue #1811, opened 2023-10-31
Do not automatically go to 'Send' screen when LN-address is entered


myxmaster:
### Describe your enhancement idea

Zeus should not automatically navigate to the 'Scend' screen when a user enters an LN-address. For pasting it is fine, but for typing manually it can be a problem, here is why:
Let's say user wants to send to fren@zeuspay.com, but now somebody made an LN-address 'fren@zeuspay.co'. The user will not be able to type the last char 'm' of the address, because Zeus recognizes a valid LN-address 'fren@zeuspay.co' and navigates to 'Send'.

shubhamkmr04:
In my opinion, we should introduce a button named `PROCEED` to go with the address to solve this issue. What do you think?

shubhamkmr04:
@myxmaster would you please check [PR](https://github.com/ZeusLN/zeus/pull/1819) for this issue?

PR #1819, opened 2023-11-02
Send: Confirm before proceeding to the address

shubhamkmr04:
# Description

This relates to issue: #1811 

_We added a `Proceed` button on the Send screen, so we will have to tap it to confirm before going with the entered address_ 

Simulator Screenshot - iPhone 14 Plus - 2023-11-02 at 23 12 18

### DIFF
--- a/views/Send.tsx
+++ b/views/Send.tsx
@@ -184,12 +184,6 @@ export default class Send extends React.Component<SendProps, SendState> {
         }
     }
 
-    async componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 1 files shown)


kaloudis:
What happens if you hit proceed and it can't load the address data?

shubhamkmr04:
> What happens if you hit proceed and it can't load the address data?

We get this error:
Simulator Screenshot - iPhone 14 Plus - 2023-11-07 at 11 51 04

kaloudis (approved):
uTACK

PR #1825, opened 2023-11-06
AddNotes: Transparent background in all themes

shubhamkmr04:
We were getting the dark background in light themes as well, so I made the background transparent for all themes.

Before:
Simulator Screenshot - iPhone 14 Plus - 2023-11-06 at 14 31 08

After:
Simulator Screenshot - iPhone 14 Plus - 2023-11-06 at 14 21 41


kaloudis (approved):
tACK.

Was OK as is but this change is a bit nicer.

PR #1869, opened 2023-11-25
Node Config: update comments for params used by c-lightning-REST

myxmaster:
# Description

When creating a new Core Lightning node, the placeholder for the host is just `localhost` although currently, the protocol must be entered too (like `https://localhost`). Now it defaults to https by adding it automatically to the host name.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance 
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [x] Core Lightning (c-lightning-REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/views/Settings/NodeConfiguration.tsx
+++ b/views/Settings/NodeConfiguration.tsx
@@ -52,9 +52,9 @@ interface NodeConfigurationProps {
 
 interface NodeConfigurationState {
     nickname: string; //
-    host: string; // lnd
-    port: string | number; // lnd
-    macaroonHex: string; // lnd
+    host: string; // lnd, c-lightning-REST
+    port: string | number; // lnd, c-lightning-REST
+    macaroonHex: string; // lnd, c-lightning-REST
     url: string; // spark, eclair
     accessKey: string; // spark
     lndhubUrl: string; // lndhub


views/Settings/NodeConfiguration.tsx
```diff
@@ -361,6 +360,16 @@ export default class NodeConfiguration extends React.Component<
...
+
+        let host = this.state.host;
+        if (
+            implementation === 'c-lightning-REST' &&
+            !host.toLowerCase().startsWith('http://') &&
+            !host.toLowerCase().startsWith('https://')
+        ) {
+            host = `https://${host}`;
+        }
+
```
kaloudis, line 372: I feel like we should handle this is the `c-lightning-REST` logic instead

kaloudis:
That's a fair point but we already handle it the other way with LND. Inconsistent behavior would make it confusing.

Furthermore if you converted LND to work this way, you would break many connections unless you include a database migration of sorts.

Lastly, I would say the only ones it would really be acceptable to be using http would be devs, and I'm sure they can figure out how to do so with prefixing http:// to the front of the host

kaloudis:
> As far as I know: Only CLN supports http at all, LND forces https. That's why it is different.

That's not the case. You can disable it for LND

https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L230

PR #1921, opened 2023-12-26
WIP: LSP support for LND REST

kaloudis:


stores/InvoicesStore.ts
```diff
@@ -121,7 +121,8 @@ export default class InvoicesStore {
         ampInvoice?: boolean,
         routeHints?: boolean,
         addressType?: string,
-        customPreimage?: string
+        customPreimage?: string,
+        noLsp?: boolean
```
kaloudis, line 125: `noLsp` is an override for the LSP. Can we come up w a better var name or this ok?

shubhamkmr04 (commented):
Screenshot_1709026349
I assume tapping on **ENABLE CERTIFICATE VERIFICATION** should take us to `Node Configuration` view instead of `Nodes` view

shubhamkmr04 (commented):
tACK

PR #1924, opened 2023-12-28
Add functionality to share Contact via QR code

shubhamkmr04:


components/CollapsedQR.tsx
```diff
@@ -56,7 +56,7 @@ function ValueText({ value, truncateLongValue }: ValueTextProps) {
 }
 
 interface CollapsedQRProps {
-    value: string;
+    value: any;
```
kaloudis, line 59: not crazy about the loosening of types here

utils/handleAnything.ts
```diff
@@ -364,6 +364,14 @@ const handleAnything = async (
                     localeString('utils.handleAnything.invalidLnurlParams')
                 );
             });
+    } else if (JSON.parse(data)?.id) {
```
kaloudis, line 367: is there a way we can better identify contacts?

shubhamkmr04: Do you mean by using any other property? Can't think of any other than 'id' for it as it is unique

kaloudis: can we turn the unique field to be `contactId`?

shubhamkmr04: Yeah, will do that.

kaloudis:
Looking pretty good right now. Are there any backwards compatibility issues we created moving from `id` to `contactId`?

shubhamkmr04:
> Looking pretty good right now. Are there any backwards compatibility issues we created moving from `id` to `contactId`?

Not in my opinion. Tested all features of contacts after this change

kaloudis:
I found some breaking changes moving from `id` to `contactId`. You need to either add checks that moves the `id` to `contactId` on the contacts or find a different approach here. I think using a `zeuscontact:` or `lncontact:` prefix instead will be a better approach.

To see the bug, create contacts with `beta2` then switch to this branch. Then try to load a contact. It will crash trying to load the banner because the contact hasn't been loaded at all. If you want to keep those old approach you need changes on `Contacts.tsx` and `ContactDetails.tsx`.

Furthermore, the other QRs are still missing prefixes. there should be `nostr:`, `lightning:`, and `bitcoin:` prefixes on the QRs where applicable

utils/handleAnything.ts
```diff
@@ -16,7 +16,7 @@ const isClipboardValue = (data: string) =>
     handleAnything(data, undefined, true);
 
 const handleAnything = async (
-    data: string,
+    data?: any,
```
kaloudis, line 19: we don't have to change the type here

kaloudis:
Screenshot 2024-01-12 at 00 04 00

let's add more spacing between favorite and qr code here

views/ContactDetails.tsx
```diff
@@ -556,13 +630,29 @@ export default class ContactDetails extends React.Component<
...
+                                    onPress={() => {
+                                        navigation.goBack();
+                                        navigation.navigate('AddContact', {
+                                            prefillContact: nostrContact,
+                                            isEdit: true,
+                                            isNostrContact:
+                                                this.state.isNostrContact
+                                        });
+                                    }}
+                                    title="Edit and save contact"
```
kaloudis, line 651: this needs to be localized

views/ContactDetails.tsx
```diff
@@ -234,6 +244,65 @@ export default class ContactDetails extends React.Component<
...
+                    // Add the 'zeuscontact:' prefix to the contactData parameter
+                    const zeusContactData =
+                        'zeuscontact:' +
+                        JSON.stringify(contactDataWithoutPhoto);
+
+                    navigation.navigate('ContactInfo', {
+                        contactData: zeusContactData,
+                        addressData: [
+                            ...addPrefixToAddresses(
+                                this.state.contact?.lnAddress,
```
kaloudis, line 279: destructure the `contact` variable out of the state as it is used multiple times

views/ContactDetails.tsx
```diff
@@ -556,13 +630,29 @@ export default class ContactDetails extends React.Component<
                                 )}
                         </ScrollView>
                         {this.state.isNostrContact && (
```
kaloudis, line 632: destructure the `isNostrContact` variable out of the state as it is used multiple times

views/ContactInfo.tsx
```diff
@@ -0,0 +1,167 @@
+import React, { useEffect, useState } from 'react';
```
kaloudis, line 1: I don't like the name of this file `ContactInfo` can be confused with `ContactDetails` - something like `ContactQRs` would be more specific, more appropriate

However, I'm unsure if we're better off transforming the `QR` view and modifying it to handle multiple QRs than creating a new one here

views/ContactDetails.tsx
```diff
@@ -234,6 +244,65 @@ export default class ContactDetails extends React.Component<
...
+                        contactDataWithoutPhoto.photo &&
+                        !contactDataWithoutPhoto.photo.startsWith('http')
+                    ) {
+                        delete contactDataWithoutPhoto.photo;
+                    }
+
+                    // Add the 'zeuscontact:' prefix to the contactData parameter
+                    const zeusContactData =
+                        'zeuscontact:' +
+                        JSON.stringify(contactDataWithoutPhoto);
```
kaloudis, line 273: you can do this as just `zeuscontact:${contactDataWithoutPhoto}`

views/ContactInfo.tsx
```diff
@@ -14,15 +14,15 @@ import Header from '../components/Header';
 import { themeColor } from '../utils/ThemeUtils';
```
kaloudis, line 14: you can delete this file now

views/ContactQR.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { themeColor } from '../utils/ThemeUtils';
+import CollapsedQR from '../components/CollapsedQR';
+
+interface ContactQRProps {
+    navigation: any;
+}
+
+interface IntroProps {
+    navigation: any;
+}
```
kaloudis, line 23: you don't need this second interface

kaloudis (approved):

PR #1939, opened 2024-01-09
Fix description field layout in AddContact view

shubhamkmr04:
Before:
Simulator Screenshot - iPhone 15 Plus - 2024-01-09 at 17 28 53

After:
Simulator Screenshot - iPhone 15 Plus - 2024-01-09 at 17 29 18


kaloudis (approved):
nice improvement

PR #1953, opened 2024-01-17
Contacts: Saving banner and photos locally

shubhamkmr04:


package.json
```diff
@@ -99,6 +99,7 @@
     "react-native-draglist": "3.5.1",
     "react-native-elements": "3.4.3",
     "react-native-encrypted-storage": "4.0.3",
+    "react-native-fs": "^2.20.0",
```
kaloudis, line 102: let's lock in the version here

utils/ContactUtils.test.ts
```diff
@@ -0,0 +1,100 @@
...
+                nostrNpub: [
+                    'npub1a3vuemsaex2rmshv3h5sr8xlxayehzkan3ucw3sxf8l2rtjyuhpshexnw4'
+                ],
+                contactId: 'mocked-uuid',
+                isFavourite: false,
+                banner: '',
+                isSelected: false
+            });
+        });
+        it('transforms contact data properly', async () => {
```
kaloudis, line 49: change the title here or put the tests in the same block

utils/ContactUtils.test.ts
```diff
@@ -0,0 +1,100 @@
+import { v4 as uuidv4 } from 'uuid';
+import ContactUtils from './ContactUtils';
+
+// Mocking the dependencies
+jest.mock('react-native-fs', () => ({
+    DocumentDirectoryPath: '/mocked/document/directory',
+    downloadFile: jest.fn(() => ({ promise: Promise.resolve() }))
+}));
```
kaloudis, line 8: is this required if we're not processing any images?

shubhamkmr04: Yes, it is not required for now, but unknowingly, we get issues in the test when this mock is removed, so I decided to keep it as it will be helpful in our future implementations.

kaloudis: Sounds good

kaloudis (approved):

PR #1961, opened 2024-01-24
RNFS: account for changing DocumentDirectoryPath

kaloudis:
This addresses an issue with Contacts images on iOS. With iOS design, the `DocumentDirectoryPath` changes with each update. This would cause our saved images to break upon update.

Related issue: https://github.com/itinance/react-native-fs/issues/528

This PR ensures that images persist across app updates.

### DIFF
--- a/utils/ContactUtils.ts
+++ b/utils/ContactUtils.ts
@@ -51,8 +51,6 @@ const transformContactData = async (contact: any) => {
             }
         }
 
-        console.log('Transformed contact:', transformedContact);
-
         if (contact?.picture) {
             console.log('Downloading image...');
             const fileName =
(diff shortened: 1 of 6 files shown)


utils/ContactUtils.ts
```diff
@@ -67,7 +67,7 @@ const transformContactData = async (contact: any) => {
                 }).promise;
 
                 console.log('Download successful!');
-                transformedContact.photo = 'file://' + filePath;
+                transformedContact.photo = 'rnfs://' + fileName;
```
shubhamkmr04, line 70: I wonder if we should also do such changes in `selectPhoto` of `AddContact.tsx` ?

kaloudis: Good call https://github.com/ZeusLN/zeus/pull/1961/commits/2e5bebf4b51cca2f70cc3bda7fc2173d8e06428d

PR #1964, opened 2024-01-28
POS: Receipt Printer

Talej:
# Description

Receipt printing functionality for android. Ended up using react-native-print package with a HTML template for receipt layout. The layout should pretty closely match the details shown on the 'Order' view for a paid order.

A print receipt button is shown on the payment completed view and also on the order view for paid orders if:

1. It's an android device
2. invoice has been paid
3. receipt printing isn't disabled in settings

react-native-print doesn't have any way of detecting if there is a printer ahead of calling the print dialog so to handle this I added the disable printer option under the POS settings.

I thought it could possibly also be useful to allow the receipt template HTML to be configuration in the POS settings - although given it is HTML this could be somewhat painful the edit on a device so wanted your input on this...

- [X] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [X] N/A

- [X] Android
- [ ] iOS

- [X] Embedded LND
- [ ] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

- [X] Contributors will need to run `yarn` after this PR is merged in
- [X] 3rd party dependencies have been modified:

### Other:


Talej:
Screenshot_20240127-135755
Screenshot_20240127-135917
Screenshot_20240127-135946

kaloudis:
I wonder if it's worth distinguishing `Print Invoice` / `Print Receipt` depending on where we are in the flow

Talej:
> I wonder if it's worth distinguishing `Print Invoice` / `Print Receipt` depending on where we are in the flow

Do you mean have a `Print Invoice` on the order screen before it is paid and `Print Receipt` after paid?

kaloudis:
> > I wonder if it's worth distinguishing `Print Invoice` / `Print Receipt` depending on where we are in the flow
> 
> Do you mean have a `Print Invoice` on the order screen before it is paid and `Print Receipt` after paid?

Precisely, and perhaps we add the payment pre-image on the receipt to distinguish them and act as a bit of payment proof

package.json
```diff
@@ -111,6 +111,7 @@
     "react-native-notifications": "5.1.0",
     "react-native-os": "aprock/react-native-os#5/head",
     "react-native-permissions": "3.8.0",
+    "react-native-print": "^0.11.0",
```
kaloudis, line 114: let's lock in the version number here

views/Order.tsx
```diff
@@ -312,6 +352,152 @@ export default class OrderView extends React.Component<OrderProps, OrderState> {
...
+            return UnitsStore.getFormattedAmount(amount, units);
+        };
+
+        const printReceipt = () => {
+            const title =
+                merchantName && merchantName.length
+                    ? merchantName
+                    : isPaid
+                    ? 'Tax Receipt'
+                    : 'Invoice';
```
kaloudis, line 393: We need these titles localized

Talej: Done!

package.json
Talej: Done!

kaloudis:
Nice job only showing the printer setting for Android, but I think it would be better to make it opt in rather than opt out as most merchants likely won't have a printer module.

Talej:
> Nice job only showing the printer setting for Android, but I think it would be better to make it opt in rather than opt out as most merchants likely won't have a printer module.

Makes total sense. Flipped the setting from disable to enable.

locales/en.json
```diff
@@ -934,9 +936,13 @@
     "pos.views.Order.totalBitcoin": "Total (Bitcoin)",
     "pos.views.Order.total": "Total",
     "pos.views.Order.paymentType": "Payment type",
+    "pos.views.Order.printReceipt": "Print Receipt",
+    "pos.views.Order.printInvoice": "Print Invoice",
     "pos.views.Settings.PointOfSale.authWarning": "Warning: no password or PIN set",
     "pos.views.Settings.PointOfSale.backendWarning": "Warning: currently only LND nodes are able to mark orders as paid",
     "pos.views.Settings.PointOfSale.currencyError": "Error: currency must be set first",
+    "pos.print.taxReceipt": "Tax Receipt",
```
kaloudis, line 944: nit: can we just make this `Receipt` instead of `Tax Receipt`

kaloudis:
@Talej can you provide an example of what these receipts look like when the merchant provides their store name in settings?

Talej:
> @Talej can you provide an example of what these receipts look like when the merchant provides their store name in settings?

proof-of-storename

IMG_1147

shubhamkmr04 (commented):
Screenshot_1708081156
Screenshot_1708081167

It would be better if we give a paddingBottom here.

shubhamkmr04 (commented):
Screenshot_1708081031
Also can we also handle the case so the user don't get this error if the device is not connected to a printer.

Talej:
> Screenshot_1708081031 Also can we also handle the case so the user don't get this error if the device is not connected to a printer.

This actually looks like the package isn't loading for you rather than anything to do with not connecting to a printer.

I have tried with various devices and even if there is no printer it still works without error (it gives an option to export as PDF). I have added in some extra checks to prevent the error but would be useful if you could debug a bit further to see why it wasn't loading.

shubhamkmr04 (approved):
tACK

kaloudis (approved):
LGTM. Great job, @Talej!

Issue #1974, opened 2024-02-04
Deactivate (or hide) SEND COINS button, when sending is not possible


myxmaster:
### Describe your enhancement idea

Example: Users on-chain balance is zero, but an on-chain address QR is scanned.
-> Zeus displays a message, explaining that no on-chain balance is available, but user can still enter an amount, choose feerate and click on the "SEND COINS" button.

Amount & feerate input, SEND COINS button, and maybe even the "sweep" icon on top right, should be deactivated/hidden.

There might be more examples where sending is not possible.

PR #2008, opened 2024-02-27
Reworked Send screen logic

myxmaster:
# Description

Fixed several problems and improved back press handling of Send screen.

First problem (#1974): When balance is 0, the send button is enabled although sending cannot succeed.
Second problem: When a destination is entered and proceed is clicked, you can change the destination afterwards to another type (BTC <-> LN), but the dialog does not adjust to the new destination type.

Fixed the first problem by disabling the send button if balance is zero.

Fixed the second problem by making the destination read-only after proceed is clicked. To change the destination, the user now can go back which now clears the input and allows entering a new destination. However, if the Send screen was entered by scanning a QR code or from the Contacts screen, the back button still closes the Send screen.

Before:
grafik

Same address, different (wrong) destination type:
grafik

After:
grafik

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:


kaloudis:
I like the idea but we have to think about what payment flows are going to look like as we target external accounts + signers in v0.9.0

kaloudis (changes requested):
This looks great, but let's remove the colon (`:`) from the `Fee` label instead of adding it to the other fields.

Let's stay consistent with the rest of the app.

kaloudis:
Let's also grey out, or completed remove the paper airplane icon on the send button when it is disabled

kaloudis (approved):

Issue #1880, opened 2023-11-28
Adding photo to contact not possible with setting "Require login after app returns from background"


myxmaster:
### Describe the bug

If you try to add a photo to a contact while "Require login after app returns from background" is enabled, Zeus will ask for PIN/Password/Biometrics when you come back from picking the photo from your phone. After you logged in, Zeus displays the default view (Balance or Keypad pane) and not the screen where you tried to add a photo, and of course the photo is not saved.

### Reproduce

- Set a Pin or Password
- Enable "Require login after app returns from background"
- Go to contacts and try to add a photo to a new or existing contact
-> After relogin Zeus is displaying default view (Balance or Keypad pane) instead of contacts with your photo

### ZEUS version

v0.8.0

### Node interface

LND (REST)

### Network

Clearnet

### Device

Samsung Galaxy S20+

### Device operating system

Android 13

PR #2010, opened 2024-02-29
Custom pictures for node config

shubhamkmr04:


views/Settings/NodeConfiguration.tsx
```diff
@@ -335,6 +341,12 @@ export default class NodeConfiguration extends React.Component<
                 newEntry
             });
         }
+        const updatedNode = {
+            ...node,
+            photo
+        };
+        console.log(node);
```
kaloudis, line 348: remove log

views/Settings/NodeConfiguration.tsx
```diff
@@ -605,7 +621,8 @@ export default class NodeConfiguration extends React.Component<
             recoveryCipherSeed,
             channelBackupsBase64,
             creatingWallet,
-            errorCreatingWallet
+            errorCreatingWallet,
+            photo
```
kaloudis, line 625: photo not used here

views/Settings/SelectNodeProfile.tsx
```diff
@@ -177,7 +197,14 @@ export default class SelectNodeProfile extends React.Component<
                     }}
                 />
 
-                <Button title="Upload picture" />
+                <Button
+                    title="Upload picture"
```
kaloudis, line 201: localize this

views/Settings/NodeConfiguration.tsx
```diff
@@ -335,6 +341,12 @@ export default class NodeConfiguration extends React.Component<
                 newEntry
             });
         }
+        const updatedNode = {
+            ...node,
+            photo
+        };
+        console.log(node);
+        updateSettings({ node: updatedNode });
```
kaloudis, line 349: unsure why you needs these lines here

kaloudis:
We have some conflicts

kaloudis:
I think we can go smaller with the image sizes. Should probably still look good at a quarter of the size I think

shubhamkmr04:
> I think we can go smaller with the image sizes. Should probably still look good at a quarter of the size I think

Reduced it by 30%, we'll have to compromise the quality if we shrink the size more than this

views/Settings/SetNodePicture.tsx
```diff
@@ -0,0 +1,254 @@
...
+                require('../../assets/images/zeus-illustration-5a.jpg'),
+                require('../../assets/images/zeus-illustration-5b.jpg'),
+                require('../../assets/images/zeus-illustration-6a.jpg'),
+                require('../../assets/images/zeus-illustration-6b.jpg'),
+                require('../../assets/images/zeus-illustration-7a.jpg'),
+                require('../../assets/images/zeus-illustration-7b.jpg'),
+                require('../../assets/images/Alby.jpg'),
+                require('../../assets/images/BTCpay.jpg'),
+                require('../../assets/images/CLN.jpg'),
+                require('../../assets/images/LND.jpg')
```
kaloudis, line 59: I want these last four to be programmatically set, depending on which backend is being used

kaloudis (changes requested):

views/Settings/SetNodePicture.tsx
shubhamkmr04: Done.

kaloudis (changes requested):
When you change the node picture, it doesn't change the state on the `NodeConfiguration` page as not saved, like it would be if you changed one of the fields.

Other than that I think it's looking good

shubhamkmr04:
> When you change the node picture, it doesn't change the state on the `NodeConfiguration` page as not saved, like it would be if you changed one of the fields.
> 
> Other than that I think it's looking good

I think we are changing state after coming to `NodeConfiguration` view in `InitProps` function line **327** and **343**

views/Settings/SetNodePicture.tsx
```diff
@@ -0,0 +1,276 @@
...
+                        }}
+                    />
+                </ScrollView>
+
+                <Button
+                    title={localeString('views.SetNodePicture.setPicture')}
+                    onPress={() => {
+                        navigation.navigate('NodeConfiguration', {
+                            photo
+                        });
```
kaloudis, line 248: here we should also set `saved: false` to update the state

views/Settings/NodeConfiguration.tsx
```diff
@@ -925,29 +956,79 @@ export default class NodeConfiguration extends React.Component<
...
+                                        node?.photo === '' ||
+                                        node?.photo === null) &&
+                                    photo === null
+                                        ? () =>
+                                              navigation.navigate(
+                                                  'SetNodePicture',
+                                                  { implementation }
+                                              )
+                                        : () =>
+                                              this.setState(
```
kaloudis, line 972: here we should set `saved: false` since the photo is changing

views/Settings/NodeConfiguration.tsx
```diff
@@ -925,29 +956,79 @@ export default class NodeConfiguration extends React.Component<
                                     message={createAccountSuccess}
                                 />
                             )}
+                        <View style={styles.container}>
+                            <TouchableOpacity
+                                onPress={
+                                    (node === null ||
+                                        node?.photo === '' ||
+                                        node?.photo === null) &&
```
kaloudis, line 964: could you simply do `!node.photo` in place of `node?.photo === '' || node?.photo === null`?

shubhamkmr04:
> Nice PR! A few ideas/general feedback:
> 
> 1. This should be fixed (happens when list of default images are displayed):
>    `VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead.`
> 2. When you pick an image and click button "SET PICTURE", it is not really set, because you still need to click on next button "SAVE NODE CONFIG". I think it should be all done after clicking the first button (set picture). (Or if you don't like that idea, maybe rename "SET PICTURE" to "CHOOSE PICTURE", because you are not really setting it here.)
> 3. When an image for a node was already set, then you go to settings again and click on that image, it is removed and the "+" icon is displayed (and the explainer "For now, the Embedded LND nodes cannot be deleted..." is gone!). I would expect to be able to just pick another image, when I click on it.
> 4. When an image is already set, it is not exactly intuitive that you can actually click on the node image in node settings to change it. This is especially problematic in case Zeus picks images automatically when configuring a new node in the future (e.g. LND image for LND nodes). I think many users will not be aware they can change it. Not sure how to solve this nicely, but I guess we need some kind of "Edit node image" label.
> 5. When there is PIN/passphrase configured and you pick an image from gallery, Zeus will ask for PIN/passphrase when you are back in Zeus and after logging in you are back in default view -> you cannot set the node image.
> 6. I think it is a lot of Zeus images, maybe there should be categories which users can expand/collapse.

Thanks for the detailed feedback! After a discussion with @kaloudis, I have updated the name to 'Choose Picture'. We will address the remaining points outlined in the feedback through subsequent PRs.

kaloudis:
>     5. When there is PIN/passphrase configured and you pick an image from gallery, Zeus will ask for PIN/passphrase when you are back in Zeus and after logging in you are back in default view -> you cannot set the node image.

Related issue here https://github.com/ZeusLN/zeus/issues/1880

kaloudis:
>     4. When an image is already set, it is not exactly intuitive that you can actually click on the node image in node settings to change it. This is especially problematic in case Zeus picks images automatically when configuring a new node in the future (e.g. LND image for LND nodes). I think many users will not be aware they can change it. Not sure how to solve this nicely, but I guess we need some kind of "Edit node image" label.

I was thinking a red circle and `-` in the top right corner of the image to remove it, but it would likely be too small. Might need to pull in a designer for this one

kaloudis (approved):

PR #2031, opened 2024-03-06
POS Enhancements

kaloudis:
This PR adds the following changes to the POS:

- Currency not set error message does not block rendering of the rest of the fields and only displays if `Square` is selected. This allows users to use the `Standalone` mode without setting a fiat currency
- Fixes a bug where the keypad and the printer could not be enabled at the same time
- Adds placeholders for when no categories or products are defined when in `Standalone` mode
- Adds loading indicator to Products and Categories view in `Standalone` mode
- Hides the fiat rate on the `Standalone POS` view if fiat rates aren't enabled
- Show charge and clear buttons even if no products are defined - this will allow people to clear orders if they're using the keypad w/o defined products


shubhamkmr04 (commented):
tACK

PR #2065, opened 2024-03-16
Refactor fetchNostrContacts

shubhamkmr04:
- The code now uses Promise.all() to wait for all asynchronous operations to complete before proceeding ensuring proper execution flow.
- Previously setState() was called inside the loop, which could result in multiple state updates and unnecessary re-renders. Now setState() is called only once after all asynchronous operations are completed, reducing unnecessary re-renders and improving performance.


kaloudis (approved):
tACK

PR #2089, opened 2024-03-27
SetNodePicture: Remove scrollview component

shubhamkmr04:
We were getting the warning `VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead.` 
Fix: Removed Scrollview and added the styling in FlatList itself


kaloudis (approved):
tACK

PR #2097, opened 2024-03-28
LSPS1 client

shubhamkmr04:
https://github.com/BitcoinAndLightningLayerSpecs/lsp/tree/main/LSPS1


stores/LSPStore.ts
```diff
@@ -270,4 +305,225 @@ export default class LSPStore {
...
+        }
+    };
+
+    @action
+    public subscribeCustomMessages = async () => {
+        if (this.customMessagesSubscriber) return;
+        let timer = 10000;
+        const timeoutId = setTimeout(() => {
+            this.error = true;
+            this.error_msg = 'Did not receive response from server';
```
kaloudis, line 373: this needs to be localized

views/Settings/LSPS1/Order.tsx
```diff
@@ -0,0 +1,515 @@
...
+                                                    : '#808000'
+                                            }
+                                        />
+                                    )}
+                                    {result.channel_expiry_blocks && (
+                                        <KeyValue
+                                            keyValue={localeString(
+                                                'views.LSPS1.channelExpiryBlocks'
+                                            )}
+                                            value={result.channel_expiry_blocks}
```
kaloudis, line 254: let's format this with commas

views/Settings/LSPS1/Order.tsx
```diff
@@ -0,0 +1,515 @@
...
+                                <View style={{ paddingHorizontal: 20 }}>
+                                    {result.announce_channel && (
+                                        <KeyValue
+                                            keyValue={localeString(
+                                                'views.OpenChannel.announceChannel'
+                                            )}
+                                            value={
+                                                result.announce_channel
+                                                    ? 'True'
+                                                    : 'False'
```
kaloudis, line 240: let's localize these - we have them defined already in our locales

views/Settings/LSPS1/OrdersPane.tsx
```diff
@@ -0,0 +1,232 @@
...
+                {isLoading ? (
+                    <View style={{ marginTop: 50 }}>
+                        <LoadingIndicator />
+                    </View>
+                ) : (
+                    <>
+                        <Header
+                            leftComponent="Back"
+                            centerComponent={{
+                                text: `LSPS1 Orders (${orders.length})`,
```
kaloudis, line 213: let's localize this

views/Settings/LSPS1/Order.tsx
```diff
@@ -0,0 +1,229 @@
...
+                    navigation={navigation}
+                />
+                {LSPStore.loading ? (
+                    <LoadingIndicator />
+                ) : (
+                    <ScrollView>
+                        {fetchOldOrder && (
+                            <View style={{ paddingHorizontal: 20 }}>
+                                <WarningMessage
+                                    message={`${LSPStore.error_msg}: Showing Previous State`}
```
kaloudis, line 214: localization needed here. Please fix the casing to `showing previous state`

views/Settings/LSPS1/Order.tsx
```diff
@@ -0,0 +1,229 @@
...
+                    error
+                );
+            });
+    }
+
+    lsps1_getorder(orderId: string) {
+        const { LSPStore } = this.props;
+        LSPStore.loading = true;
+        const peer =
+            '03e84a109cd70e57864274932fc87c5e6434c59ebb8e6e7d28532219ba38f7f6df@139.144.22.237:9735';
```
kaloudis, line 156: this being hard-coded is problematic, we should be able to pull the peer info from the existing order data so that we know which peer to query.

If we leave it as is, this is going to break if you are not using the default LSP peer all the time.

kaloudis (changes requested):

shubhamkmr04:
[e3f389](https://github.com/ZeusLN/zeus/pull/2097/commits/e3f389e5a4482d2a084824f591f673ee2a776da1):
- Localize string
- Show error if no orders are stored yet
- Fix warning `Each child in a list should have a unique key prop` on **OrdersPane** view
- Format Channel expiry block input with commas

views/Settings/LSPS1/index.tsx
```diff
@@ -255,6 +255,10 @@ export default class LSPS1 extends React.Component<LSPS1Props, LSPS1State> {
                 style={{ marginTop: -10 }}
                 onPress={() => {
                     navigation.navigate('OrdersPane');
+                    // EncryptedStorage.setItem(
+                    //     'orderResponses',
+                    //     JSON.stringify([])
+                    // );
```
kaloudis, line 261: remove this if we're not using it

stores/LSPStore.ts
```diff
@@ -499,9 +499,9 @@ export default class LSPStore {
     };
 
     @action
-    public getOrderREST(id: string) {
+    public getOrderREST(id: string, peerOrEndpoint: any) {
         this.loading = true;
-        const endpoint = `${this.getLSPS1Rest()}/api/v1/get_order?order_id=${id}`;
+        const endpoint = `${peerOrEndpoint}/api/v1/get_order?order_id=${id}`;
```
kaloudis, line 504: why the var name `peerOrEndpoint`? it's always going to be a REST endpoint on this call every time

views/Settings/LSPS1/Order.tsx
```diff
@@ -47,55 +47,59 @@ export default class Orders extends React.Component<OrderProps, OrdersState> {
...
+                        const parsedOrder = JSON.parse(order);
+                        temporaryOrder = parsedOrder;
+                        console.log('Order found in storage->', temporaryOrder);
+
+                        const peerOrEndpoint =
+                            temporaryOrder.peer || temporaryOrder.endpoint;
+
+                        BackendUtils.supportsLSPS1rest()
+                            ? LSPStore.getOrderREST(id, peerOrEndpoint)
+                            : this.lsps1_getorder(id, peerOrEndpoint);
```
kaloudis, line 75: do we need `lsps1_getorder` in the view? can we move it to the store?

views/Settings/LSPS1/Order.tsx
```diff
@@ -47,55 +47,59 @@ export default class Orders extends React.Component<OrderProps, OrdersState> {
...
-                        ).then(() => {
-                            console.log('Order updated in encrypted storage!');
-                        });
+                    if (order) {
+                        const parsedOrder = JSON.parse(order);
+                        temporaryOrder = parsedOrder;
+                        console.log('Order found in storage->', temporaryOrder);
+
+                        const peerOrEndpoint =
+                            temporaryOrder.peer || temporaryOrder.endpoint;
```
kaloudis, line 71: don't think we need to make this variable if we're checking whether REST is supported or not on the line below

stores/LSPStore.ts
```diff
@@ -525,4 +528,32 @@ export default class LSPStore {
                 this.loading = false;
             });
     }
+
+    @action
+    public lsps1_getorder(orderId: string, peer: string) {
```
kaloudis, line 533: rename to `getOrderCustomMessage`

views/Settings/LSPS1/Order.tsx
shubhamkmr04: Yes that would be good. Did the change

views/Settings/LSPS1/OrdersPane.tsx
```diff
@@ -0,0 +1,286 @@
...
+                        );
+                    } else if (BackendUtils.supportsLSPS1rest()) {
+                        selectedOrders = decodedResponses.filter(
+                            (response) => response?.endpoint
+                        );
+                    } else {
+                        selectedOrders = decodedResponses;
+                    }
+
+                    const orders = selectedOrders.map((response) => ({
```
kaloudis, line 82: I think we can clean this block up a bit by checking if `result` exists on the object and defining the order depending on it.

Something like

```
const order = response?.order?.result || response?.order;

return {
        orderId: order.order_id,
        state: order.order_state,
        createdAt: order.created_at,
        fundedAt: order.funded_at,
        lspBalanceSat: order.lsp_balance_sat
}
```

kaloudis (approved):

PR #2100, opened 2024-03-30
SetNodePicture: Refactor 'handleImageTap'

shubhamkmr04:


views/Settings/SetNodePicture.tsx
```diff
@@ -125,35 +125,14 @@ export default class SetNodePicture extends React.Component<
         );
     };
 
-    getPhoto(photo: string | null): string {
+    getPhoto(photo: string | null): object | string {
```
kaloudis, line 128: Why was it necessary to change the structure of the `getPhoto` calls? Seems like we could have just removed `handleImageTap` and made the change on line 194

kaloudis:
This approach doesn't work. Images are lost on reload

views/Settings/SetNodePicture.tsx
shubhamkmr04: We did that coz we needed the object to return with the key `uri` to fetch gallery imported photos

shubhamkmr04:
> What if we refactor `handleImageTap` to just save the image uri?
> 
> ```
> handleImageTap = async (item) => {
>         try {
>             // Fetch the local image
>             const imageUri = Image.resolveAssetSource(item).uri;
>             this.setState({ photo: imageUri });
>         } catch (error) {
>             console.error('Error resolving image uri:', error);
>         }
>     };
> ```

yup we can do that

kaloudis:
Holds up against Android upgrades. About to test iOS

kaloudis (changes requested):
Does not hold up on iOS, I believe the file paths change every upgrade.

We'll have to emulate what we do on custom images and save a copy of the image that we point to.

kaloudis:
The file storage is working correctly now, however the approach isn't quite right.

We are unnecessarily creating new copies of each image every time we tap it. The 'download' event should only trigger once the `Choose Picture` button is pressed

kaloudis (approved):
passing all my tests. nice work

Issue #1637, opened 2023-09-02
Support for Core Lightning's native CLNRest plugin?


schmitmd:
### Describe the feature

It would seem that the Core Lightning folks have created their own REST API plugin.  Do you see this as a potential feature option to be added?  It looks like it would use a Rune (part of the native core LN service if I understand correctly) rather than Macaroons for auth and permissions.  Any thoughts?

https://github.com/ElementsProject/lightning/tree/master/plugins/clnrest

schmitmd:
#1068 possibly related

kaloudis:
We want to upgrade to this and deprecate c-lightning-REST

We've got a bounty for lnmessage over at https://github.com/ZeusLN/zeus/blob/master/docs/Bounties.md#core-lightning-commandolnmessage-connection. Perhaps we'll add a smaller one for the CLNRest plugin

svpi11:
Same here. The "native clnrest plugin" of CLN integration, would be appreciated a lot !

bitcoinlizard:
"Me to."  I would pay into the bounty.  Now that Ride the Lightning is using the native plugin I could disable c-lightning-REST if Zeus migrated as well.

kaloudis:
Please reach out to support AT zeusln.com via email to coordinate donations to the bounty.

No progress has been made. Commando is somewhat close though.

sha-265:
I did something few mouths ago, but didn't finish it. If there will be a bounty maybe I will free some time to finish it.

bitcoinlizard:
I emailed support AT zeusln.com to commit to a bounty for this feature.

schmitmd:
@kaloudis you mentioned back in January that Commando was "somewhat close", has anything progressed there?

kaloudis:
CLNRest is coming in v0.9.0. Progress hasn't been made on Commando. Related issue: https://github.com/ZeusLN/zeus/issues/1068

kaloudis:
Bounty is paid out and functionality is rolling out in v0.9.0 (currently in rc1 - please test if you can)

bitcoinlizard:
> Bounty is paid out and functionality is rolling out in v0.9.0 (currently in rc1 - please test if you can)

I just tested this and it works as expected.  Great job!

ElucGeek:
Works well thanks!

PR #2228, opened 2024-06-08
Core Lightning: CLNRest backend

niteshbalusu11:
# Description

Relates to issue: #1637

Depracates c-lightning-rest and adds a new backend using core-lightning team maintained CLN Rest.
Important Note: Minimum core-lightning version to use this is `v24.02` due to breaking changes from cln team.
This PR should cover all features which the old backend was covering.
I have not tested this on iOS coz I wasn't able to get xcode to build zeus.

- [X] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [X] Android
- [ ] iOS

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -7,6 +7,7 @@ import { Hash as sha256Hash } from 'fast-sha256';
 import BigNumber from 'bignumber.js';
 import {
     getBalance,
+    getChainTransactions,
     getOffchainBalance,
     listPeers
 } from './CoreLightningRequestHandler';
@@ -177,10 +178,7 @@ export default class CLNRest {
         this.postRequest('/v1/listnodes', { id: data.id }).then((res) => {
             return res;
         });
-    getTransactions = () =>
-        this.postRequest('/v1/listfunds').then((res) => ({
-            transactions: res.outputs
-        }));
+    getTransactions = async () => await getChainTransactions();
     getChannels = async () => {
         const channels = await this.postRequest('/v1/listpeerchannels');
         return await listPeers(channels);
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -198,15 +198,15 @@ export default class CLNRest {
         if (data.utxos) {
             request = {
                 destination: data.addr,
-                feeRate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
-                satoshis: data.amount,
+                feerate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
+                satoshi: data.amount,
                 utxos: data.utxos
             };
         } else {
             request = {
                 destination: data.addr,
-                feeRate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
-                satoshis: data.amount
+                feerate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
+                satoshi: data.amount
             };
         }
         return this.postRequest('/v1/withdraw', request);
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -197,14 +197,14 @@ export default class CLNRest {
         let request: any;
         if (data.utxos) {
             request = {
-                address: data.addr,
+                destination: data.addr,
                 feeRate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
                 satoshis: data.amount,
                 utxos: data.utxos
             };
         } else {
             request = {
-                address: data.addr,
+                destination: data.addr,
                 feeRate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
                 satoshis: data.amount
             };
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -61,11 +61,8 @@ export default class CLNRest extends LND {
         );
     };
 
-    getRequest = (route: string, data?: any) =>
-        this.request(route, 'get', null, data);
     postRequest = (route: string, data?: any) =>
         this.request(route, 'post', data);
-    deleteRequest = (route: string) => this.request(route, 'delete', null);
 
     getNode = (data: any) =>
         this.postRequest('/v1/listnodes', { id: data.id }).then((res) => {
@@ -182,7 +179,7 @@ export default class CLNRest extends LND {
         return this.postRequest('/v1/close', request);
     };
     getFees = () =>
-        this.getRequest('/v1/getinfo').then((res: any) => ({
+        this.postRequest('/v1/getinfo').then((res: any) => ({
             total_fee_sum: res.fees_collected_msat / 1000
         }));
     setFees = (data: any) =>
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -1,5 +1,4 @@
 import stores from '../stores/Stores';
-import LND from './LND';
 import TransactionRequest from '../models/TransactionRequest';
 import OpenChannelRequest from '../models/OpenChannelRequest';
 import VersionUtils from '../utils/VersionUtils';
@@ -11,8 +10,13 @@ import {
     getOffchainBalance,
     listPeers
 } from './CoreLightningRequestHandler';
+import { localeString } from '../utils/LocaleUtils';
+import ReactNativeBlobUtil from 'react-native-blob-util';
+import { doTorRequest, RequestMethod } from '../utils/TorUtils';
 
-export default class CLNRest extends LND {
+const calls = new Map<string, Promise<any>>();
+
+export default class CLNRest {
     getHeaders = (rune: string): any => {
         return {
             Rune: rune
@@ -34,6 +38,91 @@ export default class CLNRest extends LND {
             );
         }
         return isSupportedVersion(version, minVersion, eosVersion);
+    };
+
+    clearCachedCalls = () => calls.clear();
+
+    restReq = async (
+        headers: Headers | any,
+        url: string,
+        method: any,
+        data?: any,
+        certVerification?: boolean,
+        useTor?: boolean
+    ) => {
+        // use body data as an identifier too, we don't want to cancel when we
+        // are making multiples calls to get all the node names, for example
+        const id = data ? `${url}${JSON.stringify(data)}` : url;
+        if (calls.has(id)) {
+            return calls.get(id);
+        }
+        // API is a bit of a mess but
+        // If tor enabled in setting, start up the daemon here
+        if (useTor === true) {
+            calls.set(
+                id,
+                doTorRequest(
+                    url,
+                    method as RequestMethod,
+                    JSON.stringify(data),
+                    headers
+                ).then((response: any) => {
+                    calls.delete(id);
+                    return response;
+                })
+            );
+        } else {
+            calls.set(
+                id,
+                ReactNativeBlobUtil.config({
+                    trusty: !certVerification
+                })
+                    .fetch(
+                        method,
+                        url,
+                        headers,
+                        data ? JSON.stringify(data) : data
+                    )
+                    .then((response: any) => {
+                        calls.delete(id);
+                        if (response.info().status < 300) {
+                            // handle ws responses
+                            if (response.data.includes('\n')) {
+                                const split = response.data.split('\n');
+                                const length = split.length;
+                                // last instance is empty
+                                return JSON.parse(split[length - 2]);
+                            }
+                            return response.json();
+                        } else {
+                            try {
+                                const errorInfo = response.json();
+                                throw new Error(
+                                    (errorInfo.error &&
+                                        errorInfo.error.message) ||
+                                        errorInfo.message ||
+                                        errorInfo.error
+                                );
+                            } catch (e) {
+                                if (
+                                    response.data &&
+                                    typeof response.data === 'string'
+                                ) {
+                                    throw new Error(response.data);
+                                } else {
+                                    throw new Error(
+                                        localeString(
+                                            'backends.LND.restReq.connectionError'
+                                        )
+                                    );
+                                }
+                            }
+                        }
+                    })
+            );
+        }
+
+        return await calls.get(id);
     };
 
     request = (route: string, method: string, data?: any, params?: any) => {
@@ -59,6 +148,26 @@ export default class CLNRest extends LND {
             certVerification,
             enableTor
         );
+    };
+
+    getURL = (
+        host: string,
+        port: string | number,
+        route: string,
+        ws?: boolean
+    ) => {
+        const hostPath = host.includes('://') ? host : `https://${host}`;
+        let baseUrl = `${hostPath}${port ? ':' + port : ''}`;
+
+        if (ws) {
+            baseUrl = baseUrl.replace('https', 'wss').replace('http', 'ws');
+        }
+
+        if (baseUrl[baseUrl.length - 1] === '/') {
+            baseUrl = baseUrl.slice(0, -1);
+        }
+
+        return `${baseUrl}${route}`;
     };
 
     postRequest = (route: string, data?: any) =>
@@ -188,7 +297,6 @@ export default class CLNRest extends LND {
             feebase: data.base_fee_msat,
             feeppm: data.fee_rate
         });
-    getRoutes = () => this.getRequest('N/A');
     getUTXOs = () => this.postRequest('/v1/listfunds');
     signMessage = (message: string) =>
         this.postRequest('/v1/signmessage', {
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -12,7 +12,7 @@ import {
     listPeers
 } from './CoreLightningRequestHandler';
 
-export default class CoreLightningRestApi extends LND {
+export default class CLNRest extends LND {
     getHeaders = (rune: string): any => {
         return {
             Rune: rune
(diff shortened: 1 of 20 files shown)


backends/CLightningREST.ts
kaloudis, line 1: thanks for the fixes here!

backends/CoreLightningRestApi.ts
kaloudis, line 1: nit: let's rename this file and class `CLNRest`

kaloudis: We can leave the name of the request handler file as is, as we might use it with other CLN interfaces

stores/SettingsStore.ts
```diff
@@ -205,12 +206,26 @@ export const INTERFACE_KEYS = [
     { key: 'Embedded LND', value: 'embedded-lnd' },
     { key: 'LND (REST)', value: 'lnd' },
     { key: 'LND (Lightning Node Connect)', value: 'lightning-node-connect' },
-    { key: 'Core Lightning (c-lightning-REST)', value: 'c-lightning-REST' },
+    { key: 'Core Lightning REST API', value: 'core-lightning-rest-api' },
```
kaloudis, line 209: nit: let's call the key `Core Lightning (CLNRest)` and the value `cln-rest`

niteshbalusu11:
Fixed and rebased

backends/CLNRest.ts
```diff
@@ -0,0 +1,282 @@
...
+import Base64Utils from '../utils/Base64Utils';
+import { Hash as sha256Hash } from 'fast-sha256';
+import BigNumber from 'bignumber.js';
+import {
+    getBalance,
+    getOffchainBalance,
+    listPeers
+} from './CoreLightningRequestHandler';
+
+export default class CLNRest extends LND {
```
vincenzopalazzo, line 15:
```suggestion
export default class CLNRest  {
```

There is no reason to extend LND, but we should re-export the method `getUrl` somewhere-else

vincenzopalazzo (changes requested):
Ok looks like that you arrive before me, I left some comments because in this way I can build on top of you

backends/CLNRest.ts
```diff
@@ -0,0 +1,282 @@
...
+    supportsOnchainReceiving = () => true;
+    supportsLightningSends = () => true;
+    supportsKeysend = () => true;
+    supportsChannelManagement = () => true;
+    supportsPendingChannels = () => false;
+    supportsMPP = () => false;
+    supportsAMP = () => false;
+    supportsCoinControl = () => this.supports('v0.8.2', undefined, 'v0.4.0');
+    supportsChannelCoinControl = () =>
+        this.supports('v0.8.2', undefined, 'v0.4.0');
```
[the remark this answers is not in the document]
niteshbalusu11: These are old functions that existed in the previous c-lightning-rest and i copied them over. Probably will leave them as is.

views/Settings/NodeConfiguration.tsx
```diff
@@ -1347,6 +1357,75 @@ export default class NodeConfiguration extends React.Component<
                                 )}
                             </>
                         )}
+                        {implementation === 'cln-rest' && (
```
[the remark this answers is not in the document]
niteshbalusu11: Because cln-rest uses runes and the old one uses a macaroon. We can just delete the old when Zeus wants to depracate it completely. Right now both are supported.

backends/CLNRest.ts
niteshbalusu11: Some views are breaking when we not extend from lnd, also all the actual execution of the fetch calls happen in the LND class so its not just getUrl. Will leave it as is.

backends/CLNRest.ts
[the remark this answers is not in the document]
kaloudis: I think you may have to delete the third parameter as that is an API version check

views/Settings/NodeConfiguration.tsx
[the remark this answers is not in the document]
vincenzopalazzo: >Because cln-rest uses runes and the old one uses a macaroon.  

are the same things a the concept level, you can just replace the string with `implementation === 'cln-rest' ? "rune" : "macaroon"` no?

backends/CLNRest.ts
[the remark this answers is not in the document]
vincenzopalazzo: can you clarify at least what `v0.8.2` is?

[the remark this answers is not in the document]
niteshbalusu11: Maybe we just mark it as true since its being supported from a very old version and the new cln-rest we are only supporting from v24.0.2 anyway?

backends/CLNRest.ts
vincenzopalazzo: >also all the actual execution of the fetch calls happen in the LND class so its not just getUrl. 

What do you mean? This approach will leave many calls unused, and you'll likely need to rewrite a lot of code as well when you support commando (or you are not supporting it?).

TL;DR: I disagree with using CLN code that relies on LND. Even though the current API design is different, the previous design wasn't bad because c-lightning-REST emulates LND. However, clnrest/commando does not.

views/Settings/NodeConfiguration.tsx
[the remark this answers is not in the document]
niteshbalusu11: I just find them cleaner to be separate because old c-lightning-rest is sharing with now anyway. Whatever Evan wants.

backends/CLNRest.ts
niteshbalusu11: We are not supporting Commando in this PR. 
The only thing the CLN class relies on LND is the actual "fetch" function to call the APIs. everything else is separate. I tried removing the LND class extension but that was breaking a couple of UI components. I didn't find the time to investigate that.

niteshbalusu11: Nvm, i figured out the issue. CLN no longer extends LND now.

sha-265:
you are too fast 😵‍💫 

anyway, this is probably a good opportunity to fix `getTransactions()` which uses wrong API method ('listtransactions`), as i did in my version:
https://github.com/sha-265/zeus/blob/c40d9c1ba20c504f5e3355cf33d9e8b4409fca09/backends/CLNRest.ts#L33

niteshbalusu11:
> anyway, this is probably a good opportunity to fix `getTransactions()` that uses wrong api method (`listfunds` which return the wallet utxos, and not transactions history), as i did in my version: https://github.com/sha-265/zeus/blob/c40d9c1ba20c504f5e3355cf33d9e8b4409fca09/backends/CLNRest.ts#L33

Thank you for catching that. I think your code is missing tag == `channel_open` and tag == `channel_close` which idk if cln also records has deposit and withdrawal. also it records chain_fee separately. 
Have to look into that.

newtonick:
I get this error when trying send onchain in signet via CLNRest (via this PR)

"unknown parameter: address, this may be caused by
a failure to autodetect key=value-style parameters.
Please try using the -k flag and explicit key=value
pairs of parameters."

image
image

backends/CLNRest.ts
```diff
@@ -0,0 +1,386 @@
...
+        });
+    getLightningBalance = () =>
+        this.postRequest('/v1/listfunds').then((res) => {
+            return getOffchainBalance(res);
+        });
+    sendCoins = (data: TransactionRequest) => {
+        let request: any;
+        if (data.utxos) {
+            request = {
+                address: data.addr,
```
newtonick, line 200: should address be destination? https://docs.corelightning.org/reference/lightning-withdraw

niteshbalusu11: Thank you for catching that. fixed.

backends/CLNRest.ts
```diff
@@ -0,0 +1,386 @@
...
+                feeRate: `${Number(data.sat_per_vbyte) * 1000}perkb`,
```
newtonick, line 201: `feeRate` should also be `feerate` and `satoshis` should also be `satoshi`. Sorry I didn't catch that when I saw address was wrong.

niteshbalusu11: No, not your fault. I missed them. I fixed them now.

sha-265:
> > anyway, this is probably a good opportunity to fix `getTransactions()` that uses wrong api method (`listfunds` which return the wallet utxos, and not transactions history), as i did in my version: https://github.com/sha-265/zeus/blob/c40d9c1ba20c504f5e3355cf33d9e8b4409fca09/backends/CLNRest.ts#L33
> 
> Thank you for catching that. I think your code is missing tag == `channel_open` and tag == `channel_close` which idk if cln also records has deposit and withdrawal. also it records chain_fee separately. Have to look into that.

i don't see this tags (at least under `wallet` account), but there is a `journal_entry` tag that should be filtered out maybe.

sha-265:
> > i don't see this tags (at least under `wallet` account), but there is a `journal_entry` tag that should be filtered out maybe.
> 
> Yeah, that's the problem, it doesn't put it under wallet. Have to figure out where it does. In the examples in the docs, it puts it under some long string account. Have to see what that is.

it's channel id according to the cln documentation:
> account (string): The account name. If the account is a channel, the channel_id.

i'm not sure this events are relevant if you want only the on-chain transactions.

niteshbalusu11:
> i'm not sure these events are relevant if you want only the on-chain transactions.

I think they are relevant because channel opens and closes are chain txs. I pushed a commit to include those. Please test if you can. I just pushed a commit to make the txs more readable so you know what the tx is for with a new notes field.

sha-265:
> > i'm not sure these events are relevant if you want only the on-chain transactions.
> 
> I think they are relevant because channel opens and closes are chain txs. I pushed a commit to include those. Please test if you can. I just pushed a commit to make the txs more readable so you know what the tx is for with a new notes field.

i don't know if there is no duplicate events in `wallet` account for channel opening (at least). 

but i think in terms of performance query with out `account` parameters would be much more heavy as it returns all the non on-chain events, that could be a lot of events, so it should be considered also.

niteshbalusu11:
> but i think in terms of performance query without account parameter, would be much more heavy as it returns all the non on-chain events

I agree. CLNs APIs don't seem to allow pagination as well which is annoying. Maybe someone can test with their mainnet node which is old and see how it performs. 
They should ideally have a simple RPC for chain txs but I'm not gonna rant about their RPC build choices here.

sha-265:
> > but i think in terms of performance query without account parameter, would be much more heavy as it returns all the non on-chain events
> 
> I agree. CLNs APIs don't seem to allow pagination as well which is annoying. Maybe someone can test with their mainnet node which is old and see how it performs. They should ideally have a simple RPC for chain txs but I'm not gonna rant about their RPC build choices here.

yep, not the best api i've come across 🤷‍♂️

other solution could be somehow to make only one query for `getTransactions()` and `getInvoices()` instead of two separated calls for the same data.

newtonick:
From what testing I have been able to do, I believe activity view needs some love. It might be helpful to start with what transactions should be included in this view. I'm not sure if this is documented somewhere (like a previous zeus GitHub issue/PR/etc). What I expect to be listed in the activity view is:

- lightning Invoices (paid, unpaid, expired)
- lightning payment (successful payments only) w/ fees
- onchain withdraw/payment (in mempool and onchain) w/ fees
- onchain receives (in mempool and onchain)

I don't expect to see channel open / close transactions but I also don't mind if they are in this view if it's communicated that they are part of an open/close channel transaction.

Right now I see multiple rows in the activity view for a single on chain payments. I have a hard time dissecting what information bkpr-listaccountevents is returning. Sorry this comment is not more helpful. I just wanted to provide feedback on what I have tested so far with this PR. Thank you for this work.

niteshbalusu11:
> What I expect to be listed in the activity view is:

lightning Invoices (paid, unpaid, expired)
lightning payment (successful payments only) w/ fees
onchain withdraw/payment (in mempool and onchain) w/ fees
onchain receives (in mempool and onchain)

This sounds good to me except expired and unpaid invoices, that could be a bit too much specially expired. 
I will leave it to @kaloudis once he tests and gives feedback on what he thinks.

newtonick:
Here is an example of 40,000 sat onchain spend (https://mutinynet.com/tx/99a918361919956cc76801b6014e233d834466dbb25b4f031825e1eb60b972e8) that shows 3 rows in the activity view for a single spend transaction. The 40,000 sat row appears first as soon as I complete the spend (while the tx is in the mempool). This shows as a credit but is actually a debit. Then once the transaction is confirmed into 1+ blocks, 2 more rows appear. A debit of the utxo I used to make the spend and then a credit of the change back to my wallet.

image

niteshbalusu11:
> Here is an example of 40,000 sat onchain spend (https://mutinynet.com/tx/99a918361919956cc76801b6014e233d834466dbb25b4f031825e1eb60b972e8) that shows 3 rows in the activity view for a single spend transaction. The 40,000 sat row appears first as soon as I complete the spend (while the tx is in the mempool). This shows as a credit but is actually a debit. Then once the transaction is confirmed into 1+ blocks, 2 more rows appear. A debit of the utxo I used to make the spend and then a credit of the change back to my wallet.
> 
> image
> 

Ok, yeah this needs to be fixed. Only the spend should show up, not the utxos spent and change received.

sha-265:
> I don't expect to see channel open / close transactions but I also don't mind if they are in this view if it's communicated that they are part of an open/close channel transaction.

i agree, and it's also solve the potential performance issue, as you can request only `wallet` account events, and not all the events for all the accounts in this node.

niteshbalusu11:
ok @newtonick can you test now? I think it should be fixed.

@sha-265 unfortunately i think we can't get around the performance issue without compromising accuracy of the info because CLN displays change outputs as deposits and its really difficult to figure out if its a real deposit or a change output from like channel opens without checking channel opens from events. I asked in the CLN group if there is an easy way to check if an address is ours in CLN so that we can easily filter out change, let's see.

sha-265:
i don't understand how tons of irrelevant events solving this, but at least you should get the same data only once and not three different times.

niteshbalusu11:
> i don't understand how tons of irrelevant events solving this, but at least you should get the same data only once and not three different times.

I don't understand what you mean? I am getting only once.

sha-265:
look here:
https://github.com/ZeusLN/zeus/blob/e89187c68c6de95487ce94267849e1c1b416e9c5/stores/ActivityStore.ts#L165-L168

i think if you are using `bkpr-listaccountevents` this three calls are overlapping.

niteshbalusu11:
> look here:
> https://github.com/ZeusLN/zeus/blob/e89187c68c6de95487ce94267849e1c1b416e9c5/stores/ActivityStore.ts#L165-L168
> 
> i think if you are using `bkpr-listaccountevents` this three calls are overlapping.

Ah, that's old code. I never touched that. Will look into it.

niteshbalusu11:
Ok I was right. It is what it is with CLN.

niteshbalusu11:
https://github.com/ElementsProject/lightning/issues/7397

Sjors:
I finally got the clnrest plugin to work (Python environments are a pain...). Happy to test all this when there's a testflight build.

backends/CLightningREST.ts
```diff
@@ -10,8 +10,7 @@ import BigNumber from 'bignumber.js';
 export default class CLightningREST extends LND {
     getHeaders = (macaroonHex: string): any => {
         return {
-            macaroon: macaroonHex,
-            encodingtype: 'hex'
```
kaloudis, line 14: can we revert this change? this will break connections for people with existing configurations

kaloudis (changes requested):

kaloudis (approved):
tACK

All things looking good. Congrats, @niteshbalusu11 you will be receiving the bounty.

We are expecting you to help address any issues that may arise during alpha and beta testing - not expecting anything major as we've already done some extensive testing

Sjors:
Just tested the https://github.com/ZeusLN/zeus/releases/tag/v0.9.0-beta1 release on iOs. It manages to connect!

I can seen my utxos on the coins screen. I can see my channels.

I can create an invoice to receive. The screen doesn't update when the invoice is paid, but afaik it didn't do that with c-lightning-rest either.

The Activity tab still crashes as it did with c-lightning-rest.

The combination of the above two things means I can't see if a payment to me succeeded.

I can pay a regular bolt11 invoice.

niteshbalusu11:
> The Activity tab still crashes as it did with c-lightning-rest.

@Sjors is your node really old with a lot of data?

Sjors:
@niteshbalusu11 the node is very old, but it does run the latest v24.05. It has tens of 10K+ transactions, mostly keysends from podcasting 2.0 listeners.

I guess it's tripping over missing field in older transactions?

niteshbalusu11:
> I guess it's tripping over missing field in older transactions?

It could also be that the response from the 10k+ txns could be very large because the RPC call right now has no way limit the records returned. Working on figuring that part out.

niteshbalusu11:
This *might* fix some issues.

https://github.com/ZeusLN/zeus/pull/2290

sha-265:
> Important Note: Minimum core-lightning version to use this is `v24.02` due to breaking changes from cln team.

I don't know how, but it's working for me with older version

Sjors:
> The screen doesn't update when the invoice is paid, but afaik it didn't do that with c-lightning-rest either.

More recently this usually does work.

> The Activity tab still crashes as it did with c-lightning-rest.

This is fixed since.

Issue #2128, opened 2024-04-18
Point of Sale: Charge button: show total in sats if currency rates disabled


kaloudis:
By default, the POS charge button will display the order total in fiat, regardless if currency rates are disabled or not.

If fiat rates are disabled, we should display the total in sats instead. Displaying the total as $0 when rates are disabled doesn't make sense.

PR #2241, opened 2024-06-14
POS: Display 'sats' in CHARGE button when fiat is not enabled

shubhamkmr04:
https://github.com/ZeusLN/zeus/issues/2128

Earlier we used to get $0 when we try to checkout POS orders if fiat are disabled in settings. instead of that we should have the `sats` as currency in that case

Before:

https://github.com/ZeusLN/zeus/assets/50761978/94713b15-54c5-41ff-8aa3-9e781c3e1886

After:

https://github.com/ZeusLN/zeus/assets/50761978/fcb1d988-0fc0-4559-a82e-1c3baac41abe


kaloudis (approved):
tACK

PR #2256, opened 2024-06-20
Currency Converter: Fix flags position in input boxes

shubhamkmr04:
Fixed the flags position in Currency Converter input boxes which broke due to our recent update to react native `0.74.2`


kaloudis (approved):
tACK

PR #2275, opened 2024-07-02
LSPS1: Change token to discount code

shubhamkmr04:


kaloudis:
Since we're rewriting the string entirely, let's rename it. Let's also delete `views.Settings.LSP.lspAccessKey` if it's not longer being used so translators don't have to translate it

locales/en.json
```diff
@@ -982,7 +982,7 @@
     "views.LSPS1.requiredChannelConfirmations": "Required channel confirmations",
     "views.LSPS1.minFundingConfirmWithingBlocks": "Min funding confirm within blocks",
     "views.LSPS1.minRequiredChannelConfirmations": "Min required channel confirmations",
-    "views.LSPS1.token": "Token",
+    "views.LSPS1.token": "Discount code",
```
kaloudis, line 985: Since we're rewriting the string entirely, let's rename it - maybe `general.discountCode`. Let's also delete `views.Settings.LSP.lspAccessKey` if it's not longer being used so translators don't have to translate it

kaloudis (changes requested):

locales/en.json
shubhamkmr04: Sure

kaloudis (approved):

PR #2283, opened 2024-07-10
ContactStore

shubhamkmr04:


kaloudis:
Somewhat on the right track, but the initial loading of contacts should happen when the wallet loads up. We don't want to read from encrypted storage every time we go to contact details, we just want to read from the store

stores/ContactStore.ts
```diff
@@ -0,0 +1,152 @@
...
+        try {
+            const contactsString = await EncryptedStorage.getItem(
+                'zeus-contacts'
+            );
+            const existingContacts: Contact[] = contactsString
+                ? JSON.parse(contactsString)
+                : [];
+
+            if (isEdit && this.prefillContact && !isNostrContact) {
+                console.log('heree', this.prefillContact);
```
kaloudis, line 48: remove this

kaloudis (changes requested):
change requested

stores/ContactStore.ts
shubhamkmr04: done

views/Settings/Contacts.tsx
```diff
@@ -463,7 +438,7 @@ export default class Contacts extends React.Component<
                             secondary
                         />
                     )}
-                    {loading ? (
+                    {ContactStore?.loading ? (
```
kaloudis, line 441: since loading is used twice (here and 405), let's destructure it on line 206

kaloudis (changes requested):

views/ContactDetails.tsx
```diff
@@ -672,8 +681,10 @@ export default class ContactDetails extends React.Component<
                                 <Button
                                     onPress={() => {
                                         navigation.goBack();
+                                        ContactStore.setPrefillContact(
```
kaloudis, line 684: since `setPrefillContact` is used twice (here and 242), let's destructure it at the top of the render function

kaloudis (changes requested):

kaloudis (approved):

PR #2290, opened 2024-07-15
CLNRest: performance improvements using sql query

niteshbalusu11:
# Description

Relates to issue: #1637

We spoke about this a bit on the telegram chat, to begin with we will limit the list of transactions to 150. Right now I was able to limit the expensive `bkpr-listaccountevents` and `listinvoices`. I couldn't restrict `listpays` because it doesn't take a limit argument. There is a `listsendpays` that can take a limit but it doesn't return bolt12 payments (classic CLN). 

- [ ] New feature
- [ ] Bug fix
- [X] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

- [X] Android
- [ ] iOS

- [ ] Embedded LND
- [ ] LND (REST)
- [X] Core Lightning (CLNRest)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -157,7 +157,7 @@ export default class CLNRest {
         route: string,
         ws?: boolean
     ) => {
-        const hostPath = host.includes('://') ? host : `http://${host}`;
+        const hostPath = host.includes('://') ? host : `https://${host}`;
         let baseUrl = `${hostPath}${port ? ':' + port : ''}`;
 
         if (ws) {
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -157,7 +157,7 @@ export default class CLNRest {
         route: string,
         ws?: boolean
     ) => {
-        const hostPath = host.includes('://') ? host : `https://${host}`;
+        const hostPath = host.includes('://') ? host : `http://${host}`;
         let baseUrl = `${hostPath}${port ? ':' + port : ''}`;
 
         if (ws) {
@@ -210,7 +210,8 @@ export default class CLNRest {
         return this.postRequest('/v1/withdraw', request);
     };
     getMyNodeInfo = () => this.postRequest('/v1/getinfo');
-    getInvoices = () => this.postRequest('/v1/listinvoices');
+    getInvoices = () =>
+        this.postRequest('/v1/listinvoices', { limit: 150, index: 'created' });
     createInvoice = (data: any) =>
         this.postRequest('/v1/invoice', {
             description: data.memo,
--- a/backends/CoreLightningRequestHandler.ts
+++ b/backends/CoreLightningRequestHandler.ts
@@ -147,7 +147,8 @@ export const getChainTransactions = async () => {
     }
 
     const getinfo = getinfoResult.value;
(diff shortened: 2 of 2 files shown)


backends/CLNRest.ts
```diff
@@ -210,7 +210,8 @@ export default class CLNRest {
         return this.postRequest('/v1/withdraw', request);
     };
     getMyNodeInfo = () => this.postRequest('/v1/getinfo');
-    getInvoices = () => this.postRequest('/v1/listinvoices');
+    getInvoices = () =>
+        this.postRequest('/v1/listinvoices', { limit: 150, index: 'created' });
```
Sjors, line 214: This seems like it might help with my 10k+ transactions :-)

(could make this paginate in a followup)

niteshbalusu11: Yeah, that's the idea.

newtonick:
I did some testing on this PR. I checked out the PR locally and built from source. I replaced this `/v1/listinvoices` call with the SQL mentioned above. I did this so the output on the activity page was identical (at least for the visible screen).

It's a noticeable performance increase. I'm running my node on the same local network my phone is connecting to via WIFI. I'm comparing to the current v0.9.0-beta1 release from yesterday to this PR built from source. The activity view loads around 6-7 seconds faster.

Here is a screen recording of the performance. The first (light green) activity view loaded is v0.9.0-beta1 (not this PR). The second (yellow) activity view loaded is this PR. I have lurker privacy mode turned on for the recording.

https://github.com/user-attachments/assets/6f808d02-b3eb-4e33-a37e-56728eea935e

niteshbalusu11:
Nice. Thank you for testing this. This again isn't the perfect solution because of the 150 records limitation but at least it makes the activity view usable.

kaloudis:
Let's try this in `v09.0-beta2`

kaloudis (approved):

niteshbalusu11:
@Sjors can you test RC2 please?

Sjors:
Just tested, I can see my transactions now! 🎉 

When I click on a (sent) lightning transaction it immediately crashes.

Clicking on on chain does behave.

I get the impression that I'm not seeing _received_ lightning transactions, though would be mostly zaps and keysend from podcast streaming.

Sjors:
By the way, I'm not a huge fan of the "€1,00 + €0,01 fee" notation. Especially for lightning fees are usually irrelevant and it just seems like visual clutter. It would be better to display fees only if you click on a transaction to see details.

What I _would_ like to see in the overview is a description. Maybe that can go where it says "lightning". And then "You sent" could say "You sent (⚡)"

niteshbalusu11:
> Just tested, I can see my transactions now! 🎉 
> 
> When I click on a (sent) lightning transaction it immediately crashes.
> 
> Clicking on on chain does behave.
> 
> I get the impression that I'm not seeing _received_ lightning transactions, though would be mostly zaps and keysend from podcast streaming.

Will look into this.

Sjors:
As of CLN v24.11 the `listpays` command has `index`, `start` and `limit` arguments. Might be good to leverage those.

PR #2292, opened 2024-07-18
Display Notes on Activity view

shubhamkmr04:
Simulator Screenshot - iPhone 15 Pro - 2024-07-23 at 20 27 16


kaloudis (changes requested):
Simulator Screenshot - iPhone 15 Plus - 2024-08-27 at 11 47 31

I think we should have a max char limit on the display preview here.

Note label is also rendered oddly here.

shubhamkmr04:
> Simulator Screenshot - iPhone 15 Plus - 2024-08-27 at 11 47 31
> 
> I think we should have a max char limit on the display preview here.
> 
> Note label is also rendered oddly here.

Simulator Screenshot - iPhone 15 Pro - 2024-09-03 at 17 00 45

Fixed the notes overlapping and the char limit to 150, you think it would work ?

kaloudis:
Looking good now, visually.

simulator_screenshot_C304922D-BC84-4C9D-92EC-6700979F63C0

views/Wallet/Wallet.tsx
```diff
@@ -345,6 +349,8 @@ export default class Wallet extends React.Component<WalletProps, WalletState> {
 
         LnurlPayStore.reset();
 
+        NotesStore?.loadNoteKeys();
```
kaloudis, line 352: I don't believe we need to load this every time we hit the wallet screen. On `connecting` should suffice

views/Activity/Activity.tsx
```diff
@@ -225,6 +227,33 @@ export default class Activity extends React.PureComponent<
             </TouchableOpacity>
         );
 
+        const getMatchingNote = (item: any) => {
+            const { NotesStore } = this.props;
+            const strippedNoteKeys = NotesStore.noteKeys.map((key) =>
```
kaloudis, line 232: going through this map and creating this object for every item in activity is both inefficient and redundant. If we need this index of stripped keys we should do it once. Perhaps in `loadNoteKeys`?

kaloudis (changes requested):

kaloudis:
We need to rebase this too, once ready. Please rewind your commits and rebase instead of doing a merge commit, please.

views/Wallet/Wallet.tsx
shubhamkmr04: Yup, we have covered that in [2393](https://github.com/ZeusLN/zeus/pull/2393)

views/AddNotes.tsx
```diff
@@ -71,8 +70,7 @@ export default class AddNotes extends React.Component<
         const { notes } = this.state;
 
         const saveNote = async () => {
-            const key: string =
-                'note-' + (payment_hash || txid || getRPreimage);
+            const key: any = payment_hash || txid || getRPreimage;
```
kaloudis, line 73: same here

views/AddNotes.tsx
```diff
@@ -119,9 +117,8 @@ export default class AddNotes extends React.Component<
                         onChangeText={(text: string) => {
                             this.setState({ notes: text });
                             if (!text) {
-                                const key: string =
-                                    'note-' +
-                                    (payment_hash || txid || getRPreimage);
+                                const key: any =
+                                    payment_hash || txid || getRPreimage;
```
kaloudis, line 121: same here. I do not understand the changes to this file

views/Invoice.tsx
```diff
@@ -91,7 +91,9 @@ export default class InvoiceView extends React.Component<InvoiceProps> {
         const EditNotesButton = () => (
             <TouchableOpacity
                 onPress={() =>
-                    navigation.navigate('AddNotes', { getRPreimage: noteKey })
+                    navigation.navigate('AddNotes', {
+                        getRPreimage: getNoteKey
```
kaloudis, line 95: I really dislike the name of the prop we're passing in here. Doesn't it work better as `noteKey` or something akin to that?

views/Payment.tsx
```diff
@@ -100,15 +100,17 @@ export default class PaymentView extends React.Component<PaymentProps> {
...
             getPaymentRequest
         } = payment;
         const date = getDisplayTime;
 
         const EditNotesButton = () => (
             <TouchableOpacity
                 onPress={() =>
-                    navigation.navigate('AddNotes', { payment_hash: noteKey })
+                    navigation.navigate('AddNotes', {
+                        payment_hash: getNoteKey
```
kaloudis, line 112: same here with the prop name here. do we handle these differently?

nit: also confused by the different casing

views/AddNotes.tsx
```diff
@@ -53,11 +53,10 @@ export default class AddNotes extends React.Component<
...
     async componentDidMount() {
-        const key: string =
-            'note-' +
-            (this.state.txid ||
-                this.state.payment_hash ||
-                this.state.getRPreimage);
+        const key: any =
+            this.state.txid ||
+            this.state.payment_hash ||
+            this.state.getRPreimage;
```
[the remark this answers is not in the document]
shubhamkmr04: we were getting type error here like `Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.` once I removed **here-** because it was already coming in getNoteKey

kaloudis (approved):
tACK

Issue #2437, opened 2024-09-24
Bug: Clearing Currency Input Causes Zero Instead of Empty Value


shubhamkmr04:
### Describe your enhancement idea

When using the currency converter, if a user clears the value from the input box, the converter sets the value to 0 instead of keeping it empty. 

**Steps to Reproduce:**

- Go to the currency converter screen.
- Enter any numeric value in any of the currency input boxes (e.g., BTC, USD, or sats).
- Clear the input box by deleting the value.
- Observe that the corresponding conversion shows 0 instead of remaining empty.

kaloudis:
tACK

PR #2438, opened 2024-09-24
Bug Fix: Prevent Currency input to display '0' when it is empty

shubhamkmr04:
https://github.com/ZeusLN/zeus/issues/2437

To fix this issue, we made the following changes:
   - We added a check that immediately stops the conversion process if the input field is cleared. This prevents the input from converting an empty value into `0`.

   - When the input is cleared, we made sure all related converted values in the other currency fields are also cleared, instead of showing incorrect conversions.

### DIFF
--- a/views/Settings/CurrencyConverter.tsx
+++ b/views/Settings/CurrencyConverter.tsx
@@ -189,7 +189,20 @@ export default class CurrencyConverter extends React.Component<
 
         const convertedValues: { [key: string]: string } = { ...inputValues };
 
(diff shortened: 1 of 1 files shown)


shubhamkmr04:
Before:

https://github.com/user-attachments/assets/4cdf60ba-bfc2-4951-a51b-7d34895749fa



After:

https://github.com/user-attachments/assets/48108c98-1293-4aa2-9330-20a723e6b5fe

kaloudis (approved):
tACK

Issue #1467, opened 2023-05-24
Activity / History CSV export


myxmaster:
### Describe the feature

It would be nice to have a CSV export for past activities :)

InnocuousFinch:
If we can export transactions, it would be great if we could also import into the wallet :)

kaloudis:
Added in v0.9.2, finally!

PR #2488, opened 2024-10-23
[feature] Activity: CSV Export

kaloudis:
# Description

Relates to issue: [**ZEUS-1467**](https://github.com/ZeusLN/zeus/issues/1467)


kaloudis:
> My bad, the 2 commas totally make sense of course, because the data always has to match the headlines and in case there is no note the 2 commas will lead to an empty "note" field when importing the csv to e.g. Excel.
> 
> But you now put everything in quotes, not sure why, I think that is unneeded.

Excel, Numbers etc should be able to grok out the values with quotes in place.

views/Activity/ActivityToCsv.tsx
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+                Alert.alert(
+                    localeString('general.success'),
+                    localeString('views.ActivityToCsv.csvDownloaded')
+                );
+                closeModal();
+            } catch (err) {
+                console.error(
+                    localeString('views.ActivityToCsv.csvDownloadFailed'),
```
kaloudis, line 163: don't need to localize this if it's only used for console errors

vatkachenko: Omg

PR #2495, opened 2024-10-28
Fix: Button Layouts in Views

shubhamkmr04:
We had these issues in button layouts on Receive view which came after our recent typescript fix after this [change](https://github.com/ZeusLN/zeus/commit/9712aad0878f2d1b8b62da94ad02f2b12a1b041e#diff-879ea016c4558f026ea4413d7927fe60b543dec7a5572514fd3224c9da3eee52R1399)

Fixed the issue by reversing the changes.

Below is one of those layout issues
Before:
Simulator Screenshot - iPhone 16 Pro - 2024-10-28 at 15 18 57

After:
Simulator Screenshot - iPhone 16 Pro - 2024-10-28 at 15 18 46

### DIFF
--- a/views/Channels/ChannelsPane.tsx
+++ b/views/Channels/ChannelsPane.tsx
@@ -44,7 +44,7 @@ interface ChannelsProps {
     SettingsStore?: SettingsStore;
 }
 
-const ColorChangingButton = (onPress: any) => {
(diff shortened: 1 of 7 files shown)


shubhamkmr04:
We still need to reverse all such changes we had done

kaloudis (approved):

PR #2539, opened 2024-11-16
deps: React Native v0.76.2

kaloudis:


kaloudis:
Might need an update from `react-native-blob-util` first

kaloudis:
`react-native-blob-util` should be ok, but `react-native-notifications` needs an update

https://github.com/wix/react-native-notifications/pull/1038

PR #2624, opened 2024-12-12
Enhance on-chain receive workflow and address type selection

myxmaster:
# Description

These changes
- improve rendering performance, when clicking `Receive` in on-chain swipeable row
  - moved selectedIndex initialization to constructor (no more renderings with wrong selectedIndex)
  - added `selectedIndex !== 2` condition to prevent unnecessary rendering
- show SettingsButton instead of ClearButton for selectedIndex=2 (on-chain), because here it makes sense to choose address type (and not going back to create another invoice)
- enhance addressType modalBox (Receive and InvoiceSettings)
  - ~~60% height~~
  - scrollable content for longer translations
  - prevents unnecessary updates when same address type is selected
  - Receive only: automatically generates new on-chain address when new address type is picked

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -1514,12 +1514,14 @@ export default class Receive extends React.Component<
                         posStatus === 'active' ||
                         hideRightHeaderComponent ? undefined : route.params
                               ?.selectedIndex === 2 ? (
-                            BackendUtils.supportsAddressTypeSelection() ? (
+                            BackendUtils.supportsAddressTypeSelection() &&
+                            account === 'default' ? (
                                 <SettingsButton />
                             ) : undefined
                         ) : haveInvoice ? (
                             selectedIndex === 2 ? (
-                                BackendUtils.supportsAddressTypeSelection() ? (
+                                BackendUtils.supportsAddressTypeSelection() &&
+                                account === 'default' ? (
                                     <SettingsButton />
                                 ) : (
                                     <ClearButton />
@@ -1529,6 +1531,7 @@ export default class Receive extends React.Component<
                             )
                         ) : !creatingInvoice &&
                           BackendUtils.supportsAddressTypeSelection() &&
+                          account === 'default' &&
                           (selectedIndex === 2 || selectedIndex === 1) ? (
                             <SettingsButton />
                         ) : undefined
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -2078,73 +2078,81 @@ export default class Receive extends React.Component<
                                         <>
                                             {BackendUtils.supportsLSPs() &&
                                                 !lspNotConfigured && (
-                                                <View
-                                                    style={{
-                                                        flexDirection: 'row',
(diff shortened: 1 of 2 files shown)


kaloudis (changes requested):
there's a bug. Go to Receive. Select an address type. Autogenerate triggers but no invoice or QR is displayed

views/Settings/InvoicesSettings.tsx
```diff
@@ -555,7 +563,7 @@ export default class InvoicesSettings extends React.Component<
                         backgroundColor: themeColor('background'),
                         borderTopLeftRadius: 20,
                         borderTopRightRadius: 20,
-                        height: 450,
+                        height: windowSize.height * 0.6,
```
kaloudis, line 566: not sure I like this

views/Receive.tsx
```diff
@@ -1507,7 +1512,8 @@ export default class Receive extends React.Component<
                         loading ||
                         watchedInvoicePaid ||
                         posStatus === 'active' ||
-                        hideRightHeaderComponent ? null : haveInvoice ? (
+                        hideRightHeaderComponent ? null : haveInvoice &&
+                          selectedIndex !== 2 ? (
```
kaloudis, line 1516: we need to redo the logic here.

If a node doesn't support address type selection, it should still show the X here. Otherwise it is not uniform.

`haveInvoice && !(selectedIndex === 2 && BackendUtils.supportsAddressTypeSelection())` should do it

kaloudis (changes requested):

kaloudis (changes requested):
There's another bug here:

When creating a new address from an external account, users are still allowed to change the address type, even though external accounts have fixed address types.

To make matters worse, the newly generated addresses come from the internal on-chain wallet but the label above the QR persists with the external account label.

shubhamkmr04 (commented):
LGTM

views/Receive.tsx
```diff
@@ -263,12 +267,13 @@ export default class Receive extends React.Component<
                 (settings?.invoices?.blindedPaths &&
                     BackendUtils.supportsBolt11BlindedRoutes()) ||
                 false,
-            enableLSP: settings?.enableLSP,
+            enableLSP: settings?.enableLSP || true,
```
kaloudis, line 270: This will cause the LSP to always be enabled

kaloudis (changes requested):

kaloudis (approved):
tACK

Issue #2602, opened 2024-12-04
Syncing wallet + Node Info: format block height values with commas


kaloudis:
Let's add commas to these values, using the `numberWithCommas` from `UnitUtils` to make them easier to read.

PR #2627, opened 2024-12-13
Format block height with commas

shubhamkmr04:
https://github.com/ZeusLN/zeus/issues/2602

### DIFF
--- a/views/NetworkInfo.tsx
+++ b/views/NetworkInfo.tsx
@@ -9,6 +9,7 @@ import Screen from '../components/Screen';
 
 import { localeString } from '../utils/LocaleUtils';
 import { themeColor } from '../utils/ThemeUtils';
+import { numberWithCommas } from '../utils/UnitsUtils';
 
 import NodeInfoStore from '../stores/NodeInfoStore';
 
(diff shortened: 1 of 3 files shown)


shubhamkmr04:
Screenshot_1734082878

@kaloudis should we also format some values here on this view ?

kaloudis:
> Screenshot_1734082878
> 
> @kaloudis should we also format some values here on this view ?

Yes, let's do it

views/NetworkInfo.tsx
```diff
@@ -72,13 +73,17 @@ export default class NetworkInfo extends React.Component<NetworkInfoProps, {}> {
 
                 <FlatList
                     data={NETWORK_INFO}
-                    renderItem={({ item }) => (
+                    renderItem={({ item, index }) => (
                         <KeyValue
                             keyValue={localeString(item.label)}
-                            value={item.value}
+                            value={
+                                index < 3
```
kaloudis, line 80: not crazy about using the index here

shubhamkmr04: pushed few changes! would this work?

PR #2660, opened 2024-12-21
Call getSettingsAndNavigate less often

myxmaster:
# Description

Before, getSettingsAndNavigate() was called everytime Wallet.tsx gained focus. Since processing getSettingsAndNavigate() takes some time, this resulted in the app not being responsive for a short time. Example: When you closed the menu and then immediately clicked on a balance slider, it was not responding.

With this PR the flow is optimized, so getSettingsAndNavigate() is only called:
- via handleFocus():
  - On initial load (using existing initialLoad state)
  - When coming from Lockscreen after logging in (using new triggerSettingsRefresh flag)
  - When POS was enabled (using new posWasEnabled flag)
- via handleAppStateChange():
  - when app comes from background (no PIN/password configured)

Additionally fixed a bug, where autonavigation into POS didn't work in this edge case:
- Configure a PIN/password, enable `Require login after app returns from background`
- Enable POS and leave menu (autonavigate works)
- Disable POS and leave menu
- Put the app in background and reopen it, log in
- Enable POS and leave menu
=> **no autonavigate is happening**

Root cause was in Lockscreen.tsx, now it is only setting posStatus to 'inactive' when POS is actually enabled.

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/views/Lockscreen.tsx
+++ b/views/Lockscreen.tsx
@@ -9,7 +9,7 @@ import {
     Text,
     View
 } from 'react-native';
-import { Route, CommonActions } from '@react-navigation/native';
+import { Route } from '@react-navigation/native';
 import { StackNavigationProp } from '@react-navigation/stack';
 
 import Button from '../components/Button';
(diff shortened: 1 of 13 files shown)


views/Lockscreen.tsx
```diff
@@ -249,8 +249,14 @@ export default class Lockscreen extends React.Component<
...
             } else {
-                setPosStatus('inactive');
+                if (
+                    (SettingsStore.settings?.pos?.posEnabled ||
+                        PosEnabled.Disabled) !== PosEnabled.Disabled
+                ) {
+                    setPosStatus('inactive');
+                }
                 this.resetAuthenticationAttempts();
+                SettingsStore.comingFromLockscreen = true;
```
kaloudis, line 259: Think I'd rather just pass in a parameter into the `proceed` call below than add another var to keep track of in the store.

views/Lockscreen.tsx
```diff
@@ -144,7 +159,7 @@ export default class Lockscreen extends React.Component<
             !deleteDuressPin
         ) {
             SettingsStore.setLoginStatus(true);
-            this.proceed('Wallet');
+            this.proceed('Wallet', true);
```
kaloudis, line 162: What does the `true` do here?

kaloudis (changes requested):

kaloudis:
Wow, much cleaner now!

views/Wallet/Wallet.tsx
```diff
@@ -196,7 +196,18 @@ export default class Wallet extends React.Component<WalletProps, WalletState> {
...
+
+        if (
+            this.state.initialLoad ||
+            SettingsStore.posWasEnabled ||
+            SettingsStore.triggerSettingsRefresh
+        ) {
+            this.getSettingsAndNavigate();
+            SettingsStore.posWasEnabled = false;
+            SettingsStore.triggerSettingsRefresh = false;
+        }
```
kaloudis, line 210: do we want to include a comment for this block?

kaloudis (approved):

PR #2728, opened 2025-01-10
Add share feature

### LINKED ISSUE #1285: Ability to copy/share QR code image
### Describe the feature

As a user, I'd like to be able to copy/share a QR code image. E.g. in Phoenix, I can long-tap a QR code image and an action pop-up appears to copy/share/enlarge the image.

Essentially, being able to copy the image would be awesome. Sharing option is just nice to have.

myxmaster:
# Description

This fixes #1285.

This adds the ability to share invoices (and on-chain addresses and lightning-addresses) via OS share dialog. The shared content includes both QR and text representation. Unfortunately it depends on the app which is used for sharing if both is actually used.

- A black/white 800x800 version of the QR will temporarily be rendered (not theme colors) for sharing and it will always use the Z logo.
- Additionally enhanced CollapsedQR (with CopyButton/ShareButton) to be used in iconOnly mode which saves some space (currently only implemented for Receive screen).
- Copy icon will give feedback if clicked (check icon for 5 seconds, see screenshot).
- ShareButton works just like CopyButton.

|grafik|grafik|grafik|
|-|-|-|

Implementation details:
I was playing around a lot, and ended up with this current approach. It looks kind of complicated, but to be able to share the QR it actually has to be rendered and captured (at least I didn't find another way). I wanted to avoid rendering a 2nd hidden black/white version of the QR whenever an invoice is generated, so I ended up with this logic:
1. User clicks ShareButton
2. ShareButton's handlePress calls its onPress prop (handleShare from CollapsedQR)
3. handleShare creates a new temporary QR ref and sets state
4. handleShare starts checking qrReady state in a loop with requestAnimationFrame
5. The state update triggers a render with the hidden QR
6. ForwardedQRCode's getRef callback fires when QR is ready, setting qrReady to true
7. handleShare resolves when qrReady becomes true
8. ShareButton's shareContent executes, capturing the QR ref to base64
9. Share dialog opens with the captured image
10. After sharing completes, tempQRRef is cleared from state

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:


shubhamkmr04 (commented):
image
Its not rendering the logo in iOS

shubhamkmr04:
> It works elsewhere on iOS though? (e.g. via Receive->Unified Invoice)

No It happens at all the places where QR can be shared, android worked perfect though

shubhamkmr04:
> No I mean, is the "Z" logo displayed in the normal QR codes on iOS? (e.g. Receive->Unified Invoice)
> 
> I'm asking because I have no idea what could be the root cause for this, if it is only happening for the shared QR.

yup its displayed in the normal QR codes generated in iOS

shubhamkmr04:
> :/
> 
> Ok. Can you maybe help me find out what the difference is? I cannot debug on iOS...

Sure, I'll take a look at the code and let you know what we can do.

components/ShareButton.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    shareContent = async () => {
+        const { value, qrRef, onShareComplete } = this.props;
+        try {
+            if (!qrRef?.current) {
+                return;
+            }
+
+            const svgElement = qrRef.current;
+            const base64Data = await captureRef(svgElement, {
```
shubhamkmr04, line 42: @myxmaster It looks like the QR logo issue on iOS happens because the capture runs before the logo fully loads. Try adding a 500ms delay here — should fix it! 
            `await new Promise((resolve) => setTimeout(resolve, 500));
`

shubhamkmr04: can't figure out any other way except adding a little delay

shubhamkmr04: yup I agree. 50ms delay is also working though if we dont find any other solution.

kaloudis: It's not ideal. Let's see if we can figure out an alternative approach.

components/CollapsedQR.tsx
```diff
@@ -277,7 +355,36 @@ export default class CollapsedQR extends React.Component<
...
+                            }
+                        />
+                    </View>
+                ) : (
+                    <CopyButton
+                        copyValue={copyValue || value}
+                        title={copyText}
+                        iconOnly={iconOnly}
+                    />
+                )}
```
kaloudis, line 387: I think it would be really nice if we changed the NFC share button to match the style of the other two new buttons - that is unless share NFC shows up in the Android menu: then it's redundant

components/ShareButton.tsx
kaloudis: Haven't found a better approach yet. Put an iOS specific patch in https://github.com/ZeusLN/zeus/pull/2728/commits/a01c684f4fce70579f372b6603c7ffafda5c4f71

components/CollapsedQR.tsx
kaloudis: NFC button can simply take the theme's highlight color when activated, then we can remove the two related locale strings if they're not used elsewhere.

Need to evaluate if we want to break it into it's own button component: `NFCButton`

kaloudis: I've pushed changes for a new `NFCButton` component in a new commit. We will reinvestigate further work on NFC at a later date.

PR #2687, opened 2024-12-30
ZEUS-2683: Storage: migrate from `react-native-encrypted-storage` to `react-native-keychain`

kaloudis:
# Description

Relates to issue: [**ZEUS-2683**](https://github.com/ZeusLN/zeus/issues/2683), [**ZEUS-2475**](https://github.com/ZeusLN/zeus/issues/2475)

This pull request is categorized as a:

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

I have tested this PR on the following platforms (please specify OS version and phone model/VM):

- [ ] Android
- [x] iOS

I have tested this PR with the following types of nodes (please specify node version and API version where appropriate):

- [ ] Embedded LND
- [x] LND (REST)
- [ ] LND (Lightning Node Connect)
- [ ] Core Lightning (CLNRest)
- [ ] LndHub
- [ ] [DEPRECATED] Core Lightning (c-lightning-REST)
- [ ] [DEPRECATED] Core Lightning (Spark)
- [ ] [DEPRECATED] Eclair

### Locales
- [ ] I’ve added new locale text that requires translations
- [ ] I’m aware that new translations should be made on the ZEUS [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Third Party Dependencies and Packages

- [ ] Contributors will need to run `yarn` after this PR is merged in
- [ ] 3rd party dependencies have been modified:
    * verify that `package.json` and `yarn.lock` have been properly updated
    * verify that dependencies are installed for both iOS and Android platforms

### Other:

- [ ] Changes were made that require an update to the README
- [ ] Changes were made that require an update to onboarding

Issue #1054, opened 2022-08-04
Remove android.enableJetifier line from gradle.properties & jetifier usage


emanuelb:
https://github.com/ZeusLN/zeus/blob/a2dcfd7f652b95a14af58bebd6493c77b3466e66/android/gradle.properties#L21
See this guide for the steps to ensure it's safe to remove above line `android.enableJetifier=true` (check that all deps are already fully migrated to AndroidX, and migrate the ones still not migrated to it)
https://medium.com/dipien/say-bye-bye-to-android-jetifier-a7e0d388f5d6

Also remove line of jetifier package:
https://github.com/ZeusLN/zeus/blob/ca19350c0071ec3d8f9a44cb832fe6cf1babcecd/package.json#L130

Compiling app result in "Jetifier is deprecated" warning.
```
Jetifier is deprecated. Future versions of react-native CLI may not run it by default.
Jetifier found 1553 file(s) to forward-jetify. Using 6 workers...
Jetifier: propose an AndroidX conversion PR to this repository: node_modules/react-native-tcp/android/src/main/java/com/peel/react/TcpSocketManager.java
Jetifier: propose an AndroidX conversion PR to this repository: node_modules/react-native-tcp/android/src/main/java/com/peel/react/TcpSockets.java
Jetifier: propose an AndroidX conversion PR to this repository: node_modules/react-native-udp/android/src/main/java/com/tradle/react/UdpSocketClient.java
Jetifier: propose an AndroidX conversion PR to this repository: node_modules/react-native-udp/android/src/main/java/com/tradle/react/UdpSockets.java
```

kaloudis:
@emanuelb are you open to making a PR to address this?

kaloudis:
@myxmaster we ready to rip this out now?

kaloudis:
I believe we're actually clear to remove this now https://github.com/ZeusLN/zeus/pull/3587

kaloudis:
https://github.com/ZeusLN/zeus/pull/3607

PR #2757, opened 2025-01-19
Check, request and handle notification permissions

myxmaster:
# Description

(Based on https://github.com/ZeusLN/zeus/pull/2687 and https://github.com/ZeusLN/zeus/pull/2752, the latter is required, otherwise the check, if notifications are needed, does not work reliably [correct ln address `enabled` flag is needed].)

This lets Zeus properly check, request and handle notification permissions.

Tested with Android 12 and Android 14 (notification permission logic changed as of Android 13). **Untested for iOS.**

Zeus now handles notification permissions across different Android versions (and iOS hopefully also works :alien:). When notifications are needed, it will either request permission directly (not possible on Android 12 and lower) or guide users to system settings with an infoModal. This happens when:
  - user enables persistent mode
  - user creates a lightning address for the first time (because `notifications=push` is default)
  - user creates a lightning address not for the first time (other node has ln address) while `notifications=push`
  - Wallet.tsx loads while persistent mode is enabled OR current node pubkey has ln address + `notifications=push`

Also:
- enabling persistent mode is only possible if user grants notification permission
- changing lightning address settings to `notifications=push` is only possible if user grants notification permission
- creating a lightning address will still be possible if user denies notification permission, but it will then set `notifications=disabled`

Additionally fixed a "Possible unhandled promise rejection" warning by adding an empty catch in `LightningAddress\index.tsx`.

Note 1:
We should use react-native-permissions everywhere, then we can finally fix this: https://github.com/ZeusLN/zeus/issues/1054. But I didn't want to put that in this same PR.

Note 2:
Added AsyncStorage and react-native-permissions mocks to handleAnything.test.ts and MigrationUtils.test.ts because these modules are required through the dependency chain (MigrationUtils imports LightningAddressStore which imports NotificationUtils).

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [ ] iOS

- [x] Embedded LND
- [x] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/components/Modals/InfoModal.tsx
+++ b/components/Modals/InfoModal.tsx
@@ -132,12 +132,9 @@ export default class InfoModal extends React.Component<InfoModalProps, {}> {
                                 <Button
                                     title={localeString('general.close')}
                                     onPress={() => {
-                                        toggleInfoModal(
-                                            undefined,
-                                            undefined,
-                                            undefined,
-                                            ModalStore.onDismiss
-                                        );
+                                        toggleInfoModal({
+                                            onDismiss: ModalStore.onDismiss
+                                        });
                                     }}
                                     secondary
                                 />
--- a/components/KeyValue.tsx
+++ b/components/KeyValue.tsx
@@ -109,11 +109,11 @@ export default class KeyValue extends React.Component<KeyValueProps, {}> {
             Key = (
                 <TouchableOpacity
                     onPress={() =>
-                        toggleInfoModal(
-                            infoModalText,
-                            infoModalLink,
-                            infoModalAdditionalButtons
-                        )
+                        toggleInfoModal({
+                            text: infoModalText,
+                            link: infoModalLink,
+                            buttons: infoModalAdditionalButtons
+                        })
                     }
                 >
                     {KeyBase}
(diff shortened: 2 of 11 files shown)


components/Modals/InfoModal.tsx
```diff
@@ -130,7 +131,14 @@ export default class InfoModal extends React.Component<InfoModalProps, {}> {
                             <View style={styles.button}>
                                 <Button
                                     title={localeString('general.close')}
-                                    onPress={() => toggleInfoModal()}
+                                    onPress={() => {
+                                        toggleInfoModal(
+                                            undefined,
+                                            undefined,
+                                            undefined,
+                                            ModalStore.onDismiss
```
kaloudis, line 139: this is getting a little crazy here. I think it's time to switch this to one param object for this func

Issue #2671, opened 2024-12-23
Activity: investigate adding paging


kaloudis:
Instead of calling all invoices, txs, and payments at once on the `Activity` view, we should use the `index_offset` parameters to do paging.

https://lightning.engineering/api-docs/api/lnd/lightning/list-payments/
https://lightning.engineering/api-docs/api/lnd/lightning/list-invoices/
https://lightning.engineering/api-docs/api/lnd/lightning/get-transactions/

The biggest challenge is putting the data together. Ideally LND would have a single Activity endpoint we could use

We could potentially be able to do something similar for CLNRest using the custom SQL calls.

Related issues:
https://github.com/ZeusLN/zeus/issues/2465
https://github.com/ZeusLN/zeus/issues/2390

kaloudis:
Took another look at where this stands. Every blocker in the original writeup has since been cleared:

**lnd**
- `ListPayments` and `ListInvoices` have always supported `index_offset` + `reversed`, with `first_index_offset`/`last_index_offset` cursors in the response.
- `GetTransactions` gained pagination (`index_offset`, `max_transactions`, plus `first_index`/`last_index` response cursors) in lnd 0.19.0 (lightningnetwork/lnd#8998). The fields are already in our vendored proto, and our embedded build has them.
- `ListInvoices` and `ListPayments` also accept `creation_date_start`/`creation_date_end` since lnd 0.16.0 (lightningnetwork/lnd#7159), so date filters can be pushed server-side. That is also the cheap fix for #2390.
- Still no unified activity endpoint, so the merge stays client-side. With per-source cursors it's a standard k-way merge: fetch newest-first pages per source, emit merged items until one source's page runs dry, then refetch only that source. Local sources (Cashu, swaps, LSP orders) are cheap and can stay fully loaded in the merge.

**Other backends**
- CLNRest already fetches invoices and payments via `/v1/sql`, so adding LIMIT/OFFSET and a timestamp WHERE clause is trivial.
- NWC's `list_transactions` takes `limit`/`offset`/`from`/`until` per NIP-47.
- LNDHub's `gettxs`/`getuserinvoices` take `limit`/`offset` (server support varies, see #3240).
- LDK Node is local, so no paging needed. LNC speaks the same lnd proto as above.

**Implementation notes**
- The Activity FlatList already sets `onEndReachedThreshold` but has no `onEndReached` handler; that's the wiring point.
[...]

PR #2771, opened 2025-01-24
Activity: LND: limit to 500 txs

kaloudis:
# Description

Based off of https://github.com/ZeusLN/zeus/pull/2751

`max_transactions` on `GetTransactions` still hasn't been cut into an LND release yet, so please wait for that to be released - then update schemas - before merging this.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

- [ ] Embedded LND
- [ ] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/backends/LND.ts
+++ b/backends/LND.ts
@@ -330,7 +330,11 @@ export default class LND {
                 : undefined,
             route_hints: data.route_hints
         });
-    getPayments = (params?: { maxPayments?: number; reversed?: boolean }) =>
+    getPayments = (
+        params: { maxPayments?: number; reversed?: boolean } = {
+            maxPayments: 500
+        }
+    ) =>
         this.getRequest(
             `/v1/payments?include_incomplete=true${
                 params?.maxPayments ? `&max_payments=${params.maxPayments}` : ''
--- a/backends/LND.ts
+++ b/backends/LND.ts
@@ -235,9 +235,13 @@ export default class LND {
 
     getTransactions = (data: any) =>
         this.getRequest(
-            data && data.start_height
-                ? `/v1/transactions?end_height=-1&start_height=${data.start_height}`
-                : '/v1/transactions?end_height=-1'
+            `/v1/transactions?end_height=-1${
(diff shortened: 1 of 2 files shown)


kaloudis:
> * `max_transactions` is accepted as a parameter but no caller ever passes it, so everything falls back to 500. Why not just hardcode 500 then?
> 
>     * The LND backend only adds `max_transactions` to the query if someone passes it in, that we never do so it ends up with no limit
> 
>     * should we let users know about this like 'you are seeing your last 500 transactions' ?

Aren't bullets 1+2 contradictory? The first one is correct. I want to leave the parameter so we have the option in case we decide to make an activity preview view.

As per point 3, I'm not sure how we can introduce that without cluttering the UI. Ideally we support paging. We may finally be able to do that with LND now, worth exploring. See https://github.com/ZeusLN/zeus/issues/2671

shubhamkmr04:
> > * `max_transactions` is accepted as a parameter but no caller ever passes it, so everything falls back to 500. Why not just hardcode 500 then?
> >   
> >   * The LND backend only adds `max_transactions` to the query if someone passes it in, that we never do so it ends up with no limit
> >   * should we let users know about this like 'you are seeing your last 500 transactions' ?
> 
> Aren't bullets 1+2 contradictory? The first one is correct. I want to leave the parameter so we have the option in case we decide to make an activity preview view.
> 
> As per point 3, I'm not sure how we can introduce that without cluttering the UI. Ideally we support paging. We may finally be able to do that with LND now, worth exploring. See #2671

ok we can leave the parameter as it is. but shouldn't we update LND to fall back to 500 like other backends?

shubhamkmr04:
why did we remove changes from EmbeddedLND.ts ?

PR #2790, opened 2025-02-01
Allow cancellation of backend calls

### LINKED ISSUE #2713: When switching nodes, data from the previous connection can be used for new connection requests
### Describe the bug

I was switching to a regtest LND node (10.0.2.2:8085) with just 2 channels and noticed multiple (more than 100) time outs like this:
`Request timed out for: https://10.0.2.2:8085/v1/graph/node/038a82d6518d9f8bbd6ca6f0c7e4b0d8fe339252e454fe5f605b6a75baf2344152`
-> all with different pub keys

I was coming from a CLNRest testnet node with lots of channels, and it is pretty clear that all those pub keys are coming from that CLNRest node. That means, we are not properly resetting data.

To reproduce, you need to connect to the CLNRest node and immediately switch to a different node, before all the backend calls for the CLNRest connection can finish.

----

Not the solution, but kind of related: Should we also clear those in ChannelsStore.ts:reset()?
```
this.aliasesById = {};
this.aliasMap.clear();
this.chanInfo = {};
```

### Reproduce

...

### ZEUS version

v0.9.5

### Node interface

Multiple (specified in comments)

### Network

Clearnet

### Device

Pixel 7 Pro emulator

### Device operating system

Android 14

### Log output

_No response_

myxmaster:
# Description

This fixes #2713.

Allow cancellation of all pending backend calls (for now only implemented for LND-based backends) or for a specific call using an `AbortSignal` (exemplarily implemented when closing Activity screen).

Todos:
- Add AbortSignals to more backend calls and abort calls when closing other Screens.
- Implement cancellation of pending backend calls for more backends:
  - CLNRest uses the same restReq method as LND and thus we should create a base class for it
  - LightningNodeConnect probably needs an implementation like for Tor requests (throw an error when result is received and call was cancelled) because I think we can't really cancel the requests.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@tradle/react-native-http": "2.0.1",
     "@types/dateformat": "5.0.0",
     "@types/humanize-duration": "3.27.1",
+    "abort-controller": "3.0.0",
     "aez": "1.0.1",
     "assert": "1.5.0",
     "asyncstorage-down": "4.2.0",
--- a/stores/NodeInfoStore.ts
+++ b/stores/NodeInfoStore.ts
@@ -4,7 +4,9 @@ import NodeInfo from '../models/NodeInfo';
 import ChannelsStore from './ChannelsStore';
 import SettingsStore from './SettingsStore';
 import { errorToUserFriendly } from '../utils/ErrorUtils';
-import BackendUtils from '../utils/BackendUtils';
+import BackendUtils, {
+    BackendRequestCancelledError
+} from '../utils/BackendUtils';
 
 import Channel from '../models/Channel';
 
@@ -84,6 +86,10 @@ export default class NodeInfoStore {
                     resolve(nodeInfo);
                 })
                 .catch((error: any) => {
+                    if (error instanceof BackendRequestCancelledError) {
+                        this.reset();
+                        return;
+                    }
                     if (this.currentRequest !== currentRequest) {
                         resolve('Old getNodeInfo call');
                         return;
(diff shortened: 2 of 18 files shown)


kaloudis (changes requested):

package.json
```diff
@@ -47,6 +47,7 @@
     "@tradle/react-native-http": "2.0.1",
     "@types/dateformat": "5.0.0",
     "@types/humanize-duration": "3.27.1",
+    "abort-controller": "^3.0.0",
```
kaloudis, line 50: we should lock the version here

kaloudis (changes requested):

stores/NodeInfoStore.ts
```diff
@@ -84,6 +86,10 @@ export default class NodeInfoStore {
                     resolve(nodeInfo);
                 })
                 .catch((error: any) => {
+                    if (error instanceof BackendRequestCancelledError) {
```
kaloudis, line 89: do all calls in the store not need the abort signal param?

myxmaster:
I think we should pick this up again.

It happens now and then that I start Zeus and it is trying to connect to a backend which I was using last time, but it is not reachable now. I then switch backends while it is trying to connect (which I know won't work). Then I am successfully connected to the other backend, but after a few seconds I still get the red error screen, and at this point I have to restart Zeus(!), because there is no way to dismiss the error (even using the wallet icon in header, then selecting the currently active, and rechable wallet again doesn't help, because Zeus doesn't reconnect to a node to which it is connected already [which makes sense]). Quite annoying.

And of course it is useful for other calls, too. I don't think it is an issue to ship this only for a handful of calls, and then see where it is important and add other calls from time to time. It is better to have at least the most important calls abortable than none.

@kaloudis if you generally agree this concept makes sense, I will work on the rebase.

PR #2795, opened 2025-02-04
fix mispelled license

pony-montana:


kaloudis (approved):
You haven't heard of the Affero Perfect Guy Life v3 license?

jk, nice catch!

PR #2828, opened 2025-02-11
improve minSendableSats handling

### LINKED ISSUE #2821: Allegedly invalid invoice when paying 1 sat to an LN address
### Describe the bug

Got this error with any ln address service when I leave the amount pre-filled to 1. 

No error if I (re)write 1 manually. 

Image

### Reproduce

On the send page fill any LN address > PROCEED > (leave all at it is) > CONFIRM

### ZEUS version

v0.9.6

### Node interface

LND (REST)

### Network

Clearnet

### Device

iPhone 15 Pro

### Device operating system

iOS 18.3

### Log output

```shell

```

myxmaster:
# Description

This fixes #2821.

Before, we were using `lnurl.minSendable` for `amount` only, but it's needed for `satAmount` also, since this is actually used for sending (`amount` is only used for displaying).

Additionally fixed another bug: When unit was set to `₿` or `Fiat` when scanning the lnurlp QR, `lnurl.minSendable` was used without conversion, resulting in `₿ 1` or `$1`.
**Note:** In `Fiat` case it will display `$0.00` in Amount input now, which is correct, but also `₿0.00 000 000` and `0sats`, which is wrong, because this conversion here uses the actual input `$0.00`. It still uses the correct satAmount (1) internally, so when user proceeds it works fine (see last screenshot), but of course we should think about how to handle this better, but not in this PR.

|Before|After|
|-|-|
|grafik|grafik|
|grafik|grafik|

It still uses `1sat`, although displayed Amount was converted to `0sats` in the view before:
grafik

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub
- [ ] [DEPRECATED] Eclair

### Locales

### Other:

### DIFF
--- a/views/LnurlPay/LnurlPay.tsx
+++ b/views/LnurlPay/LnurlPay.tsx
@@ -18,6 +18,7 @@ import LoadingIndicator from '../../components/LoadingIndicator';
 
 import InvoicesStore from '../../stores/InvoicesStore';
 import LnurlPayStore from '../../stores/LnurlPayStore';
+import UnitsStore from '../../stores/UnitsStore';
 
 import LnurlPayMetadata from './Metadata';
 
@@ -29,6 +30,7 @@ interface LnurlPayProps {
     navigation: StackNavigationProp<any, any>;
     InvoicesStore: InvoicesStore;
     LnurlPayStore: LnurlPayStore;
+    UnitsStore: UnitsStore;
     route: Route<'LnurlPay', { lnurlParams: any; amount: any; satAmount: any }>;
 }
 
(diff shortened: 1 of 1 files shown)


shubhamkmr04 (commented):
tACK

kaloudis (approved):

PR #2832, opened 2025-02-12
[swaps] Refund

shubhamkmr04:


locales/en.json
```diff
@@ -1302,5 +1302,6 @@
     "views.SwapDetails.timeoutBlockHeight": "Timeout Block Height",
     "views.SwapDetails.claimPublicKey": "Claim Public Key",
     "views.SwapDetails.swapTree": "Swap Tree",
-    "views.SwapDetails.bip21": "BIP21"
+    "views.SwapDetails.bip21": "BIP21",
```
kaloudis, line 1305: `BIP-21`, though we may use it elsewhere, might be better under general

locales/en.json
```diff
@@ -1302,5 +1302,6 @@
...
+    "views.SwapDetails.refund": "REFUND"
```
kaloudis, line 1306: don't capitalize locals. we can always capitalize them in the view using JS

locales/en.json
kaloudis: actually, this doesn't need to be translated at all. spec names shouldn't be translated

Issue #2905, opened 2025-03-14
Cannot scan QR from photo file


beinardus:
### Describe the bug

When I try to scan a clear QR code of an LNURL invoice (LNURL1...DQRSFGVP) by switching to the photo option on the scanner page and selecting the photo with the QR code, the Zeus window disappears. Then when opening Zeus from the background apps and repeat this again, Zeus disappears again and then an error report comes up:

JavascriptException: TypeError: Cannot read property "decode" of null, js engine: hermes

This is not an issue when I show this same QR code on an external display and use the camera to scan the QR.
Probably the camera uses a different engine to scan the QR.

![Image](https://github.com/user-attachments/assets/f3cee42b-198f-4004-8bef-ee58e616494c)
![Image](https://github.com/user-attachments/assets/8ac1b8a2-9793-4e0e-a11c-4320c3788032)

### Reproduce

WoS -> Receive -> Share -> Download
Zeus -> Scan -> switch to "photo" -> select file -> *BOOM*

### ZEUS version

v0.10.0

### Node interface

LndHub

### Network

Clearnet

### Device

Redme Node 10 5G

### Device operating system

MIUI 14.0.6.0

### Log output

```shell
See screenshot
```

khadar1020:
can I work on this issue

kaloudis:
Thanks for reporting. We've got a build with a fix available for testing here: https://github.com/ZeusLN/zeus/releases/tag/v0.10.1-beta1

PR #2910, opened 2025-03-16
Fix scan QR from photo

shubhamkmr04:
Issue : https://github.com/ZeusLN/zeus/issues/2905

We have updated `QRCodeScanner` logic to fix the issue with the QR code detection when paying invoices using photos. We are now using **RNQRGenerator** from **rn-qr-generator** in place of '@remobile/react-native-qrcode-local-image' to detect the QR code


kaloudis:
Need to test, but code looks good to be on first look.

But why was it breaking, even though we had the dependency pulled in locally, into `zeus_modules`? @shubhamkmr04

shubhamkmr04:
> But why was it breaking, even though we had the dependency pulled in locally, into `zeus_modules`? @shubhamkmr04

Not entirely sure why it was breaking. It might have been some compatibility issues or certain image formats we are saving in the storage.

PR #2943, opened 2025-03-29
fix: Added logic to fetch the fiatRates as the component mounts

Lordhacker756:
# Description
The problem was that the fiat rates aren't being automatically fetched when the component mounts. Instead, the component only shows an error message that requires manual intervention.

Relates to issue: **ZEUS-2492**
Added a method call that fetches the fiatRate once the component mounts

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

- [x] Embedded LND
- [] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Tools/CurrencyConverter.tsx
+++ b/views/Tools/CurrencyConverter.tsx
@@ -84,6 +84,13 @@ export default class CurrencyConverter extends React.Component<
         };
     }
 
+    fetchFiatRates = () => {
+        const { FiatStore } = this.props;
+        FiatStore?.getFiatRates().catch(() => {
+            console.log('Initial fiat rates fetch failed');
+        });
+    };
+
     componentDidMount() {
         this.retrieveInputValues();
         this.addDefaultCurrenciesToStorage();
@@ -91,6 +98,8 @@ export default class CurrencyConverter extends React.Component<
         if (selectedCurrency) {
             this.handleCurrencySelect(selectedCurrency);
         }
+
+        this.fetchFiatRates();
     }
 
     componentDidUpdate(prevProps: CurrencyConverterProps) {
--- a/views/Tools/CurrencyConverter.tsx
+++ b/views/Tools/CurrencyConverter.tsx
@@ -84,13 +84,6 @@ export default class CurrencyConverter extends React.Component<
         };
     }
 
-    fetchFiatRates = () => {
-        const { FiatStore } = this.props;
-        FiatStore?.getFiatRates().catch(() => {
-            console.log('Initial fiat rates fetch failed');
-        });
-    };
-
     componentDidMount() {
         this.retrieveInputValues();
         this.addDefaultCurrenciesToStorage();
@@ -98,8 +91,6 @@ export default class CurrencyConverter extends React.Component<
         if (selectedCurrency) {
             this.handleCurrencySelect(selectedCurrency);
         }
-
-        this.fetchFiatRates();
     }
 
     componentDidUpdate(prevProps: CurrencyConverterProps) {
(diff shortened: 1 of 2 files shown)


kaloudis:
Fiat rates should be loaded when the app starts. It's probably better to figure out why they're not loading than to just add this fix here.

views/Tools/CurrencyConverter.tsx
```diff
@@ -84,13 +84,22 @@ export default class CurrencyConverter extends React.Component<
         };
     }
 
+    fetchFiatRates = () => {
+        const { FiatStore } = this.props;
+        FiatStore?.getFiatRates().catch(() => {
+            console.log('Initial fiat rates fetch failed');
+        });
```
kaloudis, line 91: Why are we catching and consoling this here?

Lordhacker756:
Upon inspecting the issue I found out that 

The FiatStore is initialized as part of the Stores class in `Stores.ts` where the constructor initializes the `fiatStore` as per the settings but it doesn't call the `getFiatRates()`

Hence, even though the **fiatStore** is initialized the rates need to be fetched manually

We can fix it in two ways and I'd like to take your suggestion on which approach to use
1. In the constructor of `fiatStore` we could call the `this.getFiatRates()` after the `this.settingStore = settingStore`
2. We can call the `this.fiatStore.getFiatRates()` after the fiatStore is initialized in the `Stores.ts`

I'll prefer the first approach since it will keep things encapsulated

kaloudis:
> Upon inspecting the issue I found out that
> 
> The FiatStore is initialized as part of the Stores class in `Stores.ts` where the constructor initializes the `fiatStore` as per the settings but it doesn't call the `getFiatRates()`
> 
> Hence, even though the **fiatStore** is initialized the rates need to be fetched manually
> 
> We can fix it in two ways and I'd like to take your suggestion on which approach to use
> 
>     1. In the constructor of `fiatStore` we could call the `this.getFiatRates()` after the `this.settingStore = settingStore`
> 
>     2. We can call the `this.fiatStore.getFiatRates()` after the fiatStore is initialized in the `Stores.ts`
> 
> 
> I'll prefer the first approach since it will keep things encapsulated

I don't think this is strictly true. Often times the rates _are_ fetched when the user has rates enabled and can see their balance denominated in fiat.

Why does it work sometimes and not others?

Lordhacker756:
> Why does it work sometimes and not others?

when `fiatEnabled` is _true_ as you explained (when the user has rates enabled) 
The `Fiatstore.getFiatRates()` is called from the `Wallet.tsx` with this condition
```
        if (fiatEnabled) {
            FiatStore.getFiatRates();
        }
```

And this is the only place which conditionally calls the getFiatRates method, on other screens like `CurrencyConverter.tsx`, `SquarePosPane.tsx`, and `StandalonePosPane.tsx`, the getFiatRates is called on pressing their corresponding **TouchableOpacity**,

So we can see that the getFiatRates is only called, when the `fiatEnabled` is **true**, so in order to tackle this we could either call it each time the FiatStore is initialized or fetch it on the screens where required. 

How'd you suggest to approach this?

kaloudis:
After review, I think this approach is actually sufficient.

Lordhacker756:
Sure, do let me know if any other changes are required for this PR

kaloudis:
nit: next time rebase your fixes that are within the same PR

kaloudis (approved):

Issue #2699, opened 2025-01-02
POS buggy with disabled fiat currency


myxmaster:
### Describe the bug

1. Enable fiat currency and create a POS order. Then go to Settings, diable fiat currency and create another POS order -> POS order view still displays fiat values (probably cached rates?). => We need proper reloading of fiat settings for this case.
2. After disabling fiat currency, kill the app and restart. Create another POS order -> It will display `$N/A` for `Conversion rate` which makes sense, but `Total (Fiat)` is `$NaN` (should be $N/A) and `Total (Bitcoin)` is NaN sats, but that sum should be calculated.

In general I think we should think about checking if fiat is enabled when POS is used ("Open" tab and creating an order, something else?) and if fiat is not enabled, display a modal with options to enable it or leave POS view.

### Reproduce

...

### ZEUS version

v0.9.5

### Node interface

Multiple (specified in comments)

### Network

Clearnet

### Device

Pixel 7 Pro emulator

### Device operating system

Android 14

PR #2957, opened 2025-04-07
[fix] POS Bug: prevent NaN in Total sats

shubhamkmr04:
Somewhat related to https://github.com/ZeusLN/zeus/issues/2699

Previously, the Total (Bitcoin) value was displaying as **NaN**, which caused the Pay button to be disabled. This issue was due to taxSats being calculated using fiat rates, and when fiat rates were disabled, it resulted its value as **NaN**. 

This has now been fixed by updating the taxSats calculation to handle the case when fiat rates are disabled. 
Additionally, the values in fiat are not displayed now(they were $N/A anyways)when fiat rates are not enabled.

Before:
Screenshot_1744033100

After:
Screenshot_1744033110

### DIFF
--- a/views/Order.tsx
+++ b/views/Order.tsx
@@ -114,6 +114,7 @@ export default class OrderView extends React.Component<OrderProps, OrderState> {
         } = this.state;
         const { fiatRates, getRate, getSymbol }: any = FiatStore;
         const { settings } = SettingsStore;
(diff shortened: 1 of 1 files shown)


kaloudis:
Can you link the related issue?

kaloudis (approved):
LGTM

PR #3024, opened 2025-05-05
[swaps] update status while loading wallet

shubhamkmr04:


kaloudis (changes requested):
Instead of fetching the swap data directly into the `SwapsPane` view, we should make the swaps and reverse swaps data observables in the store.

stores/SwapStore.ts
```diff
@@ -375,6 +376,64 @@ export default class SwapStore {
...
+            const reverseSwaps = storedReverseSwaps
+                ? JSON.parse(storedReverseSwaps)
+                : [];
+
+            const allSwaps = [...submarineSwaps, ...reverseSwaps];
+
+            for (const swap of allSwaps) {
+                if (!swap?.id) continue;
+
+                try {
```
kaloudis, line 398: we shouldn't be checking for updates if the swap has already failed or completed

kaloudis (changes requested):

views/Swaps/Refund.tsx
```diff
@@ -105,7 +106,7 @@ export default class RefundSwap extends React.Component<
     updateSwapInStorage = async (swapId: string, txid: string) => {
```
kaloudis, line 106: Let's move this all to the store

views/Swaps/SwapDetails.tsx
```diff
@@ -424,7 +430,7 @@ export default class SwapDetails extends React.Component<
         try {
             let storedSwaps: any;
             const key = isSubmarineSwap ? 'swaps' : 'reverse-swaps';
-            storedSwaps = await EncryptedStorage.getItem(key);
+            storedSwaps = await Storage.getItem(key);
```
kaloudis, line 433: Let's move `updateSwapStatusInStorage` to the store

stores/SwapStore.ts
```diff
@@ -416,4 +412,136 @@ export default class SwapStore {
...
+            if (swapIndex === -1) {
+                throw new Error(`Swap with ID ${swapId} not found`);
+            }
+
+            // Update the swap
+            swaps[swapIndex].status = 'transaction.refunded';
+            swaps[swapIndex].txid = txid;
+
+            // Save the updated swaps back to encrypted storage
+            await Storage.setItem('swaps', JSON.stringify(swaps));
```
kaloudis, line 539: We should also be updating the swaps object in the store, no?

kaloudis (changes requested):

stores/SwapStore.ts
shubhamkmr04: Actually it will be updating by itself when we come back to the pane view

kaloudis (approved):

PR #3159, opened 2025-07-17
feature: Clover POS integration

boumba100:
# Point of Sale: Clover Integration

Claim “Point of Sale: Clover Integration Bounty” https://github.com/ZeusLN/zeus/blob/master/docs/Bounties.md#point-of-sale-clover-integration

## Description

This pull request includes code that integrates the Zeus Wallet Point-of-Sale feature with Clover.

## Features

* Products and prices are completely managed through the Clover management dashboard. When configured, the products and their prices are fetched and displayed by the Zeus POS view.
* Purchases are completely initiated through the Zeus Wallet.
* When an order is created, the Zeus wallet pushes pushes an order to the user’s clover dashboard.
* When a payment has been received, the Zeus wallet automatically marks the associated order as “paid” in the user’s Clover Dashboard.
* Tips are supported

## Configuration

The following values must be configured in the Clover POS Settings view
Screenshot from 2025-07-17 00-25-00

API Token

Reference https://docs.clover.com/dev/docs/generate-a-test-api-token

Merchant ID

Reference https://docs.clover.com/dev/docs/merchant-id-and-api-token-for-development#

### Clover Dashboard - Orders

The following represents orders that were created and paid using the Zeus wallet that are visible in the Clover Dashboard.

### Zeus Clover POS

Configuration View

Screenshot from 2025-07-17 00-25-53

The following illustrates that the clover POS view renders the products that were configured in the Clover dashboard.

Screenshot from 2025-07-17 00-26-22

The following illustrates that the clover view is able to display the “Open” orders. The status of these orders would also be reflected in the Clover dashboard

Screenshot from 2025-07-17 00-26-40

The following illustrates that the paid orders are also made visible in the Zeus wallet. The state of these orders would also be visible in the Clover dashboard

Screenshot from 2025-07-17 00-34-28


locales/en.json
```diff
@@ -912,6 +912,8 @@
     "views.Settings.POS.enablePos": "Enable POS integration",
     "views.Settings.POS.squareAccessToken": "Square Access token",
     "views.Settings.POS.squareLocationId": "Square Location ID",
+    "views.Settings.POS.cloverMerchantId": "Clover Merchant Identifier",
+    "views.Settings.POS.cloverApiToken": "Clover Api Token",
```
kaloudis: nit: API

models/Order.ts
```diff
@@ -14,6 +14,7 @@ export interface LineItem {
     name: string;
     quantity: number;
     base_price_money: BasePriceMoney;
+    productId: string;
```
kaloudis: this is only for Clover orders, so please mark it as an optional property and leave a comment specifying that it's for Clover

stores/PosStore.ts
```diff
@@ -107,6 +108,31 @@ export default class PosStore {
             preimage
         });
 
+        Storage.getItem(`pos-clover-order-${orderId}`).then(
```
kaloudis: I think we want to move this to its own function. `recordCloverPayment` OR add a parameter so we only do this lookup when in Clover mode.

utils/CloverUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+        'GET',
+        `${apiHost}/v3/merchants/${cloverMerchantId}/items?expand=taxRates`,
+        {
+            authorization: `Bearer ${cloverApiToken}`
+        }
+    );
+
+    if (response.info().status != 200) {
+        throw Error('Could not get products from clover');
+    }
```
kaloudis, line 30: does this get reflected in the UI when the error is thrown?

views/Wallet/Wallet.tsx
```diff
@@ -819,6 +820,11 @@ export default class Wallet extends React.Component<WalletProps, WalletState> {
                     {posEnabled === PosEnabled.Square && (
                         <SquarePosPane navigation={navigation} />
                     )}
+                    {posEnabled === PosEnabled.Clover && (
+                        // TODO: Put clover view
+                        // <CloverPosPane navigation={navigation} />
```
kaloudis: You can nuke this comment now

kaloudis (changes requested):

kaloudis:
Multiple tests are failing. Please remedy them.

models/Order.ts
boumba100: I updated the interface. Thanks!

stores/PosStore.ts
boumba100: Good point. I added a separate `recordCloverPayment` function and added some logic so that it is only called when Clover is configured.

utils/CloverUtils.ts
boumba100: I just added some logic in `CloverPosPane.ts` to display an error if it fails to fetch the product list.

Checkout hte following screenshot
Screenshot from 2025-07-17 23-02-58

views/Wallet/Wallet.tsx
boumba100: Good catch :)

shubhamkmr04 (commented):
Screenshot 2025-07-21 at 4 40 50 PM

we should be seeing only of the errors when no products are there

shubhamkmr04 (commented):
The first commit should also pass all the tests.

boumba100 (commented):
* Updated the CloverPosPane to display the correct message.
* Now supports hidden products

boumba100:
I also updated the previous commits to pass the following checks:

* `yarn run tsc`
* `yarn run test`
* `yarn run prettier`

shubhamkmr04 (commented):
If we have set up global tax% for standalone POS, It think it should not reflect in Clover orders. Or maybe add that field for Clover as well if we want to keep the similar tax %

stores/PosStore.ts
```diff
@@ -308,6 +337,69 @@ export default class PosStore {
...
 
+    @action
+    public saveCloverOrder = async (order: Order) => {
+        const { cloverMerchantId, cloverApiToken } =
+            this.settingsStore.settings.pos;
+
+        this.loading = true;
+        this.error = false;
+
+        const apiHost = CloverUtils.getCloverBaseUrl(this.settingsStore);
```
shubhamkmr04, line 348: nit: we can just pass `cloverDevMode` (that is needed) here instead of whole settingsStore

utils/CloverUtils.ts
```diff
@@ -0,0 +1,124 @@
+import SettingsStore from '../stores/SettingsStore';
+import Product, { PricedIn, ProductStatus } from '../models/Product';
+import ReactNativeBlobUtil from 'react-native-blob-util';
+
+const CLOVER_SANDBOX_BASE_URL = 'https://apisandbox.dev.clover.com';
```
shubhamkmr04, line 5: can we shift these to SettingsStore ?

shubhamkmr04 (commented):
We shouldn't show products in the app that are not in stock

Screenshot 2025-08-21 at 7 27 30 PM

shubhamkmr04 (commented):
We don't want to list similar products like this
Simulator Screenshot - iPhone 16 Pro Max - 2025-08-21 at 17 26 08

This would be the better approach
Screenshot 2025-08-21 at 5 29 16 PM

shubhamkmr04 (commented):
We should have a different message in case of empty search
Screenshot 2025-08-21 at 7 04 08 PM

shubhamkmr04 (commented):
Maybe we can remove **Uncategorized** heading from here as the categories don't get fetched along with products.
Screenshot 2025-08-21 at 7 09 40 PM

myxmaster:
@boumba100 still there? Would be cool to see you get the bounty :)

boumba100:
> @boumba100 still there? Would be cool to see you get the bounty :)

Hi, sorry for the super late reply on this.

I just got caught up with other stuff and have not been able to find the time to work on this.

myxmaster:
No worries, of course. Just wondering if you still plan to finish this. Would be worth it I think, lots of effort went into this already.

PR #3166, opened 2025-07-20
Update SettingsStore.ts

Michilis:
Replaced nos.lol with relay.cashumints.space, 
a Nostr relay specialized in Cashu mints and reviews.


kaloudis:
Why did you remove nos.lol?

Michilis:
Nos.lol isn’t used by either bitcoinmints.com or cashumints.space, so there’s no reason to pull Cashu data from it. Unless it’s for other stuff?

kaloudis:
> Nos.lol isn’t used by either bitcoinmints.com or cashumints.space, so there’s no reason to pull Cashu data from it. Unless it’s for other stuff?

this is not the only use case for the Nostr relays

PR #3139, opened 2025-07-07
Displaying rounded off amounts with (~) and exact amounts

cherry-1729-9090:
# Description

Relates to issue: #2786 

**Activity View (List):**

- Added `~`  symbol prefix when Lightning fees are rounded (e.g., `~ 6 sats fee` when original fee was 5.89 sats)
- Uses proper mathematical rounding (0.5 rounds up)
- Only shows "~" for actual rounded amounts, not for whole numbers

**Detail Views (Payment, Invoice, PaymentRequest):**

- Display exact amounts with full msat precision
- No rounding indicators since these views have sufficient space for precise values
- Added `showExactAmount` prop to Amount components in detail views

**Fiat Currency Handling:**

- Fiat amounts never show "~" rounding indicator (as requested)
- Continue to use "≈" for approximation in non-exact contexts
- In exact mode, fiat amounts show without approximation symbols

Before:

- Activity: [6 sats fee](vscode-file://vscode-app/c:/Users/Devi_Charan/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) (rounded, but no indication)
- Detail: [6 sats fee](vscode-file://vscode-app/c:/Users/Devi_Charan/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) (rounded, but no indication)

After:

- Activity: ~6 sats fee (clearly indicates rounding from 5.89)
- Detail: 5.89 sats fee (shows exact precision)

This enhancement improves user understanding of fee calculations while maintaining the clean interface design. The "~" symbol provides immediate visual feedback when amounts have been rounded for display purposes, while detail views offer full precision for users who need exact values.

**Before :** 
image
image

**After :**

image
image
image

This pull request is categorized as a:

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [ ] Yes
[...]

PR #3228, opened 2025-08-21
fix: add approximate symbol for fiat currencies and fix RTL placement

cherry-1729-9090:
# Description

Relates to issue: **ZEUS- #3139 **

This PR fixes RTL (right-to-left) layout issues in the Amount component that were introduced after PR #3179 was merged. The rounding indicator feature inadvertently broke the RTL/LTR logic and caused symbol misplacement issues.

## Changes Made:

1. **Restored approximate symbol (≈) for all fiat currencies** - Since fiat amounts are inherently approximate due to exchange rate volatility, all fiat displays now show the ≈ symbol to indicate this uncertainty.

2. **Fixed RTL symbol placement** - Restored proper RTL/LTR conditional rendering paths that were consolidated incorrectly in #3139 , ensuring currency symbols and indicators appear on the correct side based on text direction.

image
image

image
image
image
image

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/Amount.tsx
+++ b/components/Amount.tsx
@@ -201,83 +201,73 @@ function AmountDisplay({
             </>
         );
 
-        if (rtl) {
-            return (
-                <Row
-                    style={styles.row}
-                    accessible={accessible}
-                    accessibilityLabel={accessibilityLabel}
-                >
-                    <View style={styles.textContainer}>
-                        {feeSection}
-                        {unit === 'BTC' && roundAmount && <RoundingIndicator />}
-                        {unit === 'fiat' && <ApproximateSymbol accessible />}
-                        {unit === 'BTC' && amount !== 'N/A' && (
-                            <FiatSymbol accessible />
-                        )}
-                        {space ? <TextSpace /> : <Spacer width={1} />}
-                        <Body
-                            jumbo={jumboText}
-                            defaultSize={defaultTextSize}
-                            color={color}
-                            colorOverride={colorOverride}
-                            accessible={accessible}
-                        >
-                            {negative ? '-' : ''}
-                            {amount === 'N/A' && fiatRatesLoading ? (
-                                <LoadingIndicator size={20} />
-                            ) : unit === 'BTC' ? (
-                                formatBitcoinWithSpaces(amount)
-                            ) : (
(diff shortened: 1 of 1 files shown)


components/Amount.tsx
```diff
@@ -232,21 +201,83 @@ function AmountDisplay({
...
+                                formatBitcoinWithSpaces(amount)
+                            ) : (
+                                amount.toString()
+                            )}
+                        </Body>
+                        {feeSection}
+                    </View>
+                </Row>
+            );
+        }
```
kaloudis: I think the position of things is technically correct in this PR, but we are repeating a bunch of things.

I think it could be simpler, if we just had conditional blocks for `rtl` on the few lines that are placed differently OR componentize the big blocks that are repeated.

kaloudis (changes requested):

components/Amount.tsx
```diff
@@ -241,7 +249,25 @@ function AmountDisplay({
                 {!rtl && pending && <Pending />}
                 <View style={styles.textContainer}>
                     {rtl && feeSection}
-                    {commonContent}
+                    {indicators}
+                    {rtl ? (
+                        <>
+                            {unit === 'BTC' && amount !== 'N/A' && (
+                                <FiatSymbol accessible />
+                            )}
```
kaloudis: BTC isn't rtl, so it shouldn't be in this block. It'll never be executed.

components/Amount.tsx
```diff
@@ -241,7 +249,25 @@ function AmountDisplay({
...
+                            {spacer}
+                            {amountContent}
+                            {amount !== 'N/A' && unit === 'fiat' && (
```
kaloudis: Can't `symbols` be used here instead?

PR #3241, opened 2025-08-25

shubhamkmr04:


kaloudis:
We should be leveraging the info on the `NodeInfo` mode: `isTestNet || isRegTest || isSigNet` - or perhaps we need to add an `isMainNet` and check if it's not a hit?

kaloudis (approved):
LGTM

PR #3256, opened 2025-09-02

shubhamkmr04:


views/Settings/Support.tsx
```diff
@@ -52,7 +58,11 @@ function Support(props: SupportProps) {
                 navigation={navigation}
             />
             <FlatList
-                data={ABOUT_ITEMS}
+                data={ABOUT_ITEMS.filter(
+                    (item) =>
+                        item.path !== 'DonateToZEUS' ||
+                        nodeInfoStore.nodeInfo.isMainNet
```
kaloudis: this one sort of hurts my head. Can we do something like:

`!(item.path === 'DonateToZEUS' && !nodeInfoStore.nodeInfo.isMainNet)`

or alternatively, we can add a condition field to the ABOUT_ITEMS objects called `condition` and check if it is true, if not filter it out. condition for `DonateToZEUS` can be `nodeInfoStore.nodeInfo.isMainNet`

shubhamkmr04: Yes definitely. I'll just update that

views/Settings/Support.tsx
```diff
@@ -58,11 +61,7 @@ function Support(props: SupportProps) {
                 navigation={navigation}
             />
             <FlatList
-                data={ABOUT_ITEMS.filter(
-                    (item) =>
-                        item.path !== 'DonateToZEUS' ||
-                        nodeInfoStore.nodeInfo.isMainNet
-                )}
+                data={ABOUT_ITEMS.filter((item) => item.condition ?? true)}
```
kaloudis: I think we can change it to `(item) => !!item.condition ? item.condition : true)` and that way we don't need to set `condition: true` on all the other items

kaloudis (changes requested):

views/Settings/Support.tsx
shubhamkmr04: I think it won't filter out the donate option for testnet if we do this.

kaloudis: `(item) => item.condition !== undefined ? item.condition : true)`

PR #3264, opened 2025-09-09

shubhamkmr04:


views/DonateToZEUS.tsx
```diff
@@ -71,10 +71,14 @@ export default class DonateToZEUS extends React.Component<
                                 color: themeColor('background')
                             }}
                         />
-                        <CopyButton
-                            title={localeString('views.Receive.copyAddress')}
-                            copyValue={'tips@pay.zeusln.app'}
-                        />
+                        <View style={{ paddingTop: 4 }}>
```
kaloudis: Why doesn't this use the styles of `button`?

shubhamkmr04: The paddingTop we use for the buttons group is greater than what we paddingTop for second button

kaloudis: OK, let's revert the changes on this file and instead only increase the padding on the `button` style

kaloudis: OK, let's remove this `View` wrapper around the `CopyButton` and instead only increase the padding on the `button` style

kaloudis (approved):

Issue #3310, opened 2025-10-09
Choose Payment Method view: display balances


kaloudis:
Let's display the balance on the select payment method view so that users can make more informed decisions about which method they select.

PR #3317, opened 2025-10-14
enhancement: Display balances in payment method selection view

ajaysehwal:
# Description

Relates to issue: [ZEUS-3310](https://github.com/ZeusLN/zeus/issues/3310)

#### Before changes
Screenshot 2025-10-14 at 10 03 52 AM

#### After changes
Simulator Screenshot - iPhone 16 Pro Max - 2025-10-15 at 20 15 56

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

- [ ] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/LayerBalances/PaymentMethodList.tsx
+++ b/components/LayerBalances/PaymentMethodList.tsx
@@ -33,6 +33,13 @@ interface PaymentMethodListProps {
     lightningBalance?: number | string;
     onchainBalance?: number | string;
     ecashBalance?: number | string;
+    accounts?: Array<{
+        name: string;
+        balance: number;
+        XFP?: string;
+        watch_only?: boolean;
+        hidden?: boolean;
+    }>;
 }
 
 //  To toggle LTR/RTL change to `true`
@@ -43,6 +50,8 @@ type DataRow = {
     subtitle?: string;
     disabled?: boolean;
     balance?: number | string;
+    account?: string;
+    hidden?: boolean;
 };
 
 const Row = ({ item }: { item: DataRow }) => {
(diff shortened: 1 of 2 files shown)


components/LayerBalances/PaymentMethodList.tsx
```diff
@@ -29,6 +30,9 @@ interface PaymentMethodListProps {
     ecash?: string;
     offer?: string;
     lnurlParams?: LNURLWithdrawParams | undefined;
+    lightningBalance?: number | string;
+    onchainBalance?: number | string;
+    ecashBalance?: number | string;
```
kaloudis, line 35: Not sure this approach will work. It's possible to have multiple on-chain/external account layers. Need to think about if those should be listed separately here of it's ok as is.

ajaysehwal: @kaloudis  Should we also show all external accounts in the “Select Payment Method” view? We can display only the accounts that are not hidden/watch-only and show their balances separately for each account

kaloudis: > @kaloudis Should we also show all external accounts in the “Select Payment Method” view? We can display only the accounts that are not hidden/watch-only and show their balances separately for each account

Yes, I think this would be ideal.

PR #3383, opened 2025-11-20
fix: include routing hints when autogenerating 0 amount invoice

kaloudis:
# Description

When auto-generating invoices (hitting the `Request` button on the `Keypad` view) without an amount, we inform the user that the LSP has been disabled and that their pubkey will be exposed, because the LSP does not support 0-amount invoices. When doing so, we had a bug that would default `Routing hints` to false. In doing so, this became problematic for user that only had unannounced channels (eg. Olympus LSP channels), because this would lead to the user unable to be paid on Lightning.

This PR fixes the logic for the `routeHints` bool param in case of auto-generation and makes it uniform with the rest of the `Receive` view.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

- [x] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -556,7 +556,7 @@ export default class Receive extends React.Component<
                     : expirySeconds || '3600',
                 ampInvoice: lspIsActive ? false : ampInvoice || false,
                 blindedPaths: lspIsActive ? false : blindedPaths || false,
-                routeHints: lspIsActive ? false : routeHints || false,
+                routeHints,
                 addressType: BackendUtils.supportsAddressTypeSelection()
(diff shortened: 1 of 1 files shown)


shubhamkmr04 (approved):
tACK

PR #3437, opened 2025-12-15
docs: clarify dependency installation for npm users

divyansh-v15-06:
###Problem

While setting up the ZEUS development environment on Windows, npm install fails with an ERESOLVE dependency conflict. The issue occurs due to incompatible peer dependency requirements between:

nostr-tools@1.x (used by ZEUS)

@nostr-dev-kit/ndk@2.x (expects nostr-tools@^2.7.1)

This results in installation failure for contributors using newer npm versions, especially on Windows.

Additionally, some build scripts rely on Unix-style environment variable syntax (e.g. NODE_ENV=production), which does not work on Windows shells.

###Solution

This PR updates the README to document a known and safe workaround for contributors:

Use npm install --legacy-peer-deps to bypass strict peer dependency resolution

Clarifies that this is currently required on Windows due to upstream dependency constraints

No runtime code changes are introduced.

###Why this change

Improves onboarding for new contributors, especially Windows users

Prevents confusion when setup fails during dependency installation

Aligns documentation with real-world developer experience

Testing

Reproduced the issue on Windows 10

###Verified successful installation using:

**(npm install --legacy-peer-deps)


kaloudis:
I think we can more concisely mention `yarn` and install instructions in the Prerequisites section

divyansh-v15-06:
> I think we can more concisely mention `yarn` and install instructions in the Prerequisites section

Thanks for the suggestion @kaloudis 👍

I’ve updated the README to mention yarn and npm install instructions more concisely in the Prerequisites section, keeping it brief and avoiding duplication.

Let me know if this looks good or if you’d prefer it structured differently.

divyansh-v15-06:
> I think we can more concisely mention `yarn` and install instructions in the Prerequisites section

Thanks for the suggestion @kaloudis 🙏

I’ve moved the changes to a dedicated branch (windows-dev-docs) and updated the README to keep the yarn/npm install notes concise in the Prerequisites section as suggested.

Please let me know if this looks better now

PR #3462, opened 2025-12-19
Reproducible build optimizations

kaloudis:
## What These Fixes Address

  | Fix                          | Addresses                                   |
  |------------------------------|---------------------------------------------|
  | org.gradle.parallel=false    | Non-deterministic task execution order      |
  | reproducibleFileOrder=true   | ZIP file entry ordering                     |
  | preserveFileTimestamps=false | Variable timestamps in archive              |
  | SOURCE_DATE_EPOCH            | Timestamps in JS bundle and other artifacts |

## Trade-off

  Disabling parallel builds will increase build time (potentially 2-3x slower). If build speed is critical, you could alternatively keep parallel builds and use a post-processing step to reorder APK contents deterministically using tools like strip-nondeterminism or custom APK re-packaging.

 ## Testing

  Run ./build.sh several times and compare the SHA256 hashes. They should now match consistently.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -169,6 +169,12 @@ android {
         }
     }
 
+    // Reproducible build settings
+    tasks.withType(Zip).configureEach {
+        reproducibleFileOrder = true
+        preserveFileTimestamps = false
+    }
+
     sourceSets {
         main {
             proto {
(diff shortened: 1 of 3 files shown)


kaloudis:
tACK - hashes match non-racy v0.12.0 version bump (prev commit) outputs

Issue #3517, opened 2026-01-10
ui: `Switch` component has different styles


kaloudis:
After the React Native upgrade, some of our `Switch` components have a white head when enabled, others have a highlight-color head that matches the rail when enabled.

Let's investigate this and make the styles uniform.

kaloudis:
I think this is just an iOS feature, where changed tabs use the highlight color and unchanged tabs use white.

PR #3550, opened 2026-01-16
UI: unify Switch component usage in LNURL channel view

pradip-1111:
Description

Relates to issue: #3517 – UI: Switch component has different styles

This PR resolves a UI inconsistency introduced after the React Native upgrade where some switches used platform-default styling while others followed Zeus theme colors.

What was changed

Replaced the native React Native Switch in views/LnurlChannel.tsx with the project’s custom Switch wrapper.

Ensured the custom Switch explicitly defines thumb and track colors so enabled switches consistently follow the Zeus highlight theme.

Removed local style overrides to rely on the unified defaults provided by the shared component.

Result

All Switch components now use the same custom wrapper and display consistent, theme-aware behavior across the application, avoiding platform-specific defaults

### DIFF
--- a/views/LnurlChannel.tsx
+++ b/views/LnurlChannel.tsx
@@ -1,7 +1,7 @@
 import url from 'url';
 import * as React from 'react';
 import ReactNativeBlobUtil from 'react-native-blob-util';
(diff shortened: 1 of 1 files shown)


kaloudis:
Can you provide before and after screenshots?

PR #3588, opened 2026-01-24
Add PPM toggle for channel fees

### LINKED ISSUE #1411: PPM for channel fees
### Describe the feature

When setting channel fees, allow user to specify rates in PPM

kunal-595:
fixes #1411
Adds a toggle to switch between percentage and PPM input for fee rate when setting channel fees. Only shows for LND-based backends since CLN already uses PPM natively

### DIFF
--- a/components/FeeBreakdown.tsx
+++ b/components/FeeBreakdown.tsx
@@ -7,6 +7,7 @@ import LoadingIndicator from '../components/LoadingIndicator';
 
 import ChannelsStore from '../stores/ChannelsStore';
 import NodeInfoStore from '../stores/NodeInfoStore';
+import SettingsStore from '../stores/SettingsStore';
 
 import DateTimeUtils from '../utils/DateTimeUtils';
 import { localeString } from '../utils/LocaleUtils';
@@ -20,6 +21,7 @@ import KeyValue from './KeyValue';
 interface FeeBreakdownProps {
     ChannelsStore?: ChannelsStore;
     NodeInfoStore?: NodeInfoStore;
+    SettingsStore?: SettingsStore;
     channelId: string | any;
     channelPoint: string;
     peerDisplay?: string | any;
@@ -34,7 +36,7 @@ interface FeeBreakdownProps {
     label?: string;
 }
 
-@inject('ChannelsStore', 'NodeInfoStore')
+@inject('ChannelsStore', 'NodeInfoStore', 'SettingsStore')
 @observer
 export default class FeeBreakdown extends React.Component<
     FeeBreakdownProps,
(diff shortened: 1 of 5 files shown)


kaloudis:
I think we'd want to set up some sort of utility for this in `FeeUtils` with corresponding tests

kunal-595:
Sounds good, i will move that PPM → % logic into FeeUtils, add some unit tests, and update the store, pushing shortly.

kaloudis:
Looking mostly solid so far. Even if CLNRest uses PPM by default, you should be able to toggle between the two modes with it.

I think the next step after that would be being able to toggle between the units on the `FeeBreakdown` component

kunal-595:
That makes sense, i will hook that up in FeeBreakdown so users can switch modes on the fly, i will ping you here once i have pushed the update to the branch.

kaloudis:
@gemini-code-assist /review

kaloudis:
Your second to last commit message looks funky - consider rebasing or squashing the last two together

kunal-595:
@kaloudis I have squashed the commits and updated the commit message, please let me know if this looks good.

components/SetFeesForm.tsx
```diff
@@ -140,6 +146,34 @@ export default class SetFeesForm extends React.Component<
                     autoCorrect={false}
                 />
 
+                {(BackendUtils.isLNDBased() ||
+                    implementation === 'cln-rest') && (
```
kaloudis, line 150: is this check explicitly required? if so, why?

stores/FeeStore.ts
```diff
@@ -146,7 +148,12 @@ export default class FeeStore {
 
         // handle commas in place of decimals
         const baseFee = newBaseFee.replace(/,/g, '.');
-        const feeRate = newFeeRate.replace(/,/g, '.');
+        let feeRate = newFeeRate.replace(/,/g, '.');
+
+        // convert PPM to decimal rate for LND
```
kaloudis, line 153: perhaps note here that CLN uses PPM

components/FeeBreakdown.tsx
```diff
@@ -59,6 +62,8 @@ export default class FeeBreakdown extends React.Component<
         } = this.props;
         const { loading, chanInfo } = ChannelsStore!;
         const { nodeInfo, testnet } = NodeInfoStore!;
+        const feeRateDisplayMode =
```
kaloudis, line 65: We should be able to toggle between PPM or % display mode for both LND and CLN backends

kaloudis (changes requested):

PR #3595, opened 2026-01-25
feat: allow selecting multiple fiat currencies #3384

### LINKED ISSUE #3384: Allow selecting multiple fiat currencies
### Describe the feature

I need to make payments denominated in multiple fiat currencies. It would be nice if I could select multiple, so I don't have to go somewhere else to do the conversion.

aryanraj45:
## Description
This PR addresses issue #3384 by allowing users to select multiple fiat currencies simultaneously. 

### Key Changes:
- **SettingsStore**: Added `fiats` array to persist multiple currency selections.
- **SelectCurrency View**: Updated the UI to allow multi-select toggling with checkmarks.
- **FiatStore**: Updated fetching logic to retrieve rates for all selected currencies (supports Yadio and Zeus sources).
- **Conversion Component**: Refactored to loop through and display all active fiat conversions in the balance panes and headers.
- **AmountUtils**: Extended unit handling to support specific fiat codes (e.g., 'USD', 'INR') alongside 'sats' and 'BTC'.

## Why this is useful?
Users no longer need to switch settings back and forth to see their balance in different currencies. This is particularly useful for users dealing with multiple regional currencies or stablecoins.

## Visuals
| Currency Selection |
|---|---|

https://github.com/user-attachments/assets/1f10cd13-e1e9-48a2-ad7b-331c9b43dc1b

## Fixes
Fixes #3384

### DIFF
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1723,6 +1723,11 @@ export default class SettingsStore {
                 if (parsedSettings.preferredFiatIndex === undefined) {
                     parsedSettings.preferredFiatIndex = 0;
                 }
+                // Consolidate: fiat is always derived from fiats[preferredFiatIndex]
+                parsedSettings.fiat =
+                    parsedSettings.fiats[parsedSettings.preferredFiatIndex] ||
+                    parsedSettings.fiats[0] ||
+                    DEFAULT_FIAT;
                 this.settings = parsedSettings;
             } else {
                 console.log('attempting to load legacy settings');
@@ -1775,6 +1780,13 @@ export default class SettingsStore {
             ...existingSettings,
             ...newSetting
         };
+
+        // Consolidate: keep fiat in sync with fiats[preferredFiatIndex]
+        if (newSetting.fiats || newSetting.preferredFiatIndex !== undefined) {
+            const fiats = newSettings.fiats || [DEFAULT_FIAT];
+            const index = newSettings.preferredFiatIndex ?? 0;
+            newSettings.fiat = fiats[index] || fiats[0] || DEFAULT_FIAT;
+        }
 
         if (
             newSetting.pos?.posEnabled &&
--- a/views/Settings/SelectCurrency.tsx
+++ b/views/Settings/SelectCurrency.tsx
@@ -156,9 +156,9 @@ export default class SelectCurrency extends React.Component<
                                             newFiats.push(item.value);
                                         }
 
+                                        // fiat is auto-synced via updateSettings
                                         await updateSettings({
-                                            fiats: newFiats,
-                                            fiat: newFiats[0] || item.value
+                                            fiats: newFiats
                                         }).then(() => {
                                             getSettings();
                                         });


android/gradle/wrapper/gradle-wrapper.properties
ajaysehwal, line 1: @aryanraj45, you don’t need to change these files, so please remove this change.

aryanraj45: Thanks for the review, @ajaysehwal!

I've removed the accidental change to gradle-wrapper.properties.
I also did a quick pass to clean up the code and improve type safety (removed unnecessary non-null assertions).

ajaysehwal:
@aryanraj45, please run **yarn verify** locally before pushing your changes

aryanraj45:
@ajaysehwal  Done! I've fixed all 65+ linting and formatting errors identified by yarn verify.

Cleaned up unused variables in 
Wallet.tsx
.
Updated 
AmountUtils.test.ts
 to align with the new logic (it now correctly expects '0' instead of 'Disabled' when rates are unavailable).
Fixed formatting across all modified files.
yarn verify is now passing locally with 0 errors

aryanraj45:
@ajaysehwal  I've squashed the cleanup and refactor updates into a single commit as requested. The PR now contains exactly two commits:

The original feature request.
The cycle logic refactor + linting/type fixes.

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.0.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
```
kaloudis: why was this changed?

components/UnitToggle.tsx
```diff
@@ -19,8 +19,11 @@ interface UnitToggleProps {
 export default class UnitToggle extends React.Component<UnitToggleProps, {}> {
     render() {
         const { UnitsStore, SettingsStore, onToggle } = this.props;
-        const { changeUnits, units } = UnitsStore!;
-        const { settings } = SettingsStore!;
+
+        if (!UnitsStore || !SettingsStore) return null;
```
kaloudis: why is this change required?

utils/AmountUtils.test.ts
```diff
@@ -450,7 +450,7 @@ describe('AmountUtils', () => {
                 expect(result.error).toBe('Error fetching fiat rates');
             });
 
-            it('returns error when rate for currency is not available', () => {
+            it('returns 0 when rate for currency is not available', () => {
```
kaloudis: why does this behavior need to be changed?

utils/UnitsUtils.ts
```diff
@@ -73,7 +73,7 @@ const formatBitcoinWithSpaces = (x: string | number) => {
     return `${integerFormatted}.${firstTwo} ${rest}`.trim();
 };
 
-type Units = 'sats' | 'BTC' | 'fiat';
+type Units = 'sats' | 'BTC' | 'fiat' | string;
```
kaloudis: not a fan of this approach. Do we maybe need a type for our fiat strings?

views/Wallet/KeypadPane.tsx
kaloudis, line 1: Why do we need to make changes to this view? Shouldn't everything be covered by the `Conversion` and `UnitToggle` components?

utils/AmountUtils.test.ts
aryanraj45: This change prevents a visual flicker when switching currencies. When switching from one fiat to another, there's a brief moment where the new rate isn't loaded yet. Displaying 'Disabled' or 'N/A' causes the UI to flash, which is jarring. Returning '0' with the correct symbol ensures a smooth transition until the actual rate loads

aryanraj45:
@kaloudis Thanks for the detailed review! I've addressed all the feedback:

Reverted gradle-wrapper.properties (accidental change)
Reverted 
UnitToggle.tsx
 (removed unnecessary defensive code)
Fixed 
KeypadPane.tsx
 - removed tap-to-cycle from main amount, now only the orange button triggers currency cycling

metro_logs.txt
kaloudis, line 1: Why? Please be mindful of files you commit

utils/AmountUtils.test.ts
kaloudis: I think this approach can cause further issues. You should find another approach.

This wasn't an issue before with single fiat functionality.

stores/SettingsStore.ts
```diff
@@ -168,6 +168,8 @@ export interface Settings {
     authenticationAttempts?: number;
     fiatEnabled?: boolean;
     fiat?: string;
```
kaloudis, line 170: no need to have both `fiat` and `fiats` - best to consolidate them. Probably best to keep `fiat` and make it backwards compatible

aryanraj45: @kaloudis done Consolidated [fiat](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [fiats](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - [fiat](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is now automatically derived from [fiats[preferredFiatIndex]](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

components/Conversion.tsx
```diff
@@ -60,9 +60,13 @@ export default class Conversion extends React.Component<
         const { showRate } = this.state;
         const { units } = UnitsStore!;
         const { settings } = SettingsStore!;
-        const { fiatEnabled } = settings;
+        const { fiatEnabled, fiats, preferredFiatIndex, fiat } = settings;
+
+        // Get current fiat from multi-select array or fallback to single fiat
+        const currentFiat = fiats?.[preferredFiatIndex] || fiat || 'USD';
```
kaloudis, line 66: Potential out of bounds error here if `preferredFiatIndex` exceeds the `fiats.lengths` - need detection for this and a fallback

components/UnitToggle.tsx
```diff
@@ -12,16 +12,36 @@ interface UnitToggleProps {
...
-        const { UnitsStore, SettingsStore, onToggle } = this.props;
+        const { UnitsStore, SettingsStore, onToggle, amount } = this.props;
         const { changeUnits, units } = UnitsStore!;
-        const { settings } = SettingsStore!;
-        const { fiat } = settings;
+        const { settings, cycleFiat } = SettingsStore!;
+        const { fiat, fiats } = settings;
+
+        // Check if amount is effectively zero
+        const isAmountZero = !amount || amount === '0' || amount === '';
```
kaloudis, line 28: This doesn't catch values like '0.0', '0.00', etc. Consider:
  `const isAmountZero = !amount || parseFloat(amount) === 0;`

kaloudis:
Generally concerned about the changes here as you are a new contributor, there is complexity, and there are no tests added. 

I think you could potentially create a `FiatUtils` where a lot of the complex functionality is consolidated and tests are added.

Funcs could potentially cover:
- Fiat resolution ie. returning which is the currently selected fiat currenct (this one appears to be duplicated in a bunch of places)
- Fiat index calculation
- An amount zero check (maybe this lands in `AmountUtils`?)
- Toggling currencies on the `SelectCurrency` view
- The setting consolidation of `fiat` to `fiats`

kaloudis:
There are too many complexities and UI issues with this approach. Going to close in favor of https://github.com/ZeusLN/zeus/pull/3683 and https://github.com/ZeusLN/zeus/pull/3638

Appreciate the effort here though. Don't be discouraged

Issue #3497, opened 2026-01-03
Improve CSV export


makeasnek:
### Describe the feature

Long time Zeus user, love the wallet, here's some feedback from my efforts to pay taxes this year. I've already done all my tax stuff, so this aren't improvements I'm needing at the present time, but would be very helpful for next year and/or other people currently doing their taxes.

Improvement request #1 :
Tools -> CSV export only exports onchain transactions, misses all the lightning stuff entirely. Please make it export everything.

Improvement request #2:
When doing export from main transaction list (includes on-chain and LN), three separate files are created. A merged file would be easier to import into various accounting tools.

Improvement request #3:
While circular rebalances are noted with a note field (yay!), there is no way afaict to distinguish channel opens/closes from regular on-chain transactions. This is important for tax reporting and accounting purposes. Without a way to distinguish, people may be paying double gains/losses on these transactions.

myxmaster:
1) Tools -> Export Activity to CSV: You have 3 export options here: Invoices, Payments (this is LN), Transactions (this is on-chain). Mabye we should call it "Lightning payments" and "On-chain transactions" @kaloudis ?

2) Hm, I actually agree that it would be better to export just one file always. We could change the UI so the user can check/uncheck what to export, and then Zeus will always generate just 1 file including everything, what do you think @kaloudis ?
Of course we would have to reorganize the CSV structure for this, but we need to improve this anyway since it is kind of messy: <img width="1258" height="345" alt="Image" src="https://github.com/user-attachments/assets/b9f2a763-ff3d-45f2-ad53-23a74d457cef" />


3) That would definitely be nice, but not sure if there is a way to distinguish on-chain sends from channel opens/closes, @kaloudis ?

kaloudis:
1) I'm ok with On-chain transactions, Lightning payments, Lightning invoices

2) I don't like how you've organized it in the example above. I think if we want to consolidate things, you can put them on different pages potentially. But that might be a feature limited to Excel/Numbers files, and not pragmatic for CSVs. I generally think here we'd be adding complexity for not so much gain

3) worth looking at https://github.com/ZeusLN/zeus/issues/2444 - would really want to do the same for closes too. Though there is a lot of complexity here as the behavior varies between implementations too

myxmaster:
1. 👍 
[...]

PR #3597, opened 2026-01-25
Export Activity to CSV: rephrasing for clarity

myxmaster:
# Description

Relates to https://github.com/ZeusLN/zeus/issues/3497#issuecomment-3795672324

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


myxmaster:
Ok, I wasn't aware / thinking about that.

I guess the easiest solution is indeed to call it "Export Lightning & Cashu Invoices" and "Export Lightning & Cashu Payments" then?

shubhamkmr04 (changes requested):
Let's also change the file names as per new locales ?  we can follow "zeus_`timestamp`_ln_invoices.csv" from both tools view and from Activity view

myxmaster:
Right, filenames should be updated as well, good point. But are you guys ok with "Export Lightning & Cashu Invoices" instead of "Export Lightning Invoices" (+ same for payments)?

(I'm still thinking it would be way better to export everything into 1 file, but that requires some reorg of the columns and will take some time. So even if we go that route, until then I still think this change here makes sense and avoids misunderstandings.)

myxmaster:
After some more thoughts on this, I tend to not updating the locale strings (=close this PR).

While it does add some clarity (users will understand better which types of transactions/payments will be exported), it will still be confusing to e.g. read "Cashu", when you are not using Cashu at all. And it would be too complicated to make this conditional, since this is a temporary/quick fix anyway.

So maybe let's just keep it as is, and I will try and come up with a proposal to export everything into 1 file.

Issue #2444, opened 2024-09-29
Opening a channel & note in transaction


nbenaglia:
### Describe your enhancement idea

When a new channel is opened, the related payement (in Activity list) should have already written a payment note (ex. "Opened channel").

kaloudis:
Might be able to do this using the `pending_chan_id` in the `OpenChannel` response with LND. Worth investigating.

aryanraj45:
When opening a Lightning channel, the funding transaction now automatically gets a note 'Opened channel' in the Activity list for better transaction tracking.

Implementation:
- Import NotesStore in ChannelsStore
- Pass NotesStore to ChannelsStore constructor
- Save note after successful openChannelSync using funding_txid_str
- Added locale string views.OpenChannel.openedChannelNote

Fixes #2444"

PR #3641, opened 2026-02-01
Automatically add note when opening a Lightning channel

aryanraj45:
# Description

Relates to issue: **#2444**

When opening a Lightning channel, automatically save a note "Opened channel" for the funding transaction in the Activity list.

**Implementation:**
- Import and pass NotesStore to ChannelsStore
- After successful openChannelSync, save note with funding_txid_str
- Added locale string views.OpenChannel.openedChannelNote

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

---

**If there are any suggestions or changes needed, please let me know!**

### DIFF
--- a/stores/ChannelsStore.ts
+++ b/stores/ChannelsStore.ts
@@ -14,10 +14,13 @@ import OpenChannelRequest from '../models/OpenChannelRequest';
 import CloseChannelRequest from '../models/CloseChannelRequest';
 
 import SettingsStore from './SettingsStore';
+import NotesStore from './NotesStore';
 
 import BackendUtils from '../utils/BackendUtils';
+import Base64Utils from '../utils/Base64Utils';
 import { localeString } from '../utils/LocaleUtils';
 import { errorToUserFriendly } from '../utils/ErrorUtils';
+import Storage from '../storage';
 
 interface ChannelInfoIndex {
     [key: string]: ChannelInfo;
@@ -115,9 +118,11 @@ export default class ChannelsStore {
     @observable public haveAnnouncedChannels = false;
 
     settingsStore: SettingsStore;
+    notesStore: NotesStore;
 
-    constructor(settingsStore: SettingsStore) {
+    constructor(settingsStore: SettingsStore, notesStore: NotesStore) {
         this.settingsStore = settingsStore;
+        this.notesStore = notesStore;
 
         reaction(
             () => this.channelRequest,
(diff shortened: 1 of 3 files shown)


kaloudis:
Which wallet interfaces have you tested it against? Please don't remove items from the PR template in the future.

aryanraj45:
@kaloudis I attempted to test on testnet with Embedded LND but ran into environment issues with the embedded node setup. I've thoroughly reviewed the code logic:

The implementation uses BackendUtils.openChannelSync, available across LND, CLNRest, LNC, and Embedded LND backends
The note is saved after a successful openChannelSync response using funding_txid_str

I'm confident the logic is sound, but happy to do additional testing if you have a preferred setup or if you'd like me to troubleshoot the embedded node issue further.

Also , I apologize for removing items from the PR template I'll keep the full template in future PRs.

nbenaglia (approved):

stores/ChannelsStore.ts
```diff
@@ -1184,6 +1187,20 @@ export default class ChannelsStore {
...
+                            const noteKey = `note-${this.funding_txid_str}`;
+                            const noteValue = localeString(
+                                'views.OpenChannel.openedChannelNote'
+                            );
+                            console.log(
+                                'Auto-saving channel note:',
+                                noteKey,
+                                noteValue
+                            );
+                            this.notesStore.storeNoteKeys(noteKey, noteValue);
```
kaloudis: Incorrect usage here. `storeNotesKeys` is the array of `noteKeys`, not there context. Please read `AddNotes` for an example

stores/ChannelsStore.ts
```diff
@@ -1184,6 +1187,20 @@ export default class ChannelsStore {
                         this.channelRequest = null;
                         this.channelSuccess = true;
                         this.connectingToPeer = false;
+
+                        // Auto-save note for opened channel
+                        if (this.funding_txid_str) {
+                            const noteKey = `note-${this.funding_txid_str}`;
+                            const noteValue = localeString(
+                                'views.OpenChannel.openedChannelNote'
```
kaloudis: this note is super generic. Could we add the peer alias/pubkey or the date for more context?

kaloudis (changes requested):

kaloudis:
Before submitting PRs, it's expected that the author has actually tested the PRs. Claiming that you had issues setting up your environment is not an acceptable excuse. This PR's functionality is fundamentally broken.

Set up an environment and test as many wallet interfaces as possible. We recommend using https://lightningpolar.com/ for spinning up nodes in regtest.

aryanraj45:
Hi @kaloudis  
I've now set up Polar and tested properly. The issue was LND REST API returning funding_txid_bytes instead of funding_txid_str. Fixed it to handle both formats with proper conversion. Auto-note now works correctly  screenshots attached showing the feature working with peer pubkey and timestamp.
image

aryanraj45:
Hi @kaloudis @ajaysehwal, I have addressed the requested changes and updated the PR. Please take a look when you have time. Thank you!

stores/ChannelsStore.ts
myxmaster: I agree it would be very nice to have the node alias/pubkey in the note as well.

Not sure about the date, how relevant is that (vs confirmation block/time of the TX)?

Edit:
Just saw it was addressed already. I do think the note should be written less technical, e.g.:
"Opening Lightning channel to node {alias} (pubkey: {pubkey})"

myxmaster:
The `funding_txid_bytes` handling is a fix for a pre-existing latent bug (the observable isn't consumed by any view till now, so it had no visible impact). Ideally this would land in a separate PR, or at least a separate commit, before building the auto-note feature on top of it.

Also a small simplification for the fix worth considering: since `Base64Utils.bytesToHex` already handles both `number[]` and plain objects with number values internally, the `Uint8Array` and `{ data }` branches can be collapsed into one:

```
} else {
    hex = Base64Utils.bytesToHex(fundingBytes?.data ?? fundingBytes);
}
```

myxmaster:
After thinking about the note content/formatting a bit more, I suggest always including the pubkey and wrapping the alias in quotes for clarity, falling back to `Unknown` (existing `general.unknown` locale string) when no alias is available:

`Opening Lightning channel to "Olympus by ZEUS" (pubkey: 031b30...)`
`Opening Lightning channel to "Unknown" (pubkey: 03abcd...)`

Like [mentioned before](https://github.com/ZeusLN/zeus/pull/3641#discussion_r2821169363), the date in the note is not needed imo.

Regarding peer alias lookup:
The lookup against `this.peers` is unreliable because newly connected peer won't be in that list yet, since `getPeers` was called before the connection was made. `getNodeInfo()` already exists in `ChannelsStore`, fetches the alias directly from the network graph, populates `aliasMap`, and already falls back to `general.unknown` internally. Consider calling `await this.getNodeInfo(pubkey)` and reading the alias from `this.aliasMap.get(pubkey)` instead.

nit: `storeNoteKeys` should probably be awaited.

And finally, the same would be nice for closing channels. But the close flow is significantly more complex. The closing txid isn't captured in the response handler, there's a 6-second forced resolution for LND REST that shortcuts the response, and force closes don't resolve immediately at all. So starting with channel open only makes sense as a first step imo.

PR #3658, opened 2026-02-04
fix: prevent QR scanner from firing navigation twice

kaloudis:
# Description

Problem: When a QR code is scanned and `handleAnything` resolves, it sets `loading: false` before navigating. This can briefly remount the `QRCodeScanner` with a fresh cache, allowing the same QR to be scanned again and triggering duplicate navigation.

  Solution: Added an `isProcessing` class property that:
  1. Is checked and set synchronously (unlike `setState` which is async)
  2. Is placed at the point where final processing begins - AFTER multi-frame QR parts are collected
  3. Is reset only on errors (so users can retry), not on successful navigation

  Key points:
  - Multi-frame QRs (BC-UR, BBQR) still work because the guard is placed after their part collection logic
  - Using a class property instead of state prevents race conditions from async state updates
  - `Swap` and `RefundSwap` QR scanners are now moved to their own views based on a new `BaseQRScanner` component to improve maintainability

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [x] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/RefundSwapQRScanner.tsx
+++ b/views/RefundSwapQRScanner.tsx
@@ -23,8 +23,10 @@ export default class RefundSwapQRScanner extends BaseQRScanner {
             )
         ) {
             navigation.goBack();
-            navigation.navigate('RefundSwap', {
-                scannedAddress: value
+            navigation.navigate({
+                name: 'RefundSwap',
+                params: { scannedAddress: value },
+                merge: true
             });
             return;
         }
--- a/views/SwapsQRScanner.tsx
+++ b/views/SwapsQRScanner.tsx
@@ -27,10 +27,14 @@ export default class SwapsQRScanner extends BaseQRScanner {
             )
         ) {
             navigation.goBack();
-            navigation.navigate('Swaps', {
-                initialInvoice: value,
-                initialAmountSats: satAmount,
-                initialReverse: true
+            navigation.navigate({
+                name: 'Swaps',
+                params: {
+                    initialInvoice: value,
+                    initialAmountSats: satAmount,
+                    initialReverse: true
+                },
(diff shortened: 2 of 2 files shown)


views/RefundSwapQRScanner.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+        if (
+            AddressUtils.isValidBitcoinAddress(
+                value,
+                isTestNet || isRegTest || isSigNet
+            )
+        ) {
+            navigation.goBack();
+            navigation.navigate('RefundSwap', {
+                scannedAddress: value
```
ajaysehwal: @kaloudis, I think we should pass merge: true in the navigation.navigate options so we can preserve the existing params (like swapData) while adding scannedAddress, or perhaps pass swapData when opening the scanner and when navigating back ?

kaloudis:
> @kaloudis, the swap input values are not updating when we scan new Lightning QR again
> ScreenRecording_03-05-2026.20-31-20_1.MP4

I think this may be unrelated to this PR, something that @shubhamkmr04 may need to address

ajaysehwal (approved):
tACK

kaloudis:
Swaps issue tracked here: https://github.com/ZeusLN/zeus/issues/3801

PR #3664, opened 2026-02-05
fix: kill existing Embedded LND process before starting a new one

ajaysehwal:
# Description

### Previous behavior

Previously, when startLnd was called, the code checked whether LND was already running. If an existing process was found, startup was skipped, and in some cases this could even lead to app crashes. This behavior created edge cases where LND was technically running but not fully usable (for example, during a partial startup or after a failed initialization), leaving the app in an inconsistent state

### Current behavior

The startup flow has been updated to always ensure a clean LND start:

- Check whether an LND process is currently running
- If LND is running:
  - Log that the existing process is being stopped
  - Attempt a graceful shutdown using `stopLnd()`
  - Log any shutdown errors without blocking startup
  - Start a fresh LND process everytime

### Why this change

By avoiding the silent “already running” case and always starting LND from a clean state, startup becomes more predictable and reliable. This helps prevent the app and LND from getting out of sync, makes startup issues easier to debug, and allows faster switching between wallets.

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

- [x] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/InvoicesStore.ts
+++ b/stores/InvoicesStore.ts
@@ -205,61 +205,50 @@ export default class InvoicesStore {
             unified: true,
             customPreimage,
             noLsp
-        })
-            .then(
-                (
-                    result:
-                        | { rHash: string; paymentRequest: string }
-                        | undefined
-                ) => {
-                    if (!result?.rHash || !result?.paymentRequest) {
-                        runInAction(() => {
-                            this.loading = false;
-                            this.creatingInvoice = false;
-                        });
-                        return;
-                    }
-                    const { rHash, paymentRequest } = result;
-                    if (
-                        BackendUtils.supportsOnchainReceiving() &&
-                        !skipOnchain
-                    ) {
-                        return this.getNewAddress(
-                            addressType
-                                ? { type: addressType, unified: true }
-                                : { unified: true }
-                        )
-                            .then((onChainAddress: string) => {
-                                runInAction(() => {
-                                    this.onChainAddress = onChainAddress;
-                                    this.payment_request = paymentRequest;
-                                    this.loading = false;
-                                    this.creatingInvoice = false;
-                                });
-                                return { rHash, onChainAddress };
-                            })
-                            .catch(() => {
-                                runInAction(() => {
-                                    this.loading = false;
-                                    this.creatingInvoice = false;
-                                });
-                            });
-                    } else {
-                        runInAction(() => {
-                            this.payment_request = paymentRequest;
-                            this.creatingInvoice = false;
-                        });
-                        return { rHash };
-                    }
-                }
-            )
-            .catch((error: any) => {
+        }).then((result) => {
+            if (!result?.rHash || !result?.paymentRequest) {
                 runInAction(() => {
                     this.loading = false;
                     this.creatingInvoice = false;
                 });
-                throw error;
-            });
+                return Promise.reject(
+                    new Error(
+                        localeString(
+                            'stores.InvoicesStore.errorCreatingInvoice'
+                        )
+                    )
+                );
+            }
+            const { rHash, paymentRequest } = result;
+            if (BackendUtils.supportsOnchainReceiving() && !skipOnchain) {
+                return this.getNewAddress(
+                    addressType
+                        ? { type: addressType, unified: true }
+                        : { unified: true }
+                )
+                    .then((onChainAddress: string) => {
+                        runInAction(() => {
+                            this.onChainAddress = onChainAddress;
+                            this.payment_request = paymentRequest;
+                            this.loading = false;
+                            this.creatingInvoice = false;
+                        });
+                        return { rHash, onChainAddress };
+                    })
+                    .catch(() => {
+                        runInAction(() => {
+                            this.loading = false;
+                            this.creatingInvoice = false;
+                        });
+                    });
+            } else {
+                runInAction(() => {
+                    this.payment_request = paymentRequest;
+                    this.creatingInvoice = false;
+                });
+                return { rHash };
+            }
+        });
     };
 
     @action
@@ -329,7 +318,7 @@ export default class InvoicesStore {
                                         'stores.InvoicesStore.errorCreatingInvoice'
                                     );
                             });
-                            return Promise.reject(error);
+                            return;
                         }
                     }
                     const remotePolicy =
@@ -401,11 +390,7 @@ export default class InvoicesStore {
 
         if (this.lspStore.flow_error) {
             this.creatingInvoice = false;
-            return Promise.reject(
-                new Error(
-                    localeString('stores.InvoicesStore.errorCreatingInvoice')
-                )
-            );
+            return;
         }
 
         return BackendUtils.createInvoice(req)
@@ -429,7 +414,6 @@ export default class InvoicesStore {
                                       'stores.InvoicesStore.errorCreatingInvoice'
                                   );
                     });
-                    return;
                 }
                 const invoice = new Invoice(data);
 
@@ -514,7 +498,6 @@ export default class InvoicesStore {
                             'stores.InvoicesStore.errorCreatingInvoice'
                         );
                 });
-                throw error;
             });
     };
 
--- a/utils/LndMobileErrors.ts
+++ b/utils/LndMobileErrors.ts
@@ -1,0 +1,241 @@
+import { sleep } from './SleepUtils';
+
+/**
+ * LND error codes - use for consistent error handling across the app
+ */
+export enum LndErrorCode {
+    /** LND data folder does not exist */
+    LND_FOLDER_MISSING = 'LND_FOLDER_MISSING',
+    /** LND unlocked too quickly during wallet creation */
+    WALLET_CREATION_UNLOCKED_TOO_QUICKLY = 'WALLET_CREATION_UNLOCKED_TOO_QUICKLY',
+    /** RPC connection was closed */
+    RPC_CONNECTION_CLOSED = 'RPC_CONNECTION_CLOSED',
+    /** RPC did not become ready within timeout */
+    RPC_READY_TIMEOUT = 'RPC_READY_TIMEOUT',
+    /** LND failed to stop after max retries */
+    LND_FAILED_TO_STOP = 'LND_FAILED_TO_STOP',
+    /** LND failed to start after max retries */
+    LND_START_FAILED = 'LND_START_FAILED',
+    /** Failed to generate seed */
+    GEN_SEED_FAILED = 'GEN_SEED_FAILED',
+    /** Timeout waiting for LND to become ready */
+    LND_READY_TIMEOUT = 'LND_READY_TIMEOUT',
+    /** LND is already running (informational) */
+    LND_ALREADY_RUNNING = 'LND_ALREADY_RUNNING',
+    /** Wallet is locked - unlock required */
+    WALLET_LOCKED = 'WALLET_LOCKED',
+    /** Macaroon store is locked */
+    MACAROON_STORE_LOCKED = 'MACAROON_STORE_LOCKED',
+    /** Wallet/Unlocker already unlocked (genSeed race) */
+    GEN_SEED_UNLOCKED = 'GEN_SEED_UNLOCKED',
+    /** RPC not ready yet (transient) */
+    RPC_NOT_READY = 'RPC_NOT_READY',
+    /** Expected state when stopping LND  */
+    STOP_LND_EXPECTED = 'STOP_LND_EXPECTED',
+    /** Invalid response from LND stream */
+    STREAM_INVALID_RESPONSE = 'STREAM_INVALID_RESPONSE',
+    /** Stream ended normally (EOF) - not an error */
+    STREAM_EOF = 'STREAM_EOF'
+}
+
+/**  messages for display in UI */
+export const LND_ERROR_MESSAGES: Record<LndErrorCode, string> = {
+    [LndErrorCode.LND_FOLDER_MISSING]:
+        'LND wallet folder not found. The wallet may have been deleted.',
+    [LndErrorCode.WALLET_CREATION_UNLOCKED_TOO_QUICKLY]:
+        'Wallet creation failed: LND unlocked too quickly. Please try again.',
+    [LndErrorCode.RPC_CONNECTION_CLOSED]: 'RPC connection closed.',
+    [LndErrorCode.RPC_READY_TIMEOUT]:
+        'RPC did not become ready in time. Please try again.',
+    [LndErrorCode.LND_FAILED_TO_STOP]:
+        'LND failed to stop. Please restart the app.',
+    [LndErrorCode.LND_START_FAILED]:
+        'LND failed to start. Please restart the app.',
+    [LndErrorCode.GEN_SEED_FAILED]: 'Failed to generate wallet seed.',
+    [LndErrorCode.LND_READY_TIMEOUT]:
+        'LND did not become ready in time. Please try again.',
+    [LndErrorCode.LND_ALREADY_RUNNING]: 'LND is already running.',
+    [LndErrorCode.WALLET_LOCKED]: 'Wallet is locked. Unlock required.',
+    [LndErrorCode.MACAROON_STORE_LOCKED]:
+        'Macaroon store is locked. Please try again.',
+    [LndErrorCode.GEN_SEED_UNLOCKED]:
+        'Wallet already unlocked. Please try wallet creation again.',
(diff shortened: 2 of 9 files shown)


utils/LndMobileUtils.ts
```diff
@@ -317,103 +609,291 @@ export async function startLnd({
...
+    startArgs: {
+        args: string;
+        lndDir: string;
+        isTorEnabled: boolean;
+        isTestnet: boolean;
+    };
+    walletPassword: string;
+    unlockWallet: (password: string) => Promise<void>;
+}) {
+    while (true) {
```
kaloudis: This needs a max retry count, otherwise it'll loop forever

utils/LndMobileUtils.ts
```diff
@@ -25,7 +25,189 @@ import {
...
+        'wallet may have been deleted'
+    ],
+    [LndErrorCode.WALLET_CREATION_UNLOCKED_TOO_QUICKLY]: [],
+    [LndErrorCode.RPC_CONNECTION_CLOSED]: ['closed', 'connection closed'],
+    [LndErrorCode.RPC_READY_TIMEOUT]: [],
+    [LndErrorCode.LND_FAILED_TO_STOP]: [],
+    [LndErrorCode.GEN_SEED_FAILED]: [],
+    [LndErrorCode.LND_READY_TIMEOUT]: [],
+    [LndErrorCode.LND_ALREADY_RUNNING]: ['already started', 'already running'],
+    [LndErrorCode.WALLET_LOCKED]: ['wallet locked', 'locked'],
```
kaloudis: 'locked' here overshadows 'macaroon store is locked' below it on line 109

utils/LndMobileUtils.ts
```diff
@@ -71,19 +253,28 @@ export function checkLndStreamErrorResponse(
...
+            log.d(`Transient startup error for ${name}: ${errorDesc}`);
+            return null;
+        }
+
+        // Handle closed connection - expected when switching nodes or stopping LND
+        if (
+            matchesLndErrorCode(errorDesc, LndErrorCode.RPC_CONNECTION_CLOSED)
+        ) {
+            log.d('Connection closed (expected during node switch/stop)');
+            return null;
```
kaloudis: I think we should return something in this case, otherwise `waitForLndReady` will continue to wait for a state event even if nothing will return.

utils/LndMobileUtils.ts
```diff
@@ -317,103 +609,291 @@ export async function startLnd({
...
+            if (!isResolved) {
+                isResolved = true;
+                cleanup();
+                reject(new Error('Timeout waiting for LND to become ready'));
+            }
+        }, 60000);
+
+        const cleanup = () => {
+            if (timeout) clearTimeout(timeout);
+            LndMobileEventEmitter.removeAllListeners('SubscribeState');
```
kaloudis: why remove all listeners? wouldn't it be better to only remove the one with the specific handler ref instead?

views/Settings/Wallets.tsx
```diff
@@ -90,9 +94,48 @@ export default class Nodes extends React.Component<NodesProps, NodesState> {
         this.props.navigation.addListener('focus', this.handleFocus);
     }
 
-    componentWillUnmount() {
+    async componentWillUnmount() {
```
kaloudis: `componentWillUnmount` shouldn't be async, as React doesn't await on lifecycle methods. This logic should be in `deleteNodeConfig` or a focus event handler instead.

utils/LndMobileUtils.ts
```diff
@@ -317,103 +609,291 @@ export async function startLnd({
...
+    subscribeState,
+    walletPassword,
+    unlockWallet
+}: {
+    decodeState: (data: string) => lnrpc.SubscribeStateResponse;
+    subscribeState: () => Promise<string>;
+    walletPassword: string;
+    unlockWallet: (password: string) => Promise<void>;
+}) {
+    return new Promise(async (resolve, reject) => {
```
kaloudis: if an unhandled exception occurs after the promise settles, it becomes a silent unhandled rejection. The `isResolved` guard mitigates most risk, but after the `reject(unlockError)` on line 829, there's no return so execution falls through to break.

utils/LndMobileUtils.ts
```diff
@@ -25,7 +25,189 @@ import {
...
+        'no such file or directory',
+        "doesn't exist",
+        'not yet ready to accept calls',
+        'starting up',
+        'not yet ready',
+        'unitialized',
+        'uninitialized',
+        'connection refused',
+        'connection reset'
+    ],
```
kaloudis: Every pattern in `STOP_LND_EXPECTED` is already matched by an earlier code in the enum. `matchRawErrorToCode()` will never return `STOP_LND_EXPECTED`. Not a runtime bug since it's only used via `matchesLndErrorCode()`, but it's conceptually confusing and could cause issues if someone uses `matchRawErrorToCode` in the future.

kaloudis (changes requested):

utils/LndMobileUtils.ts
ajaysehwal: LndMobileEventEmitter doesn’t have a method to remove a specific listener. It only provides removeAllListeners for removing listeners.

shubhamkmr04:
Simulator Screenshot - iPhone 17 - 2026-02-23 at 21 46 39

also can we get rid of these errors while creating the wallet ?

shubhamkmr04:
https://github.com/user-attachments/assets/8cdaf8bf-ec1e-44af-8622-c27645c8cdb6

how are we deciding that the first wallet should be active after a wallet was deleted. maybe we stay on wallets view in this case ?

shubhamkmr04:
Simulator Screenshot - iPhone 17 - 2026-02-25 at 15 52 31

@ajaysehwal

shubhamkmr04 (approved):
tACK

utils/LndMobileUtils.ts
```diff
@@ -38,8 +36,215 @@ import {
...
+}
+
+export function isLndError(error: unknown, code?: LndErrorCode): boolean {
+    const msg = error instanceof Error ? error.message : String(error ?? '');
+    const errCode = (error as { code?: LndErrorCode })?.code;
+    if (code) {
+        return errCode === code || msg.includes(code);
+    }
+    return Object.values(LndErrorCode).some((c) => msg.includes(c));
+}
```
kaloudis: I think `LndMobileUtils` has gotten quite large. Let's consider breaking the above off into `LndMobileErrors`

utils/LndMobileUtils.ts
```diff
@@ -320,103 +651,345 @@ export async function startLnd({
...
+    subscribeState,
+    walletPassword,
+    unlockWallet
+}: {
+    decodeState: (data: string) => lnrpc.SubscribeStateResponse;
+    subscribeState: () => Promise<string>;
+    walletPassword: string;
+    unlockWallet: (password: string) => Promise<void>;
+}) {
+    return new Promise(async (resolve, reject) => {
```
kaloudis: `return new Promise(async (resolve, reject) => {`: This is the explicit construction anti-pattern. If `subscribeState()` or anything before the listener fires throws synchronously, the rejection won't propagate properly — it'll be an unhandled promise rejection.

The try/catch around subscribeState() at the end mitigates the worst case, but async errors thrown inside `stateHandler` after `isResolved` is already true could still beswallowed. Consider restructuring to avoid async inside the Promise executor.

views/Wallet/Wallet.tsx
```diff
@@ -735,48 +754,99 @@ export default class Wallet extends React.Component<WalletProps, WalletState> {
...
+                    console.log(
+                        'Transient RPC error - attempting restart:',
+                        errorMessage
+                    );
+                    setConnectingStatus(false);
+                    await new Promise((resolve) => setTimeout(resolve, 2000));
+                    setConnectingStatus(true);
+                    this.getSettingsAndNavigate();
+                    return;
+                }
```
kaloudis, line 853: Two places call `this.getSettingsAndNavigate()` (here and lines 648-658) recursively on transient RPC errorswith only a 2s delay and no retry limit.

kaloudis: If the transient error persists, this will loop infinitely (every ~2s), piling up calls. Consider adding a max retry count or exponential backoff.

utils/LndMobileUtils.ts
```diff
@@ -280,23 +494,141 @@ export async function initializeLnd({
...
+                if (!stillRunning) {
+                    log.d('LND stopped successfully');
+                    return;
+                }
+                throw new Error(
+                    `LND failed to stop after ${maxRetries} attempts`
+                );
+            },
+            maxRetries,
+            delayMs: 0,
```
kaloudis: retry() is called with `delayMs: 0` but the actual delay is manually added via await `sleep(delayMs)` inside fn. This bypasses the retry utility's built-in delay mechanism and adds delay before the first check unnecessarily. Overall this a confusing delay structure.

utils/LndMobileUtils.ts
```diff
@@ -38,8 +36,215 @@ import {
...
+    ],
+    [LndErrorCode.GEN_SEED_UNLOCKED]: [
+        'wallet already unlocked',
+        'WalletUnlocker service is no longer available'
+    ],
+    [LndErrorCode.RPC_NOT_READY]: [
+        'starting up',
+        'not yet ready to accept calls',
+        'not yet ready',
+        'unitialized',
```
kaloudis: is this typo in LND? what's the story here? correct version is below.

views/Settings/Wallets.tsx
```diff
@@ -237,6 +318,9 @@ export default class Nodes extends React.Component<NodesProps, NodesState> {
                             fontFamily: 'PPNeueMontreal-Book'
                         }
                     }}
+                    onBack={async () => {
+                        await this.handleJustDeletedWallet();
```
kaloudis: This runs every time the user presses back on the `Wallets` screen, not just after deletion.

If `justDeletedWallet` is false it returns early, but the `onBack` handler now swallows the default back behavior — does the `Header` component still navigate back after `onBack` returns? If `onBack` replaces the default behavior, pressing back after a normal visit (no deletion) would do nothing.

stores/InvoicesStore.ts
```diff
@@ -205,44 +205,61 @@ export default class InvoicesStore {
...
                 }
-            }
-        );
+            )
+            .catch((error: any) => {
+                runInAction(() => {
+                    this.loading = false;
+                    this.creatingInvoice = false;
+                });
+                throw error;
```
kaloudis: Why throw in this catch??

views/Settings/WalletConfiguration.tsx
```diff
@@ -731,23 +756,53 @@ export default class WalletConfiguration extends React.Component<
...
 
         const lndDir = uuidv4();
 
-        const response = await createLndWallet({
-            lndDir,
-            seedMnemonic: recoveryCipherSeed,
-            isTestnet: network === 'Testnet',
-            channelBackupsBase64
-        });
+        SyncStore.reset();
```
kaloudis: This second call is harmless but redundant since nothing starts sync between the two calls. (First call on line 773)

utils/LndMobileUtils.ts
```diff
@@ -38,8 +36,215 @@ import {
...
+export function isLndError(error: unknown, code?: LndErrorCode): boolean {
+    const msg = error instanceof Error ? error.message : String(error ?? '');
+    const errCode = (error as { code?: LndErrorCode })?.code;
+    if (code) {
+        return errCode === code || msg.includes(code);
+    }
+    return Object.values(LndErrorCode).some((c) => msg.includes(c));
+}
+
+/** Android needs longer delays - process cleanup is slower */
```
kaloudis: Does this comment pertain to the consts below it? It's a bit confusing

utils/LndMobileUtils.ts
```diff
@@ -320,103 +651,345 @@ export async function startLnd({
...
+            LndMobileEventEmitter.removeAllListeners('SubscribeState');
+        };
+
+        const timeout = setTimeout(() => {
+            if (!isResolved) {
+                isResolved = true;
+                cleanup();
+                reject(createLndError(LndErrorCode.LND_READY_TIMEOUT));
+            }
+        }, 60000);
```
kaloudis: Magic number 60000 for the `waitForLndReady` timeout. This should be a named constant like the other timing values at the top of the file.

views/Settings/Wallets.tsx
```diff
@@ -222,6 +254,55 @@ export default class Nodes extends React.Component<NodesProps, NodesState> {
             });
         };
 
+        const onWalletPress = async (
+            nodeIndex: number,
+            nodeActive: boolean
+        ) => {
+            if (SettingsStore.settings?.justDeletedWallet) {
+                this.handleJustDeletedWallet();
```
kaloudis: does this need an await?

kaloudis (changes requested):

utils/LndMobileUtils.ts
ajaysehwal: @kaloudis, Thanks for pointing out

utils/LndMobileErrors.ts
```diff
@@ -0,0 +1,236 @@
...
+export const LND_ERROR_PATTERNS: Record<LndErrorCode, readonly string[]> = {
+    [LndErrorCode.LND_FOLDER_MISSING]: [
+        "doesn't exist",
+        'does not exist',
+        'no such file or directory',
+        'wallet directory not found',
+        'wallet may have been deleted'
+    ],
+    [LndErrorCode.WALLET_CREATION_UNLOCKED_TOO_QUICKLY]: [],
+    [LndErrorCode.RPC_CONNECTION_CLOSED]: ['closed', 'connection closed'],
```
kaloudis: could 'closed' potential throw on other standard behavior, (e.g., "channel closed", "stream closed by peer")?

utils/LndMobileUtils.ts
```diff
@@ -320,104 +480,313 @@ export async function startLnd({
...
+                clearTimeout(timeout);
+                LndMobileEventEmitter.removeAllListeners('SubscribeState');
+            };
+
+            const settle = (fn: () => void) => {
+                cleanup();
+                fn();
+            };
+
+            const stateHandler = (event: any) => {
```
kaloudis, line 693: The `stateHandler` event listener can fire multiple times concurrently due to `await` points in `handleAsync()`. Two concurrent handlers could both call settle(). While double-resolving a promise is a no-op in JS, it could cause `cleanup()` and `waitForRpcReady()` to run twice. Fix: Add a settled boolean guard.

utils/LndMobileErrors.ts
```diff
@@ -0,0 +1,236 @@
...
+    console.log(
+        `Transient ${context} - retry ${
+            retryCount + 1
+        }/${MAX_TRANSIENT_RPC_RETRIES} in ${delayMs}ms:`,
+        errorMessage
+    );
+    setConnecting(false);
+    await sleep(delayMs);
+    setConnecting(true);
+    onRetry();
```
kaloudis: I believe this needs an `await`

utils/LndMobileErrors.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+export function getLndErrorMessage(error: unknown): string {
```
kaloudis: I believe this is a duplicate of `getErrorMessage` in `LNDMobileUtils` - let's consolidate to one

utils/LndMobileUtils.ts
```diff
@@ -602,6 +971,105 @@ export async function optimizeNeutrinoPeers(
...
+    return retry({
+        fn: async () => {
+            const seed = await genSeed(undefined);
+            if (!seed) {
+                throw createLndError(LndErrorCode.GEN_SEED_FAILED);
+            }
+            return seed;
+        },
+        maxRetries: 10,
+        delayMs: 500,
```
kaloudis: magic number here

kaloudis:
WalletConfiguration.tsx:744 still calls `restartNeeded(true)` on Android for activating an embedded-lnd wallet, while the `Wallets.tsx` switching flow has removed this requirement. Intentional?

PR #3665, opened 2026-02-05
Support LUD-18 sender nickname for LNURL-pay

preetham3127:
### Summary
Adds minimal support for LUD-18 by allowing an optional sender nickname to be included in LNURL-pay requests.

### Details
- Introduces an optional nickname input in the existing LNURL-pay UI
- Includes `payerData.name` only when the nickname is non-empty
- Omits `payerData` entirely when unset
- No persistence, no refactors, no changes outside LNURL-pay flow

### Scope
- Touches only `views/LnurlPay/LnurlPay.tsx`
- Keeps behavior fully backward-compatible

### DIFF
--- a/views/LnurlPay/LnurlPay.tsx
+++ b/views/LnurlPay/LnurlPay.tsx
@@ -45,6 +45,7 @@ interface LnurlPayState {
     satAmount: string | number;
     domain: string;
     comment: string;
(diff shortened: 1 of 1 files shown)


kaloudis:
Are there UI changes here? Can you provide screenshots?

Issue #3592, opened 2026-01-24
Export Activity to CSV -> Payments: only sent payments are exported (not received)


myxmaster:
### Describe the bug

I just noticed that we only export sent (Lightning) payments, not received payments. When exporting (on-chain) transactions we export both, send and received.

Not sure if that was on purpose, but I don't think that makes sense. We should export incoming and outgoing Lightning payments.

### Reproduce

-

### ZEUS version

v0.12.2

### Node interface

LND (REST)

### Network

Clearnet

### Device

emulator

### Device operating system

Android 16

### Log output

```shell

```

kaloudis:
Payments are outgoing, invoices are incoming

myxmaster:
Hmmm, I see. But I think technically a payment can be both directions. And an invoice can also be unpaid (no direction).

Ok this overlaps with the idea here now: https://github.com/ZeusLN/zeus/issues/3497#issuecomment-3796268205

kaloudis:
> Hmmm, I see. But I think technically a payment can be both directions. And an invoice can also be unpaid (no direction).
> 
> Ok this overlaps with the idea here now: [#3497 (comment)](https://github.com/ZeusLN/zeus/issues/3497#issuecomment-3796268205)

The problem is that models vary quite a bit. You'd have to reconcile them.

myxmaster:
Let us continue the discussion here:
https://github.com/ZeusLN/zeus/issues/3497

PR #3678, opened 2026-02-09
fix:  CSV export sanitization to prevent formula injection and preserve zero values

a-khushal:
# Description

Relates to issue: [**ZEUS-3592**](https://github.com/ZeusLN/zeus/issues/3592)

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/ActivityCsvUtils.ts
+++ b/utils/ActivityCsvUtils.ts
@@ -30,6 +30,20 @@ export const CSV_KEYS = {
     ]
 };
 
+const sanitizeCsvCell = (value: unknown): string => {
+    if (value === undefined || value === null) return '';
+    if (typeof value === 'number') return String(value);
+
+    const stringValue = String(value);
+    const escapedValue = stringValue.replace(/"/g, '""');
+
+    if (/^\s*[=+\-@]/.test(escapedValue)) {
+        return `'${escapedValue}`;
+    }
+
+    return escapedValue;
+};
+
 // Generates a formatted timestamp string for file naming.
 export const getFormattedDateTime = (): string => {
     const now = new Date();
@@ -54,7 +68,7 @@ export const convertActivityToCsv = async (
         const rows = data
             .map((item) =>
                 keysToInclude
-                    .map((field) => `"${item[field.value] || ''}"`)
+                    .map((field) => `"${sanitizeCsvCell(item[field.value])}"`)
                     .join(',')
             )
             .join('\n');
--- a/utils/ActivityCsvUtils.test.ts
+++ b/utils/ActivityCsvUtils.test.ts
@@ -141,6 +141,74 @@ describe('activityCsvUtils', () => {
             expect(result).toContain('"dest123","","","","","",""');
         });
 
+        it('preserves zero values in CSV output', async () => {
+            const mockPayments = [
+                {
+                    getDestination: 'dest123',
+                    getPaymentRequest: 'pay_req123',
+                    paymentHash: 'hash_pay1',
+                    getAmount: 0,
+                    getMemo: '',
+                    getNote: '',
+                    getDate: '2024-02-09'
+                }
+            ];
+
+            const result = await convertActivityToCsv(
+                mockPayments,
+                CSV_KEYS.payment
+            );
+
+            expect(result).toContain(
+                '"dest123","pay_req123","hash_pay1","0","","","2024-02-09"'
+            );
+        });
+
+        it('preserves negative numeric values in CSV output', async () => {
+            const mockPayments = [
+                {
+                    getDestination: 'dest123',
+                    getPaymentRequest: 'pay_req123',
+                    paymentHash: 'hash_pay1',
+                    getAmount: -21,
+                    getMemo: '',
+                    getNote: '',
+                    getDate: '2024-02-09'
+                }
+            ];
+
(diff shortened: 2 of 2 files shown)


views/Tools/ActivityExport.tsx
```diff
@@ -25,17 +25,17 @@ import {
     getFormattedDateTime,
     convertActivityToCsv,
     saveCsvFile,
-    CSV_KEYS
+    CSV_KEYS,
+    isPaymentExportActivity,
+    toPaymentCsvRow
 } from '../.././utils/ActivityCsvUtils';
```
ajaysehwal, line 30: @a-khushal , please fix this import path as well

a-khushal: Fixed.

myxmaster:
Thanks for tackling this @a-khushal.

As mentioned in the issue, the direction of merging received payments into the payments export definitely makes conceptual sense to me.

But to be honest, I think we should hold off on merging this and instead go for a more complete solution: a single unified export with user-selectable categories (On-chain transactions, Lightning payments, Cashu payments). Building that cleanly from scratch will be easier than layering on top of the existing split.

The `sanitizeCsvCell` function is definitely worth implementing though. Maybe you can make a new PR just for that, so it can get merged soon?

myxmaster:
I suggest we discuss the details of a more complete solution here: https://github.com/ZeusLN/zeus/issues/3497

a-khushal:
> Thanks for tackling this @a-khushal.
> 
> As mentioned in the issue, the direction of merging received payments into the payments export definitely makes conceptual sense to me.
> 
> But to be honest, I think we should hold off on merging this and instead go for a more complete solution: a single unified export with user-selectable categories (On-chain transactions, Lightning payments, Cashu payments). Building that cleanly from scratch will be easier than layering on top of the existing split.
> 
> The `sanitizeCsvCell` function is definitely worth implementing though. Maybe you can make a new PR just for that, so it can get merged soon?

Thanks for the feedback. I changed this PR to only the CSV sanitize fix. It now escapes quotes, blocks formula-style values (=, +, -, @), and keeps 0 values correctly, with tests. 

Also I've removed the received-payments export changes. We can discuss and build the full export update in #3497

utils/ActivityCsvUtils.ts
```diff
@@ -30,6 +30,19 @@ export const CSV_KEYS = {
     ]
 };
 
+const sanitizeCsvCell = (value: any): string => {
+    if (value === undefined || value === null) return '';
+
+    const stringValue = String(value);
+    const escapedValue = stringValue.replace(/"/g, '""');
+
+    if (/^[\s]*[=+\-@]/.test(escapedValue)) {
```
myxmaster: nit: [\s] -> \s

utils/ActivityCsvUtils.ts
```diff
@@ -54,7 +67,7 @@ export const convertActivityToCsv = async (
         const rows = data
             .map((item) =>
                 keysToInclude
-                    .map((field) => `"${item[field.value] || ''}"`)
+                    .map((field) => `"${sanitizeCsvCell(item?.[field.value])}"`)
```
myxmaster: `item?.` optional chaining is unnecessary here, `data` is always populated via `.filter()`, which never produces `undefined` elements, so `item` in `.map()` is guaranteed to be defined

utils/ActivityCsvUtils.ts
```diff
@@ -30,6 +30,19 @@ export const CSV_KEYS = {
     ]
 };
 
+const sanitizeCsvCell = (value: any): string => {
```
myxmaster: Would it make sense to type `value` as `unknown` instead of `any`? Since the function already handles `null`/`undefined` explicitly and converts everything via `String()`, `unknown` might be a safer signal here, it would force TypeScript to verify that any future access on `value` is intentional, rather than silently allowing arbitrary property access.

utils/ActivityCsvUtils.ts
a-khushal: Fixed.

utils/ActivityCsvUtils.ts
a-khushal: Fixed

utils/ActivityCsvUtils.ts
a-khushal: Fixed

utils/ActivityCsvUtils.ts
```diff
@@ -30,6 +30,19 @@ export const CSV_KEYS = {
     ]
 };
 
+const sanitizeCsvCell = (value: unknown): string => {
+    if (value === undefined || value === null) return '';
+
```
myxmaster, line 36: should we add `if (typeof value === 'number') return String(value);` here?
not sure, but can amounts be negative? we don't want to sanitize those.

Issue #3701, opened 2026-02-14
bug: Reverse swap rescue: transaction claimed


kaloudis:
When restoring a reverse swap (LN -> on-chain) with state transaction.claimed, user must be able to input an address to redeem to.

PR #3707, opened 2026-02-15
fix(swaps): Allow redeem flow for restored reverse swaps in transaction.claimed

a-khushal:
# Description

Relates to issue: [**ZEUS-3701**](https://github.com/ZeusLN/zeus/issues/3701)

the `components/NFCButton.tsx` and `views/OnChainAddresses.tsx` fixes are compatibility / build fixes required after rebase, not part of the swap feature.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [x] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


a-khushal:
@gemini-code-assist /review

a-khushal:
@gemini-code-assist /review

a-khushal:
@gemini-code-assist /review

a-khushal:
@gemini-code-assist /review

kaloudis:
@a-khushal please rebase

components/NFCButton.tsx
kaloudis, line 1: why changes here @a-khushal ?

kaloudis (changes requested):

components/NFCButton.tsx
a-khushal: it was just some build fixes from rebase, not related to this PR. 

removed them, the current commit only has the swap redeem changes now.

views/Swaps/Refund.tsx
```diff
@@ -121,6 +185,82 @@ export default class RefundSwap extends React.Component<
...
+            transactionHex,
+            lockupAddress: effectiveLockupAddress,
+            destinationAddress,
+            feeRate: Number(fee || DEFAULT_FEE_RATE),
+            minerFee: this.props.SwapStore?.claimMinerFee || 0,
+            isTestnet: nodeInfo.isTestNet
+        });
+
+        this.setState({
+            refundStatus: 'Claim transaction created successfully.',
```
kaloudis: needs localizations

views/Swaps/Refund.tsx
```diff
@@ -121,6 +185,82 @@ export default class RefundSwap extends React.Component<
...
+        const transactionHex = await this.getReverseTransactionHex(swapData);
+
+        if (!transactionHex) {
+            throw new Error('Could not fetch swap lockup transaction');
+        }
+
+        const preimageHex = await this.getReversePreimageHex(swapData);
+
+        if (!preimageHex) {
+            throw new Error('Could not derive swap preimage from rescue key');
```
kaloudis: all these error strings need localization. cross reference with existing strings too

kaloudis (changes requested):

views/Swaps/Refund.tsx
```diff
@@ -360,30 +506,47 @@ export default class RefundSwap extends React.Component<
...
+                                        destinationAddress
+                                    );
+                                }
 
                                 console.log(
-                                    'Refund transaction created successfully.'
+                                    'Swap transaction created successfully.'
                                 );
                             } catch (e) {
                                 console.error('Error in refund process:', e);
```
kaloudis, line 578: Errors thrown from `createReverseClaimTransaction` reach this catch but never surface to the user — the outer handler only `console.error`s. `createRefundTransaction` above sets `state.error` inside its own try/catch before rethrowing, so even though this outer catch is silent, refund failures still render an `ErrorMessage`. The new reverse-claim method has no inner handler, so failures like:

- `'Could not derive swap preimage from rescue key'`
- `'Could not fetch swap lockup transaction'`
- network errors from `getReverseTransactionHex` (no try/catch around the `ReactNativeBlobUtil.fetch`)
- native broadcast rejections (e.g. lockup already spent on a re-attempt)

…all leave the user with a stopped spinner and zero feedback.

Suggested fix — either wrap the new method's body in a try/catch that mirrors the refund pattern:

```ts
try {
    // ...existing body...
} catch (error: any) {
    this.setState({ loading: false, error: error.message });
    throw error;
}
```

…or surface the error here so both flows are covered:

```ts
} catch (e: any) {
    console.error('Error in refund process:', e);
    this.setState({ error: e?.message ?? String(e) });
}
```

utils/SwapUtils.ts
```diff
@@ -115,3 +115,41 @@ export const calculateLimit = (
...
+
+export const privateKeyFromSwapKeys = (
+    keys:
+        | { __D?: number[] | Uint8Array | { data?: number[] | Uint8Array } }
+        | null
+        | undefined
+): string | null => {
+    const raw = keys?.__D;
+    if (!raw) {
+        return null;
```
kaloudis, line 128: Minor diagnosability regression from the refactor: the original `privateKeyFromKeys` in `ClaimTransaction.ts` logged both branches that returned null —

```ts
if (!raw) {
    console.error('ClaimTransaction: keys.__D is missing');
    return null;
}
// ...
if (!bytes) {
    console.error('ClaimTransaction: unexpected key format', typeof raw);
    return null;
}
```

The new helper drops both `console.error` calls and just returns `null`. The only caller-side signal is the generic `'Could not derive swap private key'` thrown in `Refund.tsx`, which doesn't distinguish between "`__D` missing" vs. "`__D` present but in an unrecognized shape" vs. "no bytes parsed from the indexed-numeric-keys fallback." When a restored swap fails to redeem, this makes triage harder than it needs to be.

Suggest restoring `console.warn` (or `error`) on each null-return branch with the original context strings.

views/Swaps/Refund.tsx
```diff
@@ -121,6 +185,82 @@ export default class RefundSwap extends React.Component<
...
+        if (!effectiveLockupAddress) {
+            throw new Error('Could not find lockup address in swap data');
+        }
+
+        if (!nodeInfo) {
+            throw new Error('Node info is not available');
+        }
+
+        const privateKey =
+            swapData.refundPrivateKey || privateKeyFromSwapKeys(swapData.keys);
```
kaloudis: The `swapData.refundPrivateKey ||` fallback is dead code on this path. `refundPrivateKey` is only ever set for **submarine** swaps:

- `SwapStore.createSubmarineSwap` (`stores/SwapStore.ts:348`)
- `SwapStore.rescueSwapsFromService` (`stores/SwapStore.ts:919-925`, inside the `else { type: Submarine }` branch)

For a reverse swap — fresh or restored — `swapData.refundPrivateKey` is always `undefined`, so the `||` always falls through to `privateKeyFromSwapKeys(swapData.keys)`. This matches what `ReverseClaimTransaction.build` does in `models/ClaimTransaction.ts:93`:

```ts
const privateKey = privateKeyFromSwapKeys(swap.keys);
```

The current form is misleading because it suggests `refundPrivateKey` is a valid source of the claim key for a reverse swap, which it conceptually isn't (the user holds the *claim* key, not the refund key, on a reverse swap). Suggest simplifying to match the sibling builder:

```ts
const privateKey = privateKeyFromSwapKeys(swapData.keys);
```

PR #3612, opened 2026-01-27
fix: preserve swap amounts when scanning QR codes

shubhamkmr04:
# Description

1. Prevent swap form fields (amounts, invoice) from being reset when navigating to the QR scanner or onchain fee input and returning back
2. Rename inputFiat/outputFiat to inputDisplay/outputDisplay to correctly reflect that these fields hold display values for all unit types (sats, BTC, fiat), not just fiat

**Test plan**
1. For _submarine swaps_, scan a lightning invoice and verify both the invoice and amount fields are populated accordingly. If LSP is enabled, navigate to the LSP view and return, fields should not be cleared. 
2. For _reverse swaps_, enter an amount, then scan a bitcoin address, the amount should be preserved, the invoice field should show the scanned address, and the amount should still be editable. Navigate to the onchain fee input and return, all fields should remain intact.

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [X] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [X] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [X] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [X] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [ ] Yes
- [X] N/A

I have tested this PR on the following platforms (please specify OS version and phone model/VM):

- [X] Android
- [ ] iOS

I have tested this PR with the following types of nodes (please specify node version and API version where appropriate):

- [X] Embedded LND
- [ ] LND (REST)
- [ ] LND (Lightning Node Connect)
- [ ] Core Lightning (CLNRest)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales
- [ ] I’ve added new locale text that requires translations
- [ ] I’m aware that new translations should be made on the ZEUS [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Third Party Dependencies and Packages

- [ ] Contributors will need to run `yarn` after this PR is merged in
- [ ] 3rd party dependencies have been modified:
    * verify that `package.json` and `yarn.lock` have been properly updated
    * verify that dependencies are installed for both iOS and Android platforms

### Other:

- [ ] Changes were made that require an update to the README
[...]

PR #3739, opened 2026-02-20
fix(swaps): prevent form reset when navigating to QR scanner or fee editor

kaloudis:
# Description

Users reported that swap field values were getting cleared when scanning an invoice or changing on-chain fee values.

Root cause: The Swaps screen (views/Swaps/index.tsx:272-278) has a blur event listener that calls `resetFields()` whenever the screen loses focus. This wipes `inputSats`, `outputSats`, `invoice`, and other form state. There was only a guard for navigating to LSP fee explanation screens, but not for the QR scanner or fee customization.

Changes:

  1. `views/Swaps/index.tsx` — Renamed `isNavigatingToLSPFees` to i`sNavigatingAway` (more general) and set the flag before navigating to the QR scanner via `onScan`. Also passed an `onNavigateAway` callback to `OnchainFeeInput` so it can set the flag before navigating to `EditFee`.
  2. `components/OnchainFeeInput.tsx` — Added an optional `onNavigateAway` callbackprop, called before navigating to `EditFee`. This keeps the component generic (the prop is optional) while letting parent screens opt in to the guard.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [ ] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/OnchainFeeInput.tsx
+++ b/components/OnchainFeeInput.tsx
@@ -15,12 +15,13 @@ interface OnchainFeeInputProps {
     fee?: string;
     onChangeFee: (fee: string) => void;
     onFeeError?: (error: boolean) => void;
+    onNavigateAway?: () => void;
 }
 
 const DEFAULT_FEE = '10';
 
 export default function OnchainFeeInput(props: OnchainFeeInputProps) {
-    const { fee, onChangeFee, navigation } = props;
+    const { fee, onChangeFee, onNavigateAway, navigation } = props;
     const { settings } = settingsStore;
     const enableMempoolRates = settings?.privacy?.enableMempoolRates;
     const preferredMempoolRate =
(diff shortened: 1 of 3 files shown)


shubhamkmr04:
> Screen.Recording.2026-02-23.at.8.30.50.PM.mov 
> scanning the 5M invoice

these issues are being solved in #3612

shubhamkmr04:
> Screen.Recording.2026-02-23.at.8.30.50.PM.mov 
> scanning the 5M invoice

this one is still there

shubhamkmr04:
In reverse swap, when we scan and fill the on-chain address, entering an amount clears the address field.

This behavior makes sense for submarine swaps, since the invoice already contains the amount, and changing the amount field should clear the invoice.

But for reverse swaps, the address does not depend on the amount, so it should not be cleared when the amount is entered or modified.

PR #3769, opened 2026-02-28
fix: peer connect error handling and open channel button validation

### LINKED ISSUE #3768: 2 minor UI bugs: Open Channel / Connect Peer
#### Bug 1: Cryptic proto error when connecting peer by pubkey only
When trying to connect to a peer using only a pubkey (without providing host:port), and the peer is not discoverable via network gossip, Zeus displays a cryptic internal error message:
`proto: (line 1:180): invalid value for uint64 type: ""`

-> Instead, Zeus should display the error message that the backend is answering.

#### Bug 2: "Open Channel" / "Connect Peer" button active although required fields are empty
The buttons are enabled in situations where required inputs are missing, allowing users to trigger requests that will fail.

#### Bug 1:
1. Channels pane -> **Peers** tab
2. Tap icon "+" to open the Connect Peer screen
3. Switch Peer from "Olympus by ZEUS" to "Custom"
4. Enter a valid pubkey, leave Host empty
5. Tap **Connect Peer**

#### Bug 2:
1. Channels pane -> **Channels** tab
2. Tap icon "+" -> On-chain funded
-> **Open Channel** is active, although local amount is not set
==> this is just one example, more cases exist where the button should be inactive

### ZEUS version

v0.12.3

### Node interface

Multiple (specified in comments)

### Network

Clearnet

### Device

Android emulator 

### Device operating system

Android 16

### Log output

```shell

```

myxmaster:
# Description

This fixes https://github.com/ZeusLN/zeus/issues/3768.

#### Bug 1: Cryptic proto error when connecting peer by pubkey only
When connecting a peer without a host:port in "Connect Peer only" mode, an early-return code path in `connectPeer()` incorrectly set `channelRequest`, triggering a MobX reaction that called `openChannel()` instead of `connectPeer()`. This caused LND to receive an open-channel request with an empty `local_funding_amount` (uint64 in protobuf), producing the cryptic error `proto: invalid value for uint64 type: ""`.

The early-return (intended for the "open channel, peer already known" path) now only applies when `connectPeerOnly` is false. The actual `connectPeer` backend call proceeds and LND handles gossip lookup, returning an error if the peer is not reachable.
-> I noticed that LND does not return a nice error message, but only "EOF", but this is something that has to be fixed on LND side. Until then I don't think we need a temporary workaround on Zeus side (meaning: display a user friendly error msg), because it is an edge case anyway; usually people will provide host:port.

Also fixed a resulting unhandled promise rejection in `OpenChannel.tsx` where `connectPeer()` was called without `.catch()`.

#### Bug 2: "Open Channel" / "Connect Peer" button active although required fields are empty
Several validation state issues caused the Open Channel / Connect Peer button to remain active when required inputs were empty:
- Switching from "Olympus by ZEUS" to "Custom" did not reset `isNodePubkeyValid` to `false`
- ~~Switching back to "Olympus by ZEUS" did not restore `isNodePubkeyValid`/`isNodeHostValid` to `true`~~
- Switching back to "Olympus by ZEUS" did not restore `isNodePubkeyValid` to true; `isNodeHostValid` was refactored out of state entirely and is now derived from host in render, consistent with how `isInvalidFeeRate` is handled
- `AmountInput` initializes by calling `onAmountChange` with `satAmount = '0'` (string), so `!satAmount` never triggered -> fixed by checking `!Number(satAmount)`
- Disabling the "Use all possible funds" toggle did not reset `satAmount` in parent state, because `AmountInput.componentDidUpdate` does not call `onAmountChange` in its else-branch -> fixed by explicitly resetting `satAmount: '0'` in the toggle handler

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/ChannelsStore.ts
+++ b/stores/ChannelsStore.ts
@@ -742,7 +742,7 @@ export default class ChannelsStore {
         this.channelRequest = undefined;
         if (!silent) this.connectingToPeer = true;
 
-        if (!request.host) {
+        if (!request.host && !connectPeerOnly) {
             return await new Promise((resolve) => {
                 this.channelRequest = request;
                 resolve(true);
--- a/views/OpenChannel.tsx
+++ b/views/OpenChannel.tsx
@@ -1359,7 +1359,7 @@ export default class OpenChannel extends React.Component<
                                             },
                                             false,
                                             connectPeerOnly
-                                        );
+                                        ).catch(() => {});
                                     }}
                                     disabled={
                                         loading ||
--- a/views/OpenChannel.tsx
+++ b/views/OpenChannel.tsx
@@ -90,7 +90,6 @@ interface OpenChannelState {
     account: string;
     additionalChannels: Array<AdditionalChannel>;
     isNodePubkeyValid: boolean;
-    isNodeHostValid: boolean;
     nfcSupported: boolean;
 }
 
@@ -133,7 +132,6 @@ export default class OpenChannel extends React.Component<
             account: 'default',
             additionalChannels: [],
             isNodePubkeyValid: true,
-            isNodeHostValid: true,
             nfcSupported: false
         };
     }
@@ -344,7 +342,6 @@ export default class OpenChannel extends React.Component<
             advancedSettingsToggle,
             additionalChannels,
             isNodePubkeyValid,
-            isNodeHostValid,
             nfcSupported
         } = this.state;
         const { implementation } = SettingsStore;
@@ -363,6 +360,8 @@ export default class OpenChannel extends React.Component<
 
         const loading = connectingToPeer || openingChannel;
 
+        const isNodeHostValid =
+            host === '' || ValidationUtils.validateNodeHost(host);
         const isInvalidPeer = !isNodePubkeyValid || !isNodeHostValid;
         const isInvalidFeeRate = sat_per_vbyte === '0' || !sat_per_vbyte;
 
@@ -729,13 +728,15 @@ export default class OpenChannel extends React.Component<
                                                 'Olympus by ZEUS',
                                             node_pubkey_string:
                                                 config.lsps1Pubkey,
-                                            host: config.lsps1Host
+                                            host: config.lsps1Host,
+                                            isNodePubkeyValid: true
                                         });
                                     } else {
                                         this.setState({
                                             channelDestination: 'Custom',
                                             node_pubkey_string: '',
-                                            host: ''
+                                            host: '',
+                                            isNodePubkeyValid: false
                                         });
                                     }
                                 }}
@@ -803,11 +804,7 @@ export default class OpenChannel extends React.Component<
                                             value={host}
                                             onChangeText={(text: string) =>
                                                 this.setState({
-                                                    host: text,
-                                                    isNodeHostValid:
-                                                        ValidationUtils.validateNodeHost(
-                                                            text
-                                                        )
+                                                    host: text
                                                 })
                                             }
                                             autoCapitalize="none"
@@ -874,7 +871,8 @@ export default class OpenChannel extends React.Component<
                                                                 implementation ===
                                                                     'cln-rest'
                                                                     ? 'all'
-                                                                    : ''
+                                                                    : '',
+                                                            satAmount: ''
                                                         });
                                                     }}
                                                 />
@@ -1365,6 +1363,9 @@ export default class OpenChannel extends React.Component<
                                         loading ||
                                         (!connectPeerOnly &&
                                             isInvalidFeeRate) ||
+                                        (!connectPeerOnly &&
+                                            !fundMax &&
+                                            !Number(satAmount)) ||
                                         isInvalidPeer
                                     }
                                 />


views/OpenChannel.tsx
```diff
@@ -503,28 +503,26 @@ export default class OpenChannel extends React.Component<
...
+                                        isNodeHostValid: true
+                                    });
                                 } else {
                                     this.setState({
                                         channelDestination: 'Custom',
                                         node_pubkey_string: '',
-                                        host: ''
+                                        host: '',
+                                        isNodePubkeyValid: false,
+                                        isNodeHostValid: true
```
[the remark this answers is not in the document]
myxmaster: correct. actually, i will make a cleaner fix. will force push.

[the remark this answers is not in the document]
myxmaster: done, PR description updated

views/OpenChannel.tsx
```diff
@@ -1116,11 +1107,14 @@ export default class OpenChannel extends React.Component<
                                         },
                                         false,
                                         connectPeerOnly
-                                    );
+                                    ).catch(() => {});
```
kaloudis: should we not display an error here if an error is thrown?

myxmaster: This catch is just a minimal fix to suppress the unhandled promise rejection. I think a clean solution belongs in a separate PR (see [this issue](https://github.com/ZeusLN/zeus/issues/3814) as ToDo).

In short:
There is a deeper inconsistency, which needs to be fixed. `connectPeer` communicates errors through two channels simultaneously: MobX observables (for the UI) and `reject()` (for programmatic use), but `openChannel` only uses observables. No caller actually uses the rejection. So the clean solution would be to align both methods: always
`resolve(true/false)`, never `reject()`, and let callers react to the boolean return value.

myxmaster:
(rebased)

myxmaster:
(rebased)

views/OpenChannel.tsx
```diff
@@ -865,7 +862,8 @@ export default class OpenChannel extends React.Component<
                                                                 implementation ===
                                                                     'cln-rest'
                                                                     ? 'all'
-                                                                    : ''
+                                                                    : '',
+                                                            satAmount: '0'
```
kaloudis: why isn't this set to an empty string instead?

kaloudis (changes requested):

views/OpenChannel.tsx
myxmaster: ah you're right

myxmaster:
(rebased)

views/OpenChannel.tsx
```diff
@@ -363,6 +360,8 @@ export default class OpenChannel extends React.Component<
 
         const loading = connectingToPeer || openingChannel;
 
+        const isNodeHostValid =
```
ajaysehwal, line 363: @myxmaster, ValidationUtils.validateNodeHost returns false for empty strings, but isNodeHostValid ends up treating them as valid in isInvalidPeer. This logic seems incorrect. Could you please take another look?

myxmaster: `validateNodeHost` deliberately returns `false` for empty strings because an empty string is not a valid host address format. However, providing a host is optional in this context here (LND and CLN both support connecting to a peer by pubkey only).
For this reason, `isNodeHostValid` is computed as `host === '' || ValidationUtils.validateNodeHost(host)` -> empty is valid in this view (gossip lookup will be attempted).

kaloudis:
Concept ACK. Both fixes are correct: verified the `AmountInput.componentDidUpdate` else-branch never calls `onAmountChange`, and the derived `isNodeHostValid` matches the `isInvalidFeeRate` idiom.

One blocking item: this branch predates a fix on master, which added `isNodeHostValid` writes in `initFromProps()` and `importClipboard()`. Please rebase.

views/OpenChannel.tsx
```diff
@@ -682,13 +728,15 @@ export default class OpenChannel extends React.Component<
                                                 'Olympus by ZEUS',
                                             node_pubkey_string:
                                                 config.lsps1Pubkey,
-                                            host: config.lsps1Host
+                                            host: config.lsps1Host,
+                                            isNodePubkeyValid: true
```
kaloudis, line 732: nit: safer as `ValidationUtils.validateNodePubkey(config.lsps1Pubkey)`, consistent with `initFromProps` on master and guards an empty LSP config for the active network.

views/OpenChannel.tsx
```diff
@@ -1269,12 +1357,15 @@ export default class OpenChannel extends React.Component<
...
                                     }}
                                     disabled={
                                         loading ||
                                         (!connectPeerOnly &&
                                             isInvalidFeeRate) ||
+                                        (!connectPeerOnly &&
+                                            !fundMax &&
+                                            !Number(satAmount)) ||
                                         isInvalidPeer
                                     }
```
kaloudis, line 1370: nit: the icon color above still keys off `isInvalidFeeRate || isInvalidPeer` only, so when disabled purely by the new empty-amount guard the icon renders in the enabled color. Consider reusing the same expression as `disabled`.

stores/ChannelsStore.ts
```diff
@@ -761,7 +742,7 @@ export default class ChannelsStore {
         this.channelRequest = undefined;
         if (!silent) this.connectingToPeer = true;
 
-        if (!request.host) {
+        if (!request.host && !connectPeerOnly) {
```
kaloudis, line 745: Optional: a small store test (host/no-host x connectPeerOnly, asserting whether `channelRequest` gets set) would lock this in.

PR #3774, opened 2026-03-01
NostrContacts: fix layout and multiple bugs

myxmaster:
# Description

Multiple layout and bug fixes:
- Clean layout concept for contact rows
  - No more negative positioning/margin hacks
  - Long contacts are no longer cut off, but wrap correctly
- ~~Using `LayoutAnimation` instead of `Animated`/`Easing` (lets the OS handle layout transitions natively, no manual width interpolation needed)~~
- Fix white checkbox background
- Selected checkbox uses `themeColor('highlight')`
- `SelectButton` icon in header uses `themeColor`, correct height and is vertically centered
- `Close` icon in header removed -> `Back` icon (and hardware back button) now handle everything: first clear loaded data, then navigate back
- Fixed identification bug for selecting contacts -> now using `npub` instead of `banner` (before: when you selected a contact without banner, all contacts without banner were selected)

**Please test if behavior is fine on iOS!**

|Before|After|
|-|-|
|grafik|grafik|

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -316,7 +316,8 @@ export default class App extends React.PureComponent {
         'SendingLightning',
         'SendingOnChain',
         'CashuSendingLightning',
-        'IntroSplash'
+        'IntroSplash',
+        'NostrContacts'
     ];
 
     private handleBackPress = (navigation: any) => {
--- a/views/NostrContacts.tsx
+++ b/views/NostrContacts.tsx
@@ -46,10 +46,10 @@ interface NostrContactsProps {
 
 interface NostrContactsState {
     account: string;
-    contactsData: any[];
+    contactsData: NostrContact[];
     loading: boolean;
     isSelectionMode: boolean;
-    selectedContacts: any[];
+    selectedContacts: SelectedContact[];
     isValid: boolean;
     isValidNpub: boolean;
     isValidNip05: boolean;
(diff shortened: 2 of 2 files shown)


shubhamkmr04 (commented):
we should see a smooth transition when we toggle the checkboxes

reference:

https://github.com/user-attachments/assets/215264c8-5931-4d85-9415-bd30bff4f798

myxmaster:
ah ok, turns out this animation was never visible on android (even in a compiled release build, so not just a debugging performance issue).

so i went back to `Animated` and now it works an android as well (tested by temporarily increasing the duration), because the `Animated.View` is always mounted (no conditional rendering) and `extraData` ensures FlatList re-renders items when state changes -> so the Animated system can actually reach the views.

shubhamkmr04:
> ah ok, turns out this animation was never visible on android (even in a compiled release build, so not just a debugging performance issue).
> 
> so i went back to `Animated` and now it works an android as well (tested by temporarily increasing the duration), because the `Animated.View` is always mounted (no conditional rendering) and `extraData` ensures FlatList re-renders items when state changes -> so the Animated system can actually reach the views.

still can't see that transition in both android and iOS

myxmaster:
during debugging i cannot see any animation as well. try using `duration: 2000` to make sure it generally works.

but i did change `duration: 100` to `200`, because i think 100 ms is a bit short anyway.

myxmaster:
(rebased)

shubhamkmr04:
@myxmaster please rebase again

myxmaster:
done

shubhamkmr04:
> during debugging i cannot see any animation as well. try using `duration: 2000` to make sure it generally works.
> 
> but i did change `duration: 100` to `200`, because i think 100 ms is a bit short anyway.

changing it to 2000 does make it visible while transition but its not smooth

myxmaster:
Alright, I now tested on a physical device (not emulator). Turns out the original implementation was already problematic on older hardware (noticeable latency when entering selection mode and when tapping individual checkboxes).

Clean fix: Extracted `ContactItem` as a `React.memo` component with stable prop references, so only the one item whose `isSelected` changes actually re-renders. Combined with `useNativeDriver: true` for the slide animation, performance is now significantly improved.

Please test again.

shubhamkmr04 (approved):
tACK

kaloudis (approved):

Issue #3770, opened 2026-02-28
Add loading indicator for "Connect Peer"


myxmaster:
### Describe your enhancement idea

We should add a loading indicator to the "Connect Peer" flow. In case it takes some time (slow connection, or peer not answering etc.) it might be confusing for the user.

LSUDOKO:
hi @myxmaster 
i would like to work on this issue 
can you assign me this issue ??

myxmaster:
I cannot assign anybody, and I know that @kaloudis only assigns team members.
You are free to make a PR though, just make sure to read the [contribution guidelines](https://github.com/ZeusLN/zeus/blob/master/CONTRIBUTING.md).

PR #3785, opened 2026-03-04
fix: add loading indicator to Connect Peer flow

a-khushal:
# Description

Relates to issue: [**ZEUS-3770**](https://github.com/ZeusLN/zeus/issues/3770)

image

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- x ] I’ve run `yarn run tsc` and made sure my code compiles correctly

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -217,8 +217,7 @@ dependencies {
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.2.0-alpha01'
     implementation files("../../node_modules/react-native-tor/android/libs/sifir_android.aar")
     // gif
-    implementation 'com.facebook.fresco:fresco:3.1.3'
-    implementation 'com.facebook.fresco:animated-gif:3.1.3'
+    implementation 'com.facebook.fresco:animated-gif:3.6.0'
 
     // Pegasus
     // implementation(name:"Lndmobile", ext:"aar")
--- a/components/LightningIndicator.tsx
+++ b/components/LightningIndicator.tsx
@@ -7,7 +7,7 @@ interface LightningIndicatorProps {
     size?: number;
 }
 
-function LoadingIndicator(props: LightningIndicatorProps) {
+function LightningIndicator(props: LightningIndicatorProps) {
     const { size } = props;
 
     return (
@@ -22,4 +22,4 @@ function LoadingIndicator(props: LightningIndicatorProps) {
     );
 }
 
-export default LoadingIndicator;
+export default LightningIndicator;


kaloudis:
not sure this is the best place for it as it doesn't match the Open Channel flow. Curious to hear opinions from @myxmaster

myxmaster:
hm yeah, i think it should match the loading indicator from the Open Channel flow.

kaloudis:
@a-khushal can we get an updated screenshot or video?

a-khushal:
> @a-khushal can we get an updated screenshot or video?

@kaloudis the description has the updated screenshot

myxmaster:
Hm, I don't think we should change the loading indicator type. The lightning circle looks quite nice.

Anyway, I just looked into it and noticed that the loading indicator is not missing for connecting a peer. The actual problem seems to be that the `LightningIndicator` GIF is invisible on Android: it takes up layout space but renders nothing.

We already use a JSON-based animation elsewhere with no issues, so I suggest we convert the GIF to a Lottie JSON animation instead.

a-khushal:
image

@myxmaster this works?

myxmaster:
Yes, that does work, but I have one more suggestion:
The original GIF, and also your lottie json, have a lot of free space (transparent padding) around the actual lightning circle, inside the image file I mean. Here is a cropped version (which also reduces the file size from 545 kB to 507 kB): 
[lightning-circle.json](https://github.com/user-attachments/files/25817966/lightning-circle.json)

-> If we decide to use lottie json for this animation, we should use this cropped version and adjust usage in the files accordingly (render it `0.5` times smaller, so the effective size in the app is unchanged; and probably add margin):
- SetFeesForm
- EditFee (only here we use size 200 instead of 100)
- LnurlAuth
- LnurlChannel
- OpenChannel

@kaloudis Do you think 500 kB vs 100 kB (original GIF) is an issue? We could probably also find a way to get GIF animations working again. Another option would be to make a vector graphics lottie json, but I'm not sure I will be able to do it so it looks really similar to before.

kaloudis:
> Yes, that does work, but I have one more suggestion: The original GIF, and also your lottie json, have a lot of free space (transparent padding) around the actual lightning circle, inside the image file I mean. Here is a cropped version (which also reduces the file size from 545 kB to 507 kB): [lightning-circle.json](https://github.com/user-attachments/files/25817966/lightning-circle.json)
> 
> -> If we decide to use lottie json for this animation, we should use this cropped version and adjust usage in the files accordingly (render it `0.5` times smaller, so the effective size in the app is unchanged; and probably add margin):
> 
>     * SetFeesForm
> 
>     * EditFee (only here we use size 200 instead of 100)
> 
>     * LnurlAuth
> 
>     * LnurlChannel
> 
>     * OpenChannel
> 
> 
> @kaloudis Do you think 500 kB vs 100 kB (original GIF) is an issue? We could probably also find a way to get GIF animations working again. Another option would be to make a vector graphics lottie json, but I'm not sure I will be able to do it so it looks really similar to before.

I would prefer to keep the gif and ensure it's working. I think increasing the size of the file 5x would only be acceptable if it significantly improved UX, but that doesn't seem to be the case here.

myxmaster:
Yeah...
@a-khushal can you debug android and check how GIF usage can be fixed? otherwise please close this PR and i will make a new one.

a-khushal:
> Yeah... @a-khushal can you debug android and check how GIF usage can be fixed? otherwise please close this PR and i will make a new one.

seems like the GIF was invisible due to Fresco version mismatch, we had `animated-gif:3.1.3` but RN 0.84.0 pulls in Fresco 3.6.0 by default, so `animated-gif` should match 3.6.0. hence, I removed `fresco:3.1.3`, and updated `animated-gif to 3.6.0` in `android/app/build.gradle`. after clean and rebuild, the GIF shows up correctly like this,

image


cc @myxmaster, @kaloudis

myxmaster:
tACK

PR #3793, opened 2026-03-05
enhancement: Refresh channel data after payment sent or received

myxmaster:
# Description

Channel capacity in the channels pane was showing stale data after sending or receiving a Lightning payment, requiring a manual pull-to-refresh.

This adds a `getChannels()` call alongside the existing `getCombinedBalance()` call that already fires on payment success, so channel data is up to date by the time the user navigates to the channels pane.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

- [ ] Embedded LND
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -777,7 +777,8 @@ export default class Receive extends React.Component<
             PosStore,
             SettingsStore,
             NodeInfoStore,
-            BalanceStore
+            BalanceStore,
+            ChannelsStore
         } = this.props;
         const { orderId, orderTotal, orderTip, exchangeRate, rate, value } =
             this.state;
@@ -824,6 +825,7 @@ export default class Receive extends React.Component<
                                     Number(invoice.amt_paid_sat)
                                 );
                                 BalanceStore.getCombinedBalance();
+                                ChannelsStore.getChannels();
 
                                 if (orderId) {
                                     PosStore.recordPayment({
@@ -943,6 +945,7 @@ export default class Receive extends React.Component<
                                 if (result.settled) {
                                     setWatchedInvoicePaid(result.amt_paid_sat);
                                     BalanceStore.getCombinedBalance();
+                                    ChannelsStore.getChannels();
 
                                     if (orderId) {
                                         PosStore.recordPayment({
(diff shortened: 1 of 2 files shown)


views/Receive.tsx
kaloudis, line 1: In SendingLightning.tsx:170-172, the `ChannelsStore.getChannels()` call is wrapped in a `BackendUtils.supportsChannelManagement()` check. However, in this file, all four new `ChannelsStore.getChannels()` calls are made unconditionally.

myxmaster: Because they are all inside `if (implementation === '<backend>')` and it is therefore safe already.

But in regards to code maintenance this is not optimal. Maybe it is worth considering a minor refactor: `ChannelsStore.getChannels()` (and the surrounding post-payment logic) is currently duplicated across the different implementation branches.

We could extract a `handlePaymentReceived` helper method that centralizes all post-payment actions: `setWatchedInvoicePaid`, `getCombinedBalance`, `getChannels`, and POS recording. The `BackendUtils.supportsChannelManagement()` guard would live there too, making it obvious for anyone adding a new backend that not all operations are universally supported.

Each branch would then just call `this.handlePaymentReceived(amount, { type: 'ln', tx: ..., preimage: ... })` instead of repeating all those lines.

kaloudis: I like this idea. We have to do some similar refactoring for the start-up process in Wallet.tsx

myxmaster: Yes, but I think those refactors should be a different PR (or even 2 PRs?), or do you want this all in here?

kaloudis: definitely separately

myxmaster: 👍

but then i think we don't need to implement the `BackendUtils.supportsChannelManagement()` guard here, and i will do the refactor on top of this PR, ok?

myxmaster: https://github.com/ZeusLN/zeus/pull/3822

myxmaster: > We have to do some similar refactoring for the start-up process in Wallet.tsx

see here: https://github.com/ZeusLN/zeus/pull/3826

kaloudis (approved):

kaloudis (approved):

Issue #3878, opened 2026-03-24
No submit button for SCB backup key


kushagra0902:
### Describe the bug

A bug happened!
As of now there is no submit button for SCB backup key and thus preventing user from restoring the wallet using the key. The issue is just with the key submission and rest logic exists

### Reproduce

Reproduce the bug step by step.
Go to create or restore wallet section. Go to ~restroe wallet` section and In the key input section enter the key. No option for submitting the key is present. 

https://github.com/user-attachments/assets/7056ac58-cc71-4d3e-9fc5-96dcfead541a

### ZEUS version

None

### Node interface

Multiple (specified in comments)

### Network

Clearnet

### Device

Nothing Phone 2a

### Device operating system

Android 15

### Log output

```shell

```

kushagra0902:
I do have solved this bug in my local and will make a commit for this along with an another issue

kaloudis:
@kushagra0902 can you submit this as its own PR?

kushagra0902:
> [@kushagra0902](https://github.com/kushagra0902) can you submit this as its own PR?

done. Kindly review it and lmk the changes required if any.

kaloudis:
Fixed in https://github.com/ZeusLN/zeus/pull/3978

PR #3898, opened 2026-03-26
Added a submit button for SCB backup screen

kushagra0902:
# Description

Relates to issue: https://github.com/ZeusLN/zeus/issues/3878

SCB recovery page did not have a submit button. This made entering a SCB backup impossible for the users. This PR added the button which takes the user back to recovery page where they can add the mnemonic words and backup their wallet. The previous UI is untouched and just the button is added accordingly.

- [ ] New feature
- [ x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [ x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [ x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [ x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [ x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ x] N/A

- [ x] Android
- [ ] iOS

- [ ] Embedded LND
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

Screenshot from 2026-03-27 01-00-01


kaloudis:
Can we instead get this to work by hitting the enter key on the keyboard?

kushagra0902:
> Can we instead get this to work by hitting the enter key on the keyboard?

Done. Now both button as well as keyboard enter works fine!

myxmaster:
I was thinking about this screen for a while, and I think we should change the UI to a 2 tabs screen:
1st tab: seed word recovery
2nd tab: SCB recovery

Then have a clean screen each for the 2 tabs, with exactly the logic and buttons we need for each case. (Another example for the issues regarding the mixed logic: When you start typing in the SCB input, seed word suggestions are displayed...)

kushagra0902:
> I was thinking about this screen for a while, and I think we should change the UI to a 2 tabs screen: 1st tab: seed word recovery 2nd tab: SCB recovery
> 
> Then have a clean screen each for the 2 tabs, with exactly the logic and buttons we need for each case. (Another example for the issues regarding the mixed logic: When you start typing in the SCB input, seed word suggestions are displayed...)

Got your point and I was thinking something in similar direction:
1) I think the current flow of channel only backup is not intuitive for the users. So instead of a tab in the same screen, we should separate both things entirely. 
2) This will not disrupt the logic anyways as the SCB restore occurs only once wallet restore is done(correct me please if I am wrong). 
3) In this flow thus, once the wallet backup is done, the user can go the channel backup section in the menu and then put the SCB key there. This will also solve the recommendation error pointed above(sorry for that).


Please give your suggestions on this and I will start the work accordingly

myxmaster:
Or like this:
Restore Button -> Seed word input (like it is now) -> when all seed words are entered, and valid, a CONTINUE button becomes active, on the next screen user can provide SCB, but that is optional, then hit final RESTORE button.

kushagra0902:
> Or like this: Restore Button -> Seed word input (like it is now) -> when all seed words are entered, and valid, a CONTINUE button becomes active, on the next screen user can provide SCB, but that is optional, then hit final RESTORE button.

I think this will make restoring only the channel for a wallet already restored difficult. So I think making a different section for this. I will implement both and will make the PR for the better one so any can be used

myxmaster:
So you mean, somebody restores the wallet (without providing SCB), then later decides to use the SCB? Hm, I guess this is not impossible.

kushagra0902:
> So you mean, somebody restores the wallet (without providing SCB), then later decides to use the SCB? Hm, I guess this is not impossible.

yes, not sure if this should be done, but someone may wnat to do so

kushagra0902:
> > So you mean, somebody restores the wallet (without providing SCB), then later decides to use the SCB? Hm, I guess this is not impossible.
> 
> yes, not sure if this should be done, but someone may want to do so. Actually my bad, you were right, there is already option to restore the channel individually at disaster recovery section so need of that.

kaloudis:
> So you mean, somebody restores the wallet (without providing SCB), then later decides to use the SCB? Hm, I guess this is not impossible.

It is possible. Check out how Disaster Recovery works.

myxmaster:
Technically of course, I was trying to say, it is an unusual behavior.

Anyway, yes, there should be a way to provide SCB directly (without going through restore process with seed words at the same time). Maybe we should just add the SCB flow separately in Menu -> Tools?

kushagra0902:
> Technically of course, I was trying to say, it is an unusual behavior.
> 
> Anyway, yes, there should be a way to provide SCB directly (without going through restore process with seed words at the same time). Maybe we should just add the SCB flow separately in Menu -> Tools?

it is already there in settings -> embedded node -> disaster recovery -> initiate recovery from clipboard.

so I think your method of "giving all the words. then continue, recover" can be done along with this separate page can be done

kushagra0902:
> > Technically of course, I was trying to say, it is an unusual behavior.
> > Anyway, yes, there should be a way to provide SCB directly (without going through restore process with seed words at the same time). Maybe we should just add the SCB flow separately in Menu -> Tools?
> 
> it is already there in settings -> embedded node -> disaster recovery -> initiate recovery from clipboard.
> 
> so I think your method of "giving all the words. then continue, recover" can be done along with this separate page can be done

something like this:

https://github.com/user-attachments/assets/5c5b484c-0a76-42c4-8ddf-b1687d4a5a21

myxmaster:
> it is already there in settings -> embedded node -> disaster recovery -> initiate recovery from clipboard.

ah right, makes sense.

regarding your video in the last comment:
i think that's good!

nit: don't put the `(Optional)` in the placeholder, but in the label:
"Optional: Disaster recovery data (SCB, Base64):"

and the placeholder should probably say "Your SCB data in Base64 format". after https://github.com/ZeusLN/zeus/pull/3886 is merged "Your SCB data in Base64 or HEX format"

kushagra0902:
> > it is already there in settings -> embedded node -> disaster recovery -> initiate recovery from clipboard.
> 
> ah right, makes sense.
> 
> regarding your video in the last comment: i think that's good!
> 
> nit: don't put the `(Optional)` in the placeholder, but in the label: "Optional: Disaster recovery data (SCB, Base64):"
> 
> and the placeholder should probably say "Your SCB data in Base64 format". after #3886 is merged "Your SCB data in Base64 or HEX format"

done!

shubhamkmr04:
@kushagra0902  rebase needed

kushagra0902:
> @kushagra0902 rebase needed

done

shubhamkmr04 (commented):
- change the commit msg and author
- we only need changes in en.json

views/Settings/SeedRecovery.tsx
```diff
@@ -133,6 +133,7 @@ interface SeedRecoveryState {
     ldkPassphrase: string;
     ldkNodeDir: string;
     embeddedLdkNetwork: string;
+    step: 1 | 2;
```
kaloudis, line 136: what do these steps mean? they should be enums

kaloudis (changes requested):

shubhamkmr04:
rebase needed

kaloudis:
Closing in favor of https://github.com/ZeusLN/zeus/pull/3978

PR #3908, opened 2026-03-28
Embedded LND: v0.20.1-beta-zeus

kaloudis:


kaloudis:
Tested with Oakrerun + ziggie

Issue #3911, opened 2026-03-28
getSwapFees() called twice on startup


myxmaster:
### Describe your enhancement idea

On every app startup `getSwapFees()` is called twice:
1. Explicitly in `Wallet.tsx` after `getNodeInfo()` completes
2. Via the MobX `reaction` in `SwapStore` constructor which fires at the same moment, because `nodeInfo` being populated for the first time changes the value of `getHost`

One of the two call sites should be removed...

Generally: should swap fees even be fetched at startup at all? They're only needed when the user intends to use swaps.

a-khushal:
we should remove the duplicate `getSwapFees()` call, but not remove the fee fetching, fees are used in `Send` and `Payment Request` views, and lazy loading can make swap related ui showing up late, which might not be the best ux

myxmaster:
but fees can get old fast, so they should be updated right before usage anyway? and updating is not faster than the initial call, so i still think lazy loading makes sense.

of course a loading indicator should be displayed together with "updating current fees", so user understands whats up... it is better than a failed swap because fees were old (happens quite fast on boltz e.g.).

a-khushal:
> but fees can get old fast, so they should be updated right before usage anyway? and updating is not faster than the initial call, so i still think lazy loading makes sense.
> 
> of course a loading indicator should be displayed together with "updating current fees", so user understands whats up... it is better than a failed swap because fees were old (happens quite fast on boltz e.g.).

i’ve made this change in [2bc656d6d](https://github.com/ZeusLN/zeus/pull/3920/changes/2bc656d6dca22265362c91d767cc37a452ff9025)
it lazy loads fees in Swaps view, refreshes again before intiating swap and also it shows loading indicator + `Updating current fees...`, and blocks swap creation if refresh fails.

kaloudis:
Resolved by #4399 (commit 1bab3c57e): the explicit startup call in Wallet.tsx was removed, and the SwapStore reaction is now guarded by fetchedRatesHost so it cannot fire during boot as settings load and node info resolves. Swap rates are now only fetched from the Swaps view, with a refetch on focus when the provider changes. fetchAndUpdateSwaps still runs at startup but only polls the status of existing stored swaps and never fetches rates.

PR #3920, opened 2026-03-30
fix: remove duplicate swap fee fetch on startup

a-khushal:
# Description

Relates to issue: [**ZEUS-3911**](https://github.com/ZeusLN/zeus/issues/3911)

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

On-device
- [x] Embedded LND
- [ ] LDK Node

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -2374,6 +2374,7 @@
     "views.Swaps.rescueKey.delete": "Delete rescue key",
     "views.Swaps.intro.description": "Using the Swap service, you can move funds from on-chain to Lightning, and vice versa, without having to open or close a channel.",
     "views.Swaps.fetchFeesFailed": "Failed to fetch swap fees. Please check your connection and try again.",
+    "views.Swaps.updatingCurrentFees": "Updating fees...",
     "views.Swaps.intro.continue": "Continue",
     "views.Swaps.rescueKey.deleteConfirmation": "Are you sure you want to delete the rescue key? All swaps will also be deleted from your swap history. Please back up your keys before proceeding, as you can use them later to recover those swaps.",
     "views.SwapDetails.title": "Swap Details",
--- a/locales/en.json
+++ b/locales/en.json
@@ -2391,6 +2391,7 @@
     "views.Swaps.rescueKey.delete": "Delete rescue key",
     "views.Swaps.intro.description": "Using the Swap service, you can move funds from on-chain to Lightning, and vice versa, without having to open or close a channel.",
     "views.Swaps.fetchFeesFailed": "Failed to fetch swap fees. Please check your connection and try again.",
+    "views.Swaps.updatingCurrentFees": "Updating fees...",
     "views.Swaps.intro.continue": "Continue",
     "views.Swaps.rescueKey.deleteConfirmation": "Are you sure you want to delete the rescue key? All swaps will also be deleted from your swap history. Please back up your keys before proceeding, as you can use them later to recover those swaps.",
     "views.SwapDetails.title": "Swap Details",
--- a/locales/en.json
+++ b/locales/en.json
@@ -2466,6 +2466,7 @@
     "views.Swaps.rescueKey.delete": "Delete rescue key",
     "views.Swaps.intro.description": "Using the Swap service, you can move funds from on-chain to Lightning, and vice versa, without having to open or close a channel.",
     "views.Swaps.fetchFeesFailed": "Failed to fetch swap fees. Please check your connection and try again.",
+    "views.Swaps.updatingCurrentFees": "Updating fees...",
     "views.Swaps.intro.continue": "Continue",
     "views.Swaps.rescueKey.deleteConfirmation": "Are you sure you want to delete the rescue key? All swaps will also be deleted from your swap history. Please back up your keys before proceeding, as you can use them later to recover those swaps.",
     "views.SwapDetails.title": "Swap Details",
--- a/stores/SwapStore.ts
+++ b/stores/SwapStore.ts
@@ -1,4 +1,4 @@
-import { action, observable, computed, runInAction, reaction } from 'mobx';
+import { action, observable, computed, runInAction } from 'mobx';
 import ReactNativeBlobUtil from 'react-native-blob-util';
 import { ECPairAPI, ECPairFactory } from 'ecpair';
 import ecc from '@bitcoinerlab/secp256k1';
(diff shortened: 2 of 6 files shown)


ajaysehwal (commented):
https://github.com/a-khushal/zeus/blob/360e23986870b2b9a6ce6e186b3f2f173d2fc2fa/stores/SwapStore.ts#L54

@a-khushal, I think we should add { fireImmediately: true } to the getHost → getSwapFees reaction so the first fee load still runs when getHost doesn’t change after connect.

locales/en.json
```diff
@@ -2374,6 +2374,7 @@
     "views.Swaps.rescueKey.delete": "Delete rescue key",
     "views.Swaps.intro.description": "Using the Swap service, you can move funds from on-chain to Lightning, and vice versa, without having to open or close a channel.",
     "views.Swaps.fetchFeesFailed": "Failed to fetch swap fees. Please check your connection and try again.",
+    "views.Swaps.updatingCurrentFees": "Updating current fees...",
```
myxmaster: nit: "Updating fees..." is enough imo (otherwise it is kind of redundant)

myxmaster:
There is a regression in `Send.tsx` and `PaymentRequest.tsx`: both views consume `SwapStore.reverseInfo`/`subInfo` to decide whether to show the swap icon in header, but neither fetches up-to-date fees/data themselves. Since startup no longer preloads data from swap provider, we would not see the swap icon, although amount is inside the min/max limits for a swap (`isAmountValidToSwap` returns `false` silently because the store data is empty).

My commit https://github.com/myxmaster/zeus/commit/a0f742a5b2e37ee0828274360ae60c14e3cca701 (cherry-pick if you want) fixes this by:
- `Send.tsx`: triggering `getSwapFees()` in `componentDidUpdate` when `transactionType` resolves to 'On-chain' (lazy, only when a swap is actually relevant), and moving the `isAmountValidToSwap` check out of local state into `render()` directly so MobX reactivity kicks in when swap fees/data arrive
- `PaymentRequest.tsx`: same approach, `getSwapFees()` fire-and-forget in `componentDidMount`, and `isAmountValidToSwap()` evaluated in `render()` (not in the nested SwapButton sub-component, where MobX tracking wouldn't apply) so the button appears automatically once fees load

a-khushal:
> There is a regression in `Send.tsx` and `PaymentRequest.tsx`: both views consume `SwapStore.reverseInfo`/`subInfo` to decide whether to show the swap icon in header, but neither fetches up-to-date fees/data themselves. Since startup no longer preloads data from swap provider, we would not see the swap icon, although amount is inside the min/max limits for a swap (`isAmountValidToSwap` returns `false` silently because the store data is empty).
> 
> My commit [myxmaster@a0f742a](https://github.com/myxmaster/zeus/commit/a0f742a5b2e37ee0828274360ae60c14e3cca701) (cherry-pick if you want) fixes this by:
> 
> * `Send.tsx`: triggering `getSwapFees()` in `componentDidUpdate` when `transactionType` resolves to 'On-chain' (lazy, only when a swap is actually relevant), and moving the `isAmountValidToSwap` check out of local state into `render()` directly so MobX reactivity kicks in when swap fees/data arrive
> * `PaymentRequest.tsx`: same approach, `getSwapFees()` fire-and-forget in `componentDidMount`, and `isAmountValidToSwap()` evaluated in `render()` (not in the nested SwapButton sub-component, where MobX tracking wouldn't apply) so the button appears automatically once fees load


I cherry-picked your fix ([`a0f742a`](https://github.com/myxmaster/zeus/commit/a0f742a5b2e37ee0828274360ae60c14e3cca701)) into this branch.

myxmaster:
Ah, we should also handle the empty initial state of `reverseInfo`/`subInfo` in `isAmountValidToSwap` in both `Send.tsx`:
```
if (!reverseInfo || Object.keys(reverseInfo).length === 0) {
```

and `PaymentRequest.tsx`:
```
if (!subInfo || Object.keys(subInfo).length === 0) {
```

Otherwise the swap icon is shortly displayed before fees are loaded.

Besides that:
tACK

shubhamkmr04 (commented):
https://github.com/user-attachments/assets/26aa31dd-bef4-44fa-bff8-081bb0614fc2


here on this flow, fees are fetched 2 times, once on the PaymentRequest view and when we go on to swap the amount

stores/SwapStore.ts
```diff
@@ -30,34 +30,9 @@ import Storage from '../storage';
...
-
-interface ReverseSwapInfo {
-    fees?: {
-        percentage?: number;
-        minerFees?: {
-            claim?: number;
-            lockup?: number;
-        };
-    };
-    limits?: {
```
shubhamkmr04: I don't think we should remove this. If the user changes hosts and comes back, they still see the old data, and the values differ so much between hosts, for ZEUS the min is ~25K sats and for SwapMarket it's ~100K sats. So the user fills the swap form with the wrong service fees and min/max values displayed.

Also, on what basis are users going to change hosts if we keep this flow? A user changes hosts expecting to see updated fees right away, telling them fees changed only after they've already initiated the swap based on old rates is not a great experience imo.

a-khushal:
> here on this flow, fees are fetched 2 times, once on the PaymentRequest view and when we go on to swap the amount

yeah it's done once in `PaymentRequest` (to show swap icon) and again before swap execution (to avoid stale fees). 

the second fetch is intentional for safety against stale fee values, but we can add dedupe or TTL in swapstore to prevent multiple calls within few seconds (like 10s)

stores/SwapStore.ts
a-khushal: yes,

we can keep the current changes and also refresh fees immediately on host change or clear old fee data until new data arrieves from a different host..

kaloudis:
Thanks for working on this, and for prefiguring the lazy-loading approach here. This was superseded by #4399 (commit 1bab3c57e), which removed the startup fetch entirely: rates are now only fetched from the Swaps view, with a refetch on focus when the provider changes. Closing along with #3911.

Issue #3833, opened 2026-03-13
Use app language to determine number separators consistently across the app


myxmaster:
### The problem
Zeus currently mixes two different number formatting conventions on the same screen. A user with currency `EUR` selected sees this in the wallet balance pane:
`1,234,567.321sats`
`≈€779,54`

The comma means "thousands separator" in the sats line and "decimal separator" in the fiat line, simultaneously on the same screen. For anyone outside the English-speaking world this is quite confusing.

### Root cause
BTC and sat amounts are always formatted in English style (comma = thousands, period = decimal), hardcoded in `numberWithCommas()` and `formatBitcoinWithSpaces()` in `utils/UnitsUtils.ts`.

Fiat amounts are formatted based on the selected **currency**: We currently use `separatorSwap: true/false` in `FiatStore`, which can swap the separators to European style.

So the formatting for sats and BTC never adapts, while fiat adapts to the currency, not the language. These two decisions lead to a conflict on the same screen.

### Proposed fix
Use the app's selected **language** as the single source of truth for separator style, applied consistently to all amount types (sats, BTC, fiat).

Languages that use European-style separators -> period = thousands, comma = decimal
Languages that use English-style separators -> comma = thousands, period = decimal

This makes the choice predictable and consistent: a user who set the app to German always sees German number formatting, everywhere, regardless of which fiat currency or Bitcoin unit they are viewing.

### Scope
- `utils/UnitsUtils.ts`: `numberWithCommas`, `numberWithDecimals`, and `formatBitcoinWithSpaces` need to become locale-aware.
- `utils/AmountUtils.ts` and `components/Amount.tsx`: `getUnformattedAmount` passes `separatorSwap` through to `Amount`, which uses it to decide between `numberWithCommas` and `numberWithDecimals`. This flag would be replaced by a language-derived decision. If formatting functions become locale-aware directly, the flag concept may disappear entirely.
- `stores/FiatStore.ts`: the per-currency `separatorSwap` field would become redundant for display purposes and could be removed.
- The custom keypad (`KeypadAmountDisplay`, `PinPad`) currently adapts its decimal separator character only for fiat, driven by the currency's `separatorSwap` flag. For sats and BTC the decimal button is hardcoded to `.`. Under the new approach all three units would use the language-derived decimal separator.
- `views/Activity/Activity.tsx` and `components/CurrencyList.tsx` both read `separatorSwap` from `FiatStore` to format exchange rates and would need to be updated.
[...]

PR #3927, opened 2026-03-30
utils: use app locale for number separators

cherry-1729-9090:
# Description

Relates to issue: #3833

This PR makes number separators follow the app language across the app.
Screenshot 2026-03-31 at 2 03 53 AM
Screenshot 2026-03-31 at 2 04 58 AM

Before this change, Bitcoin and sats were always shown in English-style formatting, while fiat formatting depended on the selected currency. That could show mixed separator styles on the same screen.

With this change, the selected app language is used as the single source of truth for number separators for:
- sats
- BTC
- fiat
- exchange rates
- keypad decimal display
- related amount input and conversion flows

This keeps formatting consistent when switching between languages like English and German.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [ ] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/Amount.tsx
+++ b/components/Amount.tsx
@@ -13,11 +13,7 @@ import LoadingIndicator from './LoadingIndicator';
 
 import { localeString, formatInlineNoun } from '../utils/LocaleUtils';
 import { themeColor } from '../utils/ThemeUtils';
-import {
-    formatBitcoinWithSpaces,
-    numberWithCommas,
-    numberWithDecimals
-} from '../utils/UnitsUtils';
+import { formatBitcoinWithSpaces, numberWithCommas } from '../utils/UnitsUtils';
 import type { Units } from '../utils/UnitsUtils';
 import PrivacyUtils from '../utils/PrivacyUtils';
 import { processSatsAmount, getUnformattedAmount } from '../utils/AmountUtils';
@@ -49,7 +45,6 @@ interface AmountDisplayProps {
     accessible?: boolean;
     accessibilityLabel?: string;
     roundAmount?: boolean;
-    separatorSwap?: boolean;
     onPendingPress?: () => void;
 }
 
@@ -76,7 +71,6 @@ function AmountDisplay({
     accessible,
     accessibilityLabel,
     roundAmount = false,
-    separatorSwap = false,
     onPendingPress
 }: AmountDisplayProps) {
     if (unit === 'fiat' && !symbol) {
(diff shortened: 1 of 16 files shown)


components/Amount.tsx
```diff
@@ -370,6 +360,48 @@ interface AmountProps {
...
+
+        const hasSelectedFiatRate = FiatStore.fiatRates?.some(
+            (entry: any) => entry.code === selectedFiat
+        );
+
+        if (!hasSelectedFiatRate) {
+            FiatStore.getFiatRates();
+        }
+    };
+
```
kaloudis, line 404: why are these changes needed?

kaloudis:
I don't think this logic should apply to fiat currencies, which have their own standards for comma vs decimal formatting, that we already define very clearly in the app.

I think we should only apply this for Bitcoin and satoshi amounts. What do you think @myxmaster?

myxmaster:
First of all thanks a lot @cherry-1729-9090 for taking the time and working on this, it is a pretty important issue imo.

I did not test this yet or looked into the code.

But concept-wise, I strongly suggest keeping sats/₿ and fiat in line, meaning: when the app language is English, use English thousands and decimal separators for everything.

The formatting should follow the reader's locale, not the currency. Mixing conventions within the same screen is cognitively jarring and can lead to misreads. (Example: When you're on vacation in Europe and see € 1.021,34, would that feel intuitive to you, or would you prefer € 1,021.34 right next to your 1,755,264 sats?)

cherry-1729-9090:
Thanks a lot for the feedback, really appreciate it.
I agree with your point, separators should follow the app language (reader locale), not the currency.
So if app language is English: 1,234.5, and if it’s German: 1.234,5, consistently across sats, BTC, and fiat.

That’s the approach implemented in this PR to avoid mixed formatting on the same screen and reduce misreads.

components/Amount.tsx
cherry-1729-9090: These changes were added because, while testing this PR, I found that some places render <Amount fixedUnits="fiat" /> before the selected fiat rate is available in FiatStore, which causes the component to show N/A even though the formatting logic is correct. The new guard only runs when the component is actually rendering fiat, fiat is enabled, a fiat currency is selected, the selected rate is missing, and a fetch is not already in progress, so it is just a defensive way to avoid that missing-rate state in amount/conversion flows that do not always go through the main wallet refresh path.

PR #3985, opened 2026-04-16
feat(LSP): add waiting screen for LSPS1/LSPS7 external wallet payments

shubhamkmr04:
# Description

Relates to issue: https://github.com/ZeusLN/zeus/issues/3952

<table>
  <tr>
    <td align="center"><b>Polling order state</td>
    <td align="center"><b>success</td>
    <td align="center"><b>failure</td>
  </tr>
  <tr>
    <td align="center">
      
    </td>
    <td align="center">
      
    </td>
    <td align="center">
      
    </td>
  </tr>
</table>

- [X] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [X] N/A

- [X] Android
- [X] iOS

On-device
- [ ] LDK Node
- [X] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


locales/en.json
```diff
@@ -1633,6 +1633,15 @@
     "views.LSPS1.makePayment": "Make payment",
     "views.LSPS1.makePaymentLN": "Make payment (LN)",
     "views.LSPS1.makePaymentOnchain": "Make payment (on-chain)",
+    "views.LSPS1.payExternalWallet": "Pay from an external wallet and wait for confirmation",
+    "views.LSPS1.awaitingPayment": "Awaiting payment",
+    "views.LSPS1.checkingOrderStatus": "Checking order status...",
+    "views.LSPS1.payFromZeus": "Pay from Zeus wallet",
```
kaloudis: nit: ZEUS

kaloudis (changes requested):

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,481 @@
...
+
+        const failureFallback =
+            service === 'LSPS7'
+                ? localeString('views.LSPS7.extensionFailed')
+                : localeString('views.LSPS1.channelOpenFailed');
+
+        return (
+            <Screen>
+                <Header
+                    leftComponent="Back"
```
kaloudis, line 271: Header back and hardware back behave differently when `paymentState === 'completed'`. `handleBackPress` (L143-149) calls `popTo('Wallet')` to keep the user out of the now-stale create-order flow, but the header arrow falls through to the default `navigation.goBack()` in `Header.tsx` and lands back on the previous OrderResponse screen.

Suggest passing `onBack={paymentState === 'completed' ? this.goToWallet : undefined}` on the Header (or otherwise overriding it on the completed state) so both gestures align.

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,481 @@
...
+                    this.updateOrderInStorage(response);
+                    this.setState({ paymentState: 'completed' });
+                    this.isPolling = false;
+                    return;
+                }
+                if (result?.order_state === LSPOrderState.FAILED) {
+                    this.updateOrderInStorage(response);
+                    this.setState({
+                        paymentState: 'failed',
+                        failureMessage: LSPStore.error_msg || undefined
```
kaloudis: Dead `failureMessage` capture.

`LSPStore.error_msg` is explicitly zeroed at the start of each poll (L173), and none of the `get_order` methods in `LSPStore` set `error_msg` when a successful response carries `order_state === FAILED` (only transport-level errors do). So this read is always empty, the captured `failureMessage` is always `undefined`, and the render fallback chain at L453-457 silently does the right thing via `failureFallback`.

Either remove the dead capture, or extract a real failure reason from the response payload if the LSP returns one (`LSPOrderState` in `models/LSP.ts` doesn't currently model a `failure_reason` field).

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,481 @@
...
+                await sleep(CUSTOM_MESSAGE_WAIT_MS);
+                response = LSPStore.getExtensionOrderResponse;
+            } else {
+                if (native) {
+                    await LSPStore.lsps1GetOrderNative(orderId);
+                } else if (endpoint && BackendUtils.supportsLSPS1rest()) {
+                    await LSPStore.lsps1GetOrderREST(orderId, endpoint);
+                } else if (peer && BackendUtils.supportsLSPScustomMessage()) {
+                    LSPStore.lsps1GetOrderCustomMessage(orderId, peer);
+                    await sleep(CUSTOM_MESSAGE_WAIT_MS);
```
kaloudis: Custom-message polls have a read-after-write race.

`lsps1GetOrderCustomMessage` (L192) and `lsps7GetOrderCustomMessage` (L183) are fire-and-forget for the custom-message backends — the response arrives asynchronously via the message handler that mutates `LSPStore.getOrderResponse` / `getExtensionOrderResponse`. Sleeping a fixed 3 s and then reading the observable is racy: if the round trip exceeds 3 s, this cycle silently misses the response and waits another 5 s for the next poll.

The native and REST paths (L188, L190) are properly awaited — consider making the custom-message variants return a Promise that resolves on response arrival (or rejects on timeout) so this code can `await` them uniformly and avoid the magic-number sleep.

views/LSPS1/OrderResponse.tsx
```diff
@@ -482,17 +481,26 @@ export default class LSPS1OrderResponse extends React.Component<
...
+                                                                payment.bolt11_invoice,
+                                                            satAmount:
+                                                                payment.bolt11
+                                                                    ?.order_total_sat ||
+                                                                payment.order_total_sat ||
+                                                                payment.bolt11
+                                                                    ?.fee_total_sat ||
+                                                                payment.fee_total_sat,
+                                                            service: 'LSPS1'
+                                                        }
```
kaloudis: The `{ orderId, invoice, satAmount, service }` payload — including the four-way `bolt11?.order_total_sat || order_total_sat || bolt11?.fee_total_sat || fee_total_sat` fallback — is duplicated verbatim across four call sites:

- `views/LSPS1/OrderResponse.tsx:485-502` (this one)
- `views/LSPS1/index.tsx:1243-1260`
- `views/LSPS7/OrderResponse.tsx:357-374`
- `views/LSPS7/index.tsx:656-673`

Worth extracting a `buildPaymentAwaitParams(payment, orderId, service)` helper (e.g., in a shared LSPS utils module or on `LSPStore`) so the four sites don't drift independently as LSP payload shapes evolve.

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,481 @@
...
+import { sleep } from '../../utils/SleepUtils';
+import { themeColor } from '../../utils/ThemeUtils';
+
+import Storage from '../../storage';
+
+const POLL_INTERVAL_MS = 5000;
+const CUSTOM_MESSAGE_WAIT_MS = 3000;
+
+type PaymentAwaitState = 'polling' | 'completed' | 'failed';
+type LSPService = 'LSPS1' | 'LSPS7';
```
kaloudis: can we use enums for these?

views/LSPS1/OrderResponse.tsx
shubhamkmr04: done. fixed all the review suggested

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,474 @@
...
+import BackendUtils from '../../utils/BackendUtils';
+import handleAnything from '../../utils/handleAnything';
+import { localeString } from '../../utils/LocaleUtils';
+import { themeColor } from '../../utils/ThemeUtils';
+
+import Storage from '../../storage';
+
+const POLL_INTERVAL_MS = 5000;
+
+type PaymentAwaitState = 'polling' | 'completed' | 'failed';
```
kaloudis: this should be an enum too

kaloudis (changes requested):

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,474 @@
...
+            }
+        } catch (error) {
+            console.error('LSPPaymentAwait: poll error', error);
+        }
+
+        this.isPolling = false;
+        this.scheduleNextPoll();
+    };
+
+    private updateOrderInStorage(order: any) {
```
ajaysehwal: @shubhamkmr04, updateOrderInStorage appears to be duplicated across multiple views. Could we consolidate this logic into a single shared method to avoid duplication?

ajaysehwal (commented):
https://github.com/shubhamkmr04/zeus/blob/1adb7fa7f845e980654119e7b210931027a0e596/views/LSPS1/index.tsx#L469

nit: still loose equality here

views/LSPS/LSPPaymentAwait.tsx
```diff
@@ -0,0 +1,474 @@
...
+
+        try {
+            LSPStore.getOrderResponse = {};
+            LSPStore.getExtensionOrderResponse = {};
+            LSPStore.error = false;
+            LSPStore.error_msg = '';
+
+            let response: any;
+
+            if (service === LSPService.LSPS7) {
```
ajaysehwal, line 171: @shubhamkmr04, I think LSPS7 on LDK Node may never poll the API , because supportsLSPScustomMessage() is false → saveOrderToHistory does not set peer for LSPS7 
supportsLSPS7native() is true → lsps7GetOrderCustomMessage does not need peer
And LSPS7Order.tsx already calls lsps7GetOrderCustomMessage(id, temporaryOrder?.peer) with optional peer. Await screen is stricter than order detail and can spin forever on LDK LSPS7

ajaysehwal (approved):
tACK

PR #3996, opened 2026-04-19
Fix #2132: Restrict numeric input and prevent 0-sat LNURL-pay

### LINKED ISSUE #2132: Input Validation: Allowing Alphabetic Characters in Invoice Creation and Payment Fields
### Describe your enhancement idea

On the send view, users are able to input alphabetic characters when creating an invoice, which should not be allowed. Similarly, when attempting to pay an invoice with alphabetic characters, users receive an error message stating 'amount must be specified when paying a zero amount invoice'.

preetham3127:
Closes #2132

Two minimal fixes:

1. components/AmountInput — tighten input filter (shared component) Removes - from the onChangeText character allowlist. The regex was /[^\d.,-]/g; it is now /[^\d.,]/g. The keyboard type is already "numeric" — this closes the gap where a minus sign could slip through the software filter. One character removed from a regex.

2. views/LnurlPay/LnurlPay — disable Confirm on 0 sats The Confirm button had no guard against a zero amount, which caused a bad argument error when sendValues() was called with satAmount = 0. Added !satAmount || Number(satAmount) <= 0 to disabled, following the identical pattern already used in Send.tsx.

No new dependencies, no refactoring, no locale string additions.

### DIFF
--- a/components/AmountInput.tsx
+++ b/components/AmountInput.tsx
@@ -235,7 +235,7 @@ export default class AmountInput extends React.Component<
                                 opacity: locked ? 0.8 : 1
                             }
                         ]}
-                    >
+                >
                         <Text
                             style={[
                                 styles.amountText,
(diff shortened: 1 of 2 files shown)


shubhamkmr04:
@preetham3127 rebase against master are preferred over merge commits

PR #4020, opened 2026-04-24
fix: handle LSP non-JSON responses and clear stale error state on invoice creation

kaloudis:
# Description

When the LSP server returns non-JSON responses (such as a 503 HTML page from nginx during rate limiting or downtime), `response.json()` would throw a `SyntaxError` before the HTTP status could be checked, crashing the invoice creation flow. Both the `/fee` and `/proposal` endpoint handlers now wrap the JSON parse in a try/catch and set a meaningful `flow_error_msg` with the HTTP status.

The `getZeroConfFee` await was also missing a try/catch, so a rejected promise would propagate as an uncaught exception, skipping the `flow_error` check entirely and leaving `createInvoice` in a broken state. Additionally, `resetFee()` only cleared `zeroConfFee` but not `feeId`, `flow_error`, `flow_error_msg`, or `showLspSettings`, so errors from a previous invoice attempt would persist into the next one even on success. `showLspSettings` was also gated behind the `!errorsOnly` branch in `LSPStore.reset()`, meaning it was never cleared when the create invoice button called `reset(true)`.

Finally, when `/proposal` fails but the backend invoice was successfully created, the error state is now cleared so the unwrapped invoice can be displayed without a stale error banner alongside it.

## Test plan

- [ ] Create an invoice when LSP is healthy — should get wrapped invoice as before
- [ ] Create an invoice when LSP returns 503 on `/proposal` — should fall back to unwrapped invoice without error banner
- [ ] Create an invoice when LSP returns 503 on `/fee` — should abort with error, then retry successfully after LSP recovers with no stale errors
- [ ] Rapid successive invoice creations — verify no stale error state carries over between attempts

## PR Type

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [ ] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [ ] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [ ] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/LSPStore.ts
+++ b/stores/LSPStore.ts
@@ -89,7 +89,6 @@ export default class LSPStore {
         if (!errorsOnly) {
             this.info = {};
             this.resetFee();
-            this.showLspSettings = false;
             this.channelAcceptor = undefined;
             this.customMessagesSubscriber = undefined;
         }
@@ -97,10 +96,17 @@ export default class LSPStore {
         this.error_msg = '';
         this.flow_error = false;
         this.flow_error_msg = '';
+        this.showLspSettings = false;
     };
 
     @action
-    public resetFee = () => (this.zeroConfFee = undefined);
+    public resetFee = () => {
+        this.zeroConfFee = undefined;
+        this.feeId = undefined;
+        this.flow_error = false;
+        this.flow_error_msg = '';
+        this.showLspSettings = false;
+    };
 
     @action
     public resetLSPS1Data = () => {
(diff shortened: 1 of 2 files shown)


stores/LSPStore.ts
myxmaster: similar to line 287-293 maybe also here:
```
this.flow_error_msg = data?.message
    ? `${localeString('stores.LSPStore.error')}: ${data.message}`
    : `${localeString('stores.LSPStore.error')}: HTTP ${status}`;
```

myxmaster: add try/catch as well?

myxmaster (approved):
LGTM (but did not test)

shubhamkmr04 (approved):
tACK

PR #4035, opened 2026-05-01
fix: prevent restore stall when LndmobileStopDaemon never calls back

kaloudis:
## Summary

When restoring an embedded LND wallet via seed recovery, the app would stall indefinitely after tapping "Continue without backup." The root cause is that `LndmobileStopDaemon` on iOS never calls back if the Go daemon is not running, causing `await stopLnd(...)` to hang forever since the outer `try/catch` only catches rejections, not promises that never settle.

The fix checks `SettingsStore.embeddedLndStarted` before calling `stopLnd`, skipping the call entirely when no daemon is running. When LND is running, stale `SubscribeState` listeners are cleaned up and a shutdown delay is added before starting the new instance to avoid race conditions with the Go process still winding down. Status text is now displayed below the loading indicator during restore so users can see which step is in progress, and diagnostic logging has been added to `createLndWallet` for future debugging.

## Test plan
- [ ] Restore an embedded LND wallet via seed recovery, choosing "Continue without backup"
- [ ] Verify the restore completes successfully instead of stalling
- [ ] Verify status messages update during the flow (Stopping LND → Optimizing Neutrino peers → Starting LND → Initializing wallet)
- [ ] Test restore when an existing embedded LND wallet is already running
- [ ] Test restore on a fresh install with no prior LND instance

## PR category 
- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [x] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1090,6 +1090,7 @@
     "views.Tools.migration.import.downloadingFromOlympus": "Downloading channel backup from Olympus...",
     "views.Tools.migration.status.startingLnd": "Starting LND...",
     "views.Tools.migration.status.initializingWallet": "Initializing wallet...",
+    "views.Tools.migration.status.optimizingPeers": "Optimizing Neutrino peers...",
     "views.Tools.migration.status.importingBackup": "Importing backup...",
     "views.Tools.migration.import.olympusRestoreFailed": "Failed to restore from Olympus: {{error}}",
     "views.Tools.migration.import.restoreComplete": "Restore Complete",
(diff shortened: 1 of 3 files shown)


shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4040, opened 2026-05-01
fix: token import: add mint not working when enabling Cashu for first time

kaloudis:
# Decription

When importing a Cashu token for the first time with Cashu disabled, pressing the "Enable Cashu and add mint" button would only enable Cashu — the button text would change to "Add mint" but the mint was never actually added, requiring the user to press the button a second time. This was caused by two issues: `this.mintUrls` assignments after `await` calls in `@action`-decorated async methods were not wrapped in `runInAction()`, which could cause MobX to miss the state update and not re-render the component; and the button handler had no loading state during `initializeWallets()`, leaving the button enabled with no spinner — making it appear the operation completed prematurely — and no error handling, so if `initializeWallets()` threw, `addMint` would silently never execute.

## Changes

- **`stores/CashuStore.ts`**: Wrap all `this.mintUrls` assignments after `await` in `runInAction()` in `addMint`, `removeMint`, and `initializeWallets` (both success and fallback paths)
- **`views/Cashu/CashuToken.tsx`**: Set `CashuStore.setLoading(true)` at the start of the "Enable Cashu and add mint" button handler so the button disables immediately and the loading indicator shows during the entire operation; add `try/catch` to prevent silent failures

## Test plan

- [ ] With Cashu disabled, scan/import a Cashu token
- [ ] Press "Enable Cashu and add mint" — verify loading indicator appears immediately, button is disabled, and mint is added in a single press
- [ ] Verify the "Receive" button becomes enabled after the mint is added
- [ ] With Cashu already enabled, import a token from a new mint — verify "Add mint" still works correctly
- [ ] Remove a mint and verify mint list updates properly

## PR Type

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/CashuStore.ts
+++ b/stores/CashuStore.ts
@@ -1946,7 +1946,10 @@ export default class CashuStore {
             await this.syncCDKBalances();
 
             // Update local mintUrls from CDK (source of truth)
-            this.mintUrls = await CashuDevKit.getMintUrls();
+            const updatedMintUrls = await CashuDevKit.getMintUrls();
+            runInAction(() => {
+                this.mintUrls = updatedMintUrls;
+            });
 
             // Backup to local storage for migration on restart
             await Storage.setItem(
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4064, opened 2026-05-06
perf: improve offline wallet startup time

kaloudis:
# Description

When the device is offline, Zeus's LDK Node and Cashu init paths used to block on a series of internal HTTP timeouts — VSS dual-store build (30–60s), `node.syncWallets()` Esplora + RGS fetch (~30s+), and on a fresh recovery, the sequential Nostr relay loop (8 × 10s = up to 80s). End-to-end an offline cold-start could take 60s+ for an existing wallet and several minutes for a fresh recovery.

This PR seeds `connectivityStore.isOffline` *before* any wallet implementation init runs, then short-circuits the network-bound work that would otherwise stall on those internal timeouts. Online cold-start is unchanged apart from a one-time probe capped at ~1.5s.

## Impact

| Path | Before | After |
| --- | --- | --- |
| Initial reachability probe | ~5s probe budget | ~1.5s |
| LDK VSS dual-store build (offline) | ~30s existing / ~60s restore | skipped |
| LDK `syncWallets()` (offline) | ~30s+ Esplora + RGS fetch | skipped |
| Cashu Nostr mint restore (fresh recovery, offline) | up to ~80s sequential | skipped |
| LDK VSS build timeout when online-but-VSS-slow | 30s | 10s |
| **Offline cold-start, existing wallet** | **~60s+** | **~3s** |
| **Offline cold-start, fresh recovery** | **several minutes** | **~3s** |

Online cold-start regresses by at most the new ~1.5s `checkNow()` probe; when online, the same VSS build and sync paths run as before. As a secondary benefit, online Nostr mint restore now fans out across all relays in parallel instead of trying them one-at-a-time.

## Behaviour notes / risks

- **VSS deferral while offline.** A session built offline uses local-only SQLite. VSS resumes on the next launch once we're online; we do not currently rebuild the node mid-session on reconnect, so the same is true if the device comes back online during the session.
- **No alerts when offline.** When VSS / sync / RGS are skipped because we're offline, `vssError` / `esploraError` / `rgsError` stay `undefined`, so `Wallet.tsx` doesn't surface alerts the user already understands. Online failures alert as before.
- **Tighter VSS budget.** Existing-wallet VSS timeout drops from 30s → 10s. Users on slow connections may occasionally see a `vssError` alert where the previous 30s budget would have eventually succeeded. The graceful fallback (local-only build, retry on next launch) is unchanged.
- **Slow-network false positive on the startup probe.** The `checkNow()` budget is 1.5s across three parallel probes (block explorer, ZEUS rates endpoint, Cloudflare trace). On a connection where all three HEAD responses take longer than 1.5s, we'll declare offline and skip VSS / sync for the session. The 15s poll and NetInfo listener will then transition the session to online and fire reconnect callbacks (Cashu sweeps pending tokens, SyncStore resumes embedded-lnd sync), but LDK Node is not rebuilt mid-session — VSS / chain sync resume on the next launch.
- **Connectivity ownership.** `connectivityStore.start()` is now invoked from `Wallet.fetchData()` rather than `CashuStore.initializeWallets()`, which runs early enough for downstream wallet implementations to see a reliable `isOffline`. Cashu still registers its reconnect callback via `startConnectivityMonitoring` (now deduped against re-registration). `stop()` / `reset()` remain wired to the "disable offline check" settings toggle.

## Test plan

- [ ] **Offline cold-start, existing wallet** — disable Wi-Fi/cellular, launch app; confirm UI is interactive within ~3s and no VSS / Esplora / RGS alerts surface.
- [ ] **Online cold-start, existing wallet** — confirm startup time has not regressed beyond the ~1.5s probe.
- [ ] **Online → offline mid-session** — confirm the running session keeps working off local state; on next cold-start, offline path is taken.
- [ ] **Offline → online reconnect** — confirm `connectivityStore.onReconnect` fires and Cashu pending tokens sweep.
- [ ] **Fresh recovery, online** — verify Nostr mint restore still finds the backup with parallel fan-out.
- [ ] **Fresh recovery, offline** — verify we no longer hang on the relay loop and the wallet still loads (no mints, as expected).
- [ ] **VSS reachable but slow** — confirm the new 10s budget falls back to local with a clear `vssError` alert.
- [ ] **Disable offline check toggle** — turning the setting on should stop monitoring and clear `isOffline`; turning it off should resume monitoring.

## PR Type

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/ldknode/LdkNodeInjection.ts
+++ b/ldknode/LdkNodeInjection.ts
@@ -963,13 +963,14 @@ const initializeNode = async ({
         );
     }
 
-    // Use a longer timeout for restore-from-seed (no local DB yet) since
-    // every read falls through to VSS sequentially. Even existing nodes
-    // may have an incomplete local DB (e.g. after a previous fallback
-    // build), so the default is generous to avoid false alerts.
+    // VSS itself is gated on connectivity upstream (LdkNodeUtils.initNode skips
+    // VSS entirely when offline), so this timeout only kicks in when we're
+    // online but VSS is sluggish. A tight 10s budget for existing nodes is
+    // enough for the typical incremental sync; restore-from-seed needs longer
+    // because every read falls through to VSS sequentially.
     const localDbPath = `${storagePath}/ldk_node_data.sqlite`;
     const hasLocalDb = await RNFS.exists(localDbPath);
-    const vssBuildTimeout = hasLocalDb ? 30 : 60;
+    const vssBuildTimeout = hasLocalDb ? 10 : 60;
     await setVssBuildTimeout(vssBuildTimeout);
     console.log(
         `LDK Node: [${elapsed()}] VSS build timeout set to ${vssBuildTimeout}s (${
--- a/ldknode/LdkNodeInjection.ts
+++ b/ldknode/LdkNodeInjection.ts
@@ -932,7 +932,9 @@ const initializeNode = async ({
         );
     }
 
-    // Configure VSS (Versioned Storage Service) for cloud backup
+    // Configure VSS (Versioned Storage Service) for cloud backup.
+    // Gated on connectivity upstream (LdkNodeUtils.initNode skips VSS entirely
+    // when offline), so everything in here only runs when we're online.
     if (vssConfig && vssConfig.url && vssConfig.storeId) {
         let vssHeaders: Record<string, string> | undefined;
         try {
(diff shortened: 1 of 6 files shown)


ajaysehwal (commented):
https://github.com/user-attachments/assets/bb82a806-566d-4b62-a41e-9e694c307c33

Image

@kaloudis, LDK Node is not starting when the device is offline. It fails with the following error:

LDK Node connection failed: Error: LDK Node not running yet
    at anonymous (LdkNodeUtils.ts:474:51)
    at next (native)
    at asyncGeneratorStep (asyncToGenerator.js:3:17)
    at _next (asyncToGenerator.js:17:27)
    at tryCallOne (address at InternalBytecode.js:1:1274)
    at anonymous (address at InternalBytecode.js:1:4865)

It looks like the node startup depends on network connectivity

PR #4068, opened 2026-05-08
fix: Onboarding: prevent swipe-back from clobbering an existing wallet

kaloudis:
# Description

- Replaces `navigation.navigate('Wallet')` with `navigation.reset` in `WalletSettings` and `RecommendedSettings` so the onboarding screens are removed from the stack after wallet creation. Without this, Android's system swipe-back gesture on the Wallet view returned the user into the wallet-creation flow.
- Adds a defense-in-depth guard in `createOnboardingWallet` that bails out (logs a warning, calls `onSuccess`) when `settings.nodes` already has entries, so the onboarding path can never overwrite an existing node directory and mnemonic.

## Bug report
A user on v0.13.0 (Android, embedded LDK + Minibits ecash) created a wallet, received funds via their Lightning address, then swiped right intending to close the app. They landed back inside the wallet-creation flow with no way to cancel; on completion their balance read as zero. The original LDK node directory still existed on disk, but `settings.nodes` had been overwritten with a fresh empty node, making the funds unreachable through the UI.

## Scope
`createOnboardingWallet` is only called from the two onboarding screens. The "add another wallet" flow (`views/Settings/WalletConfiguration.tsx`) and seed recovery (`views/Settings/SeedRecovery.tsx`) call `createLndWallet` / `createLdkNodeWallet` directly and append to the `nodes` array, so they are unaffected by the guard.

## Test plan
- [ ] Fresh install → complete onboarding → verify swipe-back from Wallet does not return to onboarding

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [ ] I’ve run `yarn run prettier` and made sure my code is formatted correctly

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [ ] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/WalletCreationUtils.ts
+++ b/utils/WalletCreationUtils.ts
@@ -39,6 +39,18 @@ export async function createOnboardingWallet(params: WalletCreationParams) {
 
     const { setConnectingStatus, updateSettings, settings } = settingsStore;
 
+    // Refuse to create a new wallet if one already exists. Overwriting the
+    // `nodes` array would orphan the existing node directory and any funds it
+    // holds. Treat this as a successful no-op so callers route the user back
+    // to the Wallet view rather than into an error state.
+    if (settings?.nodes?.length) {
+        console.warn(
+            'createOnboardingWallet: existing node detected, skipping creation'
+        );
+        onSuccess();
+        return;
+    }
+
     const commonSettings = {
         privacy: {
             ...settings.privacy,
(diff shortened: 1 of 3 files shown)


shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4101, opened 2026-05-17
enhancement: LDK: wait for node readiness before fetching startup data

kaloudis:
# Description

- Wraps `LdkNode.node.start()` in `startLdkNodeWallet` with a small `retry` shell that retries on `"Node not initialized"`, and rethrows non-fee-rate `start()` failures instead of silently swallowing them.
- Adds `waitForLdkNodeReady()` helper that retries `status()` until `isRunning` is true, treating both `"Node not initialized"` and a transient `!isRunning` as retryable. Called from `Wallet.tsx` at the top of the LDK Node startup block, before `NodeInfoStore.getNodeInfo()` and the other data fetches.
- Sets `SettingsStore.walletJustCreated = true` after `createLdkNodeWallet` in `createOnboardingWallet`, matching the pattern already used by `SeedRecovery.tsx` and `WalletConfiguration.tsx`. Onboarding no longer tears down and rebuilds the freshly built node.

## Why

Users were sometimes seeing `"Node not initialized"` during LDK Node wallet startup. The native `buildNode` (iOS `LdkNodeModule.swift`, Android `LdkNodeModule.kt`) synchronously nulls the existing node reference and rebuilds on a background queue, so the field is `nil`/`null` for the entire build window (up to ~60s on a restore-from-seed with VSS). Any call into the module during that window — `nodeId`, `status`, `listChannels` — rejects with `"Node not initialized"` and surfaces as a fatal startup error in the dashboard.

Two contributing factors made this much more likely:

1. `createOnboardingWallet` was the only LDK creation path that didn't set `walletJustCreated = true`, so `Wallet.tsx` took the `!justCreated` branch on every fresh wallet — stopping and rebuilding the node it had just built and widening the race window.
2. `startLdkNodeWallet` silently swallowed any non-fee-rate `start()` failure, returning success. The downstream `getNodeInfo` call would then hit the still-unset node and produce the misleading `"Node not initialized"` error instead of the real cause.

The readiness gate mirrors what we already do for embedded LND via `waitForRpcReady` in `utils/LndMobileUtils.ts`.

## Behavior

- `start()` is retried up to 5 times at 500ms intervals on `"Node not initialized"`. Non-fee-rate failures now propagate to the caller, where `Wallet.tsx`'s LDK `catch (connectionError)` clears `connectingStatus` and calls `NodeInfoStore.handleGetNodeInfoError()`.
- `waitForLdkNodeReady()` runs `status()` via `retry` (120 attempts × 500ms = 60s ceiling). It retries on `"Node not initialized"` and on a distinct `"LDK Node not running yet"` sentinel emitted when the node exists but hasn't transitioned to running.
- The fee-rate carve-out in `startLdkNodeWallet` is preserved — those errors continue to set `esploraError` and let sync proceed.

## Out of scope (follow-ups)

- Native `buildNode` could be made atomic (build into a local, only commit to `self.node` once ready) to close the race window inside the module itself. This PR is purely consumer-side defense.

## Test plan

- [ ] Fresh wallet creation (onboarding) — verify dashboard loads without `"Node not initialized"` and that `walletJustCreated` skips the redundant stop+rebuild cycle.
- [ ] Restore-from-seed with VSS configured — verify the longer build window (up to 60s) is tolerated and the dashboard eventually loads.
- [ ] Restore-from-seed with VSS unreachable — verify fallback to local SQLite still surfaces the dashboard.
- [ ] Switch between two LDK Node wallets — verify the rebuild path between `stopLdkNode()` and `start()` doesn't leak the error.
- [ ] Force a genuine startup failure (e.g., bad Esplora URL) — verify the rethrow path produces a clean error via `handleGetNodeInfoError()` instead of a 60s phantom readiness timeout.

## PR Type

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/LdkNodeUtils.ts
+++ b/utils/LdkNodeUtils.ts
@@ -64,6 +64,13 @@ export const RGS_SERVERS_TESTNET: EsploraServer[] = [
 // Default pathfinding scores server
 export const DEFAULT_SCORER_URL = 'https://scores.zeusln.com/latest.bin';
 
+// Native-module error string emitted while `buildNode` is still rebuilding
+// the node reference on its background queue
+const LDK_NODE_NOT_INITIALIZED = 'Node not initialized';
+
+// Local sentinel for the gap between the node existing and reporting running
+const LDK_NODE_NOT_RUNNING_YET = 'LDK Node not running yet';
+
 /**
  * Get the storage directory path for LDK Node data
  */
@@ -365,7 +372,7 @@ export async function startLdkNodeWallet({
             delayMs: 500,
             shouldRetry: (e: any) => {
                 const errMsg = e?.message || e?.toString?.() || String(e);
-                return errMsg.includes('Node not initialized');
+                return errMsg.includes(LDK_NODE_NOT_INITIALIZED);
             }
         });
         nodeStarted = true;
@@ -431,8 +438,6 @@ export async function startLdkNodeWallet({
     return { vssError, esploraError, rgsError };
 }
 
-const LDK_NODE_NOT_RUNNING_YET = 'LDK Node not running yet';
-
 /**
  * Wait for the LDK Node native module to report a running node.
  *
@@ -455,7 +460,7 @@ export async function waitForLdkNodeReady(
         shouldRetry: (e: any) => {
             const errMsg = e?.message || e?.toString?.() || String(e);
             return (
-                errMsg.includes('Node not initialized') ||
+                errMsg.includes(LDK_NODE_NOT_INITIALIZED) ||
                 errMsg.includes(LDK_NODE_NOT_RUNNING_YET)
             );
         }
--- a/utils/LdkNodeUtils.ts
+++ b/utils/LdkNodeUtils.ts
@@ -431,6 +431,8 @@ export async function startLdkNodeWallet({
     return { vssError, esploraError, rgsError };
 }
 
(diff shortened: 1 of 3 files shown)


shubhamkmr04 (approved):
tACK

utils/LdkNodeUtils.ts
```diff
@@ -358,7 +359,15 @@ export async function startLdkNodeWallet({
...
 
     try {
-        await LdkNode.node.start();
+        await retry({
+            fn: () => LdkNode.node.start(),
+            maxRetries: 5,
+            delayMs: 500,
+            shouldRetry: (e: any) => {
+                const errMsg = e?.message || e?.toString?.() || String(e);
+                return errMsg.includes('Node not initialized');
```
ajaysehwal: nit: I think we should create a constant for "Node not initialized" also ?

ajaysehwal (approved):
tACK

PR #4106, opened 2026-05-19
feat(cashu): detect cashu tokens wrapped in any https URL

kaloudis:
# Description

Generalizes the early Cashu token detection in `handleAnything` from a hardcoded `https://zeusln.com/e/` prefix check to any `https://` URL whose body contains a `cashuA` / `cashuB` marker. Token extraction now goes through `CashuUtils.extractTokenString`, which already understands `?token=`, `#`, `/e/`, and other common wrapper formats.

## Context

[agi.cash](https://agi.cash/) issues redemption QRs that wrap Cashu tokens in `https://agi.cash/...` URLs — the same general pattern Zeus already uses with `https://zeusln.com/e/<token>`. Before this change, Zeus would only fast-path the ZEUS gift URL, and any other wrapper (including agi.cash) risked being misrouted into the lnurl block, where `findlnurl()` can match bech32-like substrings inside a Cashu token.

We also confirmed reciprocity: agi.cash's wallet successfully reads Cashu QRs prefixed with `https://zeusln.com/e/`, so this change aligns Zeus to accept the same wrapper shape from any host.

This is safe against the lnurl block because bech32's charset excludes uppercase `A`/`B`, so a real lnurl payload cannot contain the `cashuA`/`cashuB` marker — no false positives. Raw `cashuA…` tokens (no URL wrapper) continue to fall through to the existing late `isValidCashuTokenAsync` check. The HTTPS-only restriction preserves the existing `http://` rejection behavior.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Suggested manual test plan:

- Scan an agi.cash redemption QR → routes to the `CashuToken` view
- Scan a `https://zeusln.com/e/<token>` QR → still routes to `CashuToken` (regression)
- Scan a `https://wallet.cashu.me/?token=<token>` URL → routes to `CashuToken`
- Scan a `https://wallet.nutstash.app/#<token>` URL → routes to `CashuToken`
- Paste a raw `cashuA…` / `cashuB…` token → still handled by the existing late check
- Scan an arbitrary lnurl QR → still routes to lnurl handling (no regression)

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/handleAnything.ts
+++ b/utils/handleAnything.ts
@@ -733,11 +733,13 @@ const handleAnything = async (
             });
     } else if (
         /^https:\/\//i.test(value) &&
-        (value.includes('cashuA') || value.includes('cashuB'))
+        /cashu[AB][0-9A-Za-z_-]{10,}/.test(value)
     ) {
         // Handle web-wrapped cashu tokens (zeusln.com/e/, wallet.cashu.me/?token=,
         // wallet.nutstash.app/#, etc.). Must run before lnurl: findlnurl can match
-        // bech32-like substrings inside cashu tokens and misroute the input.
+        // bech32-like substrings inside cashu tokens and misroute the input. The
+        // base64url shape gate avoids consuming URLs that incidentally contain
+        // the literal "cashuA"/"cashuB" substring (e.g. "?ref=cashuAfrica").
         const cashuToken = CashuUtils.extractTokenString(value);
         if (CashuUtils.isValidCashuToken(cashuToken)) {
             if (isClipboardValue) return true;
--- a/utils/handleAnything.ts
+++ b/utils/handleAnything.ts
@@ -4,7 +4,7 @@ import ReactNativeBlobUtil from 'react-native-blob-util';
 
 import { nodeInfoStore, invoicesStore, settingsStore } from '../stores/Stores';
 
-import AddressUtils, { ZEUS_ECASH_GIFT_URL } from './AddressUtils';
+import AddressUtils from './AddressUtils';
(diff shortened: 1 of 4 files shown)


utils/handleAnything.ts
```diff
@@ -731,9 +731,14 @@ const handleAnything = async (
                     { cancelable: false }
                 );
             });
-    } else if (value.startsWith(ZEUS_ECASH_GIFT_URL)) {
-        // Handle zeusln.com ecash gift URLs - check before lnurl to avoid false matches
-        const cashuToken = value.replace(ZEUS_ECASH_GIFT_URL, '');
+    } else if (
+        /^https:\/\//i.test(value) &&
+        (value.includes('cashuA') || value.includes('cashuB'))
```
ajaysehwal: @kaloudis, can we add a stricter validation gate for Cashu tokens? 

For example :  `/cashu[AB][0-9A-Za-z_-]{10,}/.test(value)`

ajaysehwal (commented):
@kaloudis, After the last commit, pasting any URL containing a Cashu token now throws an error. I think we can revert this change if it's unnecessary.

Image

ajaysehwal (approved):
tACK

PR #4125, opened 2026-05-29
fix(android): apply LDK and LND persistent mode toggles without app restart

kaloudis:
# Description

Previously, toggling "Persistent Mode" for the LDK Node or Embedded LND required the user to manually restart the app for the change to take effect.

The toggle handler would call `restartNeeded()`, which presents an Alert dialog asking the user to confirm an app restart. The restart flow stops the node and then calls `ProcessPhoenix.triggerRebirth()` to relaunch the process — but if the user dismissed the dialog, or if `node.stop()` hung (Tor/VSS shutdown stalls have been observed), the foreground service was never started/stopped, leaving the user in an inconsistent state where the AsyncStorage flag said one thing but the running service reflected another.

This PR removes the restart requirement entirely. The foreground service can be controlled at runtime via static `startService`/`stopService` methods, so the toggle now reaches the native bridge directly and the change takes effect immediately.

### LDK Node (`LdkNodeModule.kt` / `LdkNodeService.kt`)

The LDK service exists purely to keep the app process alive — it can be started or stopped freely as long as the node is running. Added a `setPersistentMode(enabled)` React method that calls `LdkNodeService.startService` / `stopService` directly.

### Embedded LND (`LndMobileService.java` / `LndMobileTools.java`)

The LND service is fundamentally different — it's a bound Messenger-based IPC bridge to LND itself, so it must stay alive even when persistent mode is off. Added a new `APPLY_PERSISTENT_MODE` intent action whose `onStartCommand` branch only toggles `startForeground` / `stopForeground` + the notification, without ever calling `stopSelf`. Exposed via `LndMobileTools.setPersistentMode(enabled)`.

### UI

Both toggles (`views/Settings/EmbeddedNode/index.tsx` for LDK, `views/Settings/EmbeddedNode/Advanced.tsx` for LND) now call the new native methods after persisting to AsyncStorage. The " Restart to take effect." suffix is removed from both subtitles since no restart is required. `restartNeeded` is still imported in `Advanced.tsx` for the fee estimator / rescan / compactDb toggles, which legitimately do require an LND restart.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/android/app/src/main/java/app/zeusln/zeus/LdkNodeModule.kt
+++ b/android/app/src/main/java/app/zeusln/zeus/LdkNodeModule.kt
@@ -535,6 +535,22 @@ class LdkNodeModule(reactContext: ReactApplicationContext) : ReactContextBaseJav
         return false
     }
 
+    @ReactMethod
+    fun setPersistentMode(enabled: Boolean, promise: Promise) {
+        try {
+            if (enabled) {
+                if (node != null) {
+                    LdkNodeService.startService(reactApplicationContext)
+                }
+            } else {
+                LdkNodeService.stopService(reactApplicationContext)
+            }
+            promise.resolve(null)
+        } catch (e: Exception) {
+            promise.reject("error", errorMessage(e))
+        }
+    }
+
     // Node Lifecycle Methods
 
     @ReactMethod
--- a/android/app/src/main/java/com/zeus/LndMobileService.java
+++ b/android/app/src/main/java/com/zeus/LndMobileService.java
@@ -579,6 +579,34 @@ public class LndMobileService extends Service {
           notificationManager.notify(ONGOING_NOTIFICATION_ID, buildNotification());
         }
         return START_NOT_STICKY;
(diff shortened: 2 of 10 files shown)


kaloudis:
@gemini-code-assist /review

ajaysehwal (commented):
Image

@kaloudis, I think this could be confusing for users. After switching from one embedded node to another, the persistent notification from the previously selected node is still shown. We should probably remove or dismiss the previous node's persistent notification when switching nodes.

ajaysehwal (approved):
tACK

PR #4143, opened 2026-06-05
fix(ios): NWC Dynamic Island lifecycle when host app is dismissed

kaloudis:
# Description

Fixes three related bugs in the NWC iOS Live Activity / Dynamic Island flow that surfaced when the host app was killed (swipe-dismiss, OOM, reboot).

**1. End Session button was inert when the host app was dead.**
`NWCStopIntent.perform()` only posted a Darwin notification, which requires a live observer in the app process. When the app is killed, the notification is silently dropped and the activity is never ended. Fixed by calling `NWCLiveActivityShared.endDuplicateActivities` directly via ActivityKit from inside the intent, mirroring the pattern already used by the next / prev / toggle-mute intents. The Darwin post is preserved so the audio module still stops cleanly when the app is alive.

**2. Pressing End Session resurrected audio and a fresh island ~20s later.**
iOS launches the host app in the background to service a `LiveActivityIntent`. The NWC store's `setupIOSAppStateMonitoring` saw `AppState.currentState === 'background'` during init and auto-called `startIOSAudioKeepAlive`, which restarted audio and created a brand-new Live Activity — undoing the user's End Session tap after the ~20s it took JS to bootstrap. Removed the background-cold-launch auto-start branch. The normal foreground→background transition still arms and starts the keep-alive as before; what's gone is the speculative "we missed the transition" path that was being abused by intent-driven cold launches.

**3. Orphan island stayed visually fresh for up to an hour after force-close.**
`staleDate` was `+3600s` with a 15s refresh, so after the app died the island looked alive indefinitely. Tightened to `staleDate = 30s` and refresh = 10s (3× safety margin — tolerates short audio interruptions without false-stale flicker). Within ~30s of host death, the activity flips to a new stale presentation that is the entire island as a tap-to-dismiss surface:

- Lock screen / expanded view replaced with `NWCStaleView` — "Session ended / Tap to dismiss" + red `xmark.circle.fill`, the whole row wrapped in `NWCStopIntent`.
- Compact leading: Zeus icon dimmed to 35% opacity.
- Compact trailing: speaker icon → red `xmark.circle.fill`, bound to `NWCStopIntent` for one-tap dismissal from the collapsed island.
- Minimal: Zeus icon → red `xmark.circle.fill`.
- Keyline tint: yellow → gray.
- All playback controls (prev / mute / next, live timer) are hidden in the stale state since they would only relaunch the host app for no good reason.

### iOS lifecycle caveat
iOS provides no callback that reliably fires on user swipe-dismiss, and Live Activities are explicitly designed to outlive the host app. We cannot synchronously end an activity at force-close time. The combined fix instead makes the persisted island honest about its state (stale UI within 30s), self-dismissable (every relevant tap target ends the activity directly via ActivityKit), and non-self-healing (the host app no longer revives audio + a new island when cold-launched for an intent).

### Files
- `ios/NWCWidget/NWCWidgetIntent.swift` — `NWCStopIntent` ends activity directly.
- `ios/NWCWidget/NWCWidgetLiveActivity.swift` — new `NWCStaleView`, `context.isStale` branches across all four island presentations.
- `ios/NWCWidget/NWCLiveActivityShared.swift` — intent-push `staleDate` 60min → 30s.
- `ios/zeus/NWCActivityManager.swift` — host-push `staleDate` 60min → 30s, refresh timer 15s → 10s.
- `stores/NostrWalletConnectStore.ts` — drop `AppState === 'background'` auto-start branch in `setupIOSAppStateMonitoring`.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [x] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/ios/NWCWidget/NWCLiveActivityShared.swift
+++ b/ios/NWCWidget/NWCLiveActivityShared.swift
@@ -128,7 +128,7 @@ enum NWCLiveActivityShared {
         )
         let content = ActivityContent(
             state: state,
-            staleDate: Date().addingTimeInterval(3600)
+            staleDate: Date().addingTimeInterval(30)
         )
         await act.update(content)
         NSLog("[NWCLiveActivityShared] pushed – track=%@ muted=%d rev=%u",
--- a/ios/NWCWidget/NWCWidgetIntent.swift
+++ b/ios/NWCWidget/NWCWidgetIntent.swift
@@ -19,9 +19,7 @@ private func postDarwin(_ name: String) {
 struct NWCNextTrackIntent: LiveActivityIntent {
     static var title: LocalizedStringResource = "Next Track"
     func perform() async throws -> some IntentResult {
-        if #available(iOS 16.2, *) {
-            await NWCLiveActivityShared.applyNextTrack()
-        }
+        await NWCLiveActivityShared.applyNextTrack()
         postDarwin(kNWCNextTrack)
         return .result()
     }
@@ -31,9 +29,7 @@ struct NWCNextTrackIntent: LiveActivityIntent {
 struct NWCPrevTrackIntent: LiveActivityIntent {
     static var title: LocalizedStringResource = "Previous Track"
     func perform() async throws -> some IntentResult {
-        if #available(iOS 16.2, *) {
-            await NWCLiveActivityShared.applyPrevTrack()
-        }
+        await NWCLiveActivityShared.applyPrevTrack()
         postDarwin(kNWCPrevTrack)
         return .result()
     }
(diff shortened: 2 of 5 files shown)


ajaysehwal (approved):
tACK

ajaysehwal (approved):

PR #4146, opened 2026-06-06
fix(nwc): treat in-flight and hodl invoice payments as pending

ajaysehwal:
# Description

## Summary

Fixes NWC `pay_invoice` incorrectly returning **failed** for hodl invoices and payments still **in transit**.

Zeus **Send** already treats `IN_FLIGHT` / hodl payments as in-transit. NWC was failing them when:
- there was no preimage yet (hodl / hold invoices)
- the HTLC was still `IN_FLIGHT`
- the router timed out while the payment was still out

**Now:**
- NWC returns **success** for in-transit pays (empty preimage is OK for hodl)
- Connection activity shows **In Transit Payment** (`pending`), not **Failed Payment**
- Activity updates to **You sent** or **Failed** when the node reports the final state
- Normal settled invoices are unchanged

## Changes

- `stores/NostrWalletConnectStore.ts` — in-transit handling, `recordPendingPayment`, activity refresh
- `utils/NostrConnectUtils.ts` — `resolveLightningPaymentInTransit` and related helpers

## Test plan

- [ ] **Normal invoice** — pay via NWC → success + preimage; activity = **You sent**
- [ ] **Hodl invoice** — `lncli addholdinvoice`, pay via NWC before settle → no failure; activity = **In Transit Payment**; after settle → **You sent**

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -224,6 +224,7 @@ export default class NWCConnection extends BaseModel {
 
     private resolveMakeInvoiceActivitySats(a: ConnectionActivity): number {
         const fromSat = Math.floor(Number(a.satAmount) || 0);
+        // satAmount first: amountless invoices report 0 from bolt11
         if (fromSat > 0) return fromSat;
 
         if (a.invoice != null) {
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -222,6 +222,38 @@ export default class NWCConnection extends BaseModel {
         return Math.min(100, (this.totalSpendSats / this.maxAmountSats!) * 100);
     }
 
+    private resolveMakeInvoiceActivitySats(a: ConnectionActivity): number {
+        const fromSat = Math.floor(Number(a.satAmount) || 0);
+        if (fromSat > 0) return fromSat;
+
+        if (a.invoice != null) {
+            const inv =
+                a.invoice instanceof CashuInvoice
+                    ? a.invoice
+                    : a.invoice instanceof Invoice
+                    ? a.invoice
+                    : a.payment_source === 'cashu'
+                    ? new CashuInvoice(a.invoice)
+                    : new Invoice(a.invoice);
+            const fromInvoice = Math.floor(Number(inv.getAmount) || 0);
+            if (fromInvoice > 0) return fromInvoice;
+        }
+
+        if (a.payment != null) {
+            const pay =
+                a.payment instanceof CashuPayment
+                    ? a.payment
+                    : a.payment instanceof Payment
+                    ? a.payment
+                    : a.payment_source === 'cashu'
+                    ? new CashuPayment(a.payment)
+                    : new Payment(a.payment);
+            return Math.floor(Number(pay.getAmount) || 0);
+        }
+
+        return 0;
+    }
+
     /**
      * Sum of settled incoming invoice amounts recorded on this connection's
      * activity log (successful make_invoice only).
@@ -233,20 +265,7 @@ export default class NWCConnection extends BaseModel {
             if (a.type !== 'make_invoice' || a.status !== 'success') {
                 return sum;
             }
-            const fromInvoice =
-                a.invoice != null
-                    ? Math.floor(Number(a.invoice.getAmount) || 0)
-                    : 0;
-            const fromPayment =
-                a.payment != null
-                    ? Math.floor(Number(a.payment.getAmount) || 0)
-                    : 0;
-            const fromSat = Math.floor(Number(a.satAmount) || 0);
-            let sats = 0;
-            if (fromInvoice > 0) sats = fromInvoice;
-            else if (fromSat > 0) sats = fromSat;
-            else sats = fromPayment;
-            return sum + Math.max(0, sats);
+            return sum + Math.max(0, this.resolveMakeInvoiceActivitySats(a));
         }, 0);
     }
 
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -91,6 +91,10 @@ const PAYMENT_TIMEOUT_SECONDS = 120;
 const PAYMENT_FEE_LIMIT_SATS = 1000;
 const PAYMENT_PROCESSING_DELAY_MS = 100;
 const SAVE_CONNECTIONS_DEBOUNCE_MS = 500;
+/** Min time a pay_invoice must stay pending before getActivities polls the node */
+const PENDING_PAYMENT_STATUS_MIN_AGE_MS = 5_000;
+/** Per-connection debounce for pending pay_invoice status refresh in getActivities */
+const PENDING_PAYMENT_STATUS_REFRESH_MS = 30_000;
 
 export const DEFAULT_NOSTR_RELAYS = [
     'wss://relay.getalby.com/v1',
@@ -161,6 +165,14 @@ export default class NostrWalletConnectStore {
     private androidLogListener: EmitterSubscription | null = null;
     private _scheduledSave: ReturnType<typeof setTimeout> | null = null;
     private resetInFlight: Promise<void> | null = null;
+    private lastPendingPaymentStatusFetchByConnection = new Map<
+        string,
+        number
+    >();
+    private lastPendingInvoiceStatusFetchByConnection = new Map<
+        string,
+        number
+    >();
 
     settingsStore: SettingsStore;
     balanceStore: BalanceStore;
@@ -923,6 +935,7 @@ export default class NostrWalletConnectStore {
             }
         }
         const pendingLightningInvoiceActivities: ConnectionActivity[] = [];
+        const pendingCashuInvoiceActivities: ConnectionActivity[] = [];
 
         for (const activity of connection.activity) {
             if (activity?.invoice) {
@@ -948,13 +961,22 @@ export default class NostrWalletConnectStore {
                             !paymentRequestFirstMatchNotCashu.has(pr) &&
                             firstCashuPaidByPaymentRequest.get(pr) === true;
                         if (invoiceAlreadyPaid) {
+                            const paidInvoice = cashuInvoices.find(
+                                (inv) =>
+                                    inv.getPaymentRequest === pr && inv.isPaid
+                            );
                             runInAction(() => {
                                 activity.status = 'success';
+                                if (paidInvoice) {
+                                    activity.invoice = paidInvoice;
+                                }
                             });
                         } else if (activity.invoice.isExpired) {
                             runInAction(() => {
                                 activity.status = 'expired';
                             });
+                        } else {
+                            pendingCashuInvoiceActivities.push(activity);
                         }
                     } else {
                         pendingLightningInvoiceActivities.push(activity);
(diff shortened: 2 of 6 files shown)


utils/NostrConnectUtils.ts
kaloudis, line 1: Let's get some unit tests for these.

kaloudis (changes requested):

stores/NostrWalletConnectStore.ts
```diff
@@ -997,6 +997,50 @@ export default class NostrWalletConnectStore {
...
+        const pendingPayInvoiceActivities = connection.activity.filter(
+            (activity) =>
+                activity.type === 'pay_invoice' && activity.status === 'pending'
+        );
+        if (pendingPayInvoiceActivities.length > 0) {
+            const lightningPending = pendingPayInvoiceActivities.filter(
+                (activity) => activity.payment_source !== 'cashu'
+            );
+            if (lightningPending.length > 0) {
+                await this.paymentsStore.getPayments();
```
kaloudis: `paymentsStore.getPayments()` is a full backend round-trip (`BackendUtils.getPayments`). With this change it fires from `getActivities` every time the view refreshes and any pay_invoice activity is pending. On busy connections (LND over LNC with long histories especially) that can add noticeable latency and load. Worth considering: gate by age of the pending entry, debounce per connection, or piggyback on the existing periodic poll instead of running on every UI refresh.

stores/NostrWalletConnectStore.ts
```diff
@@ -997,6 +997,50 @@ export default class NostrWalletConnectStore {
...
+                            p.getPaymentRequest === activity.id ||
+                            (!!activity.paymentHash &&
+                                p.paymentHash === activity.paymentHash)
+                    );
+                    if (!payment) continue;
+
+                    runInAction(() => {
+                        activity.payment = new Payment(payment);
+                        if (!payment.isIncomplete) {
+                            const wasPending = activity.status === 'pending';
```
kaloudis: `wasPending` is always `true` here — `lightningPending` was already filtered for `status === 'pending'` at line 1006. The conditional can be dropped.

stores/NostrWalletConnectStore.ts
```diff
@@ -997,6 +997,50 @@ export default class NostrWalletConnectStore {
...
+                            activity.status = 'success';
+                            if (wasPending) {
+                                const amountSats =
+                                    Math.floor(Number(activity.satAmount)) ||
+                                    Math.floor(
+                                        Number(activity.payment?.getAmount) || 0
```
kaloudis: Nit: `activity.payment` was just assigned `new Payment(payment)` two lines above, so the optional chain isn't doing anything here.

stores/NostrWalletConnectStore.ts
```diff
@@ -1776,6 +1820,40 @@ export default class NostrWalletConnectStore {
...
+            });
+
+            const fees_paid =
+                this.transactionsStore.payment_fee ||
+                inTransitResult.payment?.getFee ||
+                0;
+
+            return {
+                result: {
+                    preimage: '',
```
kaloudis, line 1935: Returning `preimage: ''` as a deliberate "in-transit" signal is fine — NIP-47's `pay_invoice` response shape requires the field but doesn't explicitly forbid an empty string. Flagging because some strict NWC clients may treat an empty preimage as a failure or malformed response. A one-line comment by this return explaining the convention would help a future reader not "fix" it.

stores/NostrWalletConnectStore.ts
```diff
@@ -1960,7 +2040,17 @@ export default class NostrWalletConnectStore {
         const payment = this.cashuStore.payments?.find(
             (p) => p.getPaymentRequest === request.invoice
         );
-        if (payment) {
+
+        if (cashuInTransit) {
```
kaloudis: This branch is currently unreachable: `CashuStore.payLnInvoiceFromEcash` and `payLnInvoiceFromEcashMultiMint` never set `status` on the returned `CashuPayment` (multi-mint explicitly throws if state isn't `'PAID'`), so `cashuInvoice?.isInTransit` (which checks `status === 'IN_FLIGHT'`) is always false today. There's also no Cashu branch in the `getActivities` reconciliation loop — it filters with `payment_source !== 'cashu'`. So if `CashuStore` ever starts surfacing IN_FLIGHT melts, NWC will silently accumulate stuck-pending entries here with no path to resolution. Suggest either wiring the matching Cashu reconciliation in `getActivities` or dropping this `recordPendingPayment` call until the cashu store actually surfaces in-flight state.

kaloudis (commented):
Some inline notes from a closer pass. Direction looks good — flagging two real bugs (`paymentHash` overwrite, missing decoded-hash fallback) plus a few smaller observations.

stores/NostrWalletConnectStore.ts
```diff
@@ -2173,6 +2285,60 @@ export default class NostrWalletConnectStore {
...
+        paymentHash
+    }: {
+        rawInvoice: string;
+        connection: NWCConnection;
+        amountSats: number;
+        payment_source: ConnectionPaymentSourceType;
+        payment?: Payment | CashuPayment | null;
+        paymentHash?: string;
+    }): Promise<void> {
+        const { paymentRequest } = await NostrConnectUtils.decodeInvoiceTags(
```
kaloudis: `decodeInvoiceTags` already returns `paymentHash` — worth destructuring it and using it as a last-resort fallback so the activity is always reconcilable by hash (matches gemini-code-assist's note). For backends that don't populate `payment_request` on listed payments (LDK Node is the obvious case), the `paymentHash` match in `getActivities` is the only thing that finds the payment later.

```ts
const { paymentRequest, paymentHash: decodedHash } =
    await NostrConnectUtils.decodeInvoiceTags(rawInvoice);
// ...
paymentHash: paymentHash || decodedHash || undefined
```

stores/NostrWalletConnectStore.ts
```diff
@@ -2173,6 +2285,60 @@ export default class NostrWalletConnectStore {
...
+            rawInvoice
+        );
+        const id = paymentRequest || rawInvoice;
+        const index = connection.activity.findIndex((conn) => conn.id === id);
```
kaloudis: Nit: this is the activity array, so the parameter should be `activity` rather than `conn`.

stores/NostrWalletConnectStore.ts
```diff
@@ -2173,6 +2285,60 @@ export default class NostrWalletConnectStore {
...
+                    payment_source === 'cashu'
+                        ? new CashuPayment(payment)
+                        : new Payment(payment);
+            }
+
+            if (index !== -1) {
+                connection.activity[index] = {
+                    ...connection.activity[index],
+                    ...record
+                };
```
kaloudis, line 2476: Bug: `record` always contains the `paymentHash` key, even when the parameter is `undefined` (`{ paymentHash }` shorthand on line 2318). Spreading `...record` here then overwrites a previously-stored `paymentHash` on the existing activity with `undefined`.

Narrow trigger (a duplicate `pay_invoice` for the same invoice while the first is still pending, with the second pass not having a `payment_hash` populated), but it wipes the field `getActivities` uses to reconcile by hash. Either build `record` conditionally or filter undefined values before spreading.

utils/NostrConnectUtils.ts
```diff
@@ -867,6 +867,142 @@ export default class NostrConnectUtils {
         );
     }
 
+    static isInFlightPaymentStatus(
+        status: string | number | null | undefined
+    ): boolean {
+        return status === 'IN_FLIGHT' || status === 1 || status === '1';
```
kaloudis, line 874: Worth a one-line comment that `1` maps to `lnrpc.Payment.PaymentStatus.IN_FLIGHT` (protobuf enum) and `'1'` covers stringified variants — otherwise it reads like one of the cases is dead.

utils/NostrConnectUtils.ts
```diff
@@ -867,6 +867,142 @@ export default class NostrConnectUtils {
...
+    }
+
+    static isPaymentTimedOutMessage(message?: string | null): boolean {
+        if (!message) return false;
+        const normalized = message.toLowerCase();
+        return (
+            normalized.includes('timed out') ||
```
kaloudis, line 882: The `.includes('timed out')` check is English-only — non-English locales rely on the `=== localeString(...)` fallbacks below. Works as long as the message originates from one of those two `localeString` calls, but is fragile if the error path ever surfaces a backend-localized "timed out" equivalent in another language. Minor — flagging for future readers.

stores/NostrWalletConnectStore.ts
```diff
@@ -2173,6 +2285,60 @@ export default class NostrWalletConnectStore {
...
+        payment?: Payment | CashuPayment | null;
+        paymentHash?: string;
+    }): Promise<void> {
+        const { paymentRequest } = await NostrConnectUtils.decodeInvoiceTags(
+            rawInvoice
+        );
+        const id = paymentRequest || rawInvoice;
+        const index = connection.activity.findIndex((conn) => conn.id === id);
+        const activity = index !== -1 ? connection.activity[index] : undefined;
+        if (activity?.status === 'success') return;
```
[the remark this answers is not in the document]
kaloudis: Good direction — this is the right fix. Two notes:

- The three-tier fallback (`passedPaymentHash || decodedPaymentHash || payment?.paymentHash`) substantially narrows but doesn't fully eliminate the related spread-overwrite at lines 2329-2333. `record.paymentHash` is still a present key (just with a value computed from three sources instead of one), so spreading still clobbers a previously-stored hash in the all-undefined edge case. If you want a belt-and-suspenders fix, build `record` so `paymentHash` is only set when truthy:

  ```ts
  const record: ConnectionActivity = {
      id, type: 'pay_invoice', satAmount: amountSats,
      status: 'pending', payment_source,
      createdAt: new Date(),
      ...(paymentHash ? { paymentHash } : {})
  };
  ```

- Fallback order is fine. `passedPaymentHash` (from `transactionsStore`), `decodedPaymentHash` (from bolt11), and `payment?.paymentHash` should all resolve to the same hex string for a valid invoice, so ordering doesn't matter for correctness — live state → invoice → listed payment is reasonable.

The `passedPaymentHash` rename is a style call; either that or computing inline works.

models/NWCConnection.ts
```diff
@@ -222,6 +222,38 @@ export default class NWCConnection extends BaseModel {
         return Math.min(100, (this.totalSpendSats / this.maxAmountSats!) * 100);
     }
 
+    private resolveMakeInvoiceActivitySats(a: ConnectionActivity): number {
+        const fromSat = Math.floor(Number(a.satAmount) || 0);
```
kaloudis, line 226: Precedence here flipped from `invoice → satAmount → payment` to `satAmount → invoice → payment`. The new order is the right call for amountless invoices (LNURL-pay etc.) where the bolt11 `getAmount` is 0 but `satAmount` holds the actually-received amount — but it's a behavioral change that's easy to miss in a diff. A one-line `// satAmount first: amountless invoices report 0 from bolt11` would prevent a future "fix" reverting it.

stores/NostrWalletConnectStore.ts
```diff
@@ -161,6 +165,14 @@ export default class NostrWalletConnectStore {
     private androidLogListener: EmitterSubscription | null = null;
     private _scheduledSave: ReturnType<typeof setTimeout> | null = null;
     private resetInFlight: Promise<void> | null = null;
+    private lastPendingPaymentStatusFetchByConnection = new Map<
```
kaloudis, line 168: Nit: these Maps aren't cleared in `deleteConnection` (line 620). Each entry is ~50 bytes and bounded by total deletions in a session, so not a real leak — but a `.delete(connectionId)` for both Maps alongside the existing `unsubscribeFromConnection` teardown would be tidier.

kaloudis (commented):
Two small follow-ups from a closer pass on the last three commits. Neither blocks merge.

kaloudis (approved):

PR #4153, opened 2026-06-09
fix: validate Nostr Wallet Connect URIs before saving config

ajaysehwal:
# Description

# Description

Prevents invalid Nostr Wallet Connect (NWC) connection strings from being saved or submitted by validating them against NIP-47 before the user can continue.

Previously, NWC only checked that the value started with `nostr+walletconnect://`. Users could save empty or malformed URIs (including via QR scan, which prefills state without firing `onChangeText`). That produced broken wallet entries and late failures when `@getalby/sdk` tried to connect. The same weak prefix check existed on the ZEUS Pay NWC lightning-address create flow.

## Changes

- **`ValidationUtils.isValidNostrWalletConnectUrl()`**
  - Requires `nostr+walletconnect://` scheme
  - 64-char hex wallet-service pubkey as host (upper/lower allowed; parsed via the same `http://` host-swap as `@getalby/sdk` `parseWalletConnectUrl`)
  - Required `secret`: 64-char **lowercase** hex (matches `@getalby/sdk` `/^[0-9a-f]{64}$/`)
  - Required ≥1 `relay` params; each validated with `new URL()` and must be `ws:` or `wss:` with a non-empty host
  - Allows optional params (e.g. `lud16`) and multiple relays
  - Trims input; rejects null / undefined / empty / non-string

- **`WalletConfiguration.tsx`**
  - Derives validity from current state on every render (covers QR-prefill)
  - Disables Save while the NWC URL is invalid
  - Error text color only when the field is non-empty and invalid
  - Scoped to `implementation === 'nostr-wallet-connect'`

- **`CreateNWCLightningAddress.tsx`**
  - Reuses the same validator (replaces prefix-only check)
  - Error text color for invalid non-empty input
  - Normalizes input on change (trim + collapse whitespace)
  - Disables Test until the connection string is valid

- **Unit tests** for valid URIs, multi-relay, `lud16`, whitespace, trailing slash, uppercase pubkey, uppercase secret rejection, `ws://` local relays, paths/ports/IPv6, embedded spaces, no-slash scheme, missing params, and non-NWC schemes

## Deliberate decisions

- `ws://` cleartext relays are **accepted** for local/regtest (documented in code + tested)
- `nostr+walletconnect:` (no `//`) is **rejected** even if some SDKs accept it
- Legacy `nostrwalletconnect://` remains rejected (unchanged)
- Relay check uses `new URL()` for parity with `@getalby/sdk` at connect time

## Out of scope

- No migration of already-saved invalid NWC wallet configs (user must re-edit)

## Test plan

- [x] **Wallet config:** paste valid NWC URI → Save enabled, field not red
- [x] **Wallet config:** clear field → Save disabled, field not red
- [x] **Wallet config:** missing `secret` / `relay` / bad pubkey → Save disabled, field red
- [x] **Wallet config:** QR-scan valid NWC URI → Save enabled without typing
- [x] **Wallet config:** QR-scan invalid string → Save disabled, field red
- [x] **Create NWC LN address:** invalid string → Test disabled, field red
- [x] **Create NWC LN address:** valid string → Test enabled → after success, Create/Switch works
- [x] Save and connect a real NWC wallet (e.g. Alby) end-to-end
- [x] `yarn test` — `isValidNostrWalletConnectUrl` suite green

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/ValidationUtils.test.ts
+++ b/utils/ValidationUtils.test.ts
@@ -326,6 +326,134 @@ describe('hasValidPairingPhraseCharsAndWordcount', () => {
     });
 });
 
+describe('isValidNostrWalletConnectUrl', () => {
+    const walletPubKeyHex =
+        'b889ff5b1513b641e2a139f661a661364979c5beee91842f8f0ef42ab558e9d4';
+    const secret =
+        '71a8c14c1407c113601079c4302dab36460f0ccd0ad506f1f2dc73b5100e4f3c';
+
+    it('accepts a complete nostr wallet connect URI', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss%3A%2F%2Frelay.damus.io&secret=${secret}`
+            )
+        ).toBe(true);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('accepts optional query params such as lud16', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret}&lud16=alice@domain.com`
+            )
+        ).toBe(true);
+    });
+
+    it('accepts multiple relay parameters', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.one.com&relay=wss://relay.two.com&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('accepts URLs with surrounding whitespace', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `  nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret}  `
+            )
+        ).toBe(true);
+    });
+
+    it('accepts relay URLs with literal percent characters', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com/50%25_discount&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('rejects empty, whitespace-only, null, and undefined URLs', () => {
+        expect(ValidationUtils.isValidNostrWalletConnectUrl('')).toBe(false);
+        expect(ValidationUtils.isValidNostrWalletConnectUrl('   ')).toBe(false);
+        expect(ValidationUtils.isValidNostrWalletConnectUrl(null as any)).toBe(
+            false
+        );
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(undefined as any)
+        ).toBe(false);
+    });
+
+    it('rejects URIs missing required query params', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?secret=${secret}`
+            )
+        ).toBe(false);
+    });
+
+    it('rejects malformed pubkeys, secrets, and relays', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                'nostr+walletconnect://'
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                'nostr+walletconnect://:'
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://abc123?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com&secret=tooshort`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=https://relay.example.com&secret=${secret}`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(false);
+    });
+
+    it('rejects non-nwc URLs', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl('https://example.com')
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl('nostr+walletconnect')
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                'nostr+walletconnect:/abc123'
+            )
+        ).toBe(false);
+    });
+});
+
 describe('validateNodePubkey', () => {
     it('accepts valid 66-character hex pubkeys', () => {
         expect(
--- a/utils/ValidationUtils.test.ts
+++ b/utils/ValidationUtils.test.ts
@@ -401,6 +401,32 @@ describe('isValidNostrWalletConnectUrl', () => {
         ).toBe(false);
     });
 
+    it('accepts ws:// relays for local regtest', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=ws://127.0.0.1:8080&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('accepts relay URLs with paths, ports, and IPv6 hosts', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com/path?x=1&secret=${secret}`
+            )
+        ).toBe(true);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com:443&secret=${secret}`
+            )
+        ).toBe(true);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://[::1]:8080&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
     it('rejects relay URLs with embedded spaces', () => {
         expect(
             ValidationUtils.isValidNostrWalletConnectUrl(
@@ -410,6 +436,14 @@ describe('isValidNostrWalletConnectUrl', () => {
         expect(
             ValidationUtils.isValidNostrWalletConnectUrl(
                 `nostr+walletconnect://${walletPubKeyHex}?relay=wss://a+b.io&secret=${secret}`
+            )
+        ).toBe(false);
+    });
+
+    it('rejects no-slash nostr+walletconnect: form', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect:${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret}`
             )
         ).toBe(false);
     });
@@ -467,6 +501,11 @@ describe('isValidNostrWalletConnectUrl', () => {
         expect(
             ValidationUtils.isValidNostrWalletConnectUrl(
                 `nostr+walletconnect://${walletPubKeyHex}?relay=https://relay.example.com&secret=${secret}`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://&secret=${secret}`
             )
         ).toBe(false);
         expect(
--- a/utils/ValidationUtils.test.ts
+++ b/utils/ValidationUtils.test.ts
@@ -385,12 +385,13 @@ describe('isValidNostrWalletConnectUrl', () => {
         ).toBe(true);
     });
 
-    it('accepts uppercase hex pubkeys', () => {
+    it('rejects uppercase hex pubkeys', () => {
+        // React Native's URL polyfill preserves host case, unlike Node's URL.
         expect(
             ValidationUtils.isValidNostrWalletConnectUrl(
                 `nostr+walletconnect://${walletPubKeyHex.toUpperCase()}?relay=wss://relay.example.com&secret=${secret}`
             )
-        ).toBe(true);
+        ).toBe(false);
     });
 
     it('rejects uppercase hex secrets', () => {
@@ -444,6 +445,22 @@ describe('isValidNostrWalletConnectUrl', () => {
         expect(
             ValidationUtils.isValidNostrWalletConnectUrl(
                 `nostr+walletconnect:${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(false);
+    });
+
+    it('rejects legacy nostrwalletconnect scheme', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostrwalletconnect://${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(false);
+    });
+
+    it('rejects non-WebSocket relay URLs', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=http://relay.example.com&secret=${secret}`
             )
         ).toBe(false);
     });
--- a/utils/ValidationUtils.test.ts
+++ b/utils/ValidationUtils.test.ts
@@ -377,6 +377,43 @@ describe('isValidNostrWalletConnectUrl', () => {
         ).toBe(true);
     });
 
+    it('accepts a trailing slash after the pubkey', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}/?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('accepts uppercase hex pubkeys', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex.toUpperCase()}?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('rejects uppercase hex secrets', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://relay.example.com&secret=${secret.toUpperCase()}`
+            )
+        ).toBe(false);
+    });
+
+    it('rejects relay URLs with embedded spaces', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://a b&secret=${secret}`
+            )
+        ).toBe(false);
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}?relay=wss://a+b.io&secret=${secret}`
+            )
+        ).toBe(false);
+    });
+
     it('rejects empty, whitespace-only, null, and undefined URLs', () => {
         expect(ValidationUtils.isValidNostrWalletConnectUrl('')).toBe(false);
         expect(ValidationUtils.isValidNostrWalletConnectUrl('   ')).toBe(false);
--- a/utils/ValidationUtils.ts
+++ b/utils/ValidationUtils.ts
@@ -4,6 +4,9 @@ const HOST_REGEX =
 // in paths we have more allowed chars
 const PATH_REGEX = /^\/([a-zA-Z0-9\-._~!$&'()*+,;=]+\/?)*$/;
 const PORT_REGEX = /^:\d+$/;
+const NOSTR_WALLET_CONNECT_PREFIX = 'nostr+walletconnect://';
+const NWC_HEX_32_BYTES = /^[a-fA-F0-9]{64}$/;
+const NWC_RELAY_URL = /^wss?:\/\/.+/i;
 
 interface ValidationOptions {
     requireHttps?: boolean;
@@ -77,6 +80,34 @@ const hasValidPairingPhraseCharsAndWordcount = (phrase: string): boolean => {
     const normalizedPhrase = phrase.trim().replace(/\s+/g, ' ');
     if (!/^[a-zA-Z\s]+$/.test(normalizedPhrase)) return false;
     return normalizedPhrase.split(' ').length === 10;
+};
+
+const isValidNwcRelayUrl = (relay: string): boolean => {
+    return typeof relay === 'string' && NWC_RELAY_URL.test(relay);
+};
+
+const isValidNostrWalletConnectUrl = (url: string): boolean => {
+    if (!url || typeof url !== 'string') return false;
+    const normalizedUrl = url.trim().replace(/\s+/g, ' ');
+    if (!normalizedUrl.startsWith(NOSTR_WALLET_CONNECT_PREFIX)) return false;
+
+    const pathAndQuery = normalizedUrl.slice(
+        NOSTR_WALLET_CONNECT_PREFIX.length
+    );
+    const queryIndex = pathAndQuery.indexOf('?');
+    if (queryIndex === -1) return false;
+
+    const walletPubKeyHex = pathAndQuery.slice(0, queryIndex);
+    if (!NWC_HEX_32_BYTES.test(walletPubKeyHex)) return false;
+
+    const params = new URLSearchParams(pathAndQuery.slice(queryIndex + 1));
+    const relays = params.getAll('relay');
+    const secret = params.get('secret');
+
+    if (!secret || !NWC_HEX_32_BYTES.test(secret)) return false;
+    if (relays.length === 0 || !relays.every(isValidNwcRelayUrl)) return false;
+
+    return true;
 };
 
 const validateNodePubkey = (pubkey: string): boolean => {
@@ -160,6 +191,7 @@ const ValidationUtils = {
     hasValidRuneChars,
     hasValidMacaroonChars,
     hasValidPairingPhraseCharsAndWordcount,
+    isValidNostrWalletConnectUrl,
     validateNodePubkey,
     validateNodeHost
 };
--- a/utils/ValidationUtils.ts
+++ b/utils/ValidationUtils.ts
@@ -6,8 +6,9 @@ const PATH_REGEX = /^\/([a-zA-Z0-9\-._~!$&'()*+,;=]+\/?)*$/;
 const PORT_REGEX = /^:\d+$/;
 const NOSTR_WALLET_CONNECT_PREFIX = 'nostr+walletconnect://';
 const NWC_HEX_32_BYTES = /^[a-fA-F0-9]{64}$/;
+// @getalby/sdk checks secret against /^[0-9a-f]{64}$/; query values are not
+// case-folded like URL hosts, so uppercase hex must be rejected here.
 const NWC_SECRET_HEX = /^[0-9a-f]{64}$/;
-const NWC_RELAY_URL = /^wss?:\/\/\S+$/i;
 
 interface ValidationOptions {
     requireHttps?: boolean;
@@ -84,7 +85,17 @@ const hasValidPairingPhraseCharsAndWordcount = (phrase: string): boolean => {
 };
 
 const isValidNwcRelayUrl = (relay: string): boolean => {
-    return typeof relay === 'string' && NWC_RELAY_URL.test(relay);
+    if (typeof relay !== 'string') return false;
+    try {
+        const parsed = new URL(relay);
+        // ws:// allowed for local/regtest relays; wss:// for production.
+        return (
+            (parsed.protocol === 'wss:' || parsed.protocol === 'ws:') &&
+            !!parsed.host
+        );
+    } catch {
+        return false;
+    }
 };
 
 const isValidNostrWalletConnectUrl = (url: string): boolean => {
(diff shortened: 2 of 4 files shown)


views/Settings/WalletConfiguration.tsx
```diff
@@ -3066,6 +3067,11 @@ export default class WalletConfiguration extends React.Component<
                                                     pairingPhrase
                                                 )) ||
                                             nostrWalletConnectUrlError ||
+                                            (implementation ===
+                                                'nostr-wallet-connect' &&
+                                                !ValidationUtils.isValidNostrWalletConnectUrl(
+                                                    nostrWalletConnectUrl
+                                                )) ||
```
kaloudis: This check is what covers the QR scan path (`handleAnything` prefills `nostrWalletConnectUrl` via route params without firing `onChangeText`), which is good. But in that flow `nostrWalletConnectUrlError` stays `false`, so an invalid scanned URI renders in the normal text color while the save button is silently disabled, and the user gets no explanation.

Suggest deriving the error styling from the validator directly, e.g. making the `textColor` condition `nostrWalletConnectUrl && !ValidationUtils.isValidNostrWalletConnectUrl(nostrWalletConnectUrl)`, or validating once in `initFromProps`. Either would also let you drop the now-duplicated `nostrWalletConnectUrlError ||` a few lines above.

utils/ValidationUtils.ts
```diff
@@ -4,6 +4,9 @@ const HOST_REGEX =
 // in paths we have more allowed chars
 const PATH_REGEX = /^\/([a-zA-Z0-9\-._~!$&'()*+,;=]+\/?)*$/;
 const PORT_REGEX = /^:\d+$/;
+const NOSTR_WALLET_CONNECT_PREFIX = 'nostr+walletconnect://';
+const NWC_HEX_32_BYTES = /^[a-fA-F0-9]{64}$/;
+const NWC_RELAY_URL = /^wss?:\/\/.+/i;
```
kaloudis: `.+` matches spaces, so `wss://a b` and `wss:// ` pass. `/^wss?:\/\/\S+$/i` is a cheap tightening; parsing each relay with `new URL()` (like `@getalby/sdk` does on connect) would be stricter still.

Also flagging that `ws://` cleartext relays are accepted. That seems reasonable for local dev/regtest, but worth a conscious yes/no.

utils/ValidationUtils.ts
```diff
@@ -79,6 +82,34 @@ const hasValidPairingPhraseCharsAndWordcount = (phrase: string): boolean => {
     return normalizedPhrase.split(' ').length === 10;
 };
 
+const isValidNwcRelayUrl = (relay: string): boolean => {
+    return typeof relay === 'string' && NWC_RELAY_URL.test(relay);
+};
+
+const isValidNostrWalletConnectUrl = (url: string): boolean => {
+    if (!url || typeof url !== 'string') return false;
+    const normalizedUrl = url.trim().replace(/\s+/g, ' ');
```
kaloudis: nit: the inner `.replace(/\s+/g, ' ')` can never turn an invalid URI valid, `trim()` alone is enough here. The caller already does the collapsing before the value reaches this function.

utils/ValidationUtils.ts
```diff
@@ -79,6 +82,34 @@ const hasValidPairingPhraseCharsAndWordcount = (phrase: string): boolean => {
...
+    if (queryIndex === -1) return false;
+
+    const walletPubKeyHex = pathAndQuery.slice(0, queryIndex);
+    if (!NWC_HEX_32_BYTES.test(walletPubKeyHex)) return false;
+
+    const params = new URLSearchParams(pathAndQuery.slice(queryIndex + 1));
+    const relays = params.getAll('relay');
+    const secret = params.get('secret');
+
+    if (!secret || !NWC_HEX_32_BYTES.test(secret)) return false;
```
kaloudis: Deliberate-strictness flag: `@getalby/sdk` accepts `nsec...`-encoded secrets (nip19-decoded in the `NWCClient` constructor), so a previously scanned-and-saved wallet with an nsec secret works today but couldn't be re-saved after this change without replacing the URL. NIP-47 specifies hex, so rejecting nsec seems right; just noting it so it's a conscious decision.

kaloudis (commented):
Concept ACK

Nice catch on the missing validation here. A few things I verified while reviewing, for the record:

- `URLSearchParams` is safe on device: RN's built-in implementation is a stub whose `get`/`getAll` throw, but we load `react-native-url-polyfill/auto` in `index.js`, so the full implementation is global at runtime. Jest runs on Node's native implementation and wouldn't have caught this, so it was worth checking.
- The new save-button check is not redundant with `nostrWalletConnectUrlError`: NWC had no required-input check before (unlike lndhub/lnd/cln-rest/LNC), so a brand-new NWC wallet could previously be saved with an empty URL, and the QR scan flow prefills the field without firing `onChangeText`. The direct check closes both gaps.
- No regression for legacy `nostrwalletconnect://` URIs: both the old prefix check and `handleAnything` already rejected those, so nothing that could be pasted or scanned before is newly blocked except genuinely malformed URIs.

Left a few inline comments. The main ask is the scanned-URI UX one; the rest are nits or deliberate-decision flags.

Two cheap test additions worth considering: an uppercase-hex acceptance case (the regex allows `A-F`, and `@getalby/sdk` lowercases the pubkey via URL host, so pinning that behavior is useful), and a case documenting that the no-slash `nostr+walletconnect:` form is rejected (the SDK would accept it, so the test records the deliberate strictness).

Could you also test the QR scan path (scanning an NWC URI into wallet creation) and a real NWC save-and-connect on device? The scan path is where the inline UX comment manifests. Out of scope for this PR, but `CreateNWCLightningAddress.tsx` does a prefix-only check on the same URI type; reusing the new validator there would be a natural follow-up.

views/Settings/WalletConfiguration.tsx
```diff
@@ -1784,18 +1784,19 @@ export default class WalletConfiguration extends React.Component<
                                         }
                                         value={nostrWalletConnectUrl}
                                         autoCapitalize="none"
-                                        onChangeText={(text: string) =>
+                                        onChangeText={(text: string) => {
+                                            const nostrWalletConnectUrl = text
```
kaloudis: nit: this shadows the `nostrWalletConnectUrl` destructured from state in the render scope. Works fine, but a name like `normalizedUrl` would read better.

views/Settings/WalletConfiguration.tsx
```diff
@@ -3405,6 +3407,11 @@ export default class WalletConfiguration extends React.Component<
                                                     pairingPhrase
                                                 )) ||
                                             nostrWalletConnectUrlError ||
+                                            (implementation ===
```
kaloudis: One more item surfaced on re-review: the stale `nostrWalletConnectUrlError` flag can disable saving for other wallet interfaces. The gate keeps the unconditional `nostrWalletConnectUrlError ||` just above this new check, and the wallet interface dropdown's `onValueChange` resets `host`/`lndhubUrl` but not the NWC error flag or URL. Repro: select Nostr Wallet Connect, type a partial URI (with the stricter validator any incomplete URI now sets the flag), switch the interface to LND, fill in valid details. The save button stays disabled with no visible reason, since the NWC field is no longer rendered.

The gate shape predates this PR, but the fix lines up with the earlier styling ask: drop `nostrWalletConnectUrlError` from the gate and derive both the gate and the input styling from `ValidationUtils.isValidNostrWalletConnectUrl`, scoped to `implementation === 'nostr-wallet-connect'` (or reset the flag on interface change).

utils/ValidationUtils.ts
```diff
@@ -79,6 +82,34 @@ const hasValidPairingPhraseCharsAndWordcount = (phrase: string): boolean => {
...
+    const normalizedUrl = url.trim().replace(/\s+/g, ' ');
+    if (!normalizedUrl.startsWith(NOSTR_WALLET_CONNECT_PREFIX)) return false;
+
+    const pathAndQuery = normalizedUrl.slice(
+        NOSTR_WALLET_CONNECT_PREFIX.length
+    );
+    const queryIndex = pathAndQuery.indexOf('?');
+    if (queryIndex === -1) return false;
+
+    const walletPubKeyHex = pathAndQuery.slice(0, queryIndex);
```
kaloudis: This is the first of the two divergences from `@getalby/sdk`. `pathAndQuery.slice(0, queryIndex)` includes any path segment, so `nostr+walletconnect://<pubkey>/?relay=...&secret=...` yields `<pubkey>/` and fails the hex test.

That form works today: the SDK rewrites the scheme to `http://` and reads `url.host`, so the trailing slash is dropped, and `handleAnything.ts` only checks the `nostr+walletconnect://` prefix, so it can be scanned in. Before this PR the prefix-only check let it save too.

Two consequences worth weighing. A user scanning such a URI now gets a red field and a disabled save with no way to proceed except hand-editing the string. And because the same save gate covers edits to existing wallets, anyone who already has one of these saved can no longer change their wallet nickname or photo, since `initFromProps` loads the stored URL straight into state and the gate re-evaluates it.

Parsing through `new URL` after swapping in a special scheme, as suggested in the review body, fixes this and gets host lowercasing for free.

utils/ValidationUtils.ts
kaloudis: Second divergence, and the one that most directly undercuts the PR's premise. `[a-fA-F0-9]{64}` accepts an uppercase-hex secret, but `NWCClient.parseWalletConnectUrl` checks the secret against `/^[0-9a-f]{64}$/`, and query values are not case-folded by URL parsing the way the host is. So an uppercase secret validates here, saves, and then throws `Invalid NWC URL: invalid secret` out of `new NostrWebLNProvider(...)` in `NostrWalletConnect.initNWC`.

That throw is not caught anywhere on the path: `SettingsStore.connectNWC` sets `this.loading = true` and awaits `initNWC` with no try/catch, and `Wallet.tsx` awaits `connectNWC()` outside its try block. The result is a wallet that saves cleanly and then wedges on the connecting spinner, which is exactly the class of failure this validator exists to prevent.

The pubkey is fine either way, since `new URL` lowercases the host before the SDK's hex test, so this only needs to be tightened for the secret.

utils/ValidationUtils.ts
kaloudis: Still open from round 1, and I can now show it matters rather than just asserting it. `.+` matches spaces, so both of these validate here and then throw `Invalid relay URL` inside `parseWalletConnectUrl`:

- `relay=wss://a b`
- `relay=wss://a+b.io`, since `URLSearchParams` decodes `+` to a space

Same wedge as the uppercase-secret case: saves fine, fails at connect. `/^wss?:\/\/\S+$/i` closes both, and I verified it keeps every one of your existing positive cases passing, including the percent-encoded and literal-`%` relays.

The `ws://` question from round 1 is still worth an explicit yes or no. The SDK accepts it and it is genuinely useful for local regtest, so I lean toward keeping it, I just want it to be a decision rather than a side effect of the regex.

kaloudis (commented):
Round-2 re-review of `34e307cb0`.

The scoping commit closes both outstanding structural items:

- the QR-scan path now gets error styling, derived from the validator rather than the `onChangeText`-only flag
- the stale `nostrWalletConnectUrlError` that could disable save after switching the wallet interface away from NWC is gone, since `nostrWalletConnectUrlInvalid` is scoped to `implementation === 'nostr-wallet-connect'`

I checked the whole tree for leftover references to the removed state field and found none, and CI is green (lint / prettier / test / tsc). Nice cleanup, the derived-value shape reads much better than the flag did.

While looking at the round-1 relay-regex nit I ran the validator side by side with `NWCClient.parseWalletConnectUrl` from the pinned `@getalby/sdk@8.0.3`, which is what actually consumes this URL in `backends/NostrWalletConnect.ts`. Since the premise of the PR is "do not let a config that cannot work get saved", anywhere the two disagree is worth a look. Four disagreements:

| case | this PR | @getalby/sdk 8.0.3 |
| --- | --- | --- |
| `nostr+walletconnect://<pubkey>/?relay=...&secret=...` (trailing slash) | reject | accept |
| uppercase hex `secret` | accept | reject |
| `relay=wss://a b` | accept | reject |
| `relay=wss://a+b.io` (`+` decodes to a space) | accept | reject |

Two are new should-fixes, two are the round-1 `.+` nit with evidence attached. Details inline.

One shape that covers all four, mirroring the SDK's own scheme-swap parse so the two cannot drift apart again:

```ts
const NWC_SECRET_HEX = /^[0-9a-f]{64}$/; // the SDK rejects uppercase secrets
const NWC_RELAY_URL = /^wss?:\/\/\S+$/i;

const isValidNostrWalletConnectUrl = (url: string): boolean => {
    if (!url || typeof url !== 'string') return false;
    const normalizedUrl = url.trim();
    if (!normalizedUrl.startsWith(NOSTR_WALLET_CONNECT_PREFIX)) return false;

    let parsed: URL;
    try {
        // swap in a special scheme so the pubkey parses as a host, the way
        // NWCClient.parseWalletConnectUrl does
        parsed = new URL(
            `http://${normalizedUrl.slice(NOSTR_WALLET_CONNECT_PREFIX.length)}`
        );
    } catch {
        return false;
    }

    if (!NWC_HEX_32_BYTES.test(parsed.host)) return false;

    const relays = parsed.searchParams.getAll('relay');
    const secret = parsed.searchParams.get('secret');

    if (!secret || !NWC_SECRET_HEX.test(secret)) return false;
    if (relays.length === 0 || !relays.every(isValidNwcRelayUrl)) return false;

    return true;
};
```

I ran all 22 vectors from your existing test file against that version and they all still hold, plus the trailing-slash case flips to accept and the three false accepts flip to reject. `parsed.host` includes any port, so a `<pubkey>:8080` form still fails the hex test, same as the SDK. Take it or leave it, the two regex changes plus a trailing-slash tolerance would also do.

Worth adding as test cases now that they are concrete: trailing slash accepted, uppercase secret rejected, uppercase pubkey accepted (`new URL` lowercases the host, so the SDK accepts it), relay with an embedded space rejected.

Note the SDK has moved since my June read: 8.0.3's `parseWalletConnectUrl` enforces `/^[0-9a-f]{64}$/` on the secret before the constructor's `nsec` branch can run, so the nsec compatibility flag I raised in round 1 is moot. Rejecting nsec is now exactly what the SDK does. I have closed that thread.

Still open from round 1, both cosmetic: the dead `.replace(/\s+/g, ' ')` in the validator and the `nostrWalletConnectUrl` shadowing inside `onChangeText`.

Minor UX note, not a blocker: the field is `secureTextEntry` by default (`hidden: true`), so on the scan path the invalid-URI signal is a row of red dots. That matches how the other secret fields behave, so I am fine with it, just flagging that colour alone is a quiet signal here.

Before this can get a tACK I still need device confirmation of the two paths the change touches: scanning an NWC URI into wallet creation, and a real NWC save-and-connect. The checklist has iOS ticked but no node type, so I want to be sure the scan path was exercised.

Unchanged follow-up, out of scope here: `CreateNWCLightningAddress.tsx` still does a prefix-only check on the same URI type and would be a natural second consumer of this validator.

views/Settings/WalletConfiguration.tsx
kaloudis: Addressed in `34e307cb0`. The derived `nostrWalletConnectUrlInvalid` now feeds both the `textColor` and the save gate, and it is scoped to `implementation === 'nostr-wallet-connect'`, so the duplicated check and the stale-flag problem from my 2026-08-30 comment are both gone. Verified there are no leftover references to the removed state field. Thanks.

utils/ValidationUtils.ts
kaloudis: Still open on `34e307cb0`, `ValidationUtils.ts` is unchanged. I have attached concrete evidence in a new comment on this line: `.+` lets `wss://a b` and `wss://a+b.io` through the validator and then `@getalby/sdk` throws `Invalid relay URL` at connect time. The `ws://` yes-or-no is still unanswered too.

utils/ValidationUtils.ts
kaloudis: Still open on `34e307cb0`. One update on the reasoning: the `trim()` is now load-bearing, because the validator runs against the raw state value on the QR-scan path, which `handleAnything` prefills without normalising. The `.replace(/\s+/g, ' ')` half is still dead though, collapsing internal whitespace cannot flip an invalid URI to valid.

Related wrinkle while I was here: the validator normalises the string it validates but the normalised form is never persisted, so a scanned URI with surrounding whitespace validates and then gets saved with the whitespace intact. It happens to work, since `new URL` strips leading and trailing spaces, but validating one string and saving a different one is the kind of thing that bites later. Normalising once in `initFromProps` would make the two agree.

utils/ValidationUtils.ts
kaloudis: Closing this one, it is moot. The pinned `@getalby/sdk@8.0.3` now checks the secret against `/^[0-9a-f]{64}$/` inside `parseWalletConnectUrl`, which throws before the constructor's `nsec` branch can run, so nsec secrets no longer work through this path regardless. Rejecting them here matches the SDK exactly. My June read was against an older version of the SDK.

The hex class does still need narrowing for the secret though, see the new comment on that line.

views/Settings/WalletConfiguration.tsx
kaloudis: Still open on `34e307cb0`. Slightly more worth doing now, since `nostrWalletConnectUrlInvalid` is computed from the outer `nostrWalletConnectUrl` a thousand lines up, so having a same-named local in `onChangeText` makes the reader check twice.

utils/ValidationUtils.ts
```diff
@@ -79,6 +84,49 @@ const hasValidPairingPhraseCharsAndWordcount = (phrase: string): boolean => {
...
+        // Relay values with unencoded ? or & in the outer query can confuse
+        // URL parsing; wallets almost always percent-encode those.
+        parsed = new URL(
+            `http://${normalizedUrl.slice(NOSTR_WALLET_CONNECT_PREFIX.length)}`
+        );
+    } catch {
+        return false;
+    }
+
+    if (!NWC_HEX_32_BYTES.test(parsed.host)) return false;
```
kaloudis: This is the one remaining case where the validator accepts a URL the SDK rejects, and it only shows up on device.

`NWC_HEX_32_BYTES` allows `A-F`, on the reasoning I gave you in round 3 that `new URL` lowercases the host before the SDK's own hex test. That reasoning was measured against Node, which is what Jest runs on. In the app, `index.js` loads `react-native-url-polyfill/auto`, and `setupURLPolyfill()` replaces `globalThis.URL` with `whatwg-url-without-unicode@8.0.0-3`, whose `domainToASCII` is an identity stub (`lib/url-state-machine.js:456`). It never case-folds the host:

```
new URL('http://B889FF...').host
  Node (Jest)    -> b889ff...
  polyfill (app) -> B889FF...
```

So on device `parsed.host` keeps its case here, passes this mixed-case regex, and then fails `NWC_HEX64 = /^[0-9a-f]{64}$/` inside `parseWalletConnectUrl`. Same save-then-throw shape as the uppercase secret from round 3.

Simplest fix is to stop having two hex constants and use the lowercase-only one for both, which is what the SDK does with its single `NWC_HEX64`:

```ts
// @getalby/sdk tests both the wallet pubkey and the secret against
// /^[0-9a-f]{64}$/. React Native's URL polyfill does not case-fold hosts
// the way Node does, so uppercase hex has to be rejected in both spots.
const NWC_HEX64 = /^[0-9a-f]{64}$/;
...
if (!NWC_HEX64.test(parsed.host)) return false;
```

If you would rather keep accepting uppercase pubkeys, the URL has to be lowercased before it is persisted, which is a larger change than this PR needs.

kaloudis (commented):
Round-4 review of `0c1689103`.

All four divergences from round 3 are fixed. I re-ran the validator side by side with `NWCClient.parseWalletConnectUrl` from the pinned `@getalby/sdk@8.0.3` over a 28-case corpus to confirm:

| case | round 3 | now | SDK 8.0.3 |
| --- | --- | --- | --- |
| trailing slash after pubkey | reject | accept | accept |
| uppercase hex secret | accept | reject | reject |
| `relay=wss://a b` | accept | reject | reject |
| `relay=wss://a+b.io` | accept | reject | reject |

The scheme-swap parse reads well, and validating each relay with `new URL` plus a `ws:`/`wss:` protocol check is stricter than the SDK in the right direction. Reusing the validator in `CreateNWCLightningAddress` closes the follow-up I raised in round 1. The `.replace(/\s+/g, ' ')` is gone from the validator and the `onChangeText` shadowing is resolved.

One new should-fix, and it is a correction to something I told you in round 3.

## Uppercase pubkeys: my round-3 claim was measured on Node, not on device

I wrote that `new URL` lowercases the host before the SDK's hex test, so uppercase pubkeys were safe to accept. That is true in Jest, which runs on Node's `URL`. It is not true in the app.

`index.js` loads `react-native-url-polyfill/auto`, and `setupURLPolyfill()` replaces `globalThis.URL` with `whatwg-url-without-unicode@8.0.0-3`. That build stubs `domainToASCII` to an identity function (`lib/url-state-machine.js:456`), which is where the spec's case folding happens:

```
new URL('http://B889FF...').host
  Node (Jest)    -> b889ff...   lowercased
  polyfill (app) -> B889FF...   unchanged
```

The SDK reads `url.host` and tests it against `NWC_HEX64 = /^[0-9a-f]{64}$/`, the same lowercase-only regex it applies to the secret. Installing the polyfill globals before importing the SDK, to simulate the device:

```
URI          : nostr+walletconnect://B889FF5B...?relay=wss://relay.damus.io&secret=<lowercase hex>
PR validator : true
SDK          : THROWS -> Invalid NWC URL: invalid wallet pubkey
```

Same shape as the uppercase-secret case from round 3: validates, saves, throws at connect. Under the device polyfill this is now the only case where the validator accepts something the SDK rejects. Every other disagreement goes the safe way, with this PR stricter. Details and a suggested fix inline, plus a note on the test that currently pins the Node behaviour.

Frequency is low, since NIP-47 implementations and Zeus's own `buildWalletConnectConnectionUrl` all emit lowercase hex. What makes it worth fixing is that the test asserts the acceptance, so it reads as verified.

### Correction on the failure symptom

In round 3 I said this class of failure wedges on the connecting spinner. Re-reading the path, it does not. The throw escapes `initNWC`, then `SettingsStore.connectNWC` (no try/catch at `SettingsStore.ts:2241`), then `Wallet.tsx:1210` (awaited outside its try), and lands in the outer handler at `Wallet.tsx:521`, which clears `connecting` and shows nothing. The user gets a wallet that saved cleanly and then arrives with no balance, no transactions and no error message. `SettingsStore.loading` is also left `true`, since `connectNWC` only clears it on the success path, which holds the Receive screen in its loading state (`views/Receive.tsx:1007`) until some other settings call resets it. Quieter than I described, still exactly the failure this validator exists to prevent.

## Deliberate strictness, no action needed

Everything else where the two disagree, measured under the device polyfill. All of it is the safe direction:

| case | this PR | SDK 8.0.3 |
| --- | --- | --- |
| `relay=http://...` or `relay=https://...` | reject | accept |
| missing `secret` | reject | accept, `requireSecret` defaults to `false` |
| `nostr+walletconnect:` no-slash form | reject | accept |
| `nostrwalletconnect://` legacy scheme | reject | accept |
| `NOSTR+WALLETCONNECT://` uppercase scheme | reject | accept |
| relay value containing `#` | reject | accept |

None of these regress anything reachable. `handleAnything.ts:417` only routes on the exact lowercase `nostr+walletconnect://` prefix and the pre-PR check used the same prefix, so nothing that could previously be scanned or pasted is newly blocked. Worth recording too: none of the SDK-accepted forms above can complete a connection anyway. Without a secret there is no signing key, and an `http` URL is not a valid WebSocket target, so rejecting them matches the premise of the PR rather than working against it.

I also checked the round-trip case, since it is the one most likely to bite: a connection string generated by Zeus's own NWC service (`NostrConnectUtils.buildWalletConnectConnectionUrl`, lowercase hex from `bytesToHex`, relays from `wss://relay.getalby.com/v1` and friends) validates cleanly, both in `WalletConfiguration` and in the ZEUS Pay create flow.

Still open from round 3 and I am fine leaving it: the validator normalises the string it checks but the un-normalised form is what gets saved. It works, since both `URL` implementations strip leading and trailing spaces, so it is a tidiness point rather than a bug.

## Testing

CI is green on lint, prettier, test and tsc. The branch is 13 commits behind master.

Before a tACK I still need device confirmation of the paths this touches, which has been outstanding since round 2:

- scanning an NWC URI into wallet creation, so the QR prefill path exercises the derived error styling
- a real NWC save-and-connect end to end
- now that `CreateNWCLightningAddress` is in scope, the ZEUS Pay NWC address create flow, including a failed test-connection

The checklist has iOS ticked but no node type, so I want to be sure the scan path was actually run rather than inferred.

utils/ValidationUtils.test.ts
```diff
@@ -326,6 +326,210 @@ describe('hasValidPairingPhraseCharsAndWordcount', () => {
...
+
+    it('accepts a trailing slash after the pubkey', () => {
+        expect(
+            ValidationUtils.isValidNostrWalletConnectUrl(
+                `nostr+walletconnect://${walletPubKeyHex}/?relay=wss://relay.example.com&secret=${secret}`
+            )
+        ).toBe(true);
+    });
+
+    it('accepts uppercase hex pubkeys', () => {
```
kaloudis: This one passes in Jest but pins behaviour that is wrong on device, for the reason in my comment on `ValidationUtils.ts`. Jest uses Node's `URL`, which lowercases the host; the app uses `whatwg-url-without-unicode`, which does not, so the URI this test accepts throws `Invalid NWC URL: invalid wallet pubkey` at connect time.

Suggest flipping it to `rejects uppercase hex pubkeys` with a short comment on why, so the next reader does not re-derive the Node behaviour and reintroduce the mixed-case class.

While you are in here, two more worth pinning now that they are measured: `relay=http://...` rejected (the SDK only does `new URL(relay)` with no scheme check, so this is a deliberate divergence), and the no-slash and legacy-scheme rejections you already documented in the PR body.

views/LightningAddress/CreateNWCLightningAddress.tsx
```diff
@@ -143,6 +157,12 @@ export default class CreateNWCLightningAddress extends React.Component<
                                                             this.setState({
                                                                 nwcConnectionString:
                                                                     text
+                                                                        .trim()
+                                                                        .replace(
+                                                                            /\s+/g,
+                                                                            ' '
+                                                                        ),
+                                                                tested: false
```
kaloudis: nit: `tested: false` is unreachable. The whole `TextInput` is inside `{!tested && (...)}` a few lines up, so `onChangeText` can only fire while `tested` is already `false`. Harmless, but it reads like it is guarding a real transition.

Separate nit on the line above: collapsing internal whitespace with `.replace(/\s+/g, ' ')` cannot turn an invalid URI into a valid one, so it only changes what gets sent to the server. If the intent is to repair a line-wrapped paste, `.replace(/\s+/g, '')` would actually do that; otherwise `trim()` alone is enough. Same dead normalisation exists in `WalletConfiguration`, so matching it is defensible, just flagging that neither does anything.

views/LightningAddress/CreateNWCLightningAddress.tsx
```diff
@@ -226,7 +246,10 @@ export default class CreateNWCLightningAddress extends React.Component<
                                             }
                                         }}
                                         disabled={
-                                            !nwcConnectionString || loading
+                                            loading ||
+                                            !ValidationUtils.isValidNostrWalletConnectUrl(
```
kaloudis: nit: this runs the validator a second time per render. `nwcConnectionStringInvalid` above is deliberately a different predicate (it skips the empty case so an untouched field is not red), so you cannot reuse it directly, but hoisting the shared call reads better and makes the two predicates obviously related:

```ts
const nwcConnectionStringValid =
    ValidationUtils.isValidNostrWalletConnectUrl(nwcConnectionString);
const nwcConnectionStringInvalid =
    !!nwcConnectionString && !nwcConnectionStringValid;
...
disabled={loading || !nwcConnectionStringValid}
```

Unrelated and pre-existing, but you are the first person to touch this button in a while: neither `testNWCConnectionString(...).then(...)` nor `createNWC(...).then(...)` has a `.catch()`, and both store methods rethrow after setting `error_msg` (`LightningAddressStore.ts:502` and `:542`). A failed test or create is an unhandled rejection today. #4485 is cleaning up the same pattern in the other `LightningAddressStore` consumers but does not cover this file. Out of scope here, just noting it so it does not get lost.

utils/ValidationUtils.ts
kaloudis: Resolved in `acb77983`. The scheme-swap parse drops the trailing slash the same way the SDK does, so `nostr+walletconnect://<pubkey>/?relay=...&secret=...` now validates and existing wallets saved in that form are no longer locked out of nickname and photo edits. Verified against the SDK.

utils/ValidationUtils.ts
kaloudis: Resolved in `acb77983`. `NWC_SECRET_HEX` is lowercase-only now and matches the SDK's `NWC_HEX64` exactly, so the save-then-throw path is closed for the secret. The same narrowing is still needed for the pubkey, for a different reason: see my new comment on the `parsed.host` line, React Native's URL polyfill does not case-fold hosts the way Node does.

utils/ValidationUtils.ts
kaloudis: Resolved in `c6463182`. Each relay goes through `new URL` with a `ws:`/`wss:` protocol check, which is stricter than the SDK's bare `new URL`. Confirmed both `wss://a b` and `wss://a+b.io` are rejected now, and every positive vector including the percent-encoded and literal-`%` relays still passes. The `ws://` question is answered: accepted for local and regtest, documented in the code comment and the PR body, and pinned by a test. Good.

utils/ValidationUtils.ts
kaloudis: The dead `.replace(/\\s+/g, ' ')` is gone from the validator, thanks. The validate-normalised-but-save-raw wrinkle is still there, but I checked it and both `URL` implementations strip leading and trailing spaces, so it is cosmetic. Leaving it to you.

views/Settings/WalletConfiguration.tsx
kaloudis: Resolved. `onChangeText` reads from `text` now and there is no local `nostrWalletConnectUrl` to shadow the outer one.

views/LightningAddress/CreateNWCLightningAddress.tsx
ajaysehwal: Fixed in 8c7d469 — hoisted a single validity check: 

``` js 
const nwcConnectionStringValid =
    ValidationUtils.isValidNostrWalletConnectUrl(nwcConnectionString);
const nwcConnectionStringInvalid =
    !!nwcConnectionString && !nwcConnectionStringValid;
```

Button uses disabled={loading || !nwcConnectionStringValid};


error color still only when non-empty and invalid.Thanks for the note on the missing .catch() on test/create — agreed that’s out of scope here; will track with #4485 / a small follow-up so it doesn’t get lost.

views/LightningAddress/CreateNWCLightningAddress.tsx
ajaysehwal: Good call on both. Fixed in 8c7d469: removed the unreachable tested: false (input only mounts when !tested). Dropped the internal whitespace collapse there and left trim() only — as you noted, collapsing spaces doesn’t make an invalid NWC URI valid. WalletConfiguration still uses the older trim + collapse for consistency with nearby fields; happy to align those in a follow-up if useful

utils/ValidationUtils.test.ts
ajaysehwal: Agreed — that test was pinning Jest/Node behavior, not device behavior. 

Fixed in 8c7d4692955b53a811316b8f45ee355f4381322c: flipped to rejects uppercase hex pubkeys with a short comment on the polyfill. Also kept rejecting uppercase hex secrets. Suite is green with the raw-host check.

utils/ValidationUtils.ts
ajaysehwal: Thanks for catching this — the Node vs RN polyfill host-folding difference was easy to miss in Jest-only runs.

Fixed in 8c7d469: single NWC_HEX64 = /^[0-9a-f]{64}$/ for both pubkey and secret, and the pubkey is checked on the raw host segment (before URL parse) so uppercase is rejected under both Node and the app polyfill. Matches @getalby/sdk and closes the save-then-throw path on device.

kaloudis (approved):

PR #4176, opened 2026-06-16
fix: empty fiat inputs on Currency Converter initial render

ajaysehwal:
# Description

This PR fixes two Currency Converter issues and improves the shared TextInput layout used by the converter.

1. Empty fiat fields on initial render
Fiat fields (e.g. USD) could appear empty on first open, even when BTC and sats already had values.

Previously, fiat conversion only ran inside handleInputChange during user input. BTC ↔ sats conversion worked immediately because it does not depend on exchange rates, but fiat conversion requires fiatRates from FiatStore. If rates were not yet loaded when the screen opened—or when values were restored from storage—fiat fields stayed blank and were never backfilled.

2. Long amounts overlapping currency labels
Very long BTC/sats values could render behind the BTC label and Bitcoin icon because the amount, label, and icon were absolutely positioned on top of each other.  Here is a screenshot : 

Screenshot 2026-06-17 at 12 46 45 PM

## Changes

components/CurrencyConverterContent.tsx

- Fetch exchange rates on screen entry only when rates are not already available
- Add syncConversions() to reuse existing conversion logic and backfill fiat values once rates load or stored values are restored
- Trigger sync after loading saved values, initialising default currencies, and adding a new currency
- Show the loading indicator only when rates are missing, avoiding unnecessary loading UI when values are already populated
- Move currency code and icon into a TextInput trailing slot so long amounts no longer overlap labels 

components/TextInput.tsx

- Add optional trailing prop for fixed right-side content (label, icon, etc.)
- Lay out input, suffix, and trailing in a flex row (flex: 1 + minWidth: 0 on input)
- Render suffix in normal flex flow instead of absolute positioning 

After Changes Screen Recording : 

https://github.com/user-attachments/assets/e236a7a2-202d-4c1c-bf69-dbe892d579f4

 

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/CurrencyConverterContent.tsx
+++ b/components/CurrencyConverterContent.tsx
@@ -73,6 +73,7 @@ export default class CurrencyConverterContent extends React.Component<
     CurrencyConverterContentProps,
     CurrencyConverterContentState
 > {
+    private blurListener: (() => void) | undefined;
     constructor(props: CurrencyConverterContentProps) {
         super(props);
         this.state = {
@@ -92,6 +93,16 @@ export default class CurrencyConverterContent extends React.Component<
             this.props.FiatStore?.getFiatRates?.();
         }
         this.addDefaultCurrenciesToStorage();
+
+        this.blurListener = this.props.navigation?.addListener?.(
+            'blur',
+            this.saveInputValues
+        );
+    }
+
+    componentWillUnmount() {
+        this.blurListener?.();
+        this.saveInputValues();
     }
 
     componentDidUpdate(
@@ -232,7 +243,10 @@ export default class CurrencyConverterContent extends React.Component<
                 }
             });
 
-            this.setState({ inputValues: convertedValues });
+            this.setState(
+                { inputValues: convertedValues },
+                this.saveInputValues
+            );
             return;
         }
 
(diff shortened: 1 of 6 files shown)


shubhamkmr04 (commented):
https://github.com/user-attachments/assets/5145f4e5-bf63-4553-af7c-67d197e2d71a

we try to change the input values and go back to other view, but it doesn't save the input in this case. The values gets saved as soon as we add a currency, or delete one.

shubhamkmr04 (commented):
@ajaysehwal `right` is unused in Textinput component but we are still passing it in some places.

shubhamkmr04 (approved):
tACK

kaloudis (approved):

PR #4186, opened 2026-06-19
fix(tor): honor certVerification for LND/CLN REST over Tor (#4178)

### LINKED ISSUE #4178: v13.1.0 Attempts to validate certificate over Tor
### Describe the bug

After upgrading to v13.1.0, when connecting (via LND REST) to my node over Tor, it looks like Zeus is now attempting to validate the SSL certificate which subsequently fails due to it being self-signed.

Previous behavior on v13.0.2, it never attempted to validate the SSL certificate due to the connection being over Tor.

Since there is no option to disable SSL certificate verification when selecting the Tor option, it is now no longer possible to connect to my node over Tor on this latest version. After downgrading back to v13.0.2, I am able to connect again fine.

### Reproduce

This issue can be reproduced by connecting to a node via LND (REST) over Tor. See screenshot for image of error:

Image

### ZEUS version

v13.1.0

### Node interface

LND (REST)

### Network

Tor

### Device

Google Pixel 6

### Device operating system

GrapheneOS

### Log output

```shell

```

kaloudis:
# Description

Relates to issue: **ZEUS-4178**

Before v13.1.0 the Tor path always trusted self-signed certificates because the old `react-native-tor` library defaulted `trustSSL=true`. `react-native-nitro-tor` 0.5.3 had no such knob — TLS validation was strict, so users connecting to their own LND/CLN node over a hidden service (default LND REST + self-signed cert setup) regressed on upgrade.

This PR:

- Plumbs a `trust_invalid_certs` field through `doTorRequest` and hardcodes it to `true` on the Tor branch in `backends/LND.ts` and `backends/CLNRest.ts`.
- The `certVerification` setting is intentionally ignored over Tor: the `.onion` address authenticates the endpoint at the Tor protocol layer, and the upstream daemon's self-signed cert can never match an `.onion` hostname, so TLS validation here is a category error. The wallet config UI already hides the `certVerification` toggle when Tor is enabled, so this matches the user-visible model.
- The other `doTorRequest` callers (`utils/handleAnything.ts`, `stores/SettingsStore.ts`, `components/LayerBalances/LightningSwipeableRow.tsx`) keep the strict default — they talk to public third-party services (lnurl resolution, Olympus, Zeus servers), not user-owned nodes.

### Upstream changes

Schema/Rust support landed upstream in:
- niteshbalusu11/tor-rust-sdk#4
- niteshbalusu11/react-native-nitro-tor#10

This PR picks up the matching `react-native-nitro-tor` **0.6.0** release.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

- [x] Contributors will need to run \`yarn\` after this PR is merged in
    * verify that \`package.json\` and \`yarn.lock\` have been properly updated

### Other:

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -70,7 +70,12 @@ export default class CLNRest {
                     url,
                     method as RequestMethod,
                     JSON.stringify(data),
-                    headers
+                    headers,
+                    // Tor v3 .onion authenticates the endpoint at the
+                    // protocol layer, and the upstream daemon's self-signed
+                    // cert can't match an .onion hostname, so TLS validation
+                    // here is always redundant.
+                    true
                 ).then((response: any) => {
                     calls.delete(id);
                     return response;
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -15,7 +15,11 @@ import {
 } from './CoreLightningRequestHandler';
 import { localeString } from '../utils/LocaleUtils';
 import ReactNativeBlobUtil from 'react-native-blob-util';
-import { doTorRequest, RequestMethod } from '../utils/TorUtils';
+import {
+    doTorRequest,
+    isOnionHttpsUrl,
+    RequestMethod
+} from '../utils/TorUtils';
 
 const calls = new Map<string, Promise<any>>();
 
@@ -71,11 +75,12 @@ export default class CLNRest {
                     method as RequestMethod,
                     JSON.stringify(data),
                     headers,
-                    // Tor v3 .onion authenticates the endpoint at the
-                    // protocol layer, and the upstream daemon's self-signed
-                    // cert can't match an .onion hostname, so TLS validation
-                    // here is always redundant.
-                    true
+                    // .onion-over-Tor: bypass TLS validation (the .onion
+                    // address authenticates at the Tor layer, upstream
+                    // self-signed certs can't match anyway).
+                    // Clearnet-over-Tor: keep strict TLS because exit
+                    // nodes can MITM.
+                    isOnionHttpsUrl(url)
                 ).then((response: any) => {
                     calls.delete(id);
                     return response;
--- a/backends/LND.ts
+++ b/backends/LND.ts
@@ -1,6 +1,10 @@
 import ReactNativeBlobUtil from 'react-native-blob-util';
 import { settingsStore, nodeInfoStore } from '../stores/Stores';
-import { doTorRequest, RequestMethod } from '../utils/TorUtils';
+import {
+    doTorRequest,
+    isOnionHttpsUrl,
+    RequestMethod
+} from '../utils/TorUtils';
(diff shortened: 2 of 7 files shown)


kaloudis:
Runtime-tested end-to-end on iOS simulator + Android against a Tor v3 hidden service fronting an LND REST endpoint with a self-signed cert. Connection succeeds with the fix applied; reverting to 0.5.3 reproduces the original TLS-rejection regression.

The vendor pin (`a76fca2`) is the per-target LTO build:
- iOS slices: `lto = true` — matches the relocation form 0.5.3 ships (PAGE21/PAGOF12) so Xcode 26.5's `ld` accepts them. With LTO off/thin, rustc emits a synthetic type-15 reloc that Apple's linker rejects with `findRealAtom` / `cfiStartsArray` / `symbol index out of range` assertions.
- Android slices: `CARGO_PROFILE_RELEASE_LTO=false` override at build invocation — otherwise `cc-rs` bakes LLVM bitcode into `libz-sys`'s `adler32.o` and the NDK's `llvm-strip` rejects it.

Both upstream PRs are now also ready for review:
- niteshbalusu11/tor-rust-sdk#4
- niteshbalusu11/react-native-nitro-tor#10

Once those merge and Nitesh publishes 0.5.4, the `package.json` change here collapses to `"react-native-nitro-tor": "0.5.4"`.

backends/CLNRest.ts
```diff
@@ -70,7 +70,12 @@ export default class CLNRest {
                     url,
                     method as RequestMethod,
                     JSON.stringify(data),
-                    headers
+                    headers,
+                    // Tor v3 .onion authenticates the endpoint at the
+                    // protocol layer, and the upstream daemon's self-signed
+                    // cert can't match an .onion hostname, so TLS validation
+                    // here is always redundant.
+                    true
```
ajaysehwal: @kaloudis, this currently sets `trustInvalidCerts=true` for all Tor-routed requests. Could we scope it to `.onion` HTTPS endpoints only? Otherwise, a clearnet LND REST endpoint accessed via Tor would have TLS verification disabled, potentially exposing users to MITM attacks.

For example:

```
const trustInvalidCerts =
  typeof url === 'string' &&
  url.includes('.onion') &&
  url.startsWith('https://');

doTorRequest(
  url,
  method as RequestMethod,
  JSON.stringify(data),
  headers,
  trustInvalidCerts
);
```

ajaysehwal (approved):
tACK

PR #4195, opened 2026-06-24
refactor: Receive: deduplicate componentDidMount/render locals

### LINKED ISSUE #4193: refactor: deduplicate componentDidMount/render locals in Receive.tsx
## Context

Surfaced during review of #4192. Gemini flagged that the `lnOnly` definition is duplicated between `componentDidMount` and `render` in `views/Receive.tsx` (lines 321-329 and 1292-1300). The duplication is real but predates that PR, so we kept the bug fix minimal and are tracking the cleanup here.

The `lnOnly` block is not the only duplicate. The same pattern repeats for several adjacent locals.

## Duplicated locals between `componentDidMount` and `render` in `views/Receive.tsx`

- `lnOnly` (componentDidMount L321-329, render L1292-1300)
- `onChainOnly` (L329, L1300) — `route.params?.forceOnChain`
- `skipOnchain` (L554-556, L682-684, L1287-1289 — also recomputed in `getInvoice` / `autoGenerateInvoice` paths)
- Address-type fallback: `route.params?.addressType || settings?.invoices?.addressType || '0'` appears multiple times
- The initial `route.params` destructure (`lnurlParams`, `amount`, `autoGenerate*`, `account`, `selectedIndex`, etc.)

## Proposal

A naive `private get isLnOnly()` style getter (suggested by Gemini) handles `lnOnly` in isolation but doesn't address the broader pattern, and it hides the semantic difference between the two call sites: `componentDidMount` reads `route.params` once at mount to seed state, while `render` reads them every frame. A future change could legitimately want them to diverge.

Worth considering as part of the cleanup:
- A single private helper (e.g. `getReceiveModeFlags(props)` or similar) that derives `lnOnly`, `onChainOnly`, `skipOnchain`, and the address-type fallback from `props`/`route.params`/`SettingsStore`, and is called from both sites.
- Keeping `componentDidMount`'s mount-time snapshot semantics explicit (e.g. by storing the derived flags in state on mount, where appropriate) rather than papering over the difference with a getter.

No behavior change intended — purely a maintainability cleanup.

kaloudis:
# Description

Relates to issue: [**ZEUS-4193**](https://github.com/ZeusLN/zeus/issues/4193)

Deduplicates several locals that were redefined identically (or near-identically) across `componentDidMount`, `render`, `autoGenerateInvoice`, and `validateAddress` in `views/Receive.tsx`. Surfaced during review of #4192 (Gemini flagged the `lnOnly` duplicate, which predates that PR).

Three small refactors, one commit each, all no-behavior-change except where called out:

1. **`getReceiveModeFlags()`** — returns `{ lnOnly, onChainOnly }`. Replaces two copies (componentDidMount + render).
2. **`getSkipOnchain()`** — replaces three copies of `settings?.invoices?.defaultInvoiceType !== DefaultInvoiceType.Unified` (in `autoGenerateInvoice`, `validateAddress`, and `render`).
3. **`getAddressType()`** — returns `route.params?.addressType || settings?.invoices?.addressType || '0'`. Used at both the initial `setState` in `componentDidMount` and the per-action local that seeds `autoGenerate*` calls.

### Behavior change in (3)

Previously the `route.params?.addressType` override only applied to the immediate auto-generation in `componentDidMount`; the initial `setState` seeded `state.addressType` from settings alone, so the picker and any subsequent re-generations silently snapped back to the settings default. Both sites now share `getAddressType()`, so navigating to Receive from `OnChainAddresses` with an `addressType` in params is now reflected in state as well. This looks like a missed update when the route override was originally added (the `autoGenerate*` callers already accounted for it; only the persisted state didn't).

No behavior change in (1) or (2).

### Address-type translation fix (LND backends)

Reviewing (3) surfaced a separate pre-existing bug: the address-type picker was silently generating native segwit on every LND backend regardless of what the user picked, and the `OnChainAddresses` → `Receive` flow for `NESTED_WITNESS_PUBKEY_HASH` / `HYBRID_NESTED_WITNESS_PUBKEY_HASH` accounts produced the wrong type or errored outright.

Three different forms reach the backend — numeric strings from the picker / settings (`'0'` / `'1'` / `'4'`), lnrpc enum names, and walletrpc enum names returned by `ListAccounts` — and each transport wants something different:

- **LND REST** — grpc-gateway parses the `AddressType` query param by name. `?type=1` is silently treated as `WITNESS_PUBKEY_HASH`.
- **LNC** — same; the LNC client expects the lnrpc enum name.
- **Embedded LND** — protobufjs writes the enum via `writer.int32(...)`, which numeric-coerces string enum names to `NaN` and serializes them as `0`. Needs the numeric form.

Added `utils/LndUtils.ts` (with unit tests) exposing `toLnrpcAddressType` (name form) and `toLnrpcAddressTypeNum` (numeric form). `LND`, `LightningNodeConnect`, and `EmbeddedLND` `getNewAddress` calls now route through whichever helper their transport needs, and `HYBRID_NESTED_WITNESS_PUBKEY_HASH` is mapped to `NESTED_PUBKEY_HASH` (lnrpc has no hybrid variant; the resulting nested-segwit p2sh-p2wkh is what the receiver sees).

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [x] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/backends/EmbeddedLND.ts
+++ b/backends/EmbeddedLND.ts
@@ -80,6 +80,8 @@ import {
     verifyMessageWithAddr as verifyMsgWithAddr
 } from '../lndmobile/wallet';
 
+import { toLnrpcAddressTypeNum } from '../utils/LndUtils';
+
 export default class EmbeddedLND extends LND {
     openChannelListener: any;
 
@@ -123,7 +125,10 @@ export default class EmbeddedLND extends LND {
         reversed?: boolean;
     }) => await listPayments(params);
     getNewAddress = async (data: any) =>
-        await newAddress(data.type, data.account);
+        await newAddress(
+            toLnrpcAddressTypeNum(data?.type) as any,
+            data?.account
+        );
     getNewChangeAddress = async (data: any) =>
         await newChangeAddress(data.type, data.account);
     openChannelSync = async (data: OpenChannelRequest) =>
--- a/backends/LND.ts
+++ b/backends/LND.ts
@@ -9,6 +9,7 @@ import OpenChannelRequest from './../models/OpenChannelRequest';
 import Base64Utils from './../utils/Base64Utils';
 import VersionUtils from './../utils/VersionUtils';
 import { localeString } from './../utils/LocaleUtils';
+import { toLnrpcAddressType } from './../utils/LndUtils';
 import { Hash as sha256Hash } from 'fast-sha256';
 import BigNumber from 'bignumber.js';
 
@@ -371,7 +372,13 @@ export default class LND {
             }`
         );
 
-    getNewAddress = (data: any) => this.getRequest('/v1/newaddress', data);
+    getNewAddress = (data: any) => {
+        const params: any = { ...data };
+        const type = toLnrpcAddressType(params.type);
+        if (type !== undefined) params.type = type;
+        else delete params.type;
+        return this.getRequest('/v1/newaddress', params);
+    };
     getNewChangeAddress = (data: any) =>
         this.postRequest('/v2/wallet/address/next', data);
     openChannelSync = (data: OpenChannelRequest) => {
--- a/backends/LightningNodeConnect.ts
+++ b/backends/LightningNodeConnect.ts
@@ -10,19 +10,11 @@ import OpenChannelRequest from '../models/OpenChannelRequest';
 
 import Base64Utils from '../utils/Base64Utils';
 import { snakeize } from '../utils/DataFormatUtils';
+import { toLnrpcAddressType } from '../utils/LndUtils';
 import VersionUtils from '../utils/VersionUtils';
 
 import { Hash as sha256Hash } from 'fast-sha256';
 import BigNumber from 'bignumber.js';
-
-const ADDRESS_TYPES = [
-    'WITNESS_PUBKEY_HASH',
-    'NESTED_PUBKEY_HASH',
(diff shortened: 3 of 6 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (commented):
https://github.com/user-attachments/assets/f832b155-7ba0-4eea-86f3-a8e8ad200f73

no matter what address type we select from this view. native segwit is generate everytime

kaloudis:
@ajaysehwal @shubhamkmr04 big fixes included in https://github.com/ZeusLN/zeus/pull/4195/changes/d48a23b18a7d543bdbb926d3cee8888d1f4c30e4

Please test all three LND backends for address generation across both views. For more details see `Address-type translation fix (LND backends)` in the PR description.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4226, opened 2026-07-13
fix: deep links handled twice with 'Select wallet on start-up'

### LINKED ISSUE #4218: Deep linking: invoices are sometimes loaded twice when 'Choose wallet on start-up' is enabled

kaloudis:
# Description

Relates to issue: #4218

## Root cause

The "handle the launch deep link exactly once" invariant lived in per-instance Wallet component state (`initialLoad`), while `Linking.getInitialURL()` is global and returns the same URL on every call — and on Android, `MainActivity.onNewIntent` calls `setIntent()`, so it returns the most recent warm deep link forever after. With **Select wallet on start-up** disabled, the Wallet view mounts exactly once, so the invariant held by accident. With it enabled, the Wallet view is replaced by the wallet picker (`navigation.replace('Wallets')`) and later re-created (`popTo('Wallet')` pushes a fresh instance), so multiple instances could each consume the same initial URL:

1. A late focus/AppState trigger on the replaced instance slipped past the startup branch (`initialStart` had already been set false), fell through to `fetchData`, silently connected to the *previously selected* wallet, and opened the invoice over the wallet picker (load #1).
2. After the user picked a wallet, the fresh Wallet instance ran `fetchData` with its own `initialLoad = true` and consumed `getInitialURL()` again (load #2).

Whether both loads fired depended on event timing — hence *sometimes*.

## Fix

- **`utils/LinkingUtils.ts`**: track the last handled URL in the singleton and skip it in `handleInitialUrl`, so the initial URL is processed at most once no matter how many Wallet instances mount. Also covers the Android `setIntent` re-read of old warm links on later fresh mounts. Deliberate re-taps of the same link still work (the `'url'` event path bypasses the check), and links deferred behind the lockscreen are only marked handled once actually processed.
- **`views/Wallet/Wallet.tsx`**: gate `getSettingsAndNavigate` once the instance has replaced itself with the startup wallet picker, so a stale instance can't connect to the old wallet or consume the deep link; store the `Linking` `'url'` subscription, de-dupe registration, and remove it on unmount (it was never cleaned up, so an unmounted instance could keep handling warm deep links alongside the live one).
- Follow-up commit removes the now-vestigial `selectNodeOnStartup && initialStart` clause in `fetchData` — it was self-defeating (any connecting run set `initialStart` false ~30 lines earlier), and the invariant it aimed for is now enforced by the two mechanisms above.

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [x] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/LinkingUtils.ts
+++ b/utils/LinkingUtils.ts
@@ -9,6 +9,13 @@ import { settingsStore } from '../stores/Stores';
 class LinkingUtils {
     private shareIntentProcessed = false;
     private pendingDeepLink: string | null = null;
+    // Linking.getInitialURL() returns the same URL for the lifetime of the
+    // activity (and on Android, onNewIntent + setIntent refreshes it with
+    // each warm deep link), while the Wallet view — which consumes it — can
+    // mount more than once, e.g. when 'Select wallet on start-up' is
+    // enabled. Track the last URL actually processed so it is only ever
+    // handled once.
+    private lastHandledUrl: string | null = null;
 
     processPendingDeepLink = (
         navigation: NativeStackNavigationProp<any, any>
@@ -53,7 +60,9 @@ class LinkingUtils {
             }
 
             if (url) {
-                this.handleDeepLink(url, navigation);
+                if (url !== this.lastHandledUrl) {
+                    this.handleDeepLink(url, navigation);
+                }
                 return;
             }
             if (Platform.OS === 'android') {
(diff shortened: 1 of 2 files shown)


ajaysehwal (approved):
tACK

PR #4227, opened 2026-07-13
Privacy: configurable mempool instance

### LINKED ISSUE #3994: Allow configuring a custom Mempool.space instance for fee suggestions
### Describe your enhancement idea

In `Settings > Privacy`, we can only use the official `Mempool.space` instance for fetching fee suggestions. This is a privacy leak.

Allow selecting a custom Mempool.space instance (eg a self-hosted one) to fetch fee suggestions, in the same manner as the custom block explorer.

Depending on which fee endpoint is used here, this could be generalized to any Esplora-compatible API endpoint.

I assume the Mempool.space specific `/api/v1/fees/recommended` endpoint is used, since Zeus shows it as "recommended fees". It would be interesting to rework this to use the `/api/fee-estimates` endpoint, which is implementation-agnostic and works with all Esplora server implementations (`Mempool.space`, Blockstream's Esplora, etc.).

kaloudis:
# Description

Relates to issue: **#3994**

Allows selecting the mempool instance used for on-chain fee suggestions, raw transaction broadcasts, embedded LND sync progress, and WIF sweeps. A new dropdown always appears in `Settings > Privacy` (not gated behind the fee-suggestions toggle, since broadcasts, sweeps, and sync use the instance regardless):

- **electrs.zeusln.com** (new default) — ZEUS's own instance, now serving mempool API endpoints
- **mempool.space**
- **Custom** — reveals a text input for a self-hosted instance (addresses the privacy leak raised in #3994)

Behavior details:

- `electrs.zeusln.com` is **mainnet-only**. On testnet, the presets fall back to mempool.space's testnet3 instance (`/testnet` path). A custom instance is used verbatim on every network, so self-hosted testnet instances work.
- Custom input accepts URLs with or without a scheme (`https://` is prepended if missing); trailing slashes are stripped. An empty custom field gracefully falls back to the default preset.
- All URL resolution is centralized in `UrlUtils.getMempoolApiUrl`. Everything that consumed mempool.space data now routes through it:
  - **Tx broadcasts** — `broadcastRawTxToMempoolSpace` and the `publishTransaction` fallback (already used the helper, no changes needed)
  - **Fee suggestions** — `FeeStore.getOnchainFeesviaMempool`; this also fixes a pre-existing bug where fee suggestions on mutinynet incorrectly hit mainnet mempool.space
  - **Embedded LND sync** — `SyncStore.getBestBlockHeight` (`/blocks/tip/height`); testnet still derived from `embeddedLndNetwork` since nodeInfo isn't populated during sync
  - **WIF sweeps** — `WIFUtils.baseUrl` removed; `SweepStore`'s Esplora calls (`/address/{addr}/utxo`, `/tx/{txid}`, `/tx/{txid}/hex`) use the helper, making sweeps mutinynet-aware as well
- The broadcast button on the Raw TX Hex view now shows the effective host dynamically ("Broadcast to electrs.zeusln.com" / "Broadcast to mempool.space" / custom host) via a new `views.RawTxHex.broadcastToInstance` string with `{{instance}}` substitution.
- The "Powered by mempool.space" logo on the Edit network fee view only renders when the effective instance host is mempool.space (including the testnet fallback) — hidden for electrs.zeusln.com, mutinynet, and custom instances.
- Custom URLs that already end in `/api` are handled: the suffix is stripped before the helper appends `/api`, so pasting a full API base doesn't produce `/api/api`.
- No settings migration needed: existing users' privacy blobs simply lack the new key and resolve to the new default. **Note that this means existing installs switch from the previously hardcoded mempool.space to electrs.zeusln.com on upgrade.** This is intentional: the point of the new default is to stop sending users' on-chain data to a third party out of the box. Anyone who prefers mempool.space can select it in `Settings > Privacy`.

Verified live against `electrs.zeusln.com`: `/api/v1/fees/recommended`, `/api/tx`, `/api/blocks/tip/height`, and `/api/tx/{txid}/hex` all respond.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


kaloudis:
Added 79632ef84: the 'Powered by mempool.space' logo on the Edit network fee view now only renders when the effective instance host is mempool.space (including the testnet fallback) — hidden for electrs.zeusln.com, mutinynet, and custom instances.

kaloudis:
Added fb569ef32: SyncStore.getBestBlockHeight (embedded LND sync progress) now also routes through UrlUtils.getMempoolApiUrl, so it respects the selected instance. Verified electrs.zeusln.com serves /api/blocks/tip/height. Testnet is still derived from embeddedLndNetwork since nodeInfo isn't populated during sync.

kaloudis:
Added 1bd1bf3c5: WIF sweeps now respect the selected mempool instance. Removed WIFUtils.baseUrl entirely — SweepStore's Esplora calls (/address/{addr}/utxo, /tx/{txid}, /tx/{txid}/hex) go through UrlUtils.getMempoolApiUrl instead, which also makes sweeps mutinynet-aware. Verified electrs.zeusln.com serves the tx-hex endpoint. Remaining hardcoded mempool.space reference is only ConnectivityStore's reachability ping host (no data fetched).

ajaysehwal (commented):
Image

@kaloudis, I think we should not hide the mempool instance field behind that toggle, because broadcast/sweep/sync use the instance even when fee suggestions are off

kaloudis:
|A|B|
|-|-|
|Simulator Screenshot - iPhone 17 Pro - 2026-07-15 at 12 09 50|Simulator Screenshot - iPhone 17 Pro - 2026-07-15 at 12 09 54|

@ajaysehwal

ajaysehwal (commented):
Image

@kaloudis, I think we should validate the custom URL for both the Block Explorer and Mempool Instance fields.

ajaysehwal (commented):
Image

@kaloudis, we still able to add an invalid host url

ajaysehwal:
https://github.com/user-attachments/assets/58095b94-530e-41d8-938d-af8090886ecf

@kaloudis, does it make sense to keep the selected custom field even if the user enters an invalid URL and then goes back?

kaloudis:
@ajaysehwal good question. Pushed f3113ab25 to resolve it: if you leave the Privacy view with Custom selected but no URL saved (because the field was empty or never passed validation), the dropdown now reverts to the default preset on unmount. This applies to both the block explorer and the mempool instance.

Invalid text was already never persisted, so the only lingering state was the dropdown itself claiming Custom while every consumer silently fell back to the default. That mismatch is gone now.

Note: I also squashed the two duplicate validation commits into one (45566ac8c), so the branch was force-pushed.

utils/UrlUtils.ts
```diff
@@ -9,9 +21,38 @@ const getMempoolApiUrl = (nodeInfo: {
     isMutinynet: boolean;
     isTestNet: boolean;
 }): string => {
+    const privacy = settingsStore?.settings?.privacy;
+    const instance = privacy?.mempoolInstance || DEFAULT_MEMPOOL_INSTANCE;
```
ajaysehwal, line 29: @kaloudis, I think existing users will silently switch their Esplora host.

Stored settings don't have `privacy.mempoolInstance`, and `getSettings()` doesn't deep-merge the new default. So consumers fall back to `electrs.zeusln.com`, whereas before the default was `mempool.space`.

If this is intentional, I think we should explicitly mention in the PR that existing users will get the new default on upgrade, rather than keeping `mempool.space` until they manually change it.

utils/UrlUtils.ts
```diff
@@ -9,9 +21,38 @@ const getMempoolApiUrl = (nodeInfo: {
...
+
+    // Custom instance is used verbatim on every network
+    if (instance === 'Custom' && privacy?.customMempoolInstance) {
+        const host = withScheme(
+            privacy.customMempoolInstance.trim().replace(/\/+$/, '')
+        );
+        return `${host}/api`;
```
ajaysehwal, line 41: @kaloudis One small edge case with custom Esplora URLs:

`getMempoolApiUrl()` always appends `/api`, so if a user provides a URL that already ends with `/api`, it becomes `/api/api`.

For example:

`https://mempool.example.com/api` → `https://mempool.example.com/api/api`

This could break fees, broadcast, and WIF sweep. I think we should either strip the trailing `/api` or clearly document that the custom URL should not include it.

stores/SyncStore.ts
```diff
@@ -117,11 +118,11 @@ export default class SyncStore {
         await new Promise((resolve, reject) => {
             ReactNativeBlobUtil.fetch(
                 'get',
-                `https://mempool.space/${
-                    this.settingsStore.embeddedLndNetwork === 'Testnet'
-                        ? 'testnet/'
-                        : ''
-                }api/blocks/tip/height`
+                `${UrlUtils.getMempoolApiUrl({
+                    isMutinynet: false,
```
ajaysehwal, line 128: @kaloudis I think there’s still an issue with `SyncStore.getBestBlockHeight`.

It hardcodes `isMutinynet: false` and checks `embeddedLndNetwork === 'Testnet'`, so a Mutinynet wallet can still query the mainnet Esplora endpoint for its tip height and show incorrect sync status.

Since this call was changed in this PR, I think we should pass the actual `nodeInfo` / network here, or normalize the network value (`toLowerCase()`) and handle Mutinynet explicitly.

utils/UrlUtils.ts
kaloudis: Yes, this is intentional: the point of the new default is that existing installs also move off the hardcoded mempool.space on upgrade, without a migration writing anything to their settings. Documented it at the fallback in e00042307 and made it explicit in the PR description. Anyone who prefers mempool.space can still select it in Settings > Privacy.

utils/UrlUtils.ts
kaloudis: Good catch. Fixed in 45f403b28: a trailing /api (case-insensitive) is now stripped from the stored custom URL before the helper appends /api, so pasting the full API base works. This also covers already-saved values, and there's new unit test coverage for both scheme and scheme-less variants.

stores/SyncStore.ts
kaloudis: This can't occur in practice: SyncStore only ever runs for embedded LND (it's started from LndMobileUtils and the wallet-just-created branch in Wallet.tsx), and embedded LND is restricted to Mainnet/Testnet. Mutinynet is LDK Node-only via the separate ldkNetwork field, and its option is filtered out of the embedded LND network dropdown in WalletConfiguration; LDK Node doesn't use SyncStore at all. Casing is consistent too: every path that persists embeddedLndNetwork stores capitalized values (WalletCreationUtils uses 'Mainnet', SeedRecovery uppercases before saving), matching the pre-existing 'Testnet' check here. And as noted earlier in the thread, nodeInfo isn't populated during initial sync, which is why the network comes from settings. Recorded these invariants in a comment at the call site in dd4e79be9.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

Issue #4229, opened 2026-07-15
Open Channel List is empty after closing a channel


brianoflondon:
### Describe the bug

I'm on Version 13.1.2 (1154) on Mac 26.5.2 (25F84)

After closing a channel the Channel list for Open channels is empty. I've seen this three times now one after another. Quit and restart brings back the channel list.

### Reproduce

Co-op close a channel.

I happen to be forwarding the close amount to an off-node btc address not sure if that is important.

### ZEUS version

v13.1.2

### Node interface

LND (REST)

### Network

Clearnet

### Device

M1 Macbook Pro

### Device operating system

Mac OS 26.5.2 (25F84)

### Log output

```shell

```

kaloudis:
Thanks for the report — root cause tracked down, and a draft fix is up in #4230 (full analysis in the PR body).

TL;DR: on LND (REST), the close-channel call is a streaming endpoint that stays open until the closing tx confirms. ZEUS force-resolves the close after 6 seconds and repopulates the channel list, but the original request was left pending — and when the app's generic 30-second REST timeout rejected it, the stale error handler wiped the channel list (`channels = []`). That's why the list blanks out shortly after every co-op close and stays empty until a refetch. The delivery to an external address isn't a factor.

Until a fix ships: pull-to-refresh on the channels list (or backgrounding/foregrounding the app) restores the list without a restart.

kaloudis:
This is now fixed in ZEUS v13.1.3 - out now. Thanks for reporting.

PR #4230, opened 2026-07-15
stores: prevent stale closeChannel error from wiping channel list

kaloudis:
# Description

Relates to issue: **#4229**

Fixes the open channel list going blank after closing a channel on LND (REST).

## Root cause analysis

The channel list isn't failing to load — it's being actively wiped ~30 seconds after the close, by a stale error handler from the close call itself:

1. **LND's REST close-channel is a streaming endpoint.** `DELETE /v1/channels/{txid}/{index}` (`backends/LND.ts`) keeps the HTTP response open until the closing tx confirms on-chain. `ReactNativeBlobUtil.fetch` buffers the whole body, so this promise can never resolve during a co-op close — it just hangs.
2. **ZEUS knows this and force-resolves after 6 seconds.** `ChannelsStore.closeChannel` races the REST call against a 6s timer; the timer fires `handleChannelClose()`, the view pops to Wallet, and `getChannels()` repopulates the list. So far so good — the user sees their channels.
3. **But the underlying DELETE promise was never detached.** The `resolved` flag (added in a8f537ab8) only guards the wrong direction — it stops the 6s timer from firing after the real call settles, but doesn't stop the real call's `.then`/`.catch` from firing after the 6s fallback has already won the race.
4. **The generic 30s REST timeout guarantees a late rejection.** `restReq` races every non-Tor request against a 30s `Request timeout` (added in 44c3800968, hardened in e3c1818f71). Since the close stream stays open for at least one block, this rejection fires every single time on LND REST.
5. **The late rejection nukes the list.** `.catch` → `handleChannelCloseError` → `getChannelsError()`, which sets `channels = []` and `error = true`. The MobX reaction cascades that into empty `enrichedChannels`/`filteredChannels`. So ~24 seconds after the user is back on Wallet looking at a healthy list, it silently blanks out.
6. **Nothing refreshes it.** Channels only refetch on Wallet focus, app-state change, or pull-to-refresh — sitting on the Wallet screen, the wipe persists, which reads as "empty until restart".

The reporter's `delivery_address` detail is a red herring — any co-op close over LND (REST) reproduces this. It's effectively a regression from the Nov 2024/Apr 2025 global REST timeout interacting with the March 2024 close-error handler.

## Fix

Two minimal changes in `stores/ChannelsStore.ts`:

- **Detach late results in `closeChannel`:** the `resolved` flag is now a shared `settled` guard — the 6s forced resolution sets it too, and the real call's `.then`/`.catch` no-op once it's set. The late 30s timeout rejection can no longer mutate the store.
- **Stop wiping the list on close errors:** `handleChannelCloseError` no longer calls `getChannelsError()` — a failed close request doesn't invalidate the channel list. It now only sets `closeChannelErr` (still surfaced on the Channel view, which doesn't pop when it's set) and clears `closingChannel`. This also applies to the LNC `subscribeChannelClose` error path, where keeping the list intact is likewise correct.

## Testing notes

- Draft: not yet manually tested on device. To reproduce/verify on LND (REST): co-op close a channel, return to Wallet, wait 30+ seconds — before this patch the open channels list blanks; after, it stays populated and the channel appears under pending.
- No regression test added: stores currently have no jest coverage and `ChannelsStore` isn't importable under the current jest config without substantial mocking.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/ChannelsStore.ts
+++ b/stores/ChannelsStore.ts
@@ -691,27 +691,35 @@ export default class ChannelsStore {
         if (implementation === 'lightning-node-connect') {
             return BackendUtils.closeChannel(urlParams);
         } else {
-            let resolved = false;
+            let settled = false;
             return await Promise.race([
                 new Promise((resolve) => {
                     BackendUtils.closeChannel(urlParams)
                         .then(() => {
+                            if (settled) return;
+                            settled = true;
                             this.handleChannelClose();
-                            resolved = true;
                             resolve(true);
                         })
                         .catch((error: Error) => {
+                            if (settled) return;
+                            settled = true;
                             this.handleChannelCloseError(error);
-                            resolved = true;
                             resolve(true);
                         });
                 }),
                 // LND REST call tends to time out, so let's put a 6 second
                 // forced resolution, as true errors will typically throw
-                // before that time
+                // before that time. The settled flag ensures the call's
+                // eventual result (e.g. the 30s REST timeout on the
+                // streaming close call) can't fire handlers after the
+                // forced resolution has already succeeded
                 new Promise(async (resolve) => {
                     await new Promise((res) => setTimeout(res, 6000));
-                    if (!resolved) this.handleChannelClose();
+                    if (!settled) {
+                        settled = true;
+                        this.handleChannelClose();
+                    }
                     resolve(true);
                 })
             ]);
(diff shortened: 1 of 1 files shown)


ajaysehwal (approved):
tACK

PR #4233, opened 2026-07-21
Receive: surface Flow LSP proposal rejection as dismissable warning

kaloudis:
## Description

When the Flow LSP rejects a `/proposal` call, `InvoicesStore.createInvoice` cleared `flow_error`, `flow_error_msg`, and `showLspSettings` entirely so the unwrapped invoice could be displayed without a stale error banner. The side effect: the LSP's human-readable rejection message never reached the user — they'd get a plain invoice with no indication the LSP declined to wrap it.

This matters for upcoming events where the LSP temporarily freezes channel opens (e.g. around a reorg-risk block range for a fork): the server sends an explanatory message ("Not accepting channel open requests between blocks X and Y."), and the app silently dropped it.

## Changes

- `stores/LSPStore.ts`: new `flow_warning_msg` observable; cleared in `reset()`, `resetFee()`, and at the start of `getZeroConfInvoice()` so a stale warning never carries into a new invoice attempt.
- `stores/InvoicesStore.ts`: on `/proposal` rejection, move the LSP's message into `flow_warning_msg` before clearing the error state — the unwrapped-invoice fallback behavior is unchanged.
- `views/Receive.tsx`: render `flow_warning_msg` as a dismissable `WarningMessage` above the invoice, distinct from the red error banner.

## Testing

- `tsc --noEmit`, `prettier --check`, and `eslint` pass on the changed files.

### DIFF
--- a/stores/InvoicesStore.ts
+++ b/stores/InvoicesStore.ts
@@ -457,9 +457,12 @@ export default class InvoicesStore {
                         );
                         jit_bolt11 = response as string;
                     } catch (e: any) {
-                        // Clear error state so the unwrapped invoice
-                        // can be displayed without a stale error banner
+                        // Surface the LSP's rejection as a dismissable
+                        // warning, clearing the error state so the
+                        // unwrapped invoice can still be displayed
                         runInAction(() => {
+                            this.lspStore.flow_warning_msg =
+                                this.lspStore.flow_error_msg;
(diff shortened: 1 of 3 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4239, opened 2026-07-21
fix: prevent Security settings UI flash on open

ajaysehwal:
# Description

## 📹 Screen Recordings

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/fe08d2a1-fb26-44c3-aa8c-e2cfa4a33a3c" controls width="400"></video> | <video src="https://github.com/user-attachments/assets/127b5fa8-7f48-4cab-b246-a77080999c61" controls width="400"></video> |

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Settings/Security.tsx
+++ b/views/Settings/Security.tsx
@@ -27,6 +27,85 @@ type SecurityItem =
     | { translateKey: string; screen: string }
     | { translateKey: string; action: string };
 
+// Build list rows from in-memory settings so first paint matches other
+// settings screens (no empty → populated flash after async load).
+const buildSecurityItems = (settings: {
+    passphrase?: string;
+    pin?: string;
+    duressPassphrase?: string;
+    duressPin?: string;
+}): SecurityItem[] => {
+    // Three cases:
+    // 1) If no passphrase or pin is set, allow user to set passphrase or pin
+    // 2) If passphrase is set, allow user to change passphrase or set/change duress passphrase
+    // 3) If pin is set, allow user to change pin, delete pin, set/change duress pin
+    if (!settings.passphrase && !settings.pin) {
+        return [
+            {
+                translateKey: 'views.Settings.SetPassword.title',
+                screen: 'SetPassword'
+            },
+            {
+                translateKey: 'views.Settings.SetPin.title',
+                screen: 'SetPin'
+            }
+        ];
+    }
+    if (settings.passphrase) {
+        return [
+            {
+                translateKey: 'views.Settings.ChangePassword.title',
+                screen: 'SetPassword'
+            },
+            {
+                translateKey: 'views.Settings.SetPassword.deletePassword',
+                action: 'DeletePassword'
+            },
+            {
+                translateKey: settings.duressPassphrase
+                    ? 'views.Settings.ChangeDuressPassword.title'
+                    : 'views.Settings.SetDuressPassword.title',
+                screen: 'SetDuressPassword'
+            },
+            ...(settings.duressPassphrase
+                ? [
+                      {
+                          translateKey:
+                              'views.Settings.SetDuressPassword.deletePassword',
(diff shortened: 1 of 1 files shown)


kaloudis (commented):
Reviewed — the fix itself looks correct and matches the constructor-from-settings pattern used elsewhere. A few notes inline: one pre-existing bug in the touched flow worth folding in, plus small cleanups.

views/Settings/Security.tsx
```diff
@@ -77,83 +156,10 @@ export default class Security extends React.Component<
...
-                });
-            }
-            this.setState({
-                displaySecurityItems: items
-            });
-        }
-
         // If user tried to enable biometrics, but was forced to first set up pin or password,
         // call handleBiometricsSwitchChange again
         if (route.params?.enableBiometrics) {
```
kaloudis, line 167: Pre-existing, but since this function is being reworked: `route.params?.enableBiometrics` is never cleared and `checkSettings` runs on every focus. After enabling biometrics via the SetPin/SetPassword flow, any later re-focus of this screen instance (e.g. backing out of the Lockscreen) fires another OS biometric prompt. Clearing the param after handling it (`navigation.setParams({ enableBiometrics: undefined })`) would fix it.

views/Settings/Security.tsx
```diff
@@ -45,18 +124,18 @@ export default class Security extends React.Component<
 > {
     constructor(props: SecurityProps) {
         super(props);
+        const settings = props.SettingsStore.settings || {};
+        this.state = {
```
kaloudis: This block and the `setState` in `checkSettings` are the same settings → state derivation, differing only in the biometrics source. A shared helper (e.g. `deriveStateFromSettings(settings)`) used in both places would prevent first-paint and refresh from diverging again — the bug class this PR fixes.

views/Settings/Security.tsx
```diff
@@ -27,6 +27,85 @@ type SecurityItem =
     | { translateKey: string; screen: string }
     | { translateKey: string; action: string };
 
+// Build list rows from in-memory settings so first paint matches other
+// settings screens (no empty → populated flash after async load).
+const buildSecurityItems = (settings: {
```
kaloudis: Nit: this could be `Pick<Settings, 'passphrase' | 'pin' | 'duressPassphrase' | 'duressPin'>` (`Settings` is exported from `stores/SettingsStore.ts`) so the shape can't drift from the source of truth. Relatedly, the `|| {}` in the constructor is effectively dead — `SettingsStore.settings` is initialized inline with full defaults and is never undefined.

views/Settings/Security.tsx
```diff
@@ -27,6 +27,85 @@ type SecurityItem =
...
+                      {
+                          translateKey:
+                              'views.Settings.SetDuressPassword.deletePassword',
+                          action: 'DeleteDuressPassword'
+                      } as SecurityItem
+                  ]
+                : [])
+        ];
+    }
+    if (settings.pin) {
```
kaloudis: Nit: control can only reach here when `settings.pin` is truthy (branch 1 handled `!passphrase && !pin`, branch 2 handled `passphrase`), so this guard and the trailing `return []` are unreachable. Also, the passphrase branch uses spread + `as SecurityItem` while this one uses `push` — one idiom for both conditional appends would read better and drop the cast.

kaloudis (approved):

Issue #4036, opened 2026-05-01
bug: NWC `lookup_invoice` does not return `preimage` for paid invoices


ntheile:
### Describe the bug

When ZEUS is used as a Nostr Wallet Connect wallet service, `lookup_invoice` can return a paid/settled invoice without the payment preimage.

The invoice is correctly marked as settled, but the `preimage` field in the NIP-47 transaction response is empty. This breaks clients that rely on `lookup_invoice` to verify that an invoice was paid.

NIP-47 `lookup_invoice` responses include a `preimage` field for the payment preimage. getRPreimage only reads this.r_preimage. That works for LND-style invoice objects, but not for invoice objects whose preimage arrives as `payment_preimage` or `preimage`. ZEUS already tries to include it here:

```ts
// stores/NostrWalletConnectStore.ts
const result = NostrConnectUtils.createNip47Transaction({
    type: 'incoming',
    state,
    invoice: invoice.getPaymentRequest,
    payment_hash: invoice.getRHash || request.payment_hash!,
    amount: satsToMillisats(invoice.getAmount),
    description: invoice.getMemo,
    ...(invoice.isPaid && {
        preimage: invoice.getRPreimage
    }),
    description_hash: invoice.getDescriptionHash,
    settled_at: invoice.settleDate.getTime() / 1000,
    created_at: invoice.getCreationDate.getTime() / 1000,
    expires_at: invoice.getCreationDate.getTime() / 1000
});
```

The problem appears to be that `Invoice.getRPreimage` only reads `r_preimage`:

```ts
// models/Invoice.ts
@computed public get getRPreimage(): string {
    if (!this.r_preimage) return '';
    const preimage = this.r_preimage.data || this.r_preimage;
    return typeof preimage === 'object'
        ? Base64Utils.bytesToHex(preimage)
        : typeof preimage === 'string'
        ? preimage.includes('=')
            ? Base64Utils.base64ToHex(preimage)
            : preimage
        : '';
}
```

Some ZEUS/backend paths use different preimage field names:

```ts
// Core Lightning invoice rows map the preimage to payment_preimage.
{
    payment_preimage: invoice[8]
}

// NIP-47 lookup_invoice responses use preimage.
{
    preimage: '...'
}
```

Because `Invoice.getRPreimage` ignores `payment_preimage` and `preimage`, the `lookup_invoice` response can omit the preimage even after the invoice has been paid.

## Expected Behavior

[...]

Issue #4006, opened 2026-04-20
NWC `list_transactions` Missing `payment_hash` Field


lnbc1QWFyb24:
### Describe the bug

I am working on [Clams](https://clams.tech) a Bitcoin accounting app, and one of our users is trying to sync their Zeus wallet transactions over NWC. The `list_transactions` NWC method seems to be returning at least one payment with a missing `payment_hash` field which is required as part of the spec (and we also need it as a uniqued id for a payment).

### Reproduce

This may be hard to replicate, but try calling the `list_transactions` NWC method for an embedded LND node and see if there are any returned payments with a missing `payment_hash` field.

Note the Zeus version is v13.0.0-beta4

### ZEUS version

v13.0.0

### Node interface

Embedded LND

### Network

Clearnet

### Device

Iphone 15 Pro

### Device operating system

iOS 26.4.1

### Log output

```shell

```

ajaysehwal:
Hi @lnbc1QWFyb24, could you please try the latest version of the app? I believe this issue has been resolved with the recent NWC changes. Let us know if this issue still exists

lnbc1QWFyb24:
Thanks @ajaysehwal I will let the user know to give it a go.

lnbc1QWFyb24:
@ajaysehwal He said he is still seeing the same error on the latest beta version of Zeus.

kaloudis:
I believe this is fixed with https://github.com/zeusLN/zeus/pull/4294

v13.2.0 will be available for testing soon.

kaloudis:
v13.2.0-beta1 is now available for testing https://github.com/ZeusLN/zeus/releases/tag/v13.2.0-beta1

kaloudis:
ZEUS v13.2.0 is now available https://github.com/ZeusLN/zeus/releases/tag/v13.2.0

kaloudis:
Closing this for now. Please re-open if the issues returns.

PR #4269, opened 2026-07-24
fix(nwc): correct NIP-47 transaction field mapping

warioishere:
# Description

Relates to issue: **#4006**, **#4036**

> Opened as a draft: on-device testing on Android and iOS is still outstanding,
> and the platform/backend matrix below is unticked until then. Feedback on the
> approach is welcome in the meantime.

When ZEUS acts as an NWC wallet service, several NIP-47 transaction fields are
mapped from raw backend fields instead of the normalized model getters. The raw
field names only match LND, so anything else loses data.

**`payment_hash` is empty for non-LND invoices (#4006).**
`convertLightningDataToNip47Transactions` read `invoice.payment_hash`, which only
Core Lightning sets — LND returns `r_hash`. `Invoice.getRHash` is the normalizer
for exactly this, and the `lookup_invoice` path already used it. The model getter
itself only read `r_hash`, although `getFormattedRhash` in the same file already
did `this.r_hash || this.payment_hash`. Both getters now cover both field names.

`extractPaymentHashFromActivity` had the same gap and fell back to `activity.id`,
which holds the payment request — a BOLT11 string is not a payment hash, so it
now reports an empty string rather than a wrong value.

**`preimage` is empty for paid invoices (#4036).**
`Invoice.getRPreimage` only read `r_preimage`. Core Lightning uses
`payment_preimage` and NIP-47 transaction objects (returned by
`backends/NostrWalletConnect.ts`) use `preimage`. Both are now covered.

**`expires_at` equalled `created_at`.**
`buildNip47TransactionForLightningInvoiceLookup` set `expires_at` to the creation
date. It is now derived from Core Lightning's absolute `expires_at`, or LND's
`creation_date + expiry`, or `0` when the invoice does not expire.

The old expiry check used `invoice.timestamp`, which LND does not set, so the
comparison was `NaN` and an expired invoice was reported as `pending` forever.
It now reuses the computed expiry, so the check actually fires.

Note this makes the existing `failed` mapping reachable for the first time.
NIP-47 defines `expired` as the state for expired invoices and reserves `failed`
for payments, but `Nip47Transaction` in `@getalby/sdk` types `state` as
`"settled" | "pending" | "failed" | "accepted"`, so `expired` is not expressible
without casting around the SDK type. I kept the existing mapping rather than do
that — happy to revisit if you'd prefer it handled here.

**`settled_at` was reported for unpaid invoices.**
It was emitted unconditionally, and `Invoice.settleDate` falls back to
`timestamp`, so unpaid invoices carried a settle time. It is now `0` unless the
invoice is paid.

**The Cashu lookup reported the payment request as the preimage.**
`buildNip47TransactionForCashuInvoiceLookup` set `preimage` to the BOLT11 string
when the quote was paid. Clients use the preimage as proof of payment, so a wrong
value is worse than none; `CashuInvoice` carries no preimage, so the field is now
left empty. The same function assumed `request.payment_hash` was always present —
NIP-47 permits looking up by `invoice` alone — and now recovers the hash from the
payment request in that case.

The model getters can only turn an empty string into a real value, never the
other way around, so nothing that worked before can break. They are shared,
though, so there are two visible side effects outside NIP-47 — both surface data
that was previously dropped rather than changing anything that worked:

- `views/Invoice.tsx` renders its R hash and R preimage rows conditionally on
  these getters, so Core Lightning invoices now show those rows where they were
  previously blank.
- The `Payment Hash` column of the activity CSV export (`ActivityCsvUtils.ts`)
  is now populated for Core Lightning invoices.

I can narrow the fix to the NWC call sites instead if you'd rather leave the
shared getters alone — say the word.

### Known limitation

Cashu `make_invoice` activities now report an empty `payment_hash` in
`list_transactions` where they previously reported the payment request.
`CashuInvoice` carries a mint `quote`, not a payment hash, and
`extractPaymentHashFromActivity` is synchronous, so it cannot decode the BOLT11
to recover one. Both values are wrong per spec — an empty field is the honest
one, and it stops a BOLT11 string being consumed as a transaction identifier.

Recovering the real hash means decoding the quote's payment request, which I've
kept out of this PR to hold the diff to field mapping. I'll address it in a
follow-up PR covering the Cashu and outgoing-payment lookup paths — or fold it
in here if you'd rather not ship the intermediate state.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Eight cases were added to `utils/NostrConnectUtils.test.ts` covering LND-shaped
and Core Lightning-shaped invoices through `lookup_invoice`,
`list_transactions` and the connection-activity conversion. Six of them fail
against master; the remaining two guard paths that were already correct.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/models/Invoice.ts
+++ b/models/Invoice.ts
@@ -69,6 +69,8 @@ export default class Invoice extends BaseModel {
     public invoice_amount_msat: string | number;
     public bolt12: string;
     public payment_preimage: string;
+    // NIP-47 (Nostr Wallet Connect) transaction objects
+    public preimage?: string;
     // pay req
     public timestamp?: string | number;
     public destination?: string;
@@ -107,8 +109,11 @@ export default class Invoice extends BaseModel {
     }
 
     @computed public get getRPreimage(): string {
-        if (!this.r_preimage) return '';
-        const preimage = this.r_preimage.data || this.r_preimage;
+        // LND uses r_preimage, Core Lightning payment_preimage, NIP-47 preimage
+        const source =
+            this.r_preimage || this.payment_preimage || this.preimage;
+        if (!source) return '';
+        const preimage = source.data || source;
         return typeof preimage === 'object'
             ? Base64Utils.bytesToHex(preimage)
             : typeof preimage === 'string'
@@ -119,8 +124,10 @@ export default class Invoice extends BaseModel {
     }
 
     @computed public get getRHash(): string {
-        if (!this.r_hash) return '';
-        const hash = this.r_hash.data || this.r_hash;
+        // LND uses r_hash, Core Lightning and NIP-47 payment_hash
+        const source = this.r_hash || this.payment_hash;
+        if (!source) return '';
+        const hash = source.data || source;
         return typeof hash === 'object'
             ? Base64Utils.bytesToHex(hash)
             : typeof hash === 'string'
--- a/utils/NostrConnectUtils.test.ts
+++ b/utils/NostrConnectUtils.test.ts
@@ -37,6 +37,10 @@ import * as nostrTools from 'nostr-tools';
 
 import NostrConnectUtils from './NostrConnectUtils';
 import Payment from '../models/Payment';
+import Invoice from '../models/Invoice';
+import Base64Utils from './Base64Utils';
+import BackendUtils from './BackendUtils';
+import type { ConnectionActivity } from '../models/NWCConnection';
 
 // Stable hex values: repeat a hex digit 64 times to fill 32 bytes
 const hex64 = (c: string) => c.repeat(64);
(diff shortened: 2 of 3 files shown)


warioishere:
Flagging one deliberate behavior change that reviewers should weigh in on, also noted under "Known limitation" in the description.

Cashu `make_invoice` activities now report an empty `payment_hash` in `list_transactions` where they previously reported the payment request. `CashuInvoice` carries a mint `quote` rather than a payment hash, and `extractPaymentHashFromActivity` is synchronous, so it cannot decode the BOLT11 to recover one.

Both values are wrong per NIP-47, but a BOLT11 string is the more harmful of the two: clients use `payment_hash` as a transaction identifier, which is exactly what #4006 is about. An empty field fails loudly instead of silently poisoning a client's index.

This is not being left as-is. A follow-up PR will decode the quote's payment request so the Cashu path reports a real hash — it is grouped with the outgoing-payment lookup work because both need `lookup_invoice` to reach beyond the invoices table. I kept it out of this PR so the diff stays limited to field mapping and stays easy to revert on its own. Say the word if you'd rather have it folded in here instead.

ajaysehwal:
@warioishere, this looks solid. After on-device testing, could you mark it ready for review?

warioishere:
Thanks for looking it over.

I should be upfront, and I flagged this to Evan separately already: I won't get to the on-device matrix myself. I'm running a number of other projects and can't give this the device time it deserves — that's why it's a draft rather than sitting in your review queue. I also don't have a Core Lightning node available, which is awkward here, because CLN is really what this change is aimed at: the bug is that `payment_hash` and `payment_preimage` are read from LND's field names only, so Core Lightning invoices lose both.

So please treat this as material rather than a finished contribution — take it over, rewrite it, or cherry-pick whatever is useful, no attribution needed.

If any of you already has the rigs set up, I'd genuinely appreciate a hand with the testing. What would settle it:

- **Embedded LND** — the reporter's setup in #4006; enough to confirm `payment_hash` is no longer empty in `list_transactions`.
- **CLNRest** — the case I can't cover at all, and the one the fix targets.
- A screenshot of the invoice detail screen on CLN. `getRHash` and `getRPreimage` are shared getters, so the R hash and R preimage rows now populate for Core Lightning invoices where they used to be blank. That is a UI-visible side effect outside NWC and deserves a second opinion on whether you want it.

If someone can cover those, I'll mark it ready straight after — or push to the branch and take it over entirely, whichever is less friction for you.

Same situation applies to #4272, #4273 and #4276 and the other PRs form me, which continue the same work and are all drafts for the same reason.

warioishere:

PR #4272, opened 2026-07-25
fix(nwc): resolve lookup_invoice by invoice as well as payment_hash

warioishere:
# Description

> Opened as a draft: on-device testing on Android and iOS is still outstanding,
> and the platform/backend matrix below is unticked until then.

NIP-47 defines the `lookup_invoice` parameters as:

```
"payment_hash": "31afdf1..", // payment hash of the invoice, one of payment_hash or invoice is required
"invoice": "lnbc50n1..." // invoice to lookup
```

The Lightning path in `handleLookupInvoice` only ever read `request.payment_hash`
and returned `NOT_FOUND` when it was absent, so a spec-legal request carrying
only the payment request failed for every transaction. The Cashu path already
matched on the invoice string; only Lightning was affected.

The hash is now recovered from the payment request using the existing
`decodeInvoiceTags` helper, which the file already uses for the Cashu lookup and
for `make_invoice`. `payment_hash` still wins when both parameters are supplied.

The base64-decoding of `payment_hash` moved along with it — some clients send the
hash base64-encoded rather than as hex, which the previous inline code already
handled and which is preserved.

## Scope

This only fixes *how the request is resolved*. It does not make `lookup_invoice`
able to find outgoing payments: the lookup still queries the invoices table only,
so looking up a settled payment by its invoice will still return `NOT_FOUND` on
LND-family backends. That is a separate, larger change and I'd rather keep it in
its own PR so this one stays revertible on its own.

Independent of #4269 — the two touch different functions and can merge in either
order. They do both append a `describe` block at the end of
`utils/NostrConnectUtils.test.ts`, so whichever lands second will need a trivial
rebase there; happy to do that whenever you decide the order.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Six cases in `utils/NostrConnectUtils.test.ts` cover hex and base64
`payment_hash`, resolution from the invoice alone, precedence when both are
given, and both undecodable-input cases. The fixture is the BOLT11 reference
vector from the spec; its payment hash was confirmed by decoding it rather than
quoted from anywhere.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/NostrConnectUtils.ts
+++ b/utils/NostrConnectUtils.ts
@@ -603,7 +603,7 @@ export default class NostrConnectUtils {
      * payment_hash or invoice is required". Resolves the request to a payment
      * hash in hex, or undefined when neither parameter yields one.
      */
-    static async resolveLookupInvoicePaymentHash(
+    static async resolvePaymentHash(
         request: Nip47LookupInvoiceRequest
     ): Promise<string | undefined> {
         const paymentHash = request.payment_hash?.trim();
--- a/utils/NostrConnectUtils.ts
+++ b/utils/NostrConnectUtils.ts
@@ -23,6 +23,7 @@ import CashuInvoice from '../models/CashuInvoice';
 import CashuToken from '../models/CashuToken';
 import Transaction from '../models/Transaction';
 
+import Base64Utils from './Base64Utils';
 import { localeString } from './LocaleUtils';
 import dateTimeUtils from './DateTimeUtils';
 import Bolt11Utils from './Bolt11Utils';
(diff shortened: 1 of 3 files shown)


utils/NostrConnectUtils.ts
```diff
@@ -597,6 +598,34 @@ export default class NostrConnectUtils {
         }
     }
 
+    /**
+     * NIP-47 lookup_invoice takes either payment_hash or invoice — "one of
+     * payment_hash or invoice is required". Resolves the request to a payment
+     * hash in hex, or undefined when neither parameter yields one.
+     */
+    static async resolveLookupInvoicePaymentHash(
```
ajaysehwal: nit: Could we rename this method? We might end up using it in other places as well, so something like resolvePaymentHash would be more general than a name that implies it's only for invoice lookup.

warioishere: Agreed, renamed to `resolvePaymentHash` in f47dbe5.

It pairs well with something that came out of the review on #4288: base64-encoded hashes need normalizing before any exact comparison, since ZEUS's own NWC client sends them that way. That PR adds a `normalizePaymentHash` helper. So the two read as: normalize a value, resolve a request.

Whichever of the two lands second should have `resolvePaymentHash` delegate to `normalizePaymentHash` rather than keeping its own copy of the base64 branch. Happy to do that rebase once you decide the order.

warioishere:

PR #4273, opened 2026-07-25
fix(nwc): let lookup_invoice find outgoing payments

warioishere:
# Description

> Opened as a draft: on-device testing on Android and iOS is still outstanding.
> The per-backend table below marks exactly which claims are read off the code
> and which still need a real node — please read it before reviewing.

NIP-47 `lookup_invoice` returns `"incoming"` for invoices and `"outgoing"` for
payments; the spec is explicit that it "retrieves either type". ZEUS only ever
queried the invoices table, so a settled outgoing payment returned `NOT_FOUND`
on the LND-family backends, and on `ldk-node` — where `listPayments()` matches
outbound payments by hash and `formatPaymentAsInvoice` discards `direction` — it
came back typed `incoming` with `fees_paid: 0` and an expiry.

The handler now asks for an invoice, then for a payment, and reports `NOT_FOUND`
only when neither answers. A payment is mapped with `type: "outgoing"`, the real
fee, and `expires_at: 0`, since a payment has no expiry.

Both lookups return `undefined` when they have nothing rather than signalling
absence by throwing. That distinction matters: an earlier revision ran the
payment path from the invoice lookup's `catch` block, which silently skipped it
on any backend that reports a miss without throwing — exactly what LDK Node does.
Inferring "not an invoice" from an exception makes correctness depend on every
backend failing in the same way, and they don't.

For the same reason the invoice lookup no longer treats "is an object" as proof
of being an invoice: a response carrying neither a payment hash nor a payment
request is now a miss rather than a transaction built from an error body.

One caveat on `ldk-node`, since the wording above could be read as covering it:
this PR alone does **not** fix that backend. Its `lookupInvoice` finds the
outbound payment and returns it as an invoice, so the lookup never misses and
this fallback never runs. #4276 restricts that search to inbound payments, which
is what makes the two work together. Either merges on its own; LDK Node needs
both.

## Keeping the node round-trips down

Fetching the payment list on every lookup would be wasteful, so:

- **The fallback only runs when the invoice lookup missed.** A client polling an
  invoice it created — the common case — costs no extra request at all. Only
  requests that return `NOT_FOUND` today are affected.
- **One bounded fetch, not the default 500.** `maxPayments: 100`.
- **A 5 s window coalesces bursts.** A client polling the same hash while waiting
  for settlement causes one node round-trip, not one per poll. The window is
  deliberately short because stale payment state is what NWC clients already
  complain about; it's a single constant if you want it tuned.

Two things worth calling out in review:

**The fallback does not go through `PaymentsStore`.** `PaymentsStore.getPayments`
overwrites `this.payments` — the shared list the wallet UI renders from — and
toggles its loading flag. Routing a lookup through it with a reduced
`maxPayments` would have silently truncated the user's visible payment history on
every NWC lookup. It calls `BackendUtils.getPayments` directly and builds the
models locally instead.

**`reversed: true` is passed explicitly.** `lndmobile/index.ts` spreads it
conditionally (`...(params?.reversed && { reversed })`), so omitting it leaves the
protobuf default of `false` and embedded LND returns the *oldest* page — a
freshly settled payment would never be in the window. LND REST and LNC default it
to `true`, embedded LND does not.

## Per-backend basis for this change

Verified by reading the code in this repo:

| Backend | `maxPayments` honored | `payment_hash` shape | Fee source |
|---|---|---|---|
| LND (REST) | yes | hex string | `fee_msat` / `fee_sat` |
| Embedded LND | yes (default 1000) | hex string (lnrpc) | `fee_msat` |
| LNC | yes | hex string | `fee_msat` |
| CLNRest | **no** — SQL `limit 150` | hex string from `sendpays` | `amount_sent_msat − amount_msat` |
| LndHub | **no** — `/gettxs` | Buffer object, handled by `Payment.paymentHash` | unverified |
| LDK Node | **no** — in-memory, outbound only | string | `fee_msat` |
| NWC | yes (`limit`) | string | `fees_paid` |

So the bound is a real bound on LND-family and NWC, while CLNRest, LndHub and
LDK Node return their natural cap regardless. LDK Node is in-memory so this costs
nothing there; LndHub is the one that still fetches everything, which a targeted
lookup in a follow-up backends PR would fix.

The fee unit was checked rather than assumed: `Payment.getFee` returns **sats**
(it divides `fee_msat` down), so the mapper scales it back to msats for
`fees_paid`.

**Not verified without a node, and what I'd most like tested:** the runtime shape
of LndHub's `/gettxs` rows — whether a fee field survives into `Payment` at all,
and whether its buffer `payment_hash` matches what `lookup_invoice` is queried
with. Everything else above is read off code paths in this repo, but I have not
run any of it against a real node.

## Why a list scan rather than a targeted query

Worth being explicit, since scanning a payment list is not obviously the right
shape for a by-hash lookup: on the LND family there is no alternative in a
request/response style. `ListPaymentsRequest` in this repo's own
`proto/lightning.proto` offers `include_incomplete`, `index_offset`,
`max_payments`, `reversed` and creation-date bounds — no payment-hash filter. The
hash-targeted call is `TrackPaymentV2`, which returns a stream
(`proto/routerrpc/router.proto`), so using it for a one-shot lookup means opening
a stream, taking the first message and closing it.

Other backends could do better — Core Lightning's `listsendpays` accepts
`payment_hash` and `bolt11` directly, and LDK Node's payment list is in memory —
so a first-class `lookupPayment` per backend would beat this. That is a larger
change across all seven backends, proposed separately in #4286 rather than
smuggled in here.

## Scope

`type` for the *invoice* branch is untouched, and self-payments still resolve as
`incoming`: the invoice lookup is tried first and payments are only a fallback, so
nothing that resolves today changes.

Independent of #4269 and #4272. On its own this resolves payments by hash; with
#4272 (which resolves `lookup_invoice` by `invoice` as well) it also resolves them
by payment request. All three append to the end of
`utils/NostrConnectUtils.test.ts`, so whichever lands last needs a trivial rebase
there — happy to do that in whatever order you prefer.

Still open after this, handled separately in #4276: `lookup_invoice` fails on
`cln-rest` and `lndhub`. Correcting something I wrote earlier — the two fail for
*different* reasons. CLNRest defines no `lookupInvoice`, so `BackendUtils.call`
returns `false`. LndHub `extends LND` and therefore inherits `lookupInvoice`,
which requests `/v1/invoice/{hash}` against an LndHub server that serves no such
route. Same `NOT_FOUND` for the client, different cause and different fix.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Eleven cases in `utils/NostrConnectUtils.test.ts` cover the payment mapping
(outgoing type, no expiry, fee in msats, Core Lightning fee derivation, LndHub
buffer hash decoding, settled/failed/in-flight states) and the cache window
including the clock-moved-backwards case.

The payment mapping was extracted out of `convertLightningDataToNip47Transactions`
so `list_transactions` and `lookup_invoice` share one implementation; that part is
a pure refactor with no behavior change.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -95,6 +95,10 @@ const SAVE_CONNECTIONS_DEBOUNCE_MS = 500;
 const PENDING_PAYMENT_STATUS_MIN_AGE_MS = 5_000;
 /** Per-connection debounce for pending pay_invoice status refresh in getActivities */
 const PENDING_PAYMENT_STATUS_REFRESH_MS = 30_000;
+/** Payments fetched when lookup_invoice falls back to searching payments */
+const LOOKUP_PAYMENT_FETCH_LIMIT = 100;
+/** Window in which that fetched list is reused across lookup_invoice calls */
+const LOOKUP_PAYMENT_CACHE_MS = 5_000;
 
 export const DEFAULT_NOSTR_RELAYS = [
     'wss://relay.getalby.com/v1',
@@ -173,6 +177,11 @@ export default class NostrWalletConnectStore {
         string,
         number
     >();
+    /** Short-lived payment list reused by lookup_invoice, see getPaymentsForLookup */
+    private lookupPaymentCache: {
+        payments: Payment[];
+        fetchedAt: number;
+    } | null = null;
 
     settingsStore: SettingsStore;
     balanceStore: BalanceStore;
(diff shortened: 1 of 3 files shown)


warioishere:

PR #4275, opened 2026-07-25
build: fix vendored lnc-rn bob build + add mocha test suite

kaloudis:
# Description

Fixes the vendored lnc-rn module's broken dev tooling — surfaced while working on #4274 — and gives its dead `yarn test` script a real suite.

**Commit 1 — `build:` fix `bob build` under repo nesting**

`yarn install` inside `zeus_modules/@lightninglabs/lnc-rn` failed at the `prepare` step every time: tsc's automatic `@types` inclusion walks up parent `node_modules`, so the vendored copy parsed the zeus root's `@types/lodash`, whose `infer N extends number` syntax (TS 4.8+) the module's pinned TypeScript 4.5.5 can't parse (TS1005 — parse errors, so `skipLibCheck` doesn't help). Upstream never hits this because it isn't nested inside another repo, and upstream is on the same toolchain (TS 4.5.5 / bob 0.18.3), so there was no fix to import.

- Fix: `"typeRoots": ["./node_modules/@types"]` in the module's tsconfig.
- Also re-commits `dist/` as regenerated by the pinned toolchain so `prepare` is idempotent and installs leave a clean tree (previously every successful build dirtied the checkout). Verified determinism by hashing the diff across consecutive builds.
  - `dist/typescript` (the only dist output zeus consumes, via the package's `types` field) is **byte-identical**
  - `dist/commonjs` / `dist/module` drift is babel class-property lowering only; zeus never runs these — Metro resolves the `react-native: lib/index` field and bundles source
  - `dist/commonjs/package.json` (`{"type":"commonjs"}`) is dropped: bob 0.18.3 doesn't emit it, and the package has no `"type"` field so CJS is already the default

**Commit 2 — `tests:` add mocha suite**

Upstream lightninglabs/lnc-rn has never shipped a `test/` directory (zero commits touch that path), so `yarn test` always failed with "No test files found". New suite covers the parts of `lib/` that run in plain Node (13 passing):

- `api/createRpc`: unary calls route to `lnc.request` with capitalized method names; every entry in lnc-core's `subscriptionMethods` routes to `lnc.subscribe` — this is the Proxy all LNC RPCs in Zeus go through
- `util/credentialStore`: field storage, `isPaired` logic, `clear()`
- `util/log`: level filtering, prefix naming, `fromEnv` no-op default

Wiring: `.mocharc.yml` uses `ts-node/register/transpile-only` (no new `@types` devDeps needed); the spec pattern avoids `**` because the module's `minimatch: 9.0.7` pin makes mocha's bundled glob 7 crash on globstar (`GLOBSTAR` is a `Symbol` in minimatch 4+); root jest gains `zeus_modules` in `testPathIgnorePatterns` since its crawler only skips `node_modules` and would otherwise collect these mocha tests (root suite count unchanged at 49).

No dependency changes — no lockfile is touched.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

_Device testing pending. No runtime surface changes expected: Metro bundles the module's `lib/` source (unchanged), and the `dist/typescript` types zeus's tsc consumes are byte-identical._

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub
 
_A smoke test of the LNC backend is the relevant check before merge, out of caution — though the bundled LNC code is unchanged._

### Locales

### Other:


ajaysehwal (approved):
tACK

Issue #4216, opened 2026-07-08
NWC server: look back for past events


kaloudis:
nevent1qqspyv9smytdq0vkh4usr6xrzdqtzfpypgyytznzh5kmrt094k4tm9spz4mhxue69uhhyetvv9ujuerpd46hxtnfduhsnhdf63

<img width="414" height="348" alt="Image" src="https://github.com/user-attachments/assets/a590f042-1119-41f6-99b9-2336e01387e6" />

@ajaysehwal it would be good to explore the feasibility of loading past NWC events from before the wallet is opened.

PR #4285, opened 2026-07-28
fix(nwc): reconcile pending payments and correct the unpaid filter

warioishere:
# Description

Relates to issue: **#4216** (related), **ZEUS-0000**

> Opened as a draft: on-device testing is still outstanding.

Two `list_transactions` bugs that compound each other. Both were checked against
the NIP-47 spec text and against Alby Hub as the reference implementation, not
from memory.

## A settled payment can stay `pending` indefinitely

When a `pay_invoice` does not resolve inside `waitForPaymentCompletion`, it is
recorded with `status: 'pending'`. The only code that promotes it to `success`
lives in `getActivities`, which is called from exactly two places — the NWC
connection detail and activity screens. The NIP-47 handler never called it.

So over NWC the payment stays `pending` until a human opens that screen in the
app, while the wallet's own transaction list has shown it as paid all along.
This is a plausible part of what gets reported as delayed or stale NWC
transaction lists (#4216 covers the other half — events that arrive while the
wallet is closed).

The reconciliation block is lifted out of `getActivities` into
`reconcilePendingPayInvoiceActivities` and called from both. Behavior in the
existing UI path is unchanged; the nested filters simply collapse into one.

Reading is not the only trigger any more, because a payment that settles while
the app is closed produces no read either. Reconciliation now also runs when the
app returns to the foreground (both platform monitors) and after
`subscribeToAllConnections` finishes, so the state is caught up before a client
asks rather than because one did.

Node load is unchanged: every path goes through
`getPaymentsForPendingPayInvoiceRefresh`, which requires a pending activity to be
at least 5 s old and rate limits fetches to one per connection per 30 s, falling
back to the already-fetched payment list otherwise. A connection with nothing
pending returns immediately without touching the node, so the added triggers cost
nothing in the normal case.

## The `unpaid` filter had it backwards

NIP-47 documents the parameter as:

```
"unpaid": true, // include unpaid invoices, optional, default false
```

so it *widens* the result set. ZEUS treated it as a selector, and ignored the
documented default:

| `unpaid` | ZEUS before | NIP-47 + Alby Hub |
|---|---|---|
| absent | no filter — unpaid included | settled only |
| `true` | pending only | every state |
| `false` | settled only | settled only ✓ |

Alby Hub, as the reference wallet service, maps `unpaid` onto both
`unpaidOutgoing` and `unpaidIncoming` and then applies
`WHERE state = settled` only when neither is set, with no state filter at all
when both are (`transactions/transactions_service.go`,
`nip47/controllers/list_transactions_controller.go`). That matches the spec
reading and is what this implements.

**This changes default behavior**: a client that sends no `unpaid` parameter will
stop receiving pending and failed transactions. That is the documented default,
and it is also why the two fixes belong in one PR — with the filter corrected but
the reconciliation missing, a payment stuck at `pending` would become invisible
by default rather than merely mislabelled.

Alby Hub also accepts `unpaid_outgoing` and `unpaid_incoming` as finer-grained
extensions. Those are not in NIP-47, so they are not implemented here; happy to
add them if you want closer Alby client parity.

## What this still is not

Reconciliation on a trigger is a backstop, not a design. The wallet still learns
that a payment settled by asking afterwards rather than by being told at the
time. A proper fast path would be either a payment subscription per backend — no
backend in ZEUS has one today, only `subscribeInvoice`/`subscribeTransactions` —
or NIP-47 notifications, which ZEUS currently advertises in its info event and in
`get_info` but never publishes (#4287).

Neither removes the need for this code: the app can be killed, and a payment that
settles while it is not running produces no event for anyone to receive. So the
backstop stays either way; what is missing above it is the fast path.

## Deliberately not in this PR

For connections holding `pay_invoice`, `list_transactions` still answers from
that connection's own activity log, so payments made in the wallet UI, by other
connections, incoming invoices and on-chain transactions do not appear. That is
the larger deviation, but changing it is a permission-model decision rather than
a bug fix — it would expose the whole wallet history to any connection granted
`list_transactions`, which is what the spec implies but not necessarily what
existing users expect from the current behavior. It is also the area of #3432,
which was reverted. I would rather have that discussion separately than bundle it
into a revertible fix. Happy to open an issue for it if useful.

Worth noting for whoever picks that up: the node-backed branch calls
`paymentsStore.getPayments()`, which overwrites the shared payment list the
wallet UI renders from. Any redesign there should fetch without mutating the
store.

Independent of #4269, #4272, #4273 and #4276. All of them append to the end of
`utils/NostrConnectUtils.test.ts`, so whichever lands last needs a trivial rebase
there.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Five cases cover the filter: the documented default, explicit `false`, explicit
`true` across all states, that settled transactions are not dropped when `true`,
and that the `type` filter still composes with it. Four of the five fail against
master; the one that passes is `unpaid: false`, which was already correct.

The reconciliation change is in a store, which has no test coverage in this repo,
so it is verified by review and device testing rather than by a test.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1084,46 +1084,7 @@ export default class NostrWalletConnectStore {
             );
         }
 
-        const pendingPayInvoiceActivities = connection.activity.filter(
-            (activity) =>
-                activity.type === 'pay_invoice' && activity.status === 'pending'
-        );
-        if (pendingPayInvoiceActivities.length > 0) {
-            const lightningPending = pendingPayInvoiceActivities.filter(
-                (activity) => activity.payment_source !== 'cashu'
-            );
-            if (lightningPending.length > 0) {
-                const payments =
-                    await this.getPaymentsForPendingPayInvoiceRefresh(
-                        connectionId,
-                        lightningPending
-                    );
-                for (const activity of lightningPending) {
-                    const payment = payments.find(
-                        (p) =>
-                            p.getPaymentRequest === activity.id ||
-                            (!!activity.paymentHash &&
-                                p.paymentHash === activity.paymentHash)
-                    );
-                    if (!payment) continue;
-
-                    runInAction(() => {
-                        activity.payment = new Payment(payment);
-                        if (!payment.isIncomplete) {
-                            activity.status = 'success';
-                            const amountSats =
-                                Math.floor(Number(activity.satAmount)) ||
-                                Math.floor(Number(payment.getAmount) || 0);
-                            if (amountSats > 0) {
-                                connection.trackSpending(amountSats);
-                            }
-                        } else if (payment.isFailed) {
-                            activity.status = 'failed';
-                        }
-                    });
-                }
-            }
-        }
+        await this.reconcilePendingPayInvoiceActivities(connection);
 
         runInAction(() => {
             connection.activity = connection.activity.filter((activity) => {
@@ -1745,6 +1706,7 @@ export default class NostrWalletConnectStore {
             let nip47Transactions: Nip47Transaction[] = [];
 
             if (connection.hasPaymentPermissions()) {
+                await this.reconcilePendingPayInvoiceActivities(connection);
                 nip47Transactions = connection.activity
                     .map((activity) =>
                         NostrConnectUtils.convertConnectionActivityToNip47Transaction(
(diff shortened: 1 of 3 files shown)


warioishere:

PR #4288, opened 2026-07-28
fix(nwc): find outgoing Cashu payments in lookup_invoice

warioishere:
# Description

> Opened as a draft: on-device testing is still outstanding. This one needs a
> Cashu-configured wallet and a real melt.

On a Cashu-configured wallet, `lookup_invoice` can never resolve an outgoing
payment. `handleLookupInvoice` branches on `isCashuConfigured` before anything
else and searches `cashuStore.invoices` only — mint quotes, i.e. incoming. A
melted payment lives in `cashuStore.payments` and is never looked at. Because the
branch short-circuits, the Lightning path never runs either, so the fallbacks
that cover this for other backends do not apply.

NIP-47 defines `lookup_invoice` over both directions — `"incoming"` for invoices,
`"outgoing"` for payments — so a settled melt returning `NOT_FOUND` is a
spec violation, and it makes automated payment confirmation impossible for any
client talking to a Cashu wallet.

Payments are now searched by `payment_hash` or by payment request, reusing the
existing `findPaymentForInvoice` helper, and mapped as outgoing.

## Field mapping, and where the values come from

Checked against `stores/CashuStore.ts`, which constructs the record after a
successful melt, rather than assumed:

| NIP-47 field | Source |
|---|---|
| `payment_hash` | the decoded payment request spread into the record; `Bolt11Utils.decode` exposes it as top-level hex |
| `invoice` | `bolt11` |
| `amount` | `amount` from the melt result, scaled to msats |
| `fees_paid` | `fee` (the real `fee_paid` from the melt), scaled to msats |
| `preimage` | `payment_preimage`, set from `meltResult.preimage \|\| ''` |
| `created_at` / `settled_at` | the decoded invoice timestamp |
| `expires_at` | `0` — a payment does not expire |

Two deliberate choices:

**`state` is hardcoded to `settled`.** A melt only reaches `cashuStore.payments`
after it succeeded. Deriving the state from `Payment.isIncomplete` would be
wrong here, because that getter is preimage-based and a Cashu melt may have no
preimage at all — see below. This matches how `convertCashuDataToNip47Transactions`
already maps Cashu payments for `list_transactions`.

**An empty `preimage` is reported as empty.** Per NUT-05 the melt quote response
carries `payment_preimage` as a method-specific, optional field — ZEUS's own
`MeltQuote` interface types it `payment_preimage?: string | null` — so a mint may
legitimately return none. Reporting an empty string is correct; fabricating
something is not.

## Control flow

The Cashu branch checks the invoice list first via `findCashuInvoiceForNwcLookup`
and only then the payment list, mirroring the ordering on the Lightning side so
nothing that resolves today changes. Both checks answer with a value or
`undefined`; neither infers absence from an exception.

That means the invoice list is walked twice on a payment lookup — once to decide,
once inside the existing builder. It is an in-memory array and this keeps
`buildNip47TransactionForCashuInvoiceLookup` untouched, which avoids a conflict
with #4269. Happy to collapse it if you would rather have the extra pass gone.

## Scope

Only the Cashu lookup path. The remaining Cashu gap is the empty `payment_hash`
on `make_invoice` activities in `list_transactions`, which #4269 addresses on the
invoice side.

Independent of #4269, #4272, #4273, #4276 and #4285; all of them append to the end
of `utils/NostrConnectUtils.test.ts`, so whichever lands last needs a trivial
rebase there.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Seven cases in `utils/NostrConnectUtils.test.ts` cover matching by hash and by
payment request, the fee and amount scaling, the absent-expiry, the preimage
present and absent, a non-matching hash, and an empty payment list. The fixture
mirrors the record `CashuStore` actually builds after a melt rather than an
idealised one.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1705,12 +1705,37 @@ export default class NostrWalletConnectStore {
     ): NWCWalletServiceResponsePromise<Nip47Transaction> {
         try {
             if (this.isCashuConfigured) {
-                const result =
-                    await NostrConnectUtils.buildNip47TransactionForCashuInvoiceLookup(
+                // lookup_invoice covers both directions, and a melt never lands
+                // in the invoice list, so payments have to be searched too
+                const cashuInvoice =
+                    await NostrConnectUtils.findCashuInvoiceForNwcLookup(
                         this.cashuStore.invoices || [],
                         request
                     );
-                return { result, error: undefined };
+                if (cashuInvoice) {
+                    const result =
+                        await NostrConnectUtils.buildNip47TransactionForCashuInvoiceLookup(
+                            this.cashuStore.invoices || [],
+                            request
+                        );
+                    return { result, error: undefined };
+                }
+
+                const paymentResult =
+                    NostrConnectUtils.buildNip47TransactionForCashuPaymentLookup(
+                        this.cashuStore.payments || [],
+                        request
+                    );
+                if (paymentResult) {
+                    return { result: paymentResult, error: undefined };
+                }
+
+                return NostrConnectUtils.createNip47Error(
+                    localeString(
+                        'stores.NostrWalletConnectStore.error.invoiceNotFound'
+                    ),
+                    Nip47ErrorCode.NOT_FOUND
+                );
             }
             const rHash =
                 request.payment_hash && request.payment_hash.includes('=')
--- a/utils/NostrConnectUtils.test.ts
+++ b/utils/NostrConnectUtils.test.ts
@@ -38,6 +38,8 @@ import * as nostrTools from 'nostr-tools';
 import NostrConnectUtils from './NostrConnectUtils';
 import Payment from '../models/Payment';
 import CashuPayment from '../models/CashuPayment';
+import CashuInvoice from '../models/CashuInvoice';
+import Base64Utils from './Base64Utils';
 
 // Stable hex values: repeat a hex digit 64 times to fill 32 bytes
 const hex64 = (c: string) => c.repeat(64);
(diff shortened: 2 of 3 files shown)


ajaysehwal (commented):
Nice catch! I think we have the same normalization issue here as in the Lightning payment flow.

findPaymentForInvoice compares paymentHash using an exact ===, but payment.paymentHash is stored as hex (decoded from the BOLT11 invoice). If an NWC client sends request.payment_hash in base64, the lookup can incorrectly return NOT_FOUND even though the payment exists.

We could normalize the hash before the lookup, similar to the Lightning flow:
```js

const paymentHash =
    request.payment_hash?.includes('=')
        ? Base64Utils.base64ToHex(request.payment_hash)
        : request.payment_hash;

const payment = NostrConnectUtils.findPaymentForInvoice(
    request.invoice || '',
    payments || [],
    paymentHash
);
```

This keeps the behavior consistent across both flows and avoids format-related lookup failures

warioishere:
you're right — and it turns out to be worse than the payment path alone.

Two things I checked before fixing:

**It isn't hypothetical.** ZEUS's own NWC *client* sends the hash base64-encoded — `backends/NostrWalletConnect.ts`:

```ts
lookupInvoice = async (data: any) =>
    await this.nwc.lookupInvoice({
        payment_hash: Base64Utils.hexToBase64(data.r_hash)
    });
```

So one ZEUS wallet looking up a payment on another ZEUS wallet hits this every time. (Separately: the NIP-47 examples show hex, so the client side arguably shouldn't be sending base64 at all — but normalizing on receipt is the compatible fix either way, and it keeps working with clients that already do.)

**The invoice path has the same gap.** `findCashuInvoiceForNwcLookup` compares `decoded.paymentHash` — hex out of the BOLT11 — against the raw `request.payment_hash`, so a base64 request misses there too. That's pre-existing rather than something this PR introduced, but it sits in the branch this PR touches, and fixing only the payment side would have left the inconsistency you're pointing at.

So instead of normalizing inline I pulled it into a named helper and used it in both places:

```ts
static normalizePaymentHash(paymentHash?: string): string | undefined {
    const hash = paymentHash?.trim();
    if (!hash) return undefined;
    return hash.includes('=') ? Base64Utils.base64ToHex(hash) : hash;
}
```

Six cases added: hex passthrough, base64 decoding, blank input, plus base64 matching through both the payment lookup and the invoice finder. The invoice-finder tests use the BOLT11 reference vector from the spec, with its hash confirmed by decoding it rather than quoted.

One note for merge ordering: #4272 introduces `resolveLookupInvoicePaymentHash` for the Lightning path, which carries the same normalization inline. Whichever of the two lands second should have it delegate to this helper so there is only one copy. Happy to do that rebase.

warioishere:

PR #4292, opened 2026-07-30
fix(vss): fail restore loudly when VSS is unreachable instead of building an empty wallet

kaloudis:
# Description

Users on poor connections reported that restoring from seed (LDK Node) was not restoring their channels, while testnet VSS restores worked fine.

**Root cause:** the native VSS dual-store build has a wall-clock timeout (30s existing node / 60s restore). On timeout — or any VSS failure — `buildNode` silently fell back to building a fresh node with an empty local SQLite store. The restore flows ignored the returned `vssError`, so the wallet config was saved and the restore looked successful with zero channels. Restore time scales with node history × connection latency (state is fetched from VSS sequentially, key by key), which is why it selectively hit users with poor connections and real mainnet history, and not testnet tests.

**Follow-on damage:** the fallback build leaves a local DB containing fresh node state. Later dual-store builds read local-first, so the empty state shadows VSS even after connectivity recovers — and the fork's background bulk sync then overwrites the VSS backup's channel manager with the fresh state (unconditional puts, no user activity required). That fork-side clobber + a second masking bug (fast VSS errors like clock-skew auth rejections silently producing "successful" empty restores) are fixed in the companion PR: ZeusLN/ldk-node#3.

## Changes in this PR

- New native method `setVssFailOnError` (Android + iOS LDK Node modules). When set, a VSS build timeout/failure **rejects** (code `vss_error`) instead of falling back to an empty local store. Resets with each `createBuilder`, so existing-node startups keep the availability-first fallback.
- `failOnVssError` threaded through `initializeNode` → `initNode` → `createLdkNodeWallet`. Hard-fail builds get a 180s timeout (up from 60s): now that exceeding it produces a visible error rather than a silently wrong wallet, the window should be generous enough for slow-but-working connections.
- `createLdkNodeWallet` deletes the just-created node dir on failed init, so a partially-written local DB can't shadow VSS on retry. Every retry gets a fresh dir and the full 180s.
- `SeedRecovery` passes `failOnVssError: true` and shows a localized error (new key `views.Settings.SeedRecovery.vssRestoreFailed`) so the user knows channels were **not** restored and can retry.
- `WalletConfiguration` passes `failOnVssError` when an existing seed is imported; brand-new wallet creation (including onboarding) keeps the non-fatal warning path.
- `startLdkNodeWallet` treats a startup with a saved config but **no local DB** (e.g. config arriving on a new device via cloud keychain sync) as a de-facto restore and hard-fails on VSS errors. `Wallet.tsx` catches the failure, surfaces it via `SettingsStore.error`/`errorMsg` (standard BalancePane error + retry), and releases `fetchLock` so retry works without an app restart.

## Suggested test plan

1. Restore a mainnet seed with the VSS server blocked (firewall the VSS URL after DNS) → expect the localized error + retry, not an empty wallet.
2. Retry the same restore with VSS reachable → channels restore normally.
3. Regression: normal restore, new-wallet creation, and existing-wallet startup on both platforms.
4. With the new LDKNodeFFI binaries from ZeusLN/ldk-node#3: restore with device clock skewed ±1 day (previously a silent empty restore), and simulate a poisoned device (local-only build over a VSS store with monitors) → expect the CRITICAL log and no VSS writes.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [x] No, I’m a fool
- [ ] Yes
- [ ] N/A

(The modified `utils/LdkNodeUtils.ts` paths wrap native LDK Node module calls; the new behavior is timeout/fallback semantics inside the native build, which unit tests can't reach without mocking away the behavior under test.)

- [x] Android (emulator, Pixel-class sdk_gphone64_arm64, Android 15 / API 35)
- [x] iOS (iPhone 17 Pro simulator, iOS 26)

Detailed manual test results (10 scenarios across both platforms, with console evidence): see https://github.com/ZeusLN/zeus/pull/4292#issuecomment-5126034073

On-device
- [x] LDK Node (ldk-node `v0.7.0-zeus-vss-enhanced`, testnet, VSS: vss.zeusln.com — both platforms)
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

(Only LDK Node code paths are touched; embedded LND restore via Olympus SCB is untouched.)

### Locales

### Other:

---

**Update:** ZeusLN/ldk-node#3 has been merged and released as [`v0.7.0-zeus-vss-enhanced`](https://github.com/ZeusLN/ldk-node/releases/tag/v0.7.0-zeus-vss-enhanced); this PR now pins it (`deps: bump ldk-node to v0.7.0-zeus-vss-enhanced`). Fresh binaries verified end-to-end via `fetch-libraries.sh` checksums on both platforms. The release also carries the additive BOLT11 route-hints APIs from ZeusLN/ldk-node#2 — uniffi version unchanged and interface additive-only, so the vendored bindings need no regeneration.


kaloudis:
## Manual test results — 2026-07-29, `v0.7.0-zeus-vss-enhanced` binaries

| # | Platform | Scenario | Result | Notes |
|---|----------|----------|--------|-------|
| 1 | iOS | Normal restore from seed | ✅ | Channel restored from VSS (testnet, vss.zeusln.com) |
| 2 | iOS | Clock-skew: VSS auth timestamp offset −3 days → server rejection | ✅ | Loud localized `vssRestoreFailed` error; **no empty wallet**. Pre-fix this scenario silently produced a "successful" empty restore |
| 3 | iOS | Retry after fixing the skew | ✅ | Clean restore in ~2s — failed attempt leaves nothing behind (fresh node dir per attempt) |
| 4 | iOS | Poisoned device: deleted `monitors` + `manager` rows from `ldk_node_data.sqlite` to replicate a pre-fix fallback build | ✅ | Normal mode + CRITICAL verdict + bulk sync skipped; zero VSS writes for the session (console evidence below) |
| 5 | iOS | Re-restore after the poisoned session | ✅ | Channel came back from VSS — **backup survived**, clobber vector confirmed closed |
| 6 | iOS | New-wallet regression: create + relaunch | ✅ | Creation used non-fatal path (60s); second launch: normal mode, **no CRITICAL false-positive**, `Background bulk sync complete — 8/8 keys synced to VSS` |
| 7 | iOS | Error display parity | ✅ | `BuildError` messages now extracted via exhaustive switch; identical clean message on both platforms (re-verified after rebuild) |
| 8 | Android | Normal restore from seed | ✅ | Channel restored from VSS |
| 9 | Android | Clock-skew (as #2) | ✅ | Kotlin `vss_error` rejection path; clean `Dual store setup failed: Failed to read from store.` |
| 10 | Android | Retry after fixing the skew | ✅ | Clean restore, channel back |

Devices: iPhone 17 Pro simulator (iOS 26); Android emulator sdk_gphone64_arm64 (Android 15 / API 35).

<details>
<summary>Console evidence for the poisoned-device scenario (#4)</summary>

```
LdkNodeModule: VSS build timeout set to 30.0s
LdkNodeModule: VSS fail-on-error set to false
DualStore: Local store has 14 keys — normal mode (local-only reads)
LdkNodeModule: [timing] Node built with dual store successfully in 20ms
DualStore: CRITICAL — VSS holds 1 channel monitor(s) but the local store has none. Local state looks like a fresh node built over an existing backup; disabling all VSS writes this session to avoid overwriting the backup. Recover by restoring from seed into a new wallet.
DualStore: Bulk sync skipped — VSS pushes disabled this session
```

Subsequent re-restore (#5) entered restore mode (`180s / fail-on-error true`, `Local store is empty — entering restore mode`, `Restore mode — skipping background bulk sync`) and recovered the channel in ~2.3s.

</details>

kaloudis:
Failure error

|Android|iOS|
|-|-|
|Screenshot_1785380466|Simulator Screenshot - iPhone 17 Pro - 2026-07-29 at 23 24 05|

shubhamkmr04 (commented):
On timeout, native rejects immediately but the worker keeps running the blocking buildWithDualStoreAndFixedHeaders against config.storageDirPath (LdkNodeModule.kt:405-427; iOS workItem.cancel() is a no-op once the FFI call has started). createLdkNodeWallet's catch then RNFS.unlinks that directory while the build is still live.
If the in-flight build or the orphan node.stop() writes after the unlink, it recreates a partial DB, the shadowing the cleanup exists to prevent. An immediate retry also starts a second build over the same path; buildNode only guards against an already-built node (LdkNodeModule.kt:347-359), nothing tracks a build in progress.

kaloudis:
Good catch @shubhamkmr04, confirmed on both platforms. The timeout path rejected while the FFI build kept running, JS then unlinked the storage dir under the live build, and nothing stopped a retry from building over the same path.

Addressed in two layers:

**`6a79ac285`** closes the race in the bridge:
- Result-ownership handoff (CAS on Android, lock-guarded claim on iOS) between the caller and the build thread — whoever loses the timeout race handles cleanup, and a build that finishes just past the deadline is used instead of orphaned
- The orphaned build thread now deletes the storage dir itself once the FFI call actually returns (hard-fail/restore path only); fast VSS failures delete natively before rejecting
- `buildNode` waits up to 10s for a previous timed-out build to finish, then rejects with `build_in_progress` instead of racing it
- JS no longer unlinks on `vss_error`/`build_in_progress`: native owns that cleanup now

**ZeusLN/ldk-node#4** fixes the root enabler: vss-client already caps each HTTP request at 10s, so the long hangs were retry accumulation (100 attempts / 180s of backoff per VSS operation). Now capped at 6 attempts / 30s (~1 min worst case), so builds fail well inside the watchdog and orphaned threads become a rare edge rather than the expected timeout behavior. Merged and shipped as `v0.7.0-zeus-vss-enhanced.1`; binaries bumped in `4a2546f1c`.

Note: the earlier manual test matrix predates these commits; we need another run (restore, timeout→retry, and the failure paths on both platforms) before merge.

kaloudis:
## Manual test results — 2026-07-30, `v0.7.0-zeus-vss-enhanced.1` binaries

| # | Platform | Scenario | Result | Notes |
|---|----------|----------|--------|-------|
| 1 | iOS | Normal restore from seed | ✅ | Channel restored from VSS (testnet, vss.zeusln.com) |
| 2 | iOS | Clock-skew: VSS auth timestamp offset −3 days → server rejection | ✅ | Loud localized `vssRestoreFailed` error; **no empty wallet**. Pre-fix this scenario silently produced a "successful" empty restore |
| 3 | iOS | Retry after fixing the skew | ✅ | Clean restore in ~2s — failed attempt leaves nothing behind (fresh node dir per attempt) |
| 4 | iOS | Poisoned device: deleted `monitors` + `manager` rows from `ldk_node_data.sqlite` to replicate a pre-fix fallback build | ✅  | Normal mode + CRITICAL verdict + bulk sync skipped; zero VSS writes for the session (console evidence below) |
| 5 | iOS | Re-restore after the poisoned session | ✅  | Channel came back from VSS — **backup survived**, clobber vector confirmed closed |
| 6 | iOS | New-wallet regression: create + relaunch | ✅ | Creation used non-fatal path (60s); second launch: normal mode, **no CRITICAL false-positive**, `Background bulk sync complete — 8/8 keys synced to VSS` |
| 7 | iOS | Error display parity | ✅ | `BuildError` messages now extracted via exhaustive switch; identical clean message on both platforms (re-verified after rebuild) |
| 8 | Android | Normal restore from seed | ✅ | Channel restored from VSS |
| 9 | Android | Clock-skew (as #2) | ✅ | Kotlin `vss_error` rejection path; clean `Dual store setup failed: Failed to read from store.` |
| 10 | Android | Retry after fixing the skew | ✅ | Clean restore, channel back |

Devices: iPhone 17 Pro simulator (iOS 26); Android emulator sdk_gphone64_arm64 (Android 15 / API 35).

shubhamkmr04 (approved):
tACK

PR #4294, opened 2026-07-30
fix(nwc): fix lookup_invoice, list_transactions, and backend invoice lookup

ajaysehwal:
## Summary

This PR fixes several issues in the Nostr Wallet Connect (NIP-47) implementation around invoice/payment lookup and transaction reporting.

Previously, NWC clients such as **Amethyst**, **Primal**, and others could receive incorrect transaction data or fail to find outgoing payments after sending or receiving Lightning payments. This PR ensures `lookup_invoice` and `list_transactions` return consistent and accurate `type`, `state`, `preimage`, payment hash, and expiry information across supported backends.

## Fixes

- Fix `lookup_invoice` failing to resolve outgoing Lightning and Cashu payments.
- Fix incorrect NIP-47 transaction mapping (`expires_at`, `payment_hash`, `preimage`, etc.).
- Fix `list_transactions` default `unpaid` filtering to match the NIP-47 specification.
- Fix pending `pay_invoice` activities remaining in a `pending` state until the NWC activity screen was opened.
- Add missing or incomplete `lookupInvoice` implementations across supported backends.
- Improve transaction lookup reliability by normalizing payment hashes and refreshing payment state when needed.

### NIP-47 (`utils/NostrConnectUtils.ts`)

- Centralize invoice/payment lookup into `lookupInvoiceTransaction()`.
- Add shared transaction mappers for:
  - Lightning invoices
  - Lightning payments
  - Cashu payments
- Correctly calculate `expires_at` using `created_at + expiry`.
- Normalize payment hashes (hex ↔ base64) before matching.
- Update `lookup_invoice` to fall back to outgoing payment lookup when no invoice is found.
- Support Cashu outgoing payment lookup by payment hash or BOLT11 invoice.
- Fix `list_transactions` filtering so:
  - Default returns settled transactions only.
  - `unpaid: true` includes pending and failed transactions.

### NWC Store (`stores/NostrWalletConnectStore.ts`)

- Add lazy `getPaymentsForLookup()` with a fresh node fetch and a 5-second cache.
- Reconcile pending `pay_invoice` activities when:
  - Opening the activity screen.
  - Calling `list_transactions`.
  - The app returns to the foreground.
  - The relay reconnects.
- Persist reconciliation results and invalidate the lookup cache after payments.
- Refresh settled Lightning and Cashu invoices before opening transaction details.

### CLNRest
- Implement `lookupInvoice` using `/v1/listinvoices` (including unpaid invoices).
### LndHub
- Implement `lookupInvoice` using `/getuserinvoices` with payment hash lookup.
### LDK Node
- Restrict `lookupInvoice` to inbound invoices only, preventing outgoing payments from being incorrectly returned as invoices.

### UI

- Hide invoice expiry after an invoice has been paid.
- Only display invoice expiry in NWC activity when a valid expiry exists.
- Remove the duplicate `TextEncoder` polyfill from the NWC store.

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Expanded `utils/NostrConnectUtils.test.ts` to nearly **100 tests**, covering:

- NIP-47 transaction mapping.
- Payment hash normalization.
- Lightning and Cashu lookup paths.
- `list_transactions` unpaid filtering.
- Backend compatibility and regression cases.

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [x] LndHub / LNbits

---

### Locales
- [x] N/A — no new strings

- [x] N/A — no dependency changes

### Other

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -266,6 +266,22 @@ export default class CLNRest {
         return this.postRequest('/v1/withdraw', request);
     };
     getMyNodeInfo = () => this.postRequest('/v1/getinfo');
+    // listinvoices filters natively by payment hash, and unlike the SQL query
+    // used by getInvoices it is not restricted to paid invoices
+    lookupInvoice = (data: any) =>
+        this.postRequest('/v1/listinvoices', {
+            payment_hash: data.r_hash
+        }).then((response: any) => {
+            const invoice = response?.invoices?.[0];
+            if (!invoice) {
+                throw new Error(
+                    localeString(
+                        'stores.NostrWalletConnectStore.error.invoiceNotFound'
+                    )
+                );
+            }
+            return invoice;
+        });
     getInvoices = (data?: any) =>
         this.postRequest('/v1/sql', {
             query: `SELECT label, bolt11, bolt12, payment_hash, amount_msat, status, amount_received_msat, paid_at, payment_preimage, description, expires_at FROM invoices ORDER BY created_index DESC LIMIT ${
--- a/backends/LdkNode.ts
+++ b/backends/LdkNode.ts
@@ -1246,8 +1246,13 @@ export default class LdkNode {
      */
     lookupInvoice = async (data: { r_hash: string }): Promise<any> => {
         const payments = await LdkNodeInjection.payments.listPayments();
+        // Inbound only — an outbound payment carries the same hash but is not
+        // an invoice, and formatting it as one mislabels it as incoming
         const payment = data?.r_hash
-            ? payments.find((p) => p.kind.hash === data.r_hash)
+            ? payments.find(
+                  (p) =>
+                      p.kind.hash === data.r_hash && p.direction === 'inbound'
+              )
             : undefined;
 
         if (payment) {
--- a/backends/LndHub.ts
+++ b/backends/LndHub.ts
@@ -4,6 +4,7 @@ import LND from './LND';
 import LoginRequest from './../models/LoginRequest';
 import Base64Utils from './../utils/Base64Utils';
 import Bolt11Utils from './../utils/Bolt11Utils';
+import Invoice from './../models/Invoice';
 import { localeString } from './../utils/LocaleUtils';
 import { Hash as sha256Hash } from 'fast-sha256';
 import { ecdsaSignDERHex } from '../utils/SigningUtils';
@@ -47,7 +48,8 @@ export default class LndHub extends LND {
     lookupInvoice = (data: any) =>
         this.getRequest('/getuserinvoices').then((invoices: any) => {
             const invoice = (invoices || []).find(
-                (userInvoice: any) => userInvoice.payment_hash === data.r_hash
+                (userInvoice: any) =>
+                    new Invoice(userInvoice).getRHash === data.r_hash
             );
             if (!invoice) {
                 throw new Error(
--- a/backends/LndHub.ts
+++ b/backends/LndHub.ts
@@ -4,6 +4,7 @@ import LND from './LND';
 import LoginRequest from './../models/LoginRequest';
 import Base64Utils from './../utils/Base64Utils';
 import Bolt11Utils from './../utils/Bolt11Utils';
+import { localeString } from './../utils/LocaleUtils';
 import { Hash as sha256Hash } from 'fast-sha256';
 import { ecdsaSignDERHex } from '../utils/SigningUtils';
 
@@ -40,6 +41,23 @@ export default class LndHub extends LND {
         this.getRequest('/getuserinvoices').then((data: any) => ({
             invoices: data
         }));
+    // Overrides LND's /v1/invoice/{r_hash}, a route LndHub does not serve.
+    // LndHub has no per-invoice endpoint returning details — /checkpayment only
+    // answers paid/unpaid — so the user's invoice list is searched by hash.
+    lookupInvoice = (data: any) =>
+        this.getRequest('/getuserinvoices').then((invoices: any) => {
+            const invoice = (invoices || []).find(
+                (userInvoice: any) => userInvoice.payment_hash === data.r_hash
+            );
+            if (!invoice) {
+                throw new Error(
+                    localeString(
+                        'stores.NostrWalletConnectStore.error.invoiceNotFound'
+                    )
+                );
+            }
+            return invoice;
+        });
 
     createInvoice = (data: any) =>
         this.postRequest('/addinvoice', {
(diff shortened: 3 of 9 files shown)


warioishere:
Nice consolidation. The shared mappers and comparing via `new Invoice(userInvoice).getRHash` on LndHub are both better than what I had.

One thing I'd like to check before this goes further, in case I'm misreading it: `handleLookupInvoice` passes `lightningPayments: this.paymentsStore.payments`, and nothing on the lookup path fetches. The only `getPayments` calls in the store are in `list_transactions`, `pay_invoice` and the pending refresh.

If that's right, two cases look off to me:

- A payment still in flight when `pay_invoice` returns sits in the cache as incomplete. A client polling `lookup_invoice` for confirmation then reads that same snapshot on every call and keeps seeing `pending`, because nothing refreshes it — the same shape as the `list_transactions` bug this PR fixes.
- After a service restart the array is empty, so an outgoing lookup returns `NOT_FOUND` until something else happens to fetch.

Have I overlooked a refresh somewhere?

If not, this keeps it lazy so the common case still costs nothing:

```ts
lightningPayments: () => this.getPaymentsForLookup()
```

```ts
private async getPaymentsForLookup(): Promise<Payment[]> {
    // deliberately not via paymentsStore — that overwrites the payment list
    // the wallet UI renders from and toggles its loading flag
    const data = await BackendUtils.getPayments({
        maxPayments: 100,
        // lndmobile spreads this conditionally, so omitting it leaves the
        // protobuf default of false and embedded LND returns the oldest page
        reversed: true
    });
    return (data?.payments || []).map((p: any) => new Payment(p));
}
```

with `lookupLightningInvoiceTransaction` calling it only once the invoice lookup has missed. A short cache window on top keeps a polling client from hitting the node on every request.

Separately and much smaller: the identity check I had before the invoice mapping is gone. No backend in tree can reach it today — they all throw on a miss, and `call()` returning `false` is caught by the existing check — so it's only guarding against a future one. Fine to leave.

ajaysehwal:
> Nice consolidation. The shared mappers and comparing via `new Invoice(userInvoice).getRHash` on LndHub are both better than what I had.
> 
> One thing I'd like to check before this goes further, in case I'm misreading it: `handleLookupInvoice` passes `lightningPayments: this.paymentsStore.payments`, and nothing on the lookup path fetches. The only `getPayments` calls in the store are in `list_transactions`, `pay_invoice` and the pending refresh.
> 
> If that's right, two cases look off to me:
> 
> * A payment still in flight when `pay_invoice` returns sits in the cache as incomplete. A client polling `lookup_invoice` for confirmation then reads that same snapshot on every call and keeps seeing `pending`, because nothing refreshes it — the same shape as the `list_transactions` bug this PR fixes.
> * After a service restart the array is empty, so an outgoing lookup returns `NOT_FOUND` until something else happens to fetch.
> 
> Have I overlooked a refresh somewhere?
> 
> If not, this keeps it lazy so the common case still costs nothing:
> 
> ```ts
> lightningPayments: () => this.getPaymentsForLookup()
> ```
> 
> ```ts
> private async getPaymentsForLookup(): Promise<Payment[]> {
>     // deliberately not via paymentsStore — that overwrites the payment list
>     // the wallet UI renders from and toggles its loading flag
>     const data = await BackendUtils.getPayments({
>         maxPayments: 100,
>         // lndmobile spreads this conditionally, so omitting it leaves the
>         // protobuf default of false and embedded LND returns the oldest page
>         reversed: true
>     });
>     return (data?.payments || []).map((p: any) => new Payment(p));
> }
> ```
> 
> with `lookupLightningInvoiceTransaction` calling it only once the invoice lookup has missed. A short cache window on top keeps a polling client from hitting the node on every request.
> 
> Separately and much smaller: the identity check I had before the invoice mapping is gone. No backend in tree can reach it today — they all throw on a miss, and `call()` returning `false` is caught by the existing check — so it's only guarding against a future one. Fine to leave.

Thanks for the detailed review! You're right—I overlooked those stale/empty cache cases.

I like the lazy fetch approach after the invoice lookup misses since it keeps the common path cheap without affecting the UI state. I'll update the PR accordingly (and likely add a small cache window as well).

Good point about the identity check too. I have a few other improvements in mind as well, so I'll take another pass over this before merging.

warioishere:
Sorry for coming back minutes after the fix, we're genuinely waiting on this patch for our Marketplace :D

The lazy fetch looks right. One thing left: `lookupPaymentsCache` runs for 30s and is never invalidated — it's only ever assigned inside `getPaymentsForLookup`. So:

1. Client polls for hash X → invoice miss → payments fetched and cached, X not in them yet → NOT_FOUND, correct.
2. `pay_invoice` runs, payment succeeds.
3. Client polls again → invoice miss → cache is under 30s old → stale list without X → NOT_FOUND for a payment that definitely went through.

Same for a payment cached while still in flight: it stays `pending` for up to half a minute after it settled.

Context for why we care about this specific window: we're building against ZEUS's NWC service and need a confirmation that a payment actually happened. Our app polls `lookup_invoice` for exactly that, so we'd rather not poll for a long time or hand the user a "not found" for a payment they just made — we want the proof more or less immediately.

Dropping the cache when a payment is recorded covers it without touching the TTL — one line in `finalizePayment`:

```ts
this.lookupPaymentsCache = null;
```

Then 30s stays fine for everything else, and the one case where freshness actually matters is handled.

ajaysehwal:
@warioishere, could you take another look and see if I missed anything? I think it's working fine now. Also, could you close your PRs? I've already cherry-picked the useful commits from them.

kaloudis (changes requested):
concept ACK 44c6a107. Two things before ACK:

1) CLN `listinvoices `returns no creation timestamp, so `created_at` serializes as `null` for CLN lookups. This needs a fallback and a test with the real CLN field shape.

2) `reconcilePendingPayInvoiceActivities` can double-`trackSpending` under concurrent triggers. Re-check `activity.status === 'pending'` inside the action. 

Also please fill in the platform matrix in the PR template once you've done additional testing.

warioishere:
besides of what @kaloudis noted, lgtm, everything I raised looks addressed.

kaloudis:
Bumping this into v13.1.4

kaloudis (approved):
ACK

PR #4291, opened 2026-07-29
fix(CLN): include unpaid invoices in Activity

ajaysehwal:
# Description

In this PR, we fix CLN Activity only showing paid Lightning invoices. Unpaid and expired requests now appear with the correct amount and sort order, consistent with LND. 

## Problem
- **`CLNRest.getInvoices`** queried CLN's invoice table with `WHERE status = 'paid'`, so Activity only ever received settled invoices. Pending and expired requests never appeared.
- **Unpaid amounts showed as `0 sats`** because Activity uses `Invoice.getAmount`, which reads LND fields (`value`, `amt`) and received amounts (`amount_received_msat`). CLN unpaid invoices only expose the requested amount via `amount_msat`, which `getRequestAmount` already handles but `getAmount` did not.
- **Sorting** could be wrong for unpaid CLN invoices because CLN's SQL schema has no `created_at`; without a creation timestamp they would sort at `0` in Activity.
## Fix
- **`backends/CLNRest.ts`**
  - Remove the `WHERE status = 'paid'` filter so all invoice statuses (`unpaid`, `paid`, `expired`) are returned.
  - Derive `creation_date` from the BOLT11 timestamp so unpaid invoices sort correctly in Activity.
- **`models/Invoice.ts`**
  - Fall back to `getRequestAmount` in `getAmount` for unsettled invoices, so CLN `amount_msat` displays correctly without backend-specific mapping.
Paid invoice behavior is unchanged — received amounts are still resolved before the new fallback.

## Test plan
- [ ] Connect Zeus to a CLN REST node
- [ ] Create a Lightning invoice (e.g. 1,000 sats) and confirm it appears in **Activity** as **Requested Payment** with the correct amount
- [ ] Confirm expiry countdown displays for pending invoices
- [ ] Pay the invoice and confirm Activity updates to **You Received** after refresh
- [ ] Confirm expired unpaid invoices show as **Expired Request**
- [ ] Confirm existing paid invoice history still displays correctly with correct amounts and dates

This pull request is categorized as a:

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
[...]

PR #4300, opened 2026-08-01
fix(CLN/LND): poll lookupInvoice in the Receive payment watcher

kaloudis:
# Description

Fixes a regression introduced by #4291 (v13.1.4 milestone), plus a latent version of the same bug on LND (REST) flagged in review by @ajaysehwal.

The receive watcher in `views/Receive.tsx` polls `getInvoices({ limit: 10 })` every 5s and scans the list for the watched payment hash.

- **CLN**: before #4291, the CLN SQL query filtered `WHERE status = 'paid'`, so the window held the last 10 *paid* invoices and the watched invoice appeared at the top the moment it was paid. After #4291, the window holds the last 10 invoices of **any** status.
- **LND (REST)**: `/v1/invoices` has never filtered by paid status, so the same window problem has existed there all along.

On a busy node (POS, ZEUS Pay), if 10 or more invoices are created after the watched one before the payer pays, the watched invoice is pushed out of the result set. Every poll then returns the same 10 newer rows, the hash never matches, and `handlePaymentReceived` never fires; the success screen never shows even though the payment arrived. (The payment is still recorded correctly in Activity; only the real-time detection is missed.)

## Fix

Poll `lookupInvoice` instead, which filters natively by payment hash and includes unpaid invoices, eliminating the list window entirely rather than widening the limit.

- **CLN**: uses `CLNRest.lookupInvoice` (added in #4294), backed by `/v1/listinvoices`.
- **LND**: uses `/v1/invoice/{r_hash_str}`, which expects a hex payment hash; the watcher's `rHash` is base64url (from `getFormattedRhash`), so it is converted with `Base64Utils.base64UrlToHex`.
- Amount comparison goes through the `Invoice` model (`isPaid` / `getAmount`), which correctly handles both CLN's msat-suffixed amount strings and LND's `amt_paid_sat`.
- Lookup rejections (invoice not found, node unreachable) are swallowed so the next tick retries; the old code had no rejection handler at all.
- Match semantics (`amountPaid >= value`, `amountPaid !== 0`) and the `handlePaymentReceived` payload are unchanged.
- Embedded LND and Lightning Node Connect use invoice subscription streams and are unaffected.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -1038,33 +1038,33 @@ export default class Receive extends React.Component<
         if (implementation === 'lnd') {
             if (rHash) {
                 this.lnInterval = setInterval(() => {
-                    // only fetch the last 10 invoices
-                    BackendUtils.getInvoices({ limit: 10 }).then(
-                        (response: any) => {
-                            const invoices = response.invoices;
-                            for (let i = 0; i < invoices.length; i++) {
-                                const result = invoices[i];
-                                if (
-                                    result.r_hash
-                                        .replace(/\+/g, '-')
-                                        .replace(/\//g, '_') === rHash &&
-                                    Number(result.amt_paid_sat) >=
-                                        Number(value) &&
-                                    Number(result.amt_paid_sat) !== 0
-                                ) {
-                                    this.handlePaymentReceived(
-                                        result.amt_paid_sat,
-                                        {
-                                            type: 'ln',
-                                            tx: result.payment_request
-                                        }
-                                    );
-                                    this.clearIntervals();
-                                    break;
-                                }
+                    // Look the invoice up directly by payment hash instead of
+                    // scanning the last 10 invoices: on busy nodes the watched
+                    // invoice can be pushed out of that window before it is
+                    // paid. LND's REST lookup endpoint expects a hex hash,
+                    // while rHash is base64url here
+                    BackendUtils.lookupInvoice({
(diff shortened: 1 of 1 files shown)


ajaysehwal (commented):
@kaloudis, I think we should use the `lookupInvoice` method for LND as well.

kaloudis:
@ajaysehwal good call. The LND (REST) branch had the same window scan, and `/v1/invoices` has never filtered by paid status, so it was a latent version of the same bug (predating #4291). Pushed f442dd1dd, which switches it to `lookupInvoice` as well. One wrinkle: LND's REST lookup endpoint expects a hex payment hash while the watcher's `rHash` is base64url, so it converts with `Base64Utils.base64UrlToHex`. Embedded LND and LNC use invoice subscription streams, so they're unaffected.

ajaysehwal (commented):
tACK

PR #4301, opened 2026-08-01
fix: guard against payment double-submission in send path

kaloudis:
# Description

Adds an in-flight guard to the Lightning send path so a rapid double-tap (or a re-fired slide-to-pay) cannot dispatch a payment twice.

The keysend path is the most dangerous case: `TransactionsStore.sendPaymentInternal` generates a fresh random preimage on every invocation, so each duplicate submission has a different payment hash. That defeats every backend's same-hash duplicate protection (LND control tower, LDK `DuplicatePayment`, the REST `calls` dedup map), meaning a double-tap on a keysend can actually send the amount twice. Invoice payments were mostly saved by node-side same-hash dedup, but keysend had no protection on any backend, and neither the Send/Pay buttons nor the send path itself had any "payment in flight" condition.

## What changed

- `stores/TransactionsStore.ts`: a new dedicated `paymentInFlight` observable is set when `sendPaymentInternal` dispatches a payment, and `sendPaymentInternal` early-returns when it is already set. This is the authoritative choke point shared by every backend, so the guard holds regardless of which view initiates the payment. A dedicated flag is used instead of the pre-existing `loading` observable because `loading` is shared with unrelated operations (`getTransactions`, `sendCoins`, `broadcast`) that can run in the background (Activity refreshes, NWC `list_transactions`) and would otherwise disable the send buttons or silently drop a legitimate payment attempt.
- `views/Send.tsx` (`sendKeySendPayment`) and `views/PaymentRequest.tsx` (`triggerPayment`): bail when a payment is already in flight, which also prevents double navigation and a stale Lightning Node Connect event subscription.
- `views/Send.tsx` and `views/PaymentRequest.tsx`: disable their send buttons while `TransactionsStore.paymentInFlight`.
- `components/SwipeButton.tsx`: adds an optional `disabled` prop (read through a ref since the `PanResponder` is created once) so the slide-to-pay control is inert and dimmed while a payment is in flight. This composes with the existing `completed`-ref lock, which already prevents a second swipe after one succeeds.

`paymentInFlight` is cleared by `handlePayment`, `handlePaymentError`, and `reset()` (which runs when the wallet reconnects, not on every Wallet mount). Because a lost completion callback (for example a dropped Lightning Node Connect stream, or a hung request on a backend that gets no `timeout_seconds`) would otherwise leave the flag stuck and lock out all sends until the next reconnect, `sendPaymentInternal` also arms a timed backstop that clears the flag once the payment's timeout plus a grace period has elapsed. A sequence counter ensures the backstop only clears the payment it was armed for.

Note: the NWC service path calls `reset()` immediately before dispatching, so it intentionally bypasses the guard; the guard protects user-initiated UI sends.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### DIFF
--- a/stores/TransactionsStore.ts
+++ b/stores/TransactionsStore.ts
@@ -569,6 +569,15 @@ export default class TransactionsStore {
         amp,
         timeout_seconds
     }: SendPaymentReq) => {
+        // Guard against double-submission: if a payment is already in flight,
+        // ignore the new request. Without this a rapid double-tap (or a
+        // re-fired swipe) dispatches two payments, and keysend in particular
+        // generates a fresh preimage per call, defeating every backend's
+        // same-hash duplicate protection.
+        if (this.loading) {
+            return;
+        }
+
         this.paymentStartTime = Date.now();
         this.paymentDuration = null;
         this.loading = true;
--- a/stores/TransactionsStore.ts
+++ b/stores/TransactionsStore.ts
@@ -43,6 +43,10 @@ export interface SendPaymentReq {
     message?: string;
     amp?: boolean;
     timeout_seconds?: string;
+    // service-initiated payments (e.g. the NWC service path) neither set nor
+    // clear paymentInFlight, so the double-submission guard stays scoped to
+    // user-initiated sends
+    background?: boolean;
 }
 
 export default class TransactionsStore {
@@ -79,9 +83,14 @@ export default class TransactionsStore {
     // coin control
     @observable funded_psbt: string = '';
 
-    // identifies the payment the in-flight backstop timer was armed for,
-    // so it never clears the flag on behalf of a later payment
+    // monotonic id assigned to each dispatched payment
     private paymentSequence = 0;
+    // sequence of the payment that set paymentInFlight (null when none);
+    // completion callbacks and the backstop timer only clear the flag on
+    // behalf of that payment, so overlapping payments (e.g. a background
+    // NWC payment finishing while a user payment is still in flight)
+    // can't disarm each other's guard
+    private inFlightOwnerSeq: number | null = null;
 
     settingsStore: SettingsStore;
     nodeInfoStore: NodeInfoStore;
@@ -107,6 +116,7 @@ export default class TransactionsStore {
     public reset = () => {
         this.loading = false;
         this.paymentInFlight = false;
+        this.inFlightOwnerSeq = null;
         this.error = false;
         this.error_msg = null;
         this.transactions = [];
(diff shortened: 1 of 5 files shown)


stores/TransactionsStore.ts
ajaysehwal, line 1: I think we should only clear `paymentInFlight` when the payment reaches a terminal state.

Right now, `handlePayment` clears it even when the payment is still `IN_FLIGHT`. This could allow another payment attempt while the current HTLC is still in progress, especially for keysend payments.

Maybe we can do something like:

```ts
const stillInFlight =
    status === 'IN_FLIGHT' ||
    status === 1 ||
    payment.isInTransit;

if (!stillInFlight) {
    this.paymentInFlight = false;
}

kaloudis: Good catch. The status check alone wouldn't cover the client-side timeout paths (the LND REST forcedTimeout race resolves with no status or htlcs at all), and holding the flag without an active tracker only buys the backstop window (~60s) while a stuck HTLC can pend until CLTV expiry, so it shrinks the double-pay window rather than closing it. It would also lock out all unrelated sends in the meantime.

We're addressing this properly in #4317 by keeping the guard set on non-terminal or ambiguous outcomes and tracking the payment to a terminal state via trackPaymentV2/payment lookup. Keeping this PR scoped to accidental double-submission (double-tap / re-fired swipe).

shubhamkmr04 (commented):
NWC cuts both ways here. NostrWalletConnectStore calls transactionsStore.reset() immediately before dispatching, and reset() now clears paymentInFlight, so a remote pay request landing mid-payment doesn't just bypass the guard, it disarms it. The reverse also holds: NWC's dispatch goes through sendPaymentInternal and sets the flag, so for the duration of that remote payment the user's Pay/Send buttons are disabled and triggerPayment silently no-ops. Would a background flag on SendPaymentReq work, so service-initiated payments neither set nor clear it? That keeps the guard about user-initiated sends and leaves the reset()-on-reconnect escape hatch intact.

kaloudis:
Good catch on both counts, addressed in 71cbe41bb.

Took your suggestion: `SendPaymentReq` now has a `background` flag, and the NWC service path sets it. Background payments bypass the guard and never set `paymentInFlight`, so a remote NWC payment no longer disables the user's Pay/Send buttons.

One thing the flag alone didn't cover: the disarm you describe wasn't really the momentary `reset()` (the flag is re-armed synchronously by the next dispatch), it was that `handlePayment`/`handlePaymentError` cleared the flag unconditionally, so whichever overlapping payment completed first disarmed the guard armed for the other one. Clearing is now ownership-aware: each dispatch gets a sequence number, and the completion callbacks and backstop timer only clear the flag on behalf of the payment that set it. LNC view subscriptions can't identify their payment, so they clear unconditionally as before, and `reset()` on reconnect stays the unconditional escape hatch.

Residual, intentionally out of scope here: an NWC request landing while a user payment is in flight still calls `reset()` and stomps the shared `TransactionsStore` state (payment_hash, status, etc.), and a user send dispatched mid-NWC-payment can still confuse NWC's result resolution, which reads the shared store. That's the NWC-vs-user half of the shared-store problem, and the payment serialization queue I have in flight for the NWC store is the right place to gate it, alongside the NWC-vs-NWC serialization it already does.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4303, opened 2026-08-01
fix(nwc): serialize pay_invoice handling to prevent shared-store races

kaloudis:
# Description

Serializes NWC `pay_invoice` handling so two concurrent requests can never interleave their reads and writes of the shared singleton stores.

Both payment handlers work through global store state across awaits, and Nostr relay events dispatch handlers concurrently with no queue anywhere:

- `handleCashuPayInvoice` calls `cashuStore.getPayReq(request.invoice)`, which mutates `CashuStore.payReq` / `paymentRequest` / `meltQuote`, then re-reads that state after further awaits (`payLnInvoiceFromEcash` reads `this.payReq?.getRequestAmount`, `this.selectedMintUrl`, `this.meltQuote.quote`).
- `handleLightningPayInvoice` does `transactionsStore.reset()` then `sendPayment()` then `waitForPaymentCompletion()`, reading result state off the same shared `TransactionsStore` the UI uses.

With two overlapping `pay_invoice` requests (same or different connections), connection A's flow can melt connection B's quote (paying B's payee with an amount that was budget-validated against A's smaller invoice), and both clients can receive the same result/preimage, so one believes it paid when it didn't.

## What changed

- `stores/NostrWalletConnectStore.ts`: a store-level promise queue (`paymentQueue` + `enqueuePayment`). The `payInvoice` handler registration routes through it, so payment requests are processed strictly one at a time across all connections. A rejected handler propagates its error to its own request; the queue continues regardless.

The queue is deliberately store-level rather than per-connection: cross-connection interleaving is exactly the dangerous case. Non-payment handlers (`get_info`, `get_balance`, `make_invoice`, `lookup_invoice`, `list_transactions`, `sign_message`) are not queued; they do not participate in this race.

Throughput note: a queued request waits for the one ahead of it, which for an in-transit lightning payment can be up to the 120s completion wait. That is the correct trade for a funds path; under normal single-client sequential use nothing changes.

Not addressed here (tracked separately): budget accounting for pending payments, and passing decoded state explicitly instead of via shared stores, which would be a larger refactor of the payment flows than a fix like this should carry.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -81,6 +81,13 @@ const SERVICE_START_DELAY_MS = 2000;
 const PAYMENT_TIMEOUT_SECONDS = 120;
 const PAYMENT_FEE_LIMIT_SATS = 1000;
 const PAYMENT_PROCESSING_DELAY_MS = 100;
+/**
+ * Max time a queued pay_invoice may wait behind another payment before it
+ * is rejected with RATE_LIMITED instead of executing late. Kept well under
+ * typical NWC client timeouts so we never pay an invoice whose client has
+ * already given up on the request.
+ */
+const MAX_PAYMENT_QUEUE_WAIT_MS = 30_000;
 const SAVE_CONNECTIONS_DEBOUNCE_MS = 500;
 /** Min time a pay_invoice must stay pending before getActivities polls the node */
 const PENDING_PAYMENT_STATUS_MIN_AGE_MS = 5_000;
@@ -1293,7 +1300,16 @@ export default class NostrWalletConnectStore {
                     : this.handleLightningPayInvoice.bind(this, connection);
                 handler.payInvoice = (request: Nip47PayInvoiceRequest) =>
                     this.withGlobalHandler(connection.id, () =>
-                        this.enqueuePayment(() => payHandler(request))
+                        this.enqueuePayment(
+                            () => payHandler(request),
+                            () =>
+                                NostrConnectUtils.createNip47Error(
+                                    localeString(
+                                        'stores.NostrWalletConnectStore.error.paymentQueueWaitExceeded'
+                                    ),
+                                    Nip47ErrorCode.RATE_LIMITED
+                                )
+                        )
                     );
             }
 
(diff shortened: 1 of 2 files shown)


stores/NostrWalletConnectStore.ts
```diff
@@ -145,6 +145,8 @@ export default class NostrWalletConnectStore {
     @observable public iosAudioKeepAliveActive: boolean = false;
     @observable public iosAudioKeepAliveUptime: number = 0;
     @observable public iosAudioKeepAliveDisconnects: number = 0;
+    // Tail of the payment-handler queue; see enqueuePayment
+    private paymentQueue: Promise<unknown> = Promise.resolve();
```
ajaysehwal, line 156: nit: paymentQueue is not cleared in reset()

ajaysehwal (commented):
@kaloudis I think we should avoid queuing multiple NWC payments and fail fast if a payment is already in progress.
Right now, additional `pay_invoice` requests are queued and may execute much later if the current payment takes a long time. By then, the NWC client may have already timed out and considered the request failed.
This can cause the Nostr client to show the payment as failed or display a network error, while Zeus may still execute the queued payment later and deduct the sats.
Since NIP-47 supports `RATE_LIMITED`, I think we should return that when another payment is already running instead of adding it to the queue.

stores/NostrWalletConnectStore.ts
kaloudis: Intentional. Reassigning the tail in `reset()` cancels nothing (chained tasks still run), but it would let a post-reset payment run concurrently with a pre-reset one, which is the exact race this PR prevents. The tail must persist for the invariant to hold; it's a single self-draining promise, so nothing leaks.

kaloudis:
@ajaysehwal The harm is stale execution, not queueing itself. Normally the payment ahead settles in seconds and back-to-back zaps still work; pure fail-fast would spuriously reject those. Plan: keep the queue but stamp requests at enqueue, and if one has waited past ~30s when its turn comes, respond `RATE_LIMITED` instead of paying late. That kills the "client shows failed, sats leave later" case without breaking bursts. (Same-hash dedup already prevents duplicate bolt11 pays on client retry.) Will push that to this PR.

ajaysehwal (approved):
tACK

PR #4305, opened 2026-08-01
fix(nwc): count pending payments against connection spending budget

kaloudis:
# Description

Counts in-flight payments against NWC connection spending budgets.

`NWCConnection.canSpend` compared only the settled `totalSpendSats` against `maxAmountSats`. In-flight payments sit in the connection's activity log as `pending` records whose budget debit is deferred until reconcile flips them to `success` and calls `trackSpending`. A client could therefore fire multiple payments into hodl or slow-settling invoices (each returned as in-transit with an empty preimage after the 120s completion wait) and have every one pass `validateBudgetBeforePayment`; once they all settled, a multiple of the configured budget had been spent. Example: a 100k-sat budget allows 10 x 100k-sat payments if none settle before the last one is validated.

## What changed

- `models/NWCConnection.ts`: new `pendingSpendSats` computed that sums `pending` `pay_invoice` activity amounts. `canSpend` now checks `totalSpendSats + pendingSpendSats + amountSats <= maxAmountSats`, and `remainingBudget` (which feeds the budget-rejection error message and the connection list/details views) subtracts pending as well.
- `models/NWCConnection.test.ts`: regression tests covering the bypass scenario, the pending-to-settled handoff (no double-count), exclusion of settled/failed/expired/make_invoice activity, and unlimited connections.

No double-count is possible across the pending-to-settled transition: `reconcilePendingPayInvoiceActivities` flips the activity status and calls `trackSpending` inside a single `runInAction` (with the status guard from b65f690b8), so an amount moves atomically from the pending bucket to the settled one.

Known trade-off: a pending record that never reconciles (payment lost node-side and never appearing in the payments list) holds its amount against the budget until it resolves or the budget renews. That is the conservative direction for a spending limit; the failure mode of the old behavior was overspending.

Related: `#4303` (serializing pay_invoice handling) narrows how quickly concurrent requests can stack up pending payments, but does not close this bypass on its own since serialized in-transit payments still return before settling. The two fixes are independent.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### DIFF
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -212,9 +212,30 @@ export default class NWCConnection extends BaseModel {
         );
     }
 
+    /**
+     * Sum of in-flight (pending) pay_invoice amounts on this connection's
+     * activity log. Counted against the budget so concurrent or
+     * slow-settling (hodl) payments cannot exceed the limit while their
+     * debits are deferred to reconcile. Reconcile flips an activity to
+     * success and calls trackSpending in one atomic action, so an amount
+     * is never counted in both this and totalSpendSats.
+     */
+    @computed public get pendingSpendSats(): number {
+        if (!this.activity?.length) return 0;
+        return this.activity.reduce((sum, a) => {
+            if (a.type !== 'pay_invoice' || a.status !== 'pending') {
+                return sum;
+            }
+            return sum + Math.max(0, Math.floor(Number(a.satAmount) || 0));
+        }, 0);
+    }
+
     @computed public get remainingBudget(): number {
         if (!this.hasBudgetLimit) return Infinity;
-        return Math.max(0, this.maxAmountSats! - this.totalSpendSats);
+        return Math.max(
+            0,
+            this.maxAmountSats! - this.totalSpendSats - this.pendingSpendSats
+        );
     }
 
     @computed public get budgetUsagePercentage(): number {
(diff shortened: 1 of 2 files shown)


models/NWCConnection.ts
ajaysehwal, line 1: @kaloudis  I think the budget UI should include pending payments as well.
`remainingBudget` already includes pending amounts, but `budgetLimitReached` and `budgetUsagePercentage` still only use `totalSpendSats`. Because of this, the UI can show that there is budget remaining even when pending payments have already used the available budget. Maybe we should include pending amounts in both `budgetLimitReached` and `budgetUsagePercentage` so the UI stays consistent with `remainingBudget`.

kaloudis: Good catch. Fixed in 9ccbab70a: added an `effectiveSpendSats` computed (`totalSpendSats + pendingSpendSats`) and all budget-facing values (`canSpend`, `remainingBudget`, `budgetUsagePercentage`, `budgetLimitReached`) now derive from it, so the progress bar and warning tag stay consistent with the remaining-budget figure.

One wrinkle this surfaced: the `BudgetLimitReached` warning was sticky (only removed via the user-facing `clearWarnings` path), and pending payments can fail and free budget back up. `checkAndResetBudgetIfNeeded` now removes the warning when the limit is no longer reached, mirroring the add/remove symmetry of the wallet-balance warning. Added tests for getter consistency with pending activity and for the pending-payment-fails case.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

Issue #3065, opened 2025-05-26
Lightning payment successful, but app says “timeout”


pailakapo:
### Describe the bug

IOS app, strike invoice with no amount copied to clipboard.  Pasted into Zeus on home screen->lightning->send then edit amount and “pay this invoice”.  (Side note: when editing this amount, the number keyboard pops up with no obvious way to clear it and actually send the invoice.  The user must tap somewhere other than the keypad to remove the keyboard)

The result was payment failed (timeout).  But on transactions screen it shows successful.  (LND node, self hosted, not built in to app). And in strike the payment arrived.

### Reproduce

Reproduce the bug step by step.

### ZEUS version

v0.10.2

### Node interface

LND (REST)

### Network

Clearnet

### Device

iPhone 13pro

### Device operating system

IOS 18.4.1

### Log output

```shell

```

kaloudis:
This often happens due to the web server in front of the REST API having timeouts set. It's hard to work around. Though we have tried to make it clear when possible that a Timeout could still mean the payment is in flight.

Related:
https://github.com/ZeusLN/zeus/issues/2761

Mstone87:
> ### Describe the bug
> IOS app, strike invoice with no amount copied to clipboard. Pasted into Zeus on home screen->lightning->send then edit amount and “pay this invoice”. (Side note: when editing this amount, the number keyboard pops up with no obvious way to clear it and actually send the invoice. The user must tap somewhere other than the keypad to remove the keyboard)
> 
> The result was payment failed (timeout). But on transactions screen it shows successful. (LND node, self hosted, not built in to app). And in strike the payment arrived.
> 
> ### Reproduce
> Reproduce the bug step by step.
> 
> ### ZEUS version
> v0.10.2
> 
> ### Node interface
> LND (REST)
> 
> ### Network
> Clearnet
> 
> ### Device
>Tcl d60
> 
> ### Device operating system
>Android13
> 
> ### Log output

kaloudis:
We've made some improvements here in v13.2.0. v13.2.0-beta1 is now available for testing https://github.com/ZeusLN/zeus/releases/tag/v13.2.0-beta1

kaloudis:
ZEUS v13.2.0 is now available https://github.com/ZeusLN/zeus/releases/tag/v13.2.0

kaloudis:
Closing this for now. Please re-open if the issues returns.

Issue #2761, opened 2025-01-22
Zeus prematurely declares payment failures


chrisguida:
### Describe the bug

I've seen this happen many times kind of randomly, but I've been seeing it a lot recently, both for my own wallet and on friends' wallets.

Basically you send a payment, Zeus comes back and says "payment failed", and then the payment succeeds sometime later. It seems that Zeus is prematurely determining that the payment has failed, when it hasn't.

I've seen this while sending from:
- LNDHub with Lnbits on CLN
- LNDHub with Lnbits on LND
- Zeus directly to CLNRest/xpay on CLN

And sending to:
- LNDHub with Lnbits on CLN
- LNDHub with Lnbits on LND
- Zeus directly to CLNRest/xpay on CLN
- Phoenix

### Reproduce

Happens intermittently, not really sure how to reproduce, but it happens like 1 out of 10 times, especially on invoices that take a bit longer to route.

### ZEUS version

v0.9.4

### Node interface

LndHub

### Network

Clearnet

### Device

Android Pixel 8a

### Device operating system

Android 15

### Log output

```shell

```

Joanfxx:
> ### Describe the bug
> 
> I've seen this happen many times kind of randomly, but I've been seeing it a lot recently, both for my own wallet and on friends' wallets.
> 
> Basically you send a payment, Zeus comes back and says "payment failed", and then the payment succeeds sometime later. It seems that Zeus is prematurely determining that the payment has failed, when it hasn't.
> 
> I've seen this while sending from:
> 
> * LNDHub with Lnbits on CLN
> * LNDHub with Lnbits on LND
> * Zeus directly to CLNRest/xpay on CLN
> 
> And sending to:
> 
> * LNDHub with Lnbits on CLN
> * LNDHub with Lnbits on LND
> * Zeus directly to CLNRest/xpay on CLN
> * Phoenix
> 
> ### Reproduce
> 
> Happens intermittently, not really sure how to reproduce, but it happens like 1 out of 10 times, especially on invoices that take a bit longer to route.
> 
> ### ZEUS version
> 
> v0.9.4
> 
> ### Node interface
> 
> LndHub
> 
> ### Network
> 
> Clearnet
> 
> ### Device
> 
> Android Pixel 8a
> 
> ### Device operating system
> 
> Android 15
> 
> ### Log output

@chrisguida 

Hello, we’re here to help
Connect to the support page and use the chat icon on the page to initiate a chat with the team on there via;
[...]

PR #4306, opened 2026-08-01
fix: payment timeout races on LND REST, CLNRest, LndHub, and Tor

kaloudis:
# Description

Gives Lightning payment REST requests timeout headroom past the node-side payment deadline, so slow payments surface as the designed "Payment timed out. It may be in transit... Check the Activity view" message instead of a generic transport error that invites a retry. Retrying a payment that later settles is a double-send on keysend, where each attempt generates a fresh preimage and defeats node-side same-hash dedup.

Related issues:

- #3065 is this bug on LND (REST) clearnet verbatim: slow payment reported as timeout, later shown successful in Activity, funds received. The extra headroom here lets LND's real answer land in many of those cases, and the remainder now get the in-transit message instead of a failure claim. Full truthfulness would additionally need post-timeout reconciliation against `listPayments`, which is out of scope here.
- #2761 is the same premature-failure family across several interfaces, and both reported legs are covered here: the CLNRest leg (client timeout equal to `retry_for` discarded definitive responses) and the LndHub leg (`/payinvoice` failing at the inherited 30s transport default on slow LNbits/LndHub routes).

## The bug

**LND (REST):** `payLightningInvoice` races the REST call against a `forcedTimeout` that resolves the designed `{payment_error: paymentTimedOut}` shape at `timeout_seconds + 1`. But the REST call's own client timeout rejects with a generic `Error('Request timeout')` at exactly `timeout_seconds`, one second earlier, so the designed path was unreachable over clearnet. The fetch is never aborted, so the payment can still settle node-side after the UI reports failure.

**CLN (CLNRest):** `/v1/pay` and `/v1/keysend` block server-side for up to `retry_for` seconds, but the client timeout was exactly `retry_for * 1000` with zero headroom, so a definitive response arriving at the deadline was discarded as a generic timeout. CLN also had no in-transit message path at all.

**LndHub:** servers (LNbits included) block on `/payinvoice` until the payment resolves, but `payLightningInvoice` passed no timeout and inherited the 30s `restReq` default. Payments routing slower than 30s were reported as `Request timeout` and settled afterward (the primary interface reported in #2761). Note that unlike LND (`timeout_seconds`) and CLN (`retry_for`), the LndHub protocol has no server-side payment deadline parameter, and Zeus's payment timeout setting is neither plumbed to this backend nor exposed in its UI, so any window here is client-imposed.

**Tor:** since the react-native-nitro-tor swap, `doTorRequest` imposed a fixed 60s native timeout on every request. This reintroduced the same race over Tor for the default 60s payment timeout, and silently capped longer payment windows at 60s regardless of settings (the NWC service's 120s payment window included), losing success responses that arrive after 60s.

**Native transport (clearnet):** `ReactNativeBlobUtil` applies its own native default timeout of 60s when none is set in `config()` (OkHttp connect/read on Android, `timeoutIntervalForRequest` on iOS), and `restReq` only raced a JS-level timer. A payment holding the connection silent past 60s died natively no matter what the JS race allowed, which would beat the 61s in-transit shape at the default setting and capped longer windows at 60s on clearnet just as the Tor bug did.

## What changed

- `backends/LND.ts`: the payment REST call now gets `(timeout_seconds + 5) * 1000`, so the `forcedTimeout` shape at `+1s` reliably wins whenever the node has not answered; real responses before that still win the race. `timeout_seconds` is normalized locally (`Number(...) || 60`) as a NaN guard. `restReq` passes its `timeout` parameter through the Tor branch.
- `backends/CLNRest.ts`: `payLightningInvoice` and `sendKeysend` share a `payWithTimeout` helper mirroring the LND pattern: client timeout at `retry_for + 5` seconds, raced against the designed payment-timed-out shape at `+1s` (`handlePayment` already handles the `payment_error` field backend-agnostically). `restReq` passes its `timeout` through the Tor branch.
- `backends/LndHub.ts`: `payLightningInvoice` uses a fixed 60s payment window (65s client timeout) with the same raced payment-timed-out shape, replacing the silent 30s transport ceiling. Fixed rather than configurable because the payment timeout setting is not exposed for LndHub.
- `backends/LND.ts` / `backends/CLNRest.ts`: `restReq` passes its timeout into `ReactNativeBlobUtil.config()` so the native transport cap matches the JS race instead of silently enforcing 60s.
- `utils/TorUtils.ts`: `doTorRequest` accepts an optional per-request `timeoutMs`, defaulting to the existing 60s `REQUEST_TIMEOUT_MS`. Only the two `restReq` Tor branches pass an override, and only when a caller supplied an explicit timeout (payments); every other Tor call site keeps the 60s default.

`EmbeddedLND` and `LightningNodeConnect` override `payLightningInvoice`, so the LND change affects only remote LND (REST); LndHub's override is updated as described above.

## Behavior after the fix

| Path | Before | After |
|---|---|---|
| LND clearnet, no answer at deadline | generic "Request timeout" at 60s | designed in-transit message at 61s |
| LND Tor, default 60s timeout | generic Tor timeout at 60s | designed in-transit message at 61s |
| LND/NWC Tor, 120s timeout | request killed at 60s, late success lost | full 125s window |
| LND/NWC clearnet, 120s timeout | connection killed natively at ~60s of silence | full 125s window |
| CLN, response at the retry_for deadline | discarded as "Request timeout" | definitive result delivered |
| CLN, no answer at deadline | generic "Request timeout" | designed in-transit message |
| LndHub, payment routing longer than 30s | generic "Request timeout" at 30s | full 65s window |
| LndHub, no answer at 60s | generic "Request timeout" at 30s | designed in-transit message at 61s |

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -99,7 +99,11 @@ export default class CLNRest {
             });
 
             const fetchPromise = ReactNativeBlobUtil.config({
-                trusty: !certVerification
+                trusty: !certVerification,
+                // RNBlobUtil's native default is 60s; without this a
+                // payment request holding the connection open longer than
+                // that dies natively no matter what the race below allows
+                timeout: timeout || this.defaultTimeout
             })
                 .fetch(method, url, headers, data ? JSON.stringify(data) : data)
                 .then((response: any) => {
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -81,7 +81,10 @@ export default class CLNRest {
                     // self-signed certs can't match anyway).
                     // Clearnet-over-Tor: keep strict TLS because exit
                     // nodes can MITM.
-                    isOnionHttpsUrl(url)
+                    isOnionHttpsUrl(url),
+                    // undefined keeps doTorRequest's own default; only
+                    // request-scoped deadlines (payments) override it
+                    timeout
                 ).then((response: any) => {
                     calls.delete(id);
                     return response;
@@ -435,6 +438,45 @@ export default class CLNRest {
         return formatted;
     };
 
+    // cln blocks on /v1/pay, /v1/xpay, /v1/keysend, and /v1/xkeysend for
+    // up to retry_for seconds, so the request timeout needs headroom past
+    // that deadline or the definitive response is discarded as a generic
+    // transport timeout. Mirror LND's payLightningInvoice: race the call
+    // against the payment-timed-out shape so a slow payment surfaces as
+    // "may be in transit, check Activity" instead of a retryable-looking
+    // error (dangerous on keysend, where a retry double-sends).
+    // formatResult runs on the real response only, inside the race, so
+    // the timed-out shape is never marked status complete.
+    private payWithTimeout = (
+        route: string,
+        request: any,
+        data: any,
+        formatResult?: (result: any) => any
+    ) => {
+        const timeoutSeconds = Number(data.timeout_seconds) || 60;
+
+        const forcedTimeout = async (time_ms: number, response: any) => {
+            await new Promise((res) => setTimeout(res, time_ms));
+            return response;
+        };
+
+        let call = this.postRequest(
+            route,
+            request,
+            (timeoutSeconds + 5) * 1000
+        );
+        if (formatResult) call = call.then(formatResult);
+
+        return Promise.race([
+            forcedTimeout((timeoutSeconds + 1) * 1000, {
+                payment_error: localeString(
+                    'views.SendingLightning.paymentTimedOut'
+                )
+            }),
+            call
+        ]);
+    };
+
     payLightningInvoice = async (data: any) => {
         // pay is deprecated in CLN v26.06 and removed in v27.03; use xpay
         // (available since v24.11) when the node has it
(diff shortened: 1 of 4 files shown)


kaloudis:
Rebased on master to resolve conflicts.

Notable resolution: master picked up the CLN xpay migration (#4390), which split `payLightningInvoice`/`sendKeysend` into legacy (`pay`/`keysend`) and `xpay`/`xkeysend` paths. `payWithTimeout` now wraps all four paths. Since `formatXpayResult` synthesizes `status: 'complete'`, it is applied inside the race on the real response only, so the forced payment-timed-out shape can never be marked complete. Also kept master's rejected-call cache eviction in LND `restReq` alongside the new timeout pass-through.

`yarn verify` green locally (54 suites / 1149 tests, lint, prettier, tsc).

shubhamkmr04 (approved):
tACK

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4308, opened 2026-08-01
fix(nwc): await signature so sign_message returns the right one

kaloudis:
# Description

Fixes the NWC `sign_message` handler returning the wrong signature.

`handleSignMessage` called `messageSignStore.signMessage(request.message)` and then synchronously read `messageSignStore.signature` on the next line. But `signMessage` is fire-and-forget: it dispatches the async backend call and sets the `signature` observable later, inside a `.then()`, returning nothing. So the value read back is whatever the *previous* `sign_message` request left in the store, an empty string on the first request, or a valid signature over a *different* message on later ones.

The response then pairs `message: request.message` with that stale signature, so a verifier checking the signature against the stated message fails, or the client attributes a signature over message A to message B. This undermines every signed NWC interaction, not just an occasional one.

## What changed

- `stores/MessageSignStore.ts`: `signMessage` now returns `Promise<string | null>` that resolves to this request's signature (or `null` on failure). The `signature` observable is still updated, so the reactive `SignVerifyMessage` view is unchanged.
- `stores/NostrWalletConnectStore.ts`: `handleSignMessage` awaits `signMessage` and returns the resolved value; a `null`/empty result yields a NIP-47 `INTERNAL_ERROR` instead of a bogus signature.

Reading the resolved value rather than the shared observable also removes a secondary race: two concurrent `sign_message` requests would otherwise clobber each other's `signature` observable. `sign_message` is not on the payment serialization queue (#4303), so this matters independently.

## Verification

Verified the only two callers of `MessageSignStore.signMessage` are this handler and `views/Tools/SignVerifyMessage.tsx`; the view is an `@observer` that ignores the return value and reacts to the `signature` observable, so its behavior is unchanged.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### DIFF
--- a/stores/MessageSignStore.ts
+++ b/stores/MessageSignStore.ts
@@ -169,13 +169,19 @@ export default class MessageSignStore {
     // reading the shared `signature` observable, which may still hold a
     // previous request's result. Resolves to null on error; the
     // observable is still updated for the reactive view.
+    // `mode` overrides the shared `signingMode` observable without
+    // mutating it, so background callers don't inherit (or clobber)
+    // the Sign/Verify screen's state.
     @action
-    public signMessage = (text: string): Promise<string | null> => {
+    public signMessage = (
+        text: string,
+        mode?: 'lightning' | 'onchain'
+    ): Promise<string | null> => {
         this.loading = true;
 
         try {
             const signOperation =
-                this.signingMode === 'lightning'
+                (mode ?? this.signingMode) === 'lightning'
                     ? BackendUtils.signMessage(text)
                     : BackendUtils.signMessageWithAddr(
                           text,
--- a/stores/MessageSignStore.ts
+++ b/stores/MessageSignStore.ts
@@ -164,8 +164,13 @@ export default class MessageSignStore {
             });
     };
 
+    // Resolves to the signature so awaiting callers (e.g. the NWC
+    // sign_message handler) get THIS request's signature instead of
+    // reading the shared `signature` observable, which may still hold a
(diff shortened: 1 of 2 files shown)


ajaysehwal (commented):
@kaloudis One small thing:  `handleSignMessage` calls `signMessage()` without first setting the signing mode to `lightning`.
if the user previously used the Sign/Verify screen in **on-chain** mode, the NWC request could end up using `signMessageWithAddr` with the currently selected address instead of the node identity.
I think it would be safer to explicitly set the signing mode before signing, e.g.:

```ts
this.messageSignStore.setSigningMode('lightning');
const signature = await this.messageSignStore.signMessage(request.message);

kaloudis:
> ` [e66285a](/ZeusLN/zeus/pull/4308/commits/e66285a75414c0ae60d39d8fb76ce83597f1174f)`

Addressed in [e66285a](https://github.com/ZeusLN/zeus/pull/4308/commits/e66285a75414c0ae60d39d8fb76ce83597f1174f)

ajaysehwal (approved):
tACK

PR #4309, opened 2026-08-01
fix(swaps): verify reverse-swap invoice before it can be paid

kaloudis:
# Description

Verifies the swap host's invoice before a reverse swap can pay it.

In a reverse swap, ZEUS chooses the preimage and sends only its hash (`preimageHash`) to the swap host. The host must return a hold invoice whose payment hash equals that hash, for the requested amount. `createReverseSwap` stored and displayed the host's `invoice` verbatim with no verification, so a malicious or compromised host (custom swap hosts are supported via `settings.swaps.customHost`) could return an invoice paying itself with an unrelated payment hash. The user pays that invoice over Lightning, but the on-chain lockup ZEUS can claim with its preimage is never bound to that payment, so the Lightning funds are lost with nothing received on-chain: swap atomicity broken client-side.

## What changed

- `utils/SwapUtils.ts`: new pure `verifyReverseSwapInvoice(invoice, expectedPaymentHash, expectedAmountSats)` that decodes the invoice and asserts its `payment_hash` equals the expected hash (`sha256(preimage)`) and its amount equals the requested amount. Returns `{ valid, reason }`.
- `stores/SwapStore.ts` (`createReverseSwap`): after parsing the response and before saving/displaying/paying, run the check against `sha256(preimage)` and the requested `invoiceAmount`. On failure, set a user-facing error and abort without proceeding.
- `locales/en.json`: `views.Swaps.invalidReverseSwapInvoice` error string.
- `utils/SwapUtils.test.ts`: unit tests for the matching case, hash mismatch (the malicious-host scenario), amount mismatch, casing tolerance, and undecodable invoice, using a known regtest BOLT11 vector.

Note: an equivalent `validatePreimage` check already exists in `SwapDetails` but only on the claim path (validating the claim-details preimage against the invoice hash); it does not cover the invoice the user pays at swap creation, which is the gap this closes.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] Embedded LND
- [ ] LDK Node

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -2429,6 +2429,7 @@
     "views.Swaps.missingAmount": "Please enter an amount",
     "views.Swaps.generateInvoice": "Generate invoice",
     "views.Swaps.generateInvoiceFailed": "Failed to generate invoice",
+    "views.Swaps.invalidReverseSwapInvoice": "The swap provider returned an invoice that does not match the requested swap. For your safety, this swap has been aborted and no payment was made.",
     "views.Swaps.generateOnchainAddress": "Generate on-chain address",
     "views.Swaps.generateOnchainAddressFailed": "Failed to generate on-chain address",
     "views.Swaps.noUpdates": "No updates found!",
--- a/locales/en.json
+++ b/locales/en.json
@@ -2420,6 +2420,7 @@
     "views.Swaps.missingAmount": "Please enter an amount",
     "views.Swaps.generateInvoice": "Generate invoice",
     "views.Swaps.generateInvoiceFailed": "Failed to generate invoice",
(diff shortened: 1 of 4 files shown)


shubhamkmr04 (approved):
tACK

PR #4312, opened 2026-08-03
refactor: move Receive payment watchers into backend classes

kaloudis:
# Description

Stacked on #4300; that PR's two commits appear in this diff until it merges. Suggested by @ajaysehwal's review there, and fulfills the old `// TODO rewrite subscription logic, starting on Receive view` in `EmbeddedLND.ts`.

The Receive view's `subscribeInvoice` was a ~370 line method with six `implementation === '...'` branches mixing four different watch mechanisms (native stream, LNC event emitter, LDK event loop, REST polling) plus on-chain detection. This moves each backend's watcher into its backend class behind a uniform subscription API, so the view no longer branches on implementation strings.

## New API

Two methods on the backend classes, dispatched through `BackendUtils`, each returning an unsubscribe function:

- `watchInvoicePaid({ rHash, value }, onPaid)`
  - LND (REST) and CLNRest poll `lookupInvoice` every 5s (LND converts the base64url `rHash` to the hex its REST endpoint expects)
  - LndHub polls its invoice list scan
  - Embedded LND wraps the `SubscribeInvoices` native stream
  - Lightning Node Connect wraps its event emitter, including the `rpc error: code = Canceled` / `EOF` handling
  - LDK Node wraps `subscribeToEvents`
- `watchOnchainReceived({ address, value, numConfPreference, blockHeight }, onReceived)`
  - LND (REST) polls `getTransactions`; embedded LND and LNC wrap their transaction streams
  - LndHub gets an explicit no-op override so it does not silently inherit LND's REST poller (both `EmbeddedLND` and `LndHub` extend `LND`, so inherited watchers are overridden explicitly)

`Receive.tsx` drops 391 lines: the six-branch method becomes ~60 backend-agnostic lines, and the five watcher fields (`listener`, `listenerSecondary`, `lnInterval`, `onChainInterval`, `ldkUnsubscribe`) collapse into two unsubscribe handles. NWC still has no watcher; the view guards the dispatcher's `false` return for backends without one.

## Behavior notes

Behavior-preserving per backend (match conditions, poll intervals, payloads, and quirks like embedded LND's strict `>` confirmation check are unchanged), with three deliberate fixes:

- Unmount cleanup now actually detaches stream listeners: the old code guarded on `.stop()` but emitter subscriptions only expose `.remove()`, so embedded LND and LNC listeners leaked
- Re-subscribing replaces the previous watcher of the same kind instead of stacking intervals/listeners
- Receiving a payment stops both watchers on every backend (previously only the polling backends did this)

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Manual testing across the backend matrix is still pending, hence the draft. Every backend's watcher path was rewritten, so each of the six should get a receive smoke test (LN invoice, and on-chain for embedded LND, LNC, and LND REST) before this leaves draft.

### Locales

### Other:

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -1,4 +1,5 @@
 import { settingsStore, nodeInfoStore } from '../stores/Stores';
+import Invoice from '../models/Invoice';
 import TransactionRequest from '../models/TransactionRequest';
 import OpenChannelRequest from '../models/OpenChannelRequest';
 import VersionUtils from '../utils/VersionUtils';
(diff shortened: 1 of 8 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

Issue #4208, opened 2026-07-03
LDK Node: add support for 24 word seed phrases


kaloudis:
By default, we should stay with 12 word seed phrases for ease of use, but advanced users should have the ability to create and restore wallets with 24 seed words.

PR #4313, opened 2026-08-03
deps: consolidate nostr stack on nostr-tools 2.23.3

kaloudis:
# Description

The bundle was shipping three copies of the nostr tooling: `nostr-tools` 1.16.0 (superseded 1.x line with exact pins on old crypto: `@noble/curves` 1.1.0, `@noble/hashes` 1.3.1, `@scure/bip32` 1.3.1, `@scure/bip39` 1.2.1, `@noble/ciphers` 0.2.x), the `@nostr/tools` JSR alias at 2.17.0 (added in #4208-era to work around 1.16.0's broken NIP-04 in RN, but itself pinning curves 1.2.0 and hashes 1.3.1), and `nostr-tools` 2.23.3 via `@getalby/sdk`. Spend-capable key material (NWC wallet service keys, Zaplocker keys) was generated and signed by the frozen 1.x line, which would not pick up fixes from a routine upgrade.

This PR consolidates everything on a single `nostr-tools` 2.23.3, shared with `@getalby/sdk`. The 1.16.0 pin and the JSR alias are removed, and all of their frozen transitive crypto pins drop out of `yarn.lock`.

Changes:
* Migrate 16 source files to the nostr-tools 2.x API: `generatePrivateKey` to `generateSecretKey`, `finishEvent`/`getEventHash`+`getSignature` to `finalizeEvent`, `verifySignature` to `verifyEvent`, `relayInit` to the `Relay` class or `SimplePool` (`querySync`/`get`/`publish`), `nip44` via the nostr-tools root export
* All persisted keys remain hex strings (NWC service keys, Zaplocker `nostrPrivateKey`, connection secrets); conversion happens at call boundaries, so existing users need no migration
* The hoisted `@noble/curves` moves from 1.1.0 to 2.0.1, which accepts `Uint8Array` only; the Zaplocker schnorr sign/verify call sites are converted accordingly, `secp256k1` imports gain the `.js` suffix required by the curves 2.x exports map, and `@noble/curves` + `@noble/hashes` are pinned as direct dependencies so app code no longer depends on hoisting order
* The hoisted `@scure/bip32` similarly moves from 1.3.1 to 2.0.1 (the sole remaining copy, pulled in by nostr-tools 2.23.3) and is now pinned as a direct dependency, since `utils/VssAuthUtils.ts` and `stores/SwapStore.ts` import it directly. Derivation verified identical between 1.3.1 and 2.0.1 on every path in use: the master xpub, m/130'/0' (VSS auth key), and m/44/0/0/0/{i} (swap refund keys)
* No jest `moduleNameMapper` entries are needed: jest resolves the noble and scure subpaths natively on this tree. An earlier revision added them, but the mapper broke `@scure/bip39` by redirecting its `@noble/hashes` imports to the 2.x copy nested under nostr-tools

Flows that exercise the rewritten relay layer and deserve manual attention: Zaplocker attestation broadcast/lookup, NWC connection creation and relay ping, CLINK noffer payments, Cashu mint backup/restore over Nostr, Nostr contact import.

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Existing tests for `NostrUtils`, `NostrMintBackup`, `NostrConnectUtils`, `ClinkUtils`, and `ContactUtils` were updated for the 2.x API and pass (1094 tests, 52 suites).

- [ ] Android
- [ ] iOS

Manual on-device testing still pending; will update these boxes after runs on both platforms.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Backend testing pending, with priority on Nostr Wallet Connect (wallet service keys) and Zaplocker/ZEUS Pay flows.

### Locales

`package.json` and `yarn.lock` verified consistent (`yarn install --frozen-lockfile` compatible); JS-only change, no new native modules; `pod install` verified unchanged on iOS.

### Other:

### DIFF
--- a/views/NostrContacts.tsx
+++ b/views/NostrContacts.tsx
@@ -12,7 +12,7 @@ import {
 import { inject, observer } from 'mobx-react';
 import { CheckBox } from '@rneui/themed';
 // @ts-ignore:next-line
-import { relayInit, nip05, nip19 } from 'nostr-tools';
+import { SimplePool, nip05, nip19 } from 'nostr-tools';
 import { NativeStackNavigationProp } from '@react-navigation/native-stack';
 
 import { SharedText } from '../components/SharedTransition';
@@ -253,7 +253,7 @@ export default class NostrContacts extends React.Component<
         this.setState({ loading: true, error: '' });
         const { account } = this.state;
 
-        let pubkey: string;
+        let pubkey: string = '';
         if (this.state.isValidNpub) {
             const decoded = nip19.decode(account);
             pubkey = decoded.data.toString();
@@ -270,62 +270,51 @@ export default class NostrContacts extends React.Component<
             }
         }
 
-        const profilesEventsPromises = DEFAULT_NOSTR_RELAYS.map(
-            async (relayItem) => {
-                try {
-                    const relay = relayInit(relayItem);
-                    const tags: Array<string> = [];
+        if (!pubkey) {
+            this.setState({ loading: false });
+            return;
+        }
 
-                    relay.on('connect', () => {
-                        console.log(`connected to ${relay.url}`);
-                    });
+        const pool = new SimplePool();
 
-                    relay.on('error', (): void => {
-                        console.log(`failed to connect to ${relay.url}`);
-                    });
+        Promise.resolve()
+            .then(async () => {
+                const contactEvents = await pool.querySync(
+                    DEFAULT_NOSTR_RELAYS,
+                    {
+                        authors: [pubkey],
+                        kinds: [3]
+                    }
+                );
 
-                    await relay.connect();
-                    let eventReceived = await relay.list([
-                        {
-                            authors: [pubkey],
-                            kinds: [3]
-                        }
-                    ]);
+                let latestContactEvent: any;
 
-                    let latestContactEvent: any;
+                contactEvents.forEach((content: any) => {
+                    if (
+                        !latestContactEvent ||
+                        content.created_at > latestContactEvent.created_at
+                    ) {
+                        latestContactEvent = content;
+                    }
+                });
 
-                    eventReceived.forEach((content: any) => {
-                        if (
-                            !latestContactEvent ||
-                            content.created_at > latestContactEvent.created_at
-                        ) {
-                            latestContactEvent = content;
-                        }
-                    });
-
-                    if (!latestContactEvent) return;
-
+                const tags: Array<string> = [];
+                if (latestContactEvent) {
                     latestContactEvent.tags.forEach((tag: string) => {
                         if (tag[0] === 'p') {
                             tags.push(tag[1]);
                         }
                     });
+                }
 
-                    return relay.list([
-                        {
-                            authors: tags,
-                            kinds: [0]
-                        }
-                    ]);
-                } catch (e) {}
-            }
-        );
+                if (tags.length === 0) return [];
 
-        Promise.all(profilesEventsPromises)
-            .then((profilesEventsArrays) => {
-                const profileEvents = profilesEventsArrays
-                    .flat()
-                    .filter((event) => event !== undefined);
+                return pool.querySync(DEFAULT_NOSTR_RELAYS, {
+                    authors: tags,
+                    kinds: [0]
+                });
+            })
+            .then((profileEvents) => {
                 const newContactDataIndexByName: Record<string, NostrContact> =
                     {};
                 const newContactDataIndexByPubkey: Record<
@@ -379,6 +368,9 @@ export default class NostrContacts extends React.Component<
             })
             .catch((error) => {
                 console.error('Error fetching profiles events:', error);
+            })
+            .finally(() => {
+                pool.close(DEFAULT_NOSTR_RELAYS);
             });
     }
 
--- a/views/NostrContacts.tsx
+++ b/views/NostrContacts.tsx
@@ -27,6 +27,7 @@ import { Row } from '../components/layout/Row';
 import AddressUtils from '../utils/AddressUtils';
 import ContactUtils from '../utils/ContactUtils';
 import { localeString } from '../utils/LocaleUtils';
+import NostrUtils from '../utils/NostrUtils';
 import { themeColor } from '../utils/ThemeUtils';
 
 import Storage from '../storage';
@@ -254,8 +255,7 @@ export default class NostrContacts extends React.Component<
 
         let pubkey: string = '';
         if (this.state.isValidNpub) {
-            const decoded = nip19.decode(account);
-            pubkey = decoded.data.toString();
+            pubkey = NostrUtils.npubToHex(account) || '';
         } else if (this.state.isValidNip05) {
             try {
                 const lookup: any = await nip05.queryProfile(account);
--- a/views/NostrContacts.tsx
+++ b/views/NostrContacts.tsx
@@ -11,7 +11,6 @@ import {
 } from 'react-native';
 import { inject, observer } from 'mobx-react';
 import { CheckBox } from '@rneui/themed';
-// @ts-ignore:next-line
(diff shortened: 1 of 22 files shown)


views/NostrContacts.tsx
```diff
@@ -12,7 +12,7 @@ import {
 import { inject, observer } from 'mobx-react';
 import { CheckBox } from '@rneui/themed';
 // @ts-ignore:next-line
```
myxmaster: we don't need the `@ts-ignore` here anymore, nostr-tools 2.x types this import fine.

same here: `views/ClinkPay/ClinkPay.tsx:7`

myxmaster (commented):
1. The PR description says that jest `moduleNameMapper` entries were added, but that is not true (the commit message explains why).

2. This PR moves the hoisted `@scure/bip32` from 1.3.1 to 2.0.1: both requesters of 1.3.1 (`nostr-tools` 1.16.0 and the `@nostr/tools` JSR alias) are removed here, leaving 2.0.1 from `nostr-tools` 2.23.3 as the only copy. That affects key derivation: `VssAuthUtils.ts:51` (VSS auth key) and `SwapStore.ts:727` (Boltz refund keys). I let Claude Opus 5 check both versions against the same seed, on the paths the code actually uses: `getXpub`'s master xpub, `m/130'/0'` (VSS), and `m/44/0/0/0/{i}` (swap refund keys). All identical. (Out of scope here, but we should add tests for `SwapStore`.)
-> `@scure/bip32` should be declared/pinned in `package.json`.

3. The shim at the top of `utils/VssAuthUtils.test.ts` is dead code now. It compensated for the jest `moduleNameMapper` that #4321 removed, and `@scure/bip32` 2.0.1 doesn't import `_assert` anyway. Tests pass without it.

kaloudis:
@myxmaster very thorough pass, thank you

Everything should be addressed in https://github.com/ZeusLN/zeus/pull/4313/commits/ae005caeaf20f33a582982033ed7cb4c31260ee0

myxmaster (approved):
ACK

not tested: Zaplocker and CLINK

myxmaster:
Since https://github.com/ZeusLN/zeus/pull/4371 will probably land before this here, heads-up for the rebase, that will be needed:

https://github.com/ZeusLN/zeus/pull/4371 adds a replacement mock for `nostr-tools` in `utils/handleAnything.test.ts` that defines `relayInit`, `nip05` and `nip19` and nothing else, and its npub regression test asserts `expect(mockRelayInit).not.toHaveBeenCalled()`.

After this PR here moves `nostrProfileLookup` from `relayInit` to `SimplePool`, two things happen to that test:
- the `mockRelayInit` assertion becomes vacuous -> production code no longer calls `relayInit`, so it holds regardless of whether the guard is there. The test still protects via its `mockNip19Decode` assertion, so it does not go blind, but one of its three checks stops carrying weight.
- the mock no longer covers what the code calls. If a probe ever did reach `nostrProfileLookup`, the test would fail with "SimplePool is not a constructor" rather than a readable assertion.

Suggested rebase fix:
Replace the `relayInit` entry with a `SimplePool` stub and re-point that assertion at the constructor (or its `querySync`). Worth checking by hand afterwards (drop the guard locally and confirm the test fails for the reason you expect).

kaloudis:
Thanks for the heads-up. Rebased on master now that #4371 landed, and applied your suggested fix in 71c2c5682:

- the mock's `relayInit` entry is replaced with a `SimplePool` stub (constructor spy, plus `querySync` resolving `[]` and `close`, matching what `nostrProfileLookup` calls)
- the npub probe test now asserts `expect(mockSimplePool).not.toHaveBeenCalled()`

Also ran the hand-check you suggested: with the `isClipboardValue` guard removed locally, the test fails because the lookup path genuinely executes (the stub pool is constructed, `querySync` returns no events, and the branch throws its nostrProfileError), not with a "SimplePool is not a constructor" error. Guard restored, full suite (117 tests) green.

myxmaster:
I see `nostrProfileLookup` never actually being entered. Surfacing the error the catch swallows gives `Cannot read properties of undefined (reading 'data')`, so that's `decoded.data`, one line before the lookup call, since `mockNip19Decode` is a bare `jest.fn()`. The constructor spy agrees (`simplePoolCalls=0`), so the stub pool never gets built. Same red, same message, just coming from the branch's own try/catch.

Either way the test does its job. The mechanism just isn't quite the one you'd expect at first glance, and I didn't want to leave that unmentioned.

tACK https://github.com/ZeusLN/zeus/pull/4313/commits/71c2c56823c779312d3abe0f1e4176b2049b82ca

kaloudis:
Good catch, and thanks for tracing it properly. You're right, and my earlier hand-check description was wrong about where the red came from: with `nip19.decode` left as a bare `jest.fn()`, `decoded.data` throws before the lookup call, so the branch's own catch produces the failure and `new SimplePool()` is never reached. The constructor stub was unreachable in the regression scenario.

Fixed in 2045ce91c: the npub probe test now gives `mockNip19Decode` a real decode shape (`{ type: 'npub', data: <hex pubkey> }`). Re-ran the guard-removed hand-check with the fix in place, and the failure now genuinely comes from inside `nostrProfileLookup` (pool constructed, `querySync` resolves `[]`, branch throws its nostrProfileError), so the `mockSimplePool` assertion is live rather than decorative. Guard restored, full suite (117) green.

myxmaster (approved):
tACK 2045ce91c4932d86e7913564f95302240776b1a8

views/NostrContacts.tsx
```diff
@@ -253,7 +252,7 @@ export default class NostrContacts extends React.Component<
         this.setState({ loading: true, error: '' });
         const { account } = this.state;
 
-        let pubkey: string;
+        let pubkey: string = '';
         if (this.state.isValidNpub) {
             const decoded = nip19.decode(account);
             pubkey = decoded.data.toString();
```
ajaysehwal: @kaloudis  In `NostrContacts.tsx` and `handleAnything.ts`, we're using `decoded.data.toString()`. This works for `npub` today because `data` is a hex string, but it can break when `data` is a `Uint8Array`. I think it would be safer to centralize `npub` decoding in `NostrUtils` and use `bytesToHex()` when needed.

kaloudis: Good suggestion, applied in 5a100017b.

One nuance for the record: for npub inputs specifically this could not break. nostr-tools 2.x types `DecodedNpub.data` as `string` (hex); `Uint8Array` only appears for nsec. And both call sites sit behind `isValidNpub` guards, so `decoded.data` was always a hex string and `.toString()` was a no-op. That said, the raw `.toString()` would silently produce garbage ("130,52,31,...") if the surrounding guard ever changed, so centralizing is the right call.

Changes:
- `handleAnything.ts` and `NostrContacts.tsx` now go through `NostrUtils.npubToHex()`, which narrows on `decoded.type === 'npub'` and keeps bech32 decoding in one place (it already existed and handles the nsec/bytes case via `bytesToHex` in its sibling `nsecToHex`)
- the now-unused `nip19` import is dropped from `handleAnything.ts`; `NostrContacts.tsx` keeps it for `npubEncode`

Re-ran the guard-removed hand-check on the clipboard probe test to confirm the regression path still flows through `npubToHex` into the `SimplePool` stub (it does, and the assertions stay live), then restored the guard. Full suite (1108 tests), tsc, lint, prettier all green.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4315, opened 2026-08-03
fix: PeerItem: render stats with KeyValue to avoid Text crash

ajaysehwal:
##  Description
Fixes a React Native crash on the Peers list: `Text strings must be rendered within a <Text> component`.
`PeerItem` used patterns like `{peer.num_channels && (...)}`. When `num_channels` was `0` (or string fields were `""`), RN tried to render that falsy value as a text child of a `View`.

This PR:
- Builds peer stats into an array (only push when a value is present), then renders each row with `KeyValue` so values are always inside `<Text>`
- Skips missing stats (including ping when unavailable) instead of showing `N/A`
- Adds an optional `containerStyle` prop to `KeyValue` so PeerItem can keep its compact spacing without changing default KeyValue padding elsewhere

Simulator Screenshot - iPhone 17 Pro - 2026-08-03 at 19 55 37
Simulator Screenshot - iPhone 17 Pro - 2026-08-03 at 19 55 42

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/components/Channels/PeerItem.tsx
+++ b/components/Channels/PeerItem.tsx
@@ -25,50 +25,60 @@ export function PeerItem({
     showDisconnect
 }: PeerItemProps) {
     const title = displayName === peer.pubkey ? peer.pubkey : displayName;
-    const pingDisplay =
-        peer.ping_time != null && peer.ping_time >= 0
-            ? `${(peer.ping_time / 1000).toFixed(2)} ms`
-            : 'N/A';
 
-    const stats: Array<{ label: string; value: string | number }> = [];
-    if (pingDisplay !== 'N/A') {
+    const stats: Array<{ key: string; label: string; value: string | number }> =
+        [];
+    if (peer.ping_time != null && peer.ping_time >= 0) {
         stats.push({
+            key: 'pingTime',
             label: localeString('views.ChannelsPane.pingTime'),
-            value: pingDisplay
+            value: `${(peer.ping_time / 1000).toFixed(2)} ms`
         });
     }
     if (peer.sats_sent != null) {
-        stats.push({
-            label: localeString('views.ChannelsPane.satsSent'),
-            value: getFormattedAmount(peer.sats_sent, 'sats') ?? ''
-        });
+        const satsSent = getFormattedAmount(peer.sats_sent, 'sats');
+        if (satsSent) {
+            stats.push({
+                key: 'satsSent',
+                label: localeString('views.ChannelsPane.satsSent'),
+                value: satsSent
+            });
+        }
     }
     if (peer.sats_recv != null) {
-        stats.push({
-            label: localeString('views.ChannelsPane.satsRecv'),
-            value: getFormattedAmount(peer.sats_recv, 'sats') ?? ''
-        });
+        const satsRecv = getFormattedAmount(peer.sats_recv, 'sats');
+        if (satsRecv) {
+            stats.push({
+                key: 'satsRecv',
+                label: localeString('views.ChannelsPane.satsRecv'),
+                value: satsRecv
+            });
+        }
     }
     if (peer.num_channels != null) {
         stats.push({
+            key: 'numChannels',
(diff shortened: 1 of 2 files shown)


kaloudis (commented):
Fix looks correct: `{0 && ...}` on `num_channels` and an empty-string `address` were real crash vectors, and the KeyValue consolidation is a nice cleanup. A few nits inline.

One note: KeyValue's `Body` label and default-size value text render larger than the old 14px stat rows, so the card's look changes beyond padding. Before/after screenshots and an Android check would help.

components/Channels/PeerItem.tsx
```diff
@@ -24,13 +25,65 @@ export function PeerItem({
...
     const title = displayName === peer.pubkey ? peer.pubkey : displayName;
-    const satsSent = peer.sats_sent;
-    const satsRecv = peer.sats_recv;
     const pingDisplay =
         peer.ping_time != null && peer.ping_time >= 0
             ? `${(peer.ping_time / 1000).toFixed(2)} ms`
             : 'N/A';
 
+    const stats: Array<{ label: string; value: string | number }> = [];
+    if (pingDisplay !== 'N/A') {
```
kaloudis: nit: comparing against the `'N/A'` sentinel is roundabout now that it's never displayed. Check the condition directly and drop the `'N/A'` branch:
```ts
if (peer.ping_time != null && peer.ping_time >= 0) {
    stats.push({ ..., value: `${(peer.ping_time / 1000).toFixed(2)} ms` });
}
```

components/Channels/PeerItem.tsx
```diff
@@ -24,13 +25,65 @@ export function PeerItem({
...
+        stats.push({
+            label: localeString('views.ChannelsPane.pingTime'),
+            value: pingDisplay
+        });
+    }
+    if (peer.sats_sent != null) {
+        stats.push({
+            label: localeString('views.ChannelsPane.satsSent'),
+            value: getFormattedAmount(peer.sats_sent, 'sats') ?? ''
```
kaloudis: nit: `getFormattedAmount` always returns a string on the `'sats'` path, so `?? ''` is dead code. If it could ever fire, skipping the row beats rendering a blank value. Same below for sats received.

components/Channels/PeerItem.tsx
```diff
@@ -91,184 +144,15 @@ export function PeerItem({
...
-                                    {peer.connected
-                                        ? localeString('general.true')
-                                        : localeString('general.false')}
-                                </Text>
-                            </View>
-                        </View>
-                    )}
+                    {stats.map((stat) => (
+                        <KeyValue
+                            key={stat.label}
```
kaloudis: nit: localized strings as React keys can collide if two locale keys translate to the same string. A stable id per stat (e.g. `{ key: 'pingTime', label, value }`) would be safer.

kaloudis:
Nit (non-blocking): the guards are slightly inconsistent. `num_channels` uses `!= null` so a value of `0` renders, but `bytesSent`/`bytesRecv` use truthy checks, so a numeric `0` would hide the row while the string `"0"` shows `0 B`. Harmless in practice since the backends populate these as strings, but if you want them uniform you could use an explicit check like:

```ts
if (peer.bytesSent != null && peer.bytesSent !== '') {
```

(Plain `!= null` alone is not enough here, since an empty string would then render a blank ` B` value, which is the class of bug this PR is fixing.)

kaloudis (approved):
tACK

PR #4316, opened 2026-08-03
build: verify sha256 of FFI bindings and harden fetch-libraries.sh

kaloudis:
# Description

Supply-chain hardening for the native FFI layer.

`fetch-libraries.sh` pins sha256 hashes for every compiled native library, but the uniffi source bindings for CDK and zeus-cashu-restore were curl'd from raw.githubusercontent.com tag URLs on every install with no checksum and no signature. Git tags are mutable, so anyone with push access to cashubtc/cdk-swift or ZeusLN/zeus-cashu-restore could swap the glue code that handles the wallet mnemonic and seed without tripping any verification. Release builds were exposed too, since build.sh runs yarn install inside Docker.

This PR pins the bindings with sha256 hashes like every other artifact:

1. **New hash fields in fetch-libraries-versions.json**: `iosBindingsSha256` for cdk, `androidBindingsSha256` and `iosBindingsSha256` for zeus-cashu-restore, hashing the files at cdk-swift v0.14.2 (commit `d9959e1c53941533dce45f704193e10a12b2013e`) and zeus-cashu-restore v0.1.0 (commit `5cf5064c2adc6356aa53140519e85f2ac168a53b`). The pinned hashes match the files installs have been downloading, so there is no behavior change. Once a hash is pinned, tag mutability no longer matters: substituted content fails the build.

2. **Binding downloads use the same fail-closed verify-download-reverify structure as the binaries**, which also skips the network fetch when a verified copy is already present instead of re-downloading on every install.

3. **The rest of the script fails closed too:**
   - CDK and zeus-cashu-restore binaries are now verified unconditionally; the `[ -n "$SHA256" ]` branches that skipped verification on an empty hash field are gone
   - every hash field is validated as 64-char hex up front. This matters beyond the removed branches: macOS `sha256sum -c` exits 0 with only a warning on improperly formatted checksum lines, so a blank hash previously skipped verification silently for EVERY artifact, including LND and LDK Node
   - all curl calls now use `-f` so HTTP error pages are never written into target files

When bumping cdk or zeus-cashu-restore, update the binary hashes AND the `*BindingsSha256` fields from the same release in the same commit (comments in fetch-libraries.sh say this).

**Verification performed:**
- fresh run (bindings deleted first): all 11 checksummed artifacts download and report OK, exit 0
- tamper test: appending bytes to a downloaded binding is detected, the file is re-fetched and re-verified against the pin
- wrong-pin test (simulates a re-pointed tag): a valid-format but non-matching hash exits 1 with "CDK Swift bindings checksum failed"
- fail-closed tests: a blanked or missing sha256 field in fetch-libraries-versions.json exits 1 immediately (previously passed silently on macOS)
- `yarn verify` green (1044 tests, tsc, prettier, lint)

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

(App-level testing pending: this change only affects how build inputs are verified; all fetched files are byte-identical to current builds. Device builds on both platforms should still be done before merge per project policy.)

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/.claude/skills/zeus-build-and-env/SKILL.md
+++ b/.claude/skills/zeus-build-and-env/SKILL.md
@@ -83,9 +83,7 @@ Versions pinned at time of writing:
 | zeus-cashu-restore (`ZeusLN/zeus-cashu-restore` release) | `0.1.0` | `android/zeus-restore/zeus-cashu-restore.aar` | `ios/ZeusRestore/zeusRestoreFFI.xcframework` |
 
 Behavior and caveats (all verified by reading the script):
-- **SHA256 enforcement**: embedded-LND and LDK Node downloads hard-fail (`exit 1`) on checksum mismatch, always.
-- **Empty-hash skip caveat**: for `cdk` and `zeus-cashu-restore` ONLY, the checksum check is wrapped in `[ -n "$SHA256" ]` — if the hash field in `fetch-libraries-versions.json` is an empty string, verification is **skipped** and the script just prints the downloaded file's hash. All four hash fields are currently non-empty; never blank one to "fix" a checksum failure.
-- **Unchecked binding-source downloads**: the script also `curl`s uniffi **source** bindings with NO checksum: `ios/CashuDevKit/CashuDevKit.swift` (from cdk-swift tag), `ios/CashuDevKit/zeus_cashu_restore.swift` and `android/app/src/main/java/uniffi/zeus_cashu_restore/zeus_cashu_restore.kt` (from zeus-cashu-restore tag). Known supply-chain gap — labeled open, not planned-fixed.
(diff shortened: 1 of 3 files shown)


shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4321, opened 2026-08-04
fix(swaps): enforce BIP39 checksum on rescue key restore

kaloudis:
# Description

Restoring a submarine swap rescue key (Settings > Swaps > Rescue key > Restore) previously validated only that the 12 entered words exist in the BIP39 wordlist. The checksum was never verified before the phrase was persisted to storage, and the console logs claimed the key was "verified." A single mistyped word that is itself a valid wordlist word, or a word-order transposition, produces an invalid mnemonic with probability ~15/16 yet was accepted silently. All subsequent swap refund keys and reverse swap preimages would then be derived from a seed the user cannot reproduce from their real backup, making swap funds unrecoverable exactly when the rescue key is needed. The `rescue_key.json` file import path had the same gap (word count only), while `SwapStore.getRescuableSwaps` already validated correctly, an inconsistency that confirmed the gap was unintended.

Changes:

- Added `isValidRescueKey()` to `utils/SwapUtils.ts`: enforces exactly 12 words plus the BIP39 checksum, tolerant of irregular whitespace
- The rescue key restore button, the `rescue_key.json` import, and `SwapStore.getRescuableSwaps` now all share this single validator, so the three gates cannot drift apart again
- Invalid phrases are rejected with the existing `views.Settings.SeedRecovery.invalidChecksum` locale string instead of being silently saved; the misleading "Verifying rescue key..." / "Rescue key verified and saved!" logs are removed
- Added unit tests covering valid reference vectors, whitespace tolerance, bad checksum, single-word typo, transposition, non-wordlist words, and wrong word counts (including rejecting valid 24-word seeds, since rescue keys must be exactly 12 words)
- Removed the stale jest `moduleNameMapper` for `@noble/hashes` subpaths from `package.json`. It pinned imports to the hoisted root copy (1.7.2) while `@scure/bip39` resolves its other imports to its nested 1.8.0 copy; the version mix made `validateMnemonic` throw internally and return `false` for every input under jest, valid or not, which would have made these new tests fail. The mapper's original workaround is no longer needed: all 49 suites (1051 tests) pass without it. Metro never used this config, so runtime app behavior is unaffected.

No legitimate backup can be rejected by the new gate: `SwapStore.generateRescueKey` uses `generateMnemonic`, so every genuine ZEUS rescue key has a valid checksum.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,14 +16,7 @@
     "testPathIgnorePatterns": [
       "check-styles.test.ts",
       "zeus_modules"
-    ],
-    "moduleNameMapper": {
-      "^@noble/hashes/_assert$": "<rootDir>/node_modules/@noble/hashes/_assert.js",
-      "^@noble/hashes/pbkdf2$": "<rootDir>/node_modules/@noble/hashes/pbkdf2.js",
-      "^@noble/hashes/sha256$": "<rootDir>/node_modules/@noble/hashes/sha256.js",
-      "^@noble/hashes/sha512$": "<rootDir>/node_modules/@noble/hashes/sha512.js",
-      "^@noble/hashes/utils$": "<rootDir>/node_modules/@noble/hashes/utils.js"
-    }
+    ]
   },
   "scripts": {
     "start": "react-native start",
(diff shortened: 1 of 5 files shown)


shubhamkmr04 (approved):
tACK

PR #4322, opened 2026-08-04
fix(donations): verify LNURL invoice before silent donation payment

kaloudis:
# Description

Fixes a funds-loss vector in the donation flow reported in an internal security review.

When donations are enabled, every successful Lightning payment is followed by an automatic background donation. `loadDonationLnurl` requested an invoice from the donation LNURL endpoint and returned `invoiceData.pr` with no verification, and all three donation flows (`SendingLightning`, `CashuSendingLightning`, `MultimintPayment`) then paid that invoice silently with no confirmation screen. A compromised or malicious LNURL server could therefore return an invoice for an arbitrary amount (up to the full wallet balance) and ZEUS would pay it in the background. The `fee_limit_sat` of 100 only bounds routing fees, not the principal. On the Cashu paths, `payLnInvoiceFromEcash` prefers the decoded invoice amount over the caller-supplied amount, so those paths were equally exposed.

This PR enforces LUD-06 client-side in `loadDonationLnurl`, the single choke point for all three flows. The donation is silently skipped (existing `null` handling in all callers) unless every check passes:

* the decoded invoice amount exactly matches the requested donation amount in msat
* the invoice `description_hash` equals sha256 of the served metadata
* the callback URL stays on the donation domain (lookalike domains like `pay.zeusln.app.evil.com` are rejected)
* the requested amount is a positive integer msat value (checked before any network call; also replaces `parseFloat * 1000` with BigNumber arithmetic)

The live `pay.zeusln.app` endpoint was verified to already satisfy all of these checks: the callback is on the same domain, a request for 21000 msat returned an `lnbc210n` invoice, and its `description_hash` matched sha256 of the metadata byte for byte. That server-issued invoice is used as the happy-path test fixture, so legitimate donations are unaffected.

The interactive LNURL-pay flow is not touched: it is already mitigated by the PaymentRequest confirmation screen.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

7 new rejection tests were added (amount mismatch, no-amount invoice, undecodable invoice, off-domain callback, lookalike callback domain, tampered metadata, missing metadata, invalid requested amounts), plus the happy path against a real invoice issued by pay.zeusln.app.

- [ ] Android
- [ ] iOS

On-device testing pending: the change is confined to a pure utility function exercised before any backend call, and is fully covered by unit tests. Will follow up with a mainnet donation test on both platforms before merge.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/DonationUtils.test.ts
+++ b/utils/DonationUtils.test.ts
@@ -48,25 +48,40 @@ describe('DonationUtils', () => {
     });
 
     describe('loadDonationLnurl', () => {
+        const CALLBACK_URL =
+            'https://pay.zeusln.app/BTC/UILNURL/pay/lnaddress/tips';
+        const DONATION_METADATA =
+            '[["text/identifier","tips@pay.zeusln.app"],["text/plain","Paid to ZEUS (Order ID: )"]]';
+        // Real 21 sat (21000 msat) invoice issued by pay.zeusln.app whose
+        // description_hash is sha256(DONATION_METADATA)
+        const DONATION_INVOICE_21_SATS =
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (approved):
tACK

PR #4323, opened 2026-08-04
fix(sweep): collect UTXOs from all address types of a WIF key

kaloudis:
# Description

The WIF Sweeper (Tools > WIF Sweeper) derives four address candidates from the scanned key (p2pkh, p2sh-p2wpkh, p2wpkh, p2tr), probes each for UTXOs, and stops at the first type that has any. A key that received funds on multiple script types over its lifetime (e.g. legacy p2pkh and native segwit p2wpkh) was only partially swept, and the UI presented the partial sum as the key's full balance. A user who discarded the WIF after the "successful" sweep would permanently lose the funds remaining on the other script types. Dust on an earlier probe type could also hide a large balance on a later one.

This PR fixes that:

- `detectAddressesWithUtxos` now probes all four address types and returns every type holding UTXOs, instead of returning on the first match
- `prepareSweepInputs` builds a single mixed-input PSBT covering all supported types (p2pkh, p2sh-p2wpkh, p2wpkh), so `onChainBalance` reflects the key's full sweepable balance; all inputs share one key, so the existing single-signer finalize path handles the mixed PSBT unchanged
- Taproot (not yet sweepable): if funds exist only on p2tr, the existing `addressTypeNotSupported` error still surfaces; if p2tr funds coexist with sweepable types, the sweep proceeds and a warning shows the unswept taproot amount and tells the user to keep the private key
- Stale state (`onChainBalance`, `utxos`, taproot warning) is reset at the start of each scan so a failed second scan can't display the previous key's balance
- `SweepStore` now calls `bitcoin.initEccLib(ecc)` itself; previously its p2tr address derivation only worked because `utils/AddressUtils.ts` happened to be imported first

Added `stores/SweepStore.test.ts` with mocked esplora responses covering: multi-type UTXO collection, signing and finalizing a mixed p2pkh + p2wpkh sweep into a valid 2-input transaction, the taproot warning and taproot-only error paths, the no-UTXO error, and stale-state reset between scans.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Note: manual on-device testing is still pending. The sweep construction logic is backend-independent (mempool.space esplora), but the destination invoice generation should be exercised against at least one backend on each platform before merge.

### Locales

### Other:

### DIFF
--- a/stores/SweepStore.test.ts
+++ b/stores/SweepStore.test.ts
@@ -1,0 +1,201 @@
+import * as bitcoin from 'bitcoinjs-lib';
+import { toXOnly } from 'bitcoinjs-lib/src/psbt/bip371';
+import { getPublicKey } from '@noble/secp256k1';
+// @ts-ignore:next-line
+import { decode as wifDecode } from 'wif';
+
+import SweepStore from './SweepStore';
+
+jest.mock('./NodeInfoStore', () => ({
+    __esModule: true,
+    default: class {}
+}));
+
+jest.mock('../utils/LocaleUtils', () => ({
+    localeString: (key: string) => key
+}));
+
+jest.mock('./Stores', () => ({
+    nodeInfoStore: {
+        nodeInfo: { isTestNet: false }
+    },
+    settingsStore: {
+        settings: { privacy: {} }
+    }
+}));
+
+jest.mock('react-native', () => ({
+    Linking: {
+        canOpenURL: jest.fn(),
+        openURL: jest.fn()
+    }
+}));
+
+jest.mock('./SettingsStore', () => ({
+    DEFAULT_MEMPOOL_INSTANCE: 'electrs.zeusln.com'
+}));
+
+// Private key 0x...01 (compressed, mainnet)
+const WIF = 'KwDiBf89QgGbjEhKnhXJuH7LrciVrZi3qYjgd9M7rFU73sVHnoWn';
+
+const network = bitcoin.networks.bitcoin;
+const { privateKey } = wifDecode(WIF);
+const publicKey = Buffer.from(getPublicKey(privateKey, true));
+
+const p2pkhAddress = bitcoin.payments.p2pkh({ pubkey: publicKey, network })
+    .address!;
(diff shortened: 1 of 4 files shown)


kaloudis:
cc @a-khushal

stores/SweepStore.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        throw new Error(`unexpected fetch: ${url}`);
+    }) as any;
+}
+
+describe('SweepStore', () => {
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    it('sweeps UTXOs from every address type of the key, not just the first', async () => {
```
a-khushal: this test says every supported address type, but only checks p2pkh + p2wpkh. Can we also include p2sh-p2wpkh? 

also tthe p2sh mock option currently does nothing, so nested segwit mixed input is not tested

stores/SweepStore.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        const store = new SweepStore(mockNodeInfoStore);
+        await store.prepareSweepInputs(WIF);
+        expect(store.onChainBalance).toBe(P2WPKH_VALUE);
+        expect(store.unsweptTaprootSats).toBe(P2TR_VALUE);
+
+        mockEsplora({});
+        await store.prepareSweepInputs(WIF);
+        expect(store.sweepError).toBe(true);
+        expect(store.onChainBalance).toBe(0);
+        expect(store.unsweptTaprootSats).toBe(0);
```
a-khushal: can we also check that `utxos` and `sweptAddressTypes` are empty?

kaloudis:
> /gemini review

Gemini review plugin got discontinued

PR #4324, opened 2026-08-04
fix(receive): use plain decimal amounts in BIP21 URIs

kaloudis:
# Description

When creating a receive request of 1000 BTC or more, the BIP21 URIs behind the unified and on-chain QRs are built with `BigNumber.toFormat()`, which applies the default FORMAT and inserts comma group separators. The QR then encodes:

```
bitcoin:BC1...?amount=1,234.56789
```

BIP21 requires a plain decimal amount. Consequences:

- ZEUS cannot re-scan its own QR: our parser feeds the amount to `new BigNumber('1,234.56789')`, which is NaN, so the send flow breaks
- Strict wallets reject the URI outright
- Lenient wallets that use `parseFloat`-style parsing stop at the comma and read **1 BTC** instead of 1,234.56789 BTC, a massive underpayment to the receiver

Fix:

1. `views/Receive.tsx`: build the `amount` parameter with `toFixed()` instead of `toFormat()`. `toFixed()` always emits a plain decimal with no separators and no exponent notation (verified for dust amounts, 1000+ BTC, and 21M BTC).
2. `utils/AddressUtils.ts`: harden the BIP21 parser to de-group amounts that match unambiguous thousands formatting (`1,234.56789` -> `1234.56789`) so QRs generated by old ZEUS versions still parse. Ambiguous comma-decimal forms like `1,5` (which some locales use to mean 1.5) deliberately do NOT match and continue to fail closed as NaN, since stripping that comma would misread 1.5 BTC as 15 BTC.
3. `utils/AddressUtils.test.ts`: regression tests for both parser behaviors.

No visual changes; screenshots N/A.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

Not yet manually tested on device. The change is platform-independent string formatting; the parser paths are covered by unit tests. Will follow up with on-device verification of the receive QR round trip.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

The URI construction and parsing are backend-independent (pure string/BigNumber logic shared by all backends).

### Locales

### Other:

### DIFF
--- a/utils/AddressUtils.test.ts
+++ b/utils/AddressUtils.test.ts
@@ -115,14 +115,15 @@ describe('AddressUtils', () => {
                 satAmount: '123456789000'
             });
 
-            // ambiguous comma decimal separator must not be misread
+            // ambiguous comma decimal separator must not be misread;
+            // invalid amounts are ignored but the address is kept
             expect(
                 AddressUtils.processBIP21Uri(
                     'bitcoin:bc1q7065ezyhcd3qtqlcvwcmp9t2weaxc4sguuvlwu?amount=1,5'
                 )
             ).toEqual({
                 value: 'bc1q7065ezyhcd3qtqlcvwcmp9t2weaxc4sguuvlwu',
-                satAmount: 'NaN'
+                satAmount: undefined
             });
 
             // without fee
--- a/utils/AddressUtils.test.ts
+++ b/utils/AddressUtils.test.ts
@@ -105,6 +105,26 @@ describe('AddressUtils', () => {
                 satAmount: '170003'
             });
 
+            // comma group separators (emitted by old ZEUS versions)
+            expect(
+                AddressUtils.processBIP21Uri(
+                    'bitcoin:bc1q7065ezyhcd3qtqlcvwcmp9t2weaxc4sguuvlwu?amount=1,234.56789'
+                )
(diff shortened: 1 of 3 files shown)


ajaysehwal (commented):
@kaloudis One small improvement we could make while we're touching this path:

In bitcoinQrParser, an invalid amount can currently end up as the string 'NaN'. Since 'NaN' is truthy, processBIP21Uri then sets satAmount to 'NaN'.

Image

For example, scanning a malformed BIP21 URI with an amount like amount=1,5 can prefill the amount field with NaN. I think it would be better to ignore invalid amounts and leave the field empty, while still keeping the parsed address

kaloudis:
@ajaysehwal addressed in [f25556e](https://github.com/ZeusLN/zeus/pull/4324/commits/f25556ebdee0b369e0db3446d74fb7292d99c86a)

ajaysehwal (approved):
tACK

PR #4327, opened 2026-08-04
fix(security): fully wipe wallet data on duress and lockout

kaloudis:
# Description

The duress PIN/passphrase wipe (`deleteNodes`) and the automatic 5-failed-attempts wipe (`authenticationFailure`) in `views/Lockscreen.tsx` only set `nodes: undefined` in the settings blob. Every piece of seed-bearing material survived the "wipe":

- the swap rescue key (`swaps-rescue-key`)
- the per-node Cashu seed phrase (`${nodeDir}-cashu-seed-phrase`), which is identical to the wallet mnemonic on 12-word LDK wallets
- the CDK ecash SQLite database (spendable bearer proofs)
- the on-disk LND/LDK node data directories (channel + wallet state)

An attacker who watches the user enter the duress PIN under coercion and then takes the device can recover the entire wallet from the keychain and on-disk state, even though the app shows an empty, wiped wallet. The feature provides false assurance to exactly the users who rely on it.

## Fix

Both wipe handlers now route through `clearAllData()`, which already cleared the keychain, Cashu seeds, CDK database, and swap rescue key. `clearAllData()` itself had the same node-directory gap (the Tools "Clear All Data" action left the node dirs behind too), so this adds node-data-directory deletion there, stopping the node first (`deleteLndWallet` stops LND internally; LDK is stopped via `stopLdkNode()` before unlinking). Routing both duress paths through the single `clearAllData()` avoids maintaining a second wipe list in `Lockscreen` that drifts out of sync.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Behavior change worth a maintainer decision

The 5-failed-attempts wipe is now genuinely destructive of on-disk wallet state, not just the settings pointer. That matches the security intent, but a user with no external seed backup now loses funds permanently on a lockout, whereas previously the data directory survived and was in principle recoverable. The 2026-08-01 audit listed "duress-wipe scope" as a known/accepted exclusion, so this crosses into keychain/settings-storage and security-feature territory that should get explicit sign-off before merge.

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

New `utils/DataClearUtils.test.ts` asserts `clearAllData()` stops and deletes each configured node's data directory (embedded-LND and LDK, single and multi-node, default `lndDir` fallback, ordering of stop-before-unlink, and the empty/missing-settings guards). Verified as a real regression guard: the positive-assertion tests fail when the node-directory deletion is removed.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

> Draft: not yet exercised on a device. Needs a two-platform run of the duress-PIN and 5-failed-attempts flows, asserting `swaps-rescue-key`, the `${nodeDir}-cashu-seed-phrase` entry, the CDK database, and the LDK/LND node directories are all gone afterward. The background-relock case (node actually running when the duress PIN is entered) is the most important path to exercise, since the startup lockscreen has no running node.

### Locales
- [x] N/A

### DIFF
--- a/utils/DataClearUtils.ts
+++ b/utils/DataClearUtils.ts
@@ -56,6 +56,7 @@ import {
 } from '../utils/SwapUtils';
 import { deleteLndWallet } from './LndMobileUtils';
 import { deleteLdkNodeWallet, stopLdkNode } from './LdkNodeUtils';
+import { sleep } from './SleepUtils';
 
 const KEY_PREFIX = 'zeus:';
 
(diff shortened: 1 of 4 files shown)


views/Lockscreen.tsx
```diff
@@ -449,10 +455,15 @@ export default class Lockscreen extends React.Component<
...
+    authenticationFailure = async () => {
         const { SettingsStore } = this.props;
         const { updateSettings } = SettingsStore;
 
+        // Fully wipe wallet data on repeated failed logins, then clear the
+        // local auth secrets. clearAllData() removes the node data dirs, Cashu
+        // seeds + CDK db, and swap rescue key the settings wipe leaves behind.
+        await clearAllData();
+
         updateSettings({
```
ajaysehwal: @kaloudis I think there’s an issue with the data wipe flow in `Lockscreen.tsx`. Both `deleteNodes` and `authenticationFailure` call `clearAllData()` and then `updateSettings(...)`.
The problem is that after `clearAllData()` removes the stored settings, `updateSettings()` can fall back to the existing in-memory `this.settings`, merge the changes, and write those settings back to storage.
So the wipe may not actually clear everything — `nodes` gets removed, but other settings still in memory, including `pin` / `duressPin`, can be written back.
The **Clear All Data** flow already avoids this by calling `clearAllData()` and restarting without updating settings afterward. I think we should follow a similar approach here and avoid writing settings again after the wipe.

views/Lockscreen.tsx
ajaysehwal, line 1: nit: this.deleteNodes() / this.authenticationFailure() not awaited
https://github.com/ZeusLN/zeus/blob/8c79c0ab621fbff7bf9380da7ac2faa9f0c1475c/views/Lockscreen.tsx#L349

kaloudis: [837b505](https://github.com/ZeusLN/zeus/pull/4327/commits/837b505ca5d8a792a81d7786817cd80f89fc02a4)

views/Lockscreen.tsx
kaloudis: [0d0632e](https://github.com/ZeusLN/zeus/pull/4327/commits/0d0632e0039587681511bbb13707ba94d00ef63e)

utils/DataClearUtils.ts
```diff
@@ -247,6 +249,31 @@ async function clearCashuDataForNode(lndDir: string) {
...
+ * separately or seed-bearing wallet state survives a "wipe".
+ */
+async function clearNodeDataDirectories(settings: any): Promise<void> {
+    if (!settings?.nodes || !Array.isArray(settings.nodes)) return;
+
+    for (const node of settings.nodes) {
+        try {
+            if (node.implementation === 'embedded-lnd') {
+                // deleteLndWallet stops LND before unlinking the directory
+                await deleteLndWallet(node.lndDir || 'lnd');
```
ajaysehwal: @kaloudis One more thing: `deleteLndWallet()` and `clearNodeDataDirectories()` currently swallow directory deletion failures (`log.e` / `console.warn`) and continue with the wipe. The app then clears the keychain and restarts as if everything succeeded.

This means that if deleting the LND/LDK directories fails, seed-related data could remain on disk even though the UI appears to have been fully wiped.

I think it would be safer if the wipe failed closed. For example, `clearNodeDataDirectories()` could return an error (or rethrow) when any directory deletion fails, so `Lockscreen` doesn't continue or restart until the cleanup actually succeeds.

kaloudis: During duress, the last thing we want to do is fail loudly or impede the rest of the deletion process. In [1dde3f3](https://github.com/ZeusLN/zeus/pull/4327/commits/1dde3f3953cbcbf7144acb380f70f7eb323c5155) we've add explicit retry to help out here.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):

ajaysehwal (approved):

PR #4328, opened 2026-08-04
fix(security): purge orphaned key material on wallet deletion and wipe

kaloudis:
# Description

Builds on #4327 (KEY-005), now merged. This PR started as KEY-006 from the 2026-08-01 security audit (four classes of secrets surviving deletion and wipes) and grew to eight findings: every fix was verified on-device on both platforms, and the device testing itself surfaced four additional bugs, including one that had silently neutered Android LND directory deletion since 2023.

### From the security audit (KEY-006)

1. **Cached account xprvs.** `views/Settings/SeedQRExport.tsx` cached the wallet's BIP49 and BIP84 **master** extended private keys under `<pubkey>-extended-private-keys`. No code path ever deleted this key.
2. **LDK-node Cashu seed.** `clearAllData` only iterated `node.lndDir` for Cashu data, but LDK nodes namespace Cashu keys by `ldkNodeDir` (`CashuStore.getNodeDir`). So `<ldkNodeDir>-cashu-seed-phrase` (identical to the wallet mnemonic on 12-word LDK wallets) survived a full wipe.
3. **LNC pairing keys.** LNC credentials live under `lnc-rn:<sha256(pairingPhrase)>` and `...:host` (`backends/LNC/credentialStore.ts`). `clearAllData` cleared only the literal `lnc-rn` prefix, never the hashed keys.
4. **CDK ecash databases.** `clearCDKDatabase` unlinked only the literal `cashu_wallet.db`, but since v13.0.0 (e14d0e9ed) the native modules create **per-wallet** databases named `cashu_wallet_<sha256(mnemonic)[0..8]>.db`. Every wallet's CDK database (spendable proofs, mint keys, transaction history) survived `clearAllData`, the duress wipe, and wallet deletion. CDK also opens SQLite with `journal_mode=WAL` (cdk-sqlite `common.rs`), so `-wal`/`-shm` sidecars hold recent proof data next to each db, and the db is not encrypted at rest.

### Found during this PR's on-device testing

5. **Android LND directories were never deleted at all.** Two independent bugs in `LndMobileTools.java` each made deletion a no-op: `deleteRecursive`'s only `fileOrDirectory.delete()` call lived inside a `HyperLog.d(...)` statement and was removed along with logging in 54e94665e (2023-09-04), leaving a pure directory walk; and `deleteLndDirectory` built its path as `"--lnddir=" + filesDir + ...`, the LND startup flag pasted into a file path, then resolved success unconditionally. Deleting an embedded wallet left `wallet.db` (encrypted seed), admin macaroons, and `channel.backup` behind, and the merged #4327 duress wipe silently skipped LND directories on Android, its retry loop defeated by the unconditional success. iOS was unaffected.
6. **Cashu clear list had drifted behind CashuStore.** Nine node-dir-namespaced key suffixes written by CashuStore were missing from the clear list, including `offline-pending-tokens` and `offline-spent-tokens` (actual ecash token data); proven on-device by `<nodeDir>-cashu-original-seed-version` surviving a fixed-build deletion. Separately, `clearCashuDataForNode` cleared `-cashu-mintUrls` and then read that same key to enumerate legacy pre-CDK per-mint keys, so that legacy cleanup had always been a no-op.
7. **Async writers resurrected settings after the wipe.** Clear storage wiped everything, then the app restarted into the wallet view with all wallets intact: RNRestart reloads the JS root but the dying context keeps running briefly, `SettingsStore.getSettings` falls back to in-memory settings on a storage miss, and `updateSettings` re-persists the full blob. Device logs showed the resurrecting write landing ~20ms after the wipe finished.
8. **NWC key material was never in the wipe list.** `zeus-nwc-service-keys` (nostr secret keys for the NWC service role) survived a full wipe; client keys and connection configs were equally uncovered.

On iOS, keychain items also survive app uninstall (demonstrated in testing below), compounding all of the above.

## Fixes, by commit

| Commit | Fix |
|---|---|
| 794160723 | **xprv cache:** never persist derived xprvs (recompute per export); delete legacy caches at the read site. **LNC:** reconstruct and clear hashed keys from each node's pairing phrase. **LDK Cashu:** clear both `lndDir` and `ldkNodeDir` namespaces. **`deleteNodeConfig`:** clear the deleted node's keys via shared `clearNodeKeychainData(node)`. |
| 7deb7b6af | Legacy configs without explicit dirs: mirror `getNodeDir()`'s `'lnd'`/`'ldk'` fallbacks, gated by implementation so remote nodes never clear namespaces owned by embedded wallets. |
| ee33ce64a | Clear keychain material before `updateSettings` removes the node config; a crash between the steps would otherwise orphan the hashed LNC keys permanently. |
| 49389b831 | **CDK full wipe:** sweep the CDK directory for the `cashu_wallet` prefix (legacy db, all per-wallet dbs, WAL/SHM sidecars). **Single deletion:** `clearCDKDatabaseForNode` reconstructs the exact filename hash from the stored cashu seed before the keychain clear wipes it; remote nodes excluded so a shared default namespace can never destroy another wallet's proofs. |
| ff1e0f159 | **Android:** restore the `delete()` in `deleteRecursive` (null-guarded, returns success), drop the `--lnddir=` prefix, reject on partial failure. The legacy `"lnd"` branch now deletes only LND artifacts since the files-dir root holds other wallets' data. Single deletion routes through shared `deleteNodeDataDirectoryWithRetry` (3x/500ms) instead of ignoring the result. |
| 3bafdb633 | Exported `CASHU_KEY_SUFFIXES`, complete against CashuStore, with a drift-guard test that greps `stores/CashuStore.ts` for `-cashu-` literals and fails CI when a key is written but not cleared. Legacy per-mint mint list read before clearing so that cleanup actually runs. |
| 31df05f88 | `Storage.blockWrites()` latch: `setItem` becomes a warning no-op for the rest of the JS context once a wipe starts; the post-restart reload starts fresh. Kills the settings-resurrection class for both wipe paths. |
| 36004ff3c | All six `NostrWalletConnectStore` keys, rating counters, and the bare `lnurlpay:` entry added to the wipe list. |
| 55294fc84 | Legacy xprv cache also purged in `NodeInfoStore.getNodeInfo`'s success path, where the pubkey is definitionally known; the read-site purge silently no-oped when the export screen was opened before the node finished connecting (caught by the upgrade-path device test) and could never help users who don't open that screen. |
| ed6a6d104 | **Android orphan kill:** full wipe deletes react-native-keychain's backing stores wholesale (`files/datastore/RN_KEYCHAIN.preferences_pb` plus legacy `shared_prefs/RN_KEYCHAIN.xml`) as its final clearing action, removing the orphaned entries no config-driven clear can reach. Ordered last because react-native-keychain caches the full preference map and a later write would re-persist the orphans; the write latch holds writers until restart. iOS unchanged (system keychain, no file). |

## Known residuals (documented, not fixed here)

- **Pre-v13 shared `cashu_wallet.db`:** one wallet's proofs cannot be selectively removed from the legacy shared file; it dies on last-wallet removal or full wipe. Post-v13 per-wallet databases fully fix this for current wallets.
- **Historical orphans:** devices in the field hold LND directories and keychain entries from wallets deleted by older (broken) builds; their configs are gone, so config-driven cleanup cannot find them. Orphaned *keychain entries* are now killed on Android by the wholesale backing-store deletion (ed6a6d104); orphaned *LND directories* (a files-dir sweep on wipe) and iOS keychain orphans (needs a native delete-all-in-access-group call) remain follow-ups.
- **`lnurlpay:<hash>` metadata:** written with no enumerating index; a write-side index is a follow-up.
- **Cosmetic:** the CDK native module holds its SQLite handle across the wipe and re-creates an empty, schemaless db skeleton (verified: no tables). Avoiding it needs a native close/dispose API on both platforms.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

`utils/DataClearUtils.test.ts` extended to 36 tests: LDK-namespaced Cashu and hashed LNC clearing, legacy default-namespace fallbacks with the remote-node guard, CDK sweep (including partial-failure resilience) and targeted per-wallet deletion (exact native hash derivation, remote/missing-seed refusal), the `CASHU_KEY_SUFFIXES` drift guard against CashuStore's source, the write latch engaging before the first key removal, deletion retry behavior, and NWC key clearing. Verified as real regression guards (assertions fail when the corresponding clearing is removed).

- [x] Android (emulator, sdk_gphone64_arm64, Android 15)
- [x] iOS (simulator, iPhone 17 Pro, iOS 18)

On-device
- [x] LDK Node
- [x] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

> Device testing performed on both platforms, all passes:
> - **Single-wallet deletion** (Android + iOS): deleted wallet's LND dir, CDK db + WAL/SHM sidecars, and cashu keychain keys removed; surviving wallet's ecash balance intact and spent successfully afterward. Restore-from-seed after deletion recovered the deleted wallet's full ecash balance (deterministic db naming confirmed on-device).
> - **Full wipe** (Android + iOS): every `cashu_wallet*` file swept (including orphans from years of accumulated dev state), node dirs deleted, hashed LNC keys cleared, no settings resurrection (iOS keychain went from 96 live entries to 12 orphaned dynamic entries). Post-wipe boot lands on intro and writes only a default settings blob, identical to a fresh install.
> - **Duress wipe** (Android + iOS): same assertions as full wipe, all pass.
> - **iOS uninstall/reinstall** (simulator): all keychain entries survive app deletion, confirming the audit's persistence concern; because the wipe ran first, the survivors contain no key material, and the Keychain Recovery scanner on the fresh install reports "No Lost Data Found".
> - Residual keychain survivors are exclusively pre-existing orphans from old-build deletions and unenumerable hash-keyed dynamic entries, both documented above.
> - **Upgrade path** (Android): pre-PR build wrote the `<pubkey>-extended-private-keys` cache under the real pubkey; PR build purged it on the wallet's next connect, zero occurrences left in the keychain. This test caught and fixed the read-site purge's dependency on nodeInfo being loaded (55294fc84).
> - **LNC credential lifecycle** (Android, regtest litd, public mailbox): pairing writes the hashed `lnc-rn` key pair; single-wallet deletion clears both, verified live. Re-pairing with the consumed phrase fails by design (LNC phrases are one-time; the destroyed static keys were the only resumption path), which is the desired property: after deletion the session is unrecoverable from the device. A brand-new litd session pairs cleanly afterward. A failed pairing attempt does persist credential material, but under a saved wallet config, so deleting that config clears it.
> - **LND REST** (Android, regtest): a connected REST wallet writes no node-namespaced keychain material (credentials live in the settings blob; the Cashu capability gate holds, no `lnd-cashu-*` keys appear). Deletion removes the config cleanly with zero keychain-entry churn and the surviving LNC wallet still connects; the remote-node guards (no default-namespace clearing, no targeted CDK delete) behave as designed.
> - **Android keychain wholesale delete** (emulator): full wipe removed `RN_KEYCHAIN.preferences_pb` entirely, killing three orphaned cashu-seed-phrase families that had survived every prior wipe; a second wipe pass over the already-deleted file ran cleanly (no crash, no file resurrection), demonstrating idempotency. Fresh boot regenerates the store with only the default settings blob; first EncryptedStorage access after the wipe paid a one-time ~28s androidx-crypto keyset regeneration on the emulator.
> - Dev-mode observation: RNRestart occasionally reloads the JS root without relaunching the process, leaving stale in-memory UI over wiped storage until a manual restart; the write latch holds throughout, so nothing re-persists. A hard process restart (e.g. ProcessPhoenix, already a dependency) for the wipe paths is a follow-up candidate.
> - Authoritative uninstall/keychain check on a physical iOS device remains a nice-to-have before release.

### Locales
- [x] N/A

### DIFF
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1973,28 +1973,60 @@ export default class SettingsStore {
         return settings;
     }
 
-    public updateSettings = async (newSetting: any) => {
+    // Serializes updateSettings calls. Each update is a read-merge-write
+    // against the persisted blob; without ordering, a caller that read
+    // before another's write can commit a stale snapshot over it, e.g. a
+    // background update straddling a wallet deletion writes the old nodes
+    // array back, resurrecting the deleted node's seed and password.
+    private updateSettingsQueue: Promise<unknown> = Promise.resolve();
+
+    public updateSettings = (
+        newSetting: any | ((currentSettings: Settings) => any)
+    ): Promise<Settings> => {
+        const task = this.updateSettingsQueue.then(() =>
+            this.applySettingsUpdate(newSetting)
+        );
+        // Keep the queue alive after a rejected update; the caller still
+        // sees the rejection through `task`.
+        this.updateSettingsQueue = task.catch(() => undefined);
+        return task;
+    };
+
+    private applySettingsUpdate = async (
+        newSetting: any | ((currentSettings: Settings) => any)
+    ) => {
         this.settingsUpdateInProgress = true;
-        const existingSettings = await this.getSettings();
-        const newSettings = {
-            ...existingSettings,
-            ...newSetting
-        };
+        try {
+            const existingSettings = await this.getSettings();
+            // Functional updates read the settings inside the critical
+            // section, so callers whose new value depends on the current
+            // one (delete node X from the array) cannot act on a snapshot
+            // taken before earlier queued writes landed.
+            const resolvedSetting =
+                typeof newSetting === 'function'
+                    ? newSetting(existingSettings)
+                    : newSetting;
+            const newSettings = {
+                ...existingSettings,
+                ...resolvedSetting
+            };
 
-        if (
-            newSetting.pos?.posEnabled &&
-            newSetting.pos.posEnabled !== PosEnabled.Disabled
-        ) {
-            this.posWasEnabled = true;
+            if (
+                resolvedSetting.pos?.posEnabled &&
+                resolvedSetting.pos.posEnabled !== PosEnabled.Disabled
+            ) {
+                this.posWasEnabled = true;
+            }
+
+            await this.setSettings(newSettings);
+            this.triggerSettingsRefresh = true;
+
+            // Update store's node properties from latest settings
+            this.updateNodeProperties(newSettings);
+            return newSettings;
+        } finally {
+            this.settingsUpdateInProgress = false;
         }
-
-        await this.setSettings(newSettings);
-        this.triggerSettingsRefresh = true;
-
-        // Update store's node properties from latest settings
-        this.updateNodeProperties(newSettings);
-        this.settingsUpdateInProgress = false;
-        return newSettings;
     };
 
     // LNDHub
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1965,12 +1965,20 @@ export default class SettingsStore {
         return this.settings;
     };
 
-    public async setSettings(settings: any) {
+    // Returns whether the settings were persisted. Storage.setItem returns
+    // false while the data-wipe write latch is engaged (and the keychain
+    // call itself is typed `false | Result`); in-memory settings only
+    // update when the write landed, so a blocked write cannot leave the
+    // store advertising state that will not survive the imminent restart.
+    public async setSettings(settings: any): Promise<boolean> {
         this.loading = true;
-        await Storage.setItem(STORAGE_KEY, settings);
-        this.settings = settings;
+        const persisted =
+            (await Storage.setItem(STORAGE_KEY, settings)) !== false;
+        if (persisted) {
+            this.settings = settings;
+        }
         this.loading = false;
-        return settings;
+        return persisted;
     }
 
     // Serializes updateSettings calls. Each update is a read-merge-write
@@ -2018,7 +2026,13 @@ export default class SettingsStore {
                 this.posWasEnabled = true;
             }
 
-            await this.setSettings(newSettings);
+            const persisted = await this.setSettings(newSettings);
+            if (!persisted) {
+                // Write latch engaged (data wipe in progress): nothing was
+                // persisted or kept in memory, so skip the derived-state
+                // updates that would advertise the unsaved settings.
+                return this.settings;
+            }
             this.triggerSettingsRefresh = true;
 
             // Update store's node properties from latest settings
--- a/utils/DataClearUtils.ts
+++ b/utils/DataClearUtils.ts
@@ -195,6 +195,45 @@ async function clearKey(key: string) {
 }
 
 /**
+ * Deletes react-native-keychain's Android backing stores wholesale: the
+ * Jetpack DataStore file (current versions) and the legacy shared-prefs XML
+ * (older versions). This is the only way to remove orphaned entries that
+ * config-driven clearing can never reach: keys from wallets deleted by older
+ * builds and unenumerable hash-keyed dynamic entries. Android-only; on iOS
+ * the keychain is system-managed with no file to delete.
+ *
+ * MUST run as the final step of a full wipe: react-native-keychain caches
+ * the full preference map in memory, and any keychain write after this
+ * deletion would re-persist that map, resurrecting the orphans. Storage's
+ * write latch blocks the known writers until the post-wipe restart.
+ */
+async function clearKeychainBackingStore(): Promise<void> {
+    if (Platform.OS !== 'android') return;
+    const dirs = ReactNativeBlobUtil.fs.dirs;
+    const appRoot = `${dirs.DocumentDir}/..`;
+    const targets = [
+        `${appRoot}/files/datastore/RN_KEYCHAIN.preferences_pb`,
+        `${appRoot}/shared_prefs/RN_KEYCHAIN.xml`
+    ];
+    for (const path of targets) {
+        try {
+            if (await ReactNativeBlobUtil.fs.exists(path)) {
+                await ReactNativeBlobUtil.fs.unlink(path);
+                console.log(
+                    '[ClearData] Keychain backing store deleted:',
+                    path
+                );
+            }
+        } catch (e) {
+            console.warn(
+                `[ClearData] Error deleting keychain backing store ${path}:`,
+                e
+            );
+        }
+    }
+}
+
+/**
  * Directory holding the CDK SQLite database files
  * iOS: Application Support; Android: files dir
  */
@@ -608,6 +647,11 @@ export async function clearAllData(): Promise<void> {
     for (const key of additionalKeys) {
         await clearKey(key);
     }
+
+    // 8. Android: delete the keychain backing stores wholesale, killing the
+    // orphaned entries no config-driven clear can reach. Keep this the LAST
+    // clearing action (see clearKeychainBackingStore).
+    await clearKeychainBackingStore();
 
     console.log('[ClearData] All data cleared successfully');
 }
--- a/utils/DataClearUtils.ts
+++ b/utils/DataClearUtils.ts
@@ -312,6 +312,18 @@ export async function clearNodeKeychainData(node: any): Promise<void> {
     if (node.ldkNodeDir) {
         await clearCashuDataForNode(node.ldkNodeDir);
     }
+    // Legacy configs can predate explicit node dirs (see the `lndDir || 'lnd'`
+    // fallbacks in WalletConfiguration/getNodeDir). Mirror getNodeDir()'s
+    // defaults so those wallets' Cashu keys are cleared too. Gate on
+    // implementation: only one config can occupy a default namespace per
+    // implementation, and remote nodes must never clear the 'lnd'/'ldk'
+    // namespaces owned by embedded wallets.
+    if (node.implementation === 'embedded-lnd' && !node.lndDir) {
+        await clearCashuDataForNode('lnd');
+    }
+    if (node.implementation === 'ldk-node' && !node.ldkNodeDir) {
+        await clearCashuDataForNode('ldk');
+    }
     if (node.pairingPhrase) {
         await clearLncCredentials(node.pairingPhrase);
     }
--- a/utils/DataClearUtils.ts
+++ b/utils/DataClearUtils.ts
@@ -2,7 +2,7 @@ import * as Keychain from 'react-native-keychain';
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import EncryptedStorage from 'react-native-encrypted-storage';
 import ReactNativeBlobUtil from 'react-native-blob-util';
-import { Platform } from 'react-native';
+import { BackHandler, Platform } from 'react-native';
 import Storage from '../storage';
 
 import {
@@ -137,6 +137,34 @@ const STORAGE_KEYS = [
     'backup-complete',
     'backup-complete-v2'
 ];
+
+/**
+ * Pins the user to the current screen while a wipe runs. clearAllData()
+ * takes tens of seconds on device, and both wipe surfaces sit on screens
+ * the user can otherwise leave mid-wipe: hardware/gesture back either
+ * exits the app (Lockscreen, via the App-level loginRequired handler) or
+ * pops back into an app whose data is being destroyed underneath it
+ * (Tools). Swallows hardware back presses (registered after the App-level
+ * handler, so it runs first), blocks removal of the screen from the
+ * navigation stack, and disables the iOS back-swipe gesture. Returns a
+ * release function for unmount hygiene; in practice the guard holds until
+ * the post-wipe restart tears the JS context down.
+ */
+export function blockNavigationDuringWipe(navigation: any): () => void {
+    navigation.setOptions({ gestureEnabled: false });
+    const backSubscription = BackHandler.addEventListener(
+        'hardwareBackPress',
+        () => true
+    );
+    const removeBeforeRemove = navigation.addListener(
+        'beforeRemove',
+        (e: any) => e.preventDefault()
+    );
+    return () => {
+        backSubscription.remove();
+        removeBeforeRemove();
+    };
+}
 
 /**
  * Clears a key from all possible locations:
--- a/utils/DataClearUtils.ts
+++ b/utils/DataClearUtils.ts
@@ -287,49 +287,74 @@ export async function clearCDKDatabaseForNode(node: any): Promise<void> {
 }
 
 /**
+ * Suffixes of every node-dir-namespaced Cashu storage key written by
+ * CashuStore (`<nodeDir>-cashu-<suffix>`). Kept complete by the drift guard
+ * in DataClearUtils.test.ts, which extracts the `-cashu-` literals from
+ * stores/CashuStore.ts and fails when a new key is written but not cleared.
+ */
+export const CASHU_KEY_SUFFIXES = [
+    'mintUrls',
+    'selectedMintUrl',
+    'selectedMintUrls',
+    'multiMintSelectedUrls',
+    'totalBalanceSats',
+    'invoices',
+    'payments',
+    'received-tokens',
+    'sent-tokens',
+    'offline-pending-tokens',
+    'offline-spent-tokens',
+    'seed-version',
+    'seed-phrase',
+    'seed',
+    'original-seed-version',
+    'v1-restore-done',
+    'dismissedUpgradeThreshold',
+    'randomizeMintSelection',
+    'nostrMintBackupTimestamp'
+];
+
+/**
  * Clears Cashu data for a specific node directory
  */
 async function clearCashuDataForNode(lndDir: string) {
-    // Clear app-level Cashu storage keys
-    const cashuKeys = [
-        `${lndDir}-cashu-mintUrls`,
-        `${lndDir}-cashu-selectedMintUrl`,
-        `${lndDir}-cashu-totalBalanceSats`,
-        `${lndDir}-cashu-invoices`,
-        `${lndDir}-cashu-payments`,
-        `${lndDir}-cashu-received-tokens`,
-        `${lndDir}-cashu-sent-tokens`,
-        `${lndDir}-cashu-seed-version`,
-        `${lndDir}-cashu-seed-phrase`,
-        `${lndDir}-cashu-seed`
-    ];
-
-    for (const key of cashuKeys) {
-        await clearKey(key);
-    }
-
-    // Try to clear per-mint wallet keys (legacy storage)
-    // Note: After CDK migration, mintUrls are stored in CDK database, not local storage
-    // These keys may still exist from before migration
+    // Read the legacy pre-CDK mint list BEFORE clearing: the loop below
+    // removes the mintUrls key this read depends on
+    let legacyMintUrls: string[] = [];
     try {
         const mintUrlsJson = await Storage.getItem(`${lndDir}-cashu-mintUrls`);
         if (mintUrlsJson) {
-            const mintUrls = JSON.parse(mintUrlsJson);
-            if (Array.isArray(mintUrls)) {
-                for (const mintUrl of mintUrls) {
-                    // walletId format: ${lndDir}==${mintUrl}
-                    const walletId = `${lndDir}==${mintUrl}`;
-                    const walletKeys = [
-                        `${walletId}-mintInfo`,
-                        `${walletId}-counter`,
-                        `${walletId}-proofs`,
-                        `${walletId}-balance`,
-                        `${walletId}-pubkey`
-                    ];
-                    for (const walletKey of walletKeys) {
-                        await clearKey(walletKey);
-                    }
-                }
+            const parsed = JSON.parse(mintUrlsJson);
+            if (Array.isArray(parsed)) {
+                legacyMintUrls = parsed;
+            }
+        }
+    } catch (e) {
+        console.warn(
+            `[ClearData] Error reading Cashu mint list for ${lndDir}:`,
+            e
+        );
+    }
+
+    // Clear app-level Cashu storage keys
+    for (const suffix of CASHU_KEY_SUFFIXES) {
+        await clearKey(`${lndDir}-cashu-${suffix}`);
+    }
+
+    // Clear per-mint wallet keys (legacy pre-CDK storage)
+    try {
+        for (const mintUrl of legacyMintUrls) {
+            // walletId format: ${lndDir}==${mintUrl}
+            const walletId = `${lndDir}==${mintUrl}`;
+            const walletKeys = [
+                `${walletId}-mintInfo`,
+                `${walletId}-counter`,
+                `${walletId}-proofs`,
+                `${walletId}-balance`,
+                `${walletId}-pubkey`
+            ];
+            for (const walletKey of walletKeys) {
+                await clearKey(walletKey);
             }
         }
     } catch (e) {
--- a/utils/DataClearUtils.ts
+++ b/utils/DataClearUtils.ts
@@ -367,6 +367,30 @@ async function deleteNodeDataDirectory(node: any): Promise<boolean> {
 }
 
 /**
+ * Stops and deletes a single node's on-disk data directory, retrying on
+ * failure. The likely cause of failure is a file handle still held by a node
+ * that hasn't finished shutting down, which clears on its own after a moment.
+ * Returns whether the directory was removed. Exported so single-wallet
+ * deletion (WalletConfiguration.deleteNodeConfig) shares the wipe path's
+ * retry behavior instead of assuming the first attempt succeeded.
+ */
+export async function deleteNodeDataDirectoryWithRetry(
+    node: any
+): Promise<boolean> {
+    for (let attempt = 1; attempt <= NODE_DIR_DELETE_ATTEMPTS; attempt++) {
+        if (await deleteNodeDataDirectory(node)) return true;
+
+        if (attempt < NODE_DIR_DELETE_ATTEMPTS) {
+            await sleep(NODE_DIR_RETRY_DELAY_MS);
+        }
+    }
+    console.warn(
+        `[ClearData] Gave up deleting node data directory after ${NODE_DIR_DELETE_ATTEMPTS} attempts`
+    );
+    return false;
+}
+
+/**
  * Clears the LNC pairing credentials for a given pairing phrase.
  * Keys are namespaced by sha256(pairingPhrase), so they can only be removed
  * when the phrase is known (it lives in the node config).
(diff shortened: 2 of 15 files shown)


ajaysehwal (commented):
@kaloudis i think  `<pubkey>-extended-private-keys` isn't removed by `clearAllData()` / `clearNodeKeychainData()` because the pubkey isn't available in the node config.
This means deleting a single wallet can leave its xprv behind. On a full wipe, Android eventually removes it through the backing-store cleanup, but on iOS it can still remain.
I think we should explicitly clean up these xprv entries during wallet deletion/full wipe.

utils/DataClearUtils.ts
```diff
@@ -276,6 +454,104 @@ async function deleteNodeDataDirectory(node: any): Promise<boolean> {
...
+    if (
+        !opts?.skipXprvPurge &&
+        node.implementation === 'embedded-lnd' &&
+        Array.isArray(node.seedPhrase) &&
+        node.seedPhrase.length > 0
+    ) {
+        try {
+            const nodeId = await deriveEmbeddedNodeId(
+                node.seedPhrase,
+                node.embeddedLndNetwork?.toLowerCase?.() === 'testnet'
```
ajaysehwal: @kaloudis One edge case: `clearNodeKeychainData()` currently uses coin type `1` only when `embeddedLndNetwork === 'testnet'`. Other non-mainnet networks that also use coin type `1` could derive the wrong pubkey, causing us to miss the actual xprv cache key during cleanup.

It may be safer to determine the coin type based on whether the network is mainnet vs non-mainnet, rather than checking only for `testnet`.

kaloudis:
> @kaloudis i think `<pubkey>-extended-private-keys` isn't removed by `clearAllData()` / `clearNodeKeychainData()` because the pubkey isn't available in the node config. This means deleting a single wallet can leave its xprv behind. On a full wipe, Android eventually removes it through the backing-store cleanup, but on iOS it can still remain. I think we should explicitly clean up these xprv entries during wallet deletion/full wipe.

Good catch, confirmed. The entry was only reachable via the connect-time purge, which misses wallets deleted without reconnecting after upgrade, and on iOS there's no keychain backing store to delete and no enumeration API for internet credentials in react-native-keychain.

Fixed in e05f0f7e7 by re-deriving the pubkey from the wallet's seed at cleanup time: aezeed entropy -> BIP32 master -> m/1017'/coinType'/6'/0/0 (lnd's node identity key). `clearNodeKeychainData` now clears `<derived pubkey>-extended-private-keys` for embedded LND configs, so single-wallet deletion covers it on both platforms and full wipes cover it on iOS. Android full wipes skip the scrypt-heavy derivation since the backing-store deletion added earlier in this PR already removes every entry wholesale, and that path runs on the duress wipe where added seconds per wallet matter. If derivation fails (non-default aezeed passphrase) the purge is skipped safely: the old export screen failed the same way before its cache write, so no entry exists for those wallets.

The derivation is pinned by golden vectors in `utils/AezeedUtils.test.ts` generated from lnd v0.19.2-beta's own Go code (aezeed + hdkeychain), covering mnemonic -> entropy and entropy -> node id for both coin types.

Verified on-device (Android emulator, mainnet embedded LND wallet):

1. Checked out master so the pre-fix bundle ran, opened Seed QR Export on a fresh embedded wallet, and confirmed the legacy cache landed on disk: `<pubkey>-extended-private-keys` visible in `files/datastore/RN_KEYCHAIN.preferences_pb` via `adb exec-out run-as ... | strings`.
2. Switched the active wallet to a different one, flipped to the PR branch, and reloaded. Confirmed the entry was still on disk, i.e. the doomed wallet never connected on the fixed build, which is exactly the scenario you flagged (connect-time purge can't help).
3. Deleted the doomed wallet from the wallet config screen. The entry was removed, which also proves the offline-derived pubkey matches the real node identity on a live wallet, and the surviving wallet's keys were untouched.
4. Regression: a fresh embedded wallet's Seed QR Export still renders both xprv tabs (the decrypt logic moved to `utils/AezeedUtils.ts`), and no `extended-private-keys` entry is ever written by the fixed build.

ajaysehwal:
@kaloudis One more thing I noticed in the wallet deletion flow. If `deleteNodeDataDirectoryWithRetry()` fails, we only log a warning and continue with navigation. This can make it look like the wallet was successfully deleted even though its wallet/channel data may still remain on disk.
I think we should show an error and avoid completing the deletion flow until the directory is successfully removed.

utils/DataClearUtils.ts
kaloudis: Good point. Embedded LND only supports Mainnet and Testnet today, so the 'testnet' check happens to cover every wallet that currently exists, but btcd's HDCoinType is 1 for testnet, signet, and regtest alike, so the mainnet-vs-non-mainnet rule is the right shape and makes this future-proof for any network we add later.

Fixed in 5f90e53d6: the derivation now uses coin type 1 for any non-mainnet network, following the same `(network || 'mainnet').toLowerCase() !== 'mainnet'` rule WalletConfiguration already applies to `embeddedLndNetwork`, with a missing value defaulting to mainnet so legacy configs without the field stay on coin type 0. Tests cover Testnet, mutinynet, and the missing-field default.

kaloudis:
> @kaloudis One more thing I noticed in the wallet deletion flow. If `deleteNodeDataDirectoryWithRetry()` fails, we only log a warning and continue with navigation. This can make it look like the wallet was successfully deleted even though its wallet/channel data may still remain on disk. I think we should show an error and avoid completing the deletion flow until the directory is successfully removed.

Should be surfaced now in https://github.com/ZeusLN/zeus/pull/4328/commits/9395f75b94c9c8a711f8fc19a742be9f6f2a871e - duress / `clearAllData` keeps its log-and-continue behavior.

ajaysehwal (commented):
@kaloudis `deleteNodeDataDirectoryWithRetry()` failures are now surfaced with a blocking alert, which is good. But if xprv-cache key derivation fails in `clearNodeKeychainData()`, the error is caught and cleanup continues silently.
Since the cached xprv is one of the most sensitive items we're trying to remove, I think this failure should at least have proper logging/telemetry, or ideally be surfaced to the user as well.

ajaysehwal:
We have a CI drift guard for `CASHU_KEY_SUFFIXES`, but `clearCDKDatabaseForNode()` also depends on matching the native CDK filename hashing logic (`sha256(seedWords.join(' ')).slice(0, 16)`). If the hashing convention in `CashuDevKitModule.kt/.swift` changes later, single-wallet CDK cleanup could silently stop finding the correct database.
I think it would be good to add a similar drift-guard test for the CDK filename hash logic

shubhamkmr04 (commented):
Storage.blockWrites() is step 0 of clearAllData() and there's no corresponding release. Lockscreen handles this correctly by calling RNRestart.Restart() in a finally, but views/Tools/index.tsx restarts only inside the try — its catch shows an alert and returns. So if the wipe throws anywhere after step 0, the user dismisses the dialog and keeps using an app where every Storage.setItem is a silent no-op for the rest of the session: settings changes and new wallets appear to work in memory and persist nothing, with only a console.warn as the signal. Moving that restart into a finally would match Lockscreen's behavior.

shubhamkmr04:
getItem in storage/index.ts has no try/catch, and clearKeychainBackingStore() unlinks RN_KEYCHAIN.preferences_pb while react-native-keychain's DataStore is still live. Writes are latched but reads aren't, so a read from the still-running JS context after step 8 would reject rather than return false. Your device testing suggests this is fine in practice — a try/catch returning false would just make it robust regardless of DataStore's post-unlink behavior.

shubhamkmr04:
Minor test hygiene: jest.restoreAllMocks() in the "skips the scrypt-heavy derivation" test body and in the backing-store afterEach also restores the module-level console.log spy, so every later test spams [ClearData] output when you run the suite. jest.replaceProperty(Platform, 'OS', 'android') also leaks into subsequent tests if an assertion throws before the restore line. Moving both into afterEach handles it.

kaloudis:
> We have a CI drift guard for `CASHU_KEY_SUFFIXES`, but `clearCDKDatabaseForNode()` also depends on matching the native CDK filename hashing logic (`sha256(seedWords.join(' ')).slice(0, 16)`). If the hashing convention in `CashuDevKitModule.kt/.swift` changes later, single-wallet CDK cleanup could silently stop finding the correct database. I think it would be good to add a similar drift-guard test for the CDK filename hash logic

Agreed, done in d0f8ea5dc

kaloudis:
> Storage.blockWrites() is step 0 of clearAllData() and there's no corresponding release. Lockscreen handles this correctly by calling RNRestart.Restart() in a finally, but views/Tools/index.tsx restarts only inside the try — its catch shows an alert and returns. So if the wipe throws anywhere after step 0, the user dismisses the dialog and keeps using an app where every Storage.setItem is a silent no-op for the rest of the session: settings changes and new wallets appear to work in memory and persist nothing, with only a console.warn as the signal. Moving that restart into a finally would match Lockscreen's behavior.

ixed in e46267e0b: the restart now runs in a `finally`

kaloudis:
> Minor test hygiene: jest.restoreAllMocks() in the "skips the scrypt-heavy derivation" test body and in the backing-store afterEach also restores the module-level console.log spy, so every later test spams [ClearData] output when you run the suite. jest.replaceProperty(Platform, 'OS', 'android') also leaks into subsequent tests if an assertion throws before the restore line. Moving both into afterEach handles it.

Fixed in d0f8ea5dc. `Platform.OS` replacements now go through a tracked handle restored in a top-level `afterEach`, and both `jest.restoreAllMocks()` calls are gone, so the module-level console spies survive the whole suite and a throwing assertion can no longer leak the platform override into later tests.

kaloudis:
> getItem in storage/index.ts has no try/catch, and clearKeychainBackingStore() unlinks RN_KEYCHAIN.preferences_pb while react-native-keychain's DataStore is still live. Writes are latched but reads aren't, so a read from the still-running JS context after step 8 would reject rather than return false. Your device testing suggests this is fine in practice — a try/catch returning false would just make it robust regardless of DataStore's post-unlink behavior.

Disagree. Would rather leave `getItem` propagating errors:

1. A blanket try/catch returning `false` converts any transient keychain read failure in normal operation into "no data", which is precisely the storage-miss failure mode the write latch exists to defend against. A rejected read is noisy but honest. 

A read latch does not work either: `clearAllData` itself reads through `Storage.getItem` after step 0 (the settings blob at step 1, and each node's cashu seed phrase during CDK cleanup), so reads must stay live during the wipe.

2. The actual exposure window (reads from the dying JS context after the backing store unlink) is now as small as it can be: with e46267e0b

views/Settings/WalletConfiguration.tsx
```diff
@@ -793,15 +823,30 @@ export default class WalletConfiguration extends React.Component<
                 justDeletedWallet: active
             });
```
ajaysehwal, line 847: @kaloudis I think there’s still a concurrency issue with wallet deletion.
`updateSettings()` does a read → merge → write without serialization. So if another settings update starts before the wallet deletion and finishes afterward, it can write the stale `nodes` array back and resurrect the deleted wallet configuration. This could also restore sensitive fields like the seed/password after the wallet data has already been cleaned up.

kaloudis: Check out https://github.com/ZeusLN/zeus/pull/4328/commits/5fc679ea4a3c4e22317799fcff52a62e6ac6edfb

views/Settings/WalletConfiguration.tsx
```diff
@@ -740,21 +755,16 @@ export default class WalletConfiguration extends React.Component<
...
 
     deleteNodeConfig = async () => {
+        if (this.state.deletingWallet) return;
         this.setState({ deletingWallet: true });
         const { SettingsStore, navigation } = this.props;
         const { updateSettings, embeddedLndStarted, settings } = SettingsStore;
         const { index, implementation, lndDir, ldkNodeDir, active } =
             this.state;
         const { nodes } = settings;
+        const deletedNode = index != null ? nodes?.[index] : undefined;
```
ajaysehwal, line 766: @kaloudis I think there’s still a race condition in the wallet deletion flow.
`deletedNode` and its `index` are captured before the cleanup starts, but the settings queue is only acquired afterward. During the cleanup, another settings operation could change the `nodes` array, causing the same index to point to a different wallet. In that case, we could delete wallet A’s data but remove wallet B’s config, while A’s sensitive config remains.
I think the exclusive store lock should cover the entire deletion flow, from cleanup through the settings update. Also, it would be safer to identify the wallet using a stable ID instead of the array index.

kaloudis: I think the exclusive lock is a bit overkill but in https://github.com/ZeusLN/zeus/pull/4328/commits/1d3751a6736b8781354a78738a2b03a0b5c35285 I've added functionality that relocates the node to delete using content matching with lodash

ajaysehwal (commented):
https://github.com/user-attachments/assets/a6d12732-004e-40f3-9dca-c9adc3f39039

@kaloudis I think we should show a loading state when the user clicks **Clear All Data**.
The cleanup can take some time depending on how much data is stored in the wallet, so without a loading indicator it may look like the button didn't work or the app is stuck. A loading state would make it clear that the wipe is still in progress.

kaloudis:
Good call @ajaysehwal, thanks for testing. Added in bd1e2849e: the Tools view now shows a loading modal with a "Clearing data..." message while `clearAllData()` runs, so the wipe no longer looks like the button did nothing. It reuses the existing `ChannelBackupLoadingModal` (already a generic spinner + message overlay in this view), appears as soon as you confirm the alert, and stays up until the unconditional restart.

Note the duress/lockout wipe paths in Lockscreen intentionally stay indicator-free, since those shouldn't advertise that a wipe is in progress.

stores/SettingsStore.ts
ajaysehwal, line 1: @kaloudis i think one small issue with `setSettings()` during the wipe flow. `Storage.setItem()` can return `false` when `blockWrites()` is active, but `setSettings()` still updates `this.settings` in memory. This can leave the in-memory settings different from what is actually persisted, and potentially make deleted node configs appear to still exist until restart. I think we should only update `this.settings` when the storage write succeeds:
```ts
const persisted = await Storage.setItem(STORAGE_KEY, settings);

if (persisted !== false) {
    this.settings = settings;
}

return persisted !== false ? settings : this.settings;

ajaysehwal:
@kaloudis One edge case with the xprv cleanup:`clearNodeKeychainData()` uses `decodeAezeedEntropy`, which only works with the default aezeed passphrase. So wallets using a custom passphrase may still have their `<pubkey>-extended-private-keys` entry left in the iOS keychain after single-wallet deletion.

I don't think this is introduced by the PR, and the default-passphrase case is already improved. But it may be worth documenting this as a known limitation, or using the runtime `nodeId` from `NodeInfoStore` when the wallet is active to also identify and remove the cache entry.

stores/SettingsStore.ts
kaloudis: Agreed, fixed in 58ee40597. `setSettings` now only updates `this.settings` when the write landed and returns whether it persisted.

Two things beyond the suggested snippet:

1. Gating `setSettings` alone wasn't enough: `applySettingsUpdate` unconditionally set `triggerSettingsRefresh` and called `updateNodeProperties(newSettings)` afterward, which would have derived node-property observables from settings that were now neither persisted nor in memory. Both are skipped for an unpersisted update, and `updateSettings` returns the current settings instead, as you suggested.

2. The check is `!== false` rather than a truthiness/`=== true` check on purpose: the empty-value path in `Storage.setItem` routes through `removeItem`, whose underlying `resetInternetCredentials` resolves `void`, and a successful `setInternetCredentials` resolves a `Result` object. Only the latch (and the keychain call's typed `false` failure path) produce `false`.

Severity note for the record: the latch is only ever engaged inside `clearAllData`, and since e46267e0b both callers restart unconditionally in a `finally`, so the divergence window was bounded to the seconds of the wipe behind the loading modal. Still worth closing: an in-memory update that pretends a blocked write succeeded is the mirror image of the "noisy but honest" argument I used for `getItem` earlier in this thread.

Regression test added in `stores/SettingsStore.test.ts` ("keeps memory consistent with disk when the write is blocked"): with `setItem` returning `false` once, the returned settings, `store.settings`, `triggerSettingsRefresh`, and the persisted blob all keep the pre-update values, and the next update flows through normally. Verified it fails on the previous commit.

kaloudis:
> @kaloudis One edge case with the xprv cleanup:`clearNodeKeychainData()` uses `decodeAezeedEntropy`, which only works with the default aezeed passphrase. So wallets using a custom passphrase may still have their `<pubkey>-extended-private-keys` entry left in the iOS keychain after single-wallet deletion.

Fair question, but this case is vacuous: no such entry can exist for a custom-passphrase wallet, on either platform.

The only code that ever wrote `<pubkey>-extended-private-keys` was `SeedQRExport.tsx`, and in every version since the cache was introduced (ba70ce58f) it decrypted with `AEZEED_DEFAULT_PASSPHRASE` only. aezeed decryption is authenticated, so for a custom-passphrase seed `aez.decrypt` returns null and the screen sets an error state and returns before the `Storage.setItem` cache write is reached. The purge derivation and the cache write fail on exactly the same set of wallets, which is what makes skipping on derivation failure safe. (Custom-passphrase embedded wallets do exist via the restore flow's `walletPassphrase`, so the concern was worth checking; they just never get the cache.)

Your fallback suggestion is also already in the PR: `NodeInfoStore.getNodeInfo` purges `${nodeInfo.nodeId}-extended-private-keys` at connect time using the runtime nodeId, which is passphrase-independent. That backstop plus the derivation invariant covers both halves.

I've extended the comment at the derivation-failure skip in `clearNodeKeychainData` to note the connect-time backstop (aa7414b49), so the reasoning is documented at the call site.

ajaysehwal (commented):
https://github.com/user-attachments/assets/5cb4aca8-498d-46ec-b350-ab571343d4f4

@kaloudis, I noticed that after multiple incorrect PIN/password attempts, the app redirects to **“Zeus is starting your node”**, with the wallet configuration visible at the top. This seems concerning because it will expose the wallet configuration/seed backup flow even though authentication has failed. We should make sure the wallet config and any sensitive information remain protected after failed authentication attempts.

kaloudis:
Great catch @ajaysehwal, thank you for the video. Confirmed the issue.

**Root cause:** both lockscreen wipe paths (duress credentials and 5 failed attempts) called `SettingsStore.setLoginStatus(true)` before starting the wipe. That flag predates this PR and was harmless on master, where the "wipe" was a single settings write followed by an immediate navigation to the intro screen. This PR made the wipe thorough (keychain sweep, CDK db, node data dirs), so it now runs for tens of seconds before the restart lands. During that whole window `loggedIn: true` disarmed every auth gate in the app: `loginRequired()` went false, so a back-press popped the lockscreen instead of exiting the app, and the Wallet view rendered its connecting screen ("ZEUS is starting your node") with the wallet header live. From there the wallets list and wallet configuration were one tap away, exactly as your recording shows.

**Fix (pushed in a226423):**
1. `setLoginStatus(true)` is removed from both wipe paths. `loggedIn` stays false for the entire wipe, so every gate holds: back-press exits the app, the Wallet view renders nothing even if the lockscreen were somehow dismissed, and `protectedNavigation` stays armed.
2. The lockscreen now shows a neutral loading spinner while the wipe runs, instead of the dead login form. Deliberately indistinct so it discloses nothing about the duress mechanism.
3. `onAttemptLogIn` is guarded against re-entry, so a second submit during the wipe cannot start a concurrent wipe or mutate settings mid-wipe.

One residual to be aware of: if the user force-exits during the wipe (back-press now exits by design), the wipe halts part-way. That fails closed: the settings blob with the PIN/password is cleared late in the wipe order, so the next launch either lands back on the lockscreen (and the next failed attempt resumes the wipe) or on a fresh install state. Same exposure as any process kill mid-wipe, which no JS-side ordering can fully prevent.

Would appreciate a re-test of your repro: 5 wrong entries should now leave you on a spinner until the app restarts into the fresh install state, with nothing tappable in between.

ajaysehwal (commented):
https://github.com/user-attachments/assets/05ecb774-8d39-4bfd-9970-966bbdc8bc8d

@kaloudis, on Android, after reaching the maximum failed attempts, we can't redirect back to the Intro view.
Also, during the wipe/loading state, we should disable slide navigation so the user can't navigate away while the wipe is still in progress.

kaloudis:
> @kaloudis, on Android, after reaching the maximum failed attempts, we can't redirect back to the Intro view.
> Also, during the wipe/loading state, we should disable slide navigation so the user can't navigate away while the wipe is still in progress.

Confirmed both. Fixed in 46812944c.

**Why Android never reached the Intro view:** `react-native-restart`'s `Restart()` is broken on Android under the New Architecture. Its `ReactInstanceManager` path throws (`ReactApplication.getReactNativeHost()` raises in bridgeless mode), so it silently falls back to `Activity.recreate()`, which keeps the OS process and the JS runtime alive. After the wipe, the pre-wipe settings blob (node configs, pins) was still live in `SettingsStore` (`getSettings()` keeps in-memory state on a storage miss), so `loginRequired()` stayed true and the recreated UI landed back on the lockscreen prompting for a PIN that had just been wiped, with the write latch still engaged. iOS was unaffected because RNRestart performs a genuine JS reload there.

Post-wipe restarts now go through a new `restartApp()` helper: on Android it uses the native ProcessPhoenix full-process rebirth (`LndMobileTools.restartApp`) that the existing restart prompts already use, so the relaunch is a cold boot that reads the wiped disk and lands on the intro screen. A process rebirth also cannot resurrect anything from the live keychain DataStore cache or the MobX stores, which closes that whole class of concern on Android. iOS keeps the RNRestart reload.

**Navigation during the wipe:** added `blockNavigationDuringWipe()`, engaged by both lockscreen wipe paths and the Tools Clear All Data flow. It swallows hardware/gesture back presses (which previously exited the app from the lockscreen via the `loginRequired` handler, or popped the Tools screen mid-wipe), blocks removal of the screen from the navigation stack, and disables the iOS back-swipe gesture, holding until the restart lands. Covered by unit tests in DataClearUtils.test.ts.

One note: the video in your comment 404s for me (the earlier two play fine), so the diagnosis above is from code inspection of the restart path. If it shows something beyond landing back on the lockscreen instead of the intro view, let me know.

The other `RNRestart.Restart()` call sites on master (channel migration, seed recovery, LSPS1 settings, fetch-lock restart alert) degrade the same way on Android; they are out of scope here and I will open a follow-up issue for them.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4331, opened 2026-08-04
fix(lnc): route payment results through handlePayment on every path

kaloudis:
# Description

Fixes an internal security review finding (M17): on lightning-node-connect, the NWC wallet service never observes payment results, and the donation flow misreads them.

## The bug

On LNC, `sendPaymentV2` is a streaming RPC. `LightningNodeConnect.payLightningInvoice` returned the raw event name (`'routerrpc.Router.SendPaymentV2'`), and `TransactionsStore.sendPaymentInternal` passed it through untouched, leaving result handling to whichever caller remembered to attach a `NativeEventEmitter` listener. Only the Send and PaymentRequest views did, via byte-identical `subscribePayment` copies.

Consequences on an LNC backend:

- **NWC wallet service** (`supportsNostrWalletConnectService` is true for LNC): `NostrWalletConnectStore` discards `sendPayment`'s return value, so no listener is ever attached. Every NWC payment, including instantly settled ones, ran out the full 120s window and answered in-transit with an empty preimage; `TransactionsStore.loading` stayed stuck true until the next reset. Every payment's amount also escaped the pending-budget check window until reconcile, so this fix shrinks the residual LNC exposure of the pending-budget work (#4305) to genuinely slow payments only.
- **Donation flow**: `sendPaymentSilently` had the same pass-through branch. `SendingLightning` awaited it, received the event-name string, misread it as success, and computed `result.amount_msat / 1000` = NaN.
- The view listeners themselves leaked (never called `.remove()`), passed no payment sequence (so they could clear another payment's in-flight guard from #4301), and called `handlePaymentError` with a raw string instead of an `Error`.

## What changed

- `backends/LightningNodeConnect.ts`: `payLightningInvoice` now resolves with the payment's terminal result, the same promise contract as every other backend. All concurrent `SendPaymentV2` calls share one event channel with no request correlation, so events are matched to their payment by `payment_hash`: read from the keysend request, or decoded from the bolt11 invoice. AMP payments skip correlation (lnd derives per-attempt hashes) and fall back to first-terminal-event, the previous behavior. If no terminal event arrives within `timeout_seconds` plus grace, the promise resolves with the designed in-transit `payment_error` shape rather than rejecting, since a hard failure invites a retry that double-sends on keysend. The subscription is always removed once settled.
- `utils/LncPayUtils.ts` (new): pure per-event classification (EOF/IN_FLIGHT/hash-mismatch skip, terminal, error) and expected-hash derivation, with unit tests.
- `stores/TransactionsStore.ts`: both LNC special-case branches deleted; LNC now flows through the generic `handlePayment(result, seq)` / `handlePaymentError(err, seq)` path.
- `views/Send.tsx`, `views/PaymentRequest.tsx`: dead `subscribePayment` copies, listener fields, and the bogus `.stop()` cleanup removed. `views/Tools/Sweep.tsx` carried a third, never-invoked copy, also removed.
- `views/Tools/Rebalance.tsx`: intentionally untouched. It calls the backend directly but only used its own subscriber when the result was a string, which no longer happens; the resolved object flows into its existing `handleLncPaymentResult`. The guard branch is retained as defensive code.

Known caveat (status quo, not a regression): two concurrent payments that both lack a derivable hash (e.g. two AMP payments) both accept the first terminal event. Uncorrelated native error strings on the shared channel are attributed to all listening payments.

Keysend on LNC flows through the same `payLightningInvoice` path, so one fix covers both.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Pending. Planned matrix before ready-for-review:

- LNC user payment: bolt11 success/failure/timeout, keysend, AMP keysend (exercises the no-correlation fallback)
- NWC-over-LNC: instant-settle payment returns a real preimage instead of in-transit after 120s; budget settles
- Rebalance on LNC (direct backend caller)
- Donation flow on LNC (amount no longer NaN)
- Regression pass on LND (REST)

Pending, see above.

### DIFF
--- a/backends/LightningNodeConnect.ts
+++ b/backends/LightningNodeConnect.ts
@@ -401,7 +401,7 @@ export default class LightningNodeConnect {
                     } else if (decision.kind === 'error') {
                         settle(() => reject(decision.error));
                     }
-                    // 'ignore': EOF, IN_FLIGHT, or another payment's event
+                    // 'ignore': EOF, a non-terminal status, or another payment's event
                 }
             );
         });
--- a/backends/LightningNodeConnect.ts
+++ b/backends/LightningNodeConnect.ts
@@ -1,4 +1,8 @@
-import { NativeModules, NativeEventEmitter } from 'react-native';
+import {
+    NativeModules,
+    NativeEventEmitter,
+    EmitterSubscription
+} from 'react-native';
 
 import LNC from '../zeus_modules/@lightninglabs/lnc-rn';
 import { lnrpc, walletrpc } from '../zeus_modules/@lightninglabs/lnc-core';
@@ -10,6 +14,11 @@ import OpenChannelRequest from '../models/OpenChannelRequest';
 
 import Base64Utils from '../utils/Base64Utils';
 import { snakeize } from '../utils/DataFormatUtils';
+import {
+    decideLncPayEvent,
+    deriveExpectedPaymentHash
+} from '../utils/LncPayUtils';
+import { localeString } from '../utils/LocaleUtils';
 import { toLnrpcAddressType } from '../utils/LndUtils';
 import VersionUtils from '../utils/VersionUtils';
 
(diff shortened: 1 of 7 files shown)


kaloudis:
## Device testing update

Environment: iOS Simulator (iPhone 17 Pro, iOS 26.4), Polar regtest network. LNC wallet connected through litd 0.16.0-alpha; peers running LND 0.20.0-beta, bitcoind 30.0.

Completed from the planned matrix:

- [x] LNC bolt11 payment, success path
- [x] LNC bolt11 payment, failure path (no route): error surfaced through handlePaymentError, no stuck loading state
- [x] LNC keysend (payment_hash correlation from the request)
- [x] LNC AMP keysend (exercises the no-correlation first-terminal-event fallback)
- [x] LNC hodl invoice timeout: payment resolved with the in-transit shape after timeout_seconds plus grace, no hang and no hard rejection
- [x] Rebalance on LNC (direct backend caller, existing handleLncPaymentResult path)
- [x] Regression pass on LND 0.20.0-beta (REST), bolt11 payment

Remaining, to be tested on mainnet:

- [ ] NWC-over-LNC instant settle (real preimage returned promptly, pending budget settles)
- [ ] Donation flow on LNC (amount no longer NaN)

shubhamkmr04 (commented):
`decideLncPayEvent` treats everything except `IN_FLIGHT` as terminal, but `INITIATED` (lightning.proto:4359 — "created and has not attempted any HTLCs") is non-terminal, and lnd streams it on `SendPaymentV2` whenever `no_inflight_updates` is false — our default on LNC, since we only set it for Tor and AMP. The old view listener had the same test but never removed the listener, so the later `SUCCEEDED` still corrected the screen; here `settle()` removes the subscription and resolves on `INITIATED`, so the real result is lost. `handlePayment` then sets `error = true` with an empty `payment_error`, and NWC answers `FAILED_TO_PAY_INVOICE` for a payment that settles. Suggest an explicit non-terminal set, keeping the "anything else is terminal" fallback:

```ts
const NON_TERMINAL = new Set(['IN_FLIGHT', 'INITIATED', 'UNKNOWN']);
if (NON_TERMINAL.has(result.status)) return { kind: 'ignore' };

kaloudis:
Good catch, thanks. Confirmed against the vendored proto (INITIATED = 4, "created and has not attempted any HTLCs") and TransactionsStore: we only set no_inflight_updates for Tor and AMP, so the default LNC path streams it. The old view listener never removed its subscription, so the later SUCCEEDED self-corrected; with settle() tearing the subscription down, the misclassification would have been permanent, and NWC-over-LNC would have answered FAILED_TO_PAY_INVOICE for a payment that settles.

Fixed in d1f0d0f18 with your suggested non-terminal set (UNKNOWN, IN_FLIGHT, INITIATED), keeping the anything-else-is-terminal fallback, plus jest cases for INITIATED and UNKNOWN. Note the timeout backstop already covers a stream that dies after only non-terminal events, so ignoring INITIATED can't strand the promise.

shubhamkmr04 (approved):
tACK

PR #4334, opened 2026-08-04
fix(nwc): debit budget on settled payments missing from the payments list

kaloudis:
# Description

Fixes the remaining missed-accounting path from external security report FND-004 (NWC budget enforcement).

The report's headline finding, a check-then-act race where N parallel `pay_invoice` requests could each pass the budget check before any debits landed, was already closed by #4303 (per-service payment queue) and #4305 (pending amounts reserved inside `canSpend`). Verification against master also showed its reconcile-miss claim is now fail-closed rather than an escape: an unreconciled pending activity stays counted in `pendingSpendSats` indefinitely.

One claim survived verification as a real residual hole. In `handleLightningPayInvoice`, `finalizePayment`, the only place a settled spend is debited via `trackSpending`, was gated on finding the settled payment in the refreshed payments list:

```ts
if (payment) {
    await this.finalizePayment({ ... });
}
```

If the payment settled (preimage in `TransactionsStore`) but the list lookup missed, the handler returned success with a valid preimage while recording **no activity at all**: neither `totalSpendSats` nor `pendingSpendSats` ever saw the amount, so the spend permanently escaped the connection budget. Miss windows: payments lists windowed at 100 entries on busy nodes, node indexing lag immediately after settlement, backends with incomplete `listPayments` (CLN's missing-entries lists are a known issue), or a failed refresh. The cashu path had the same `if (payment)` gate after the melt had already been confirmed successful.

## What changed

- `stores/NostrWalletConnectStore.ts`: budget accounting no longer depends on the payments-list lookup. The lightning success path finalizes whenever a preimage is in hand (`payment || preimage`); the cashu path finalizes unconditionally once the melt has succeeded. When the list lookup misses, the activity record is synthesized from request-side state (invoice, payment hash, preimage, amount, fee).
- `utils/NostrConnectUtils.ts`: new `buildSettledPaymentFallback()` builds that settled `Payment`/`CashuPayment` record. Passing `decoded: null` into `finalizePayment` was not an option; the `BaseModel` constructor throws on null data.
- `utils/NostrConnectUtils.test.ts`: unit tests for the fallback builder (lightning and cashu shapes, settled-payment semantics, optional-field tolerance).

When the lookup succeeds, behavior is byte-identical to before. The fallback only fires in the window where the spend previously escaped accounting.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Pending. Planned before ready-for-review:

- NWC pay_invoice happy path (list lookup succeeds): unchanged behavior, budget debited once
- Forced list miss (busy node or CLN backend): success response still debits budget and records a success activity with correct amount/preimage
- Cashu pay_invoice with and without the melt appearing in cashu payments
- Budget exhaustion after repeated payments confirms no double-counting with reconcile

Pending, see above.

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2050,11 +2050,25 @@ export default class NostrWalletConnectStore {
                 paymentHash
             );
 
-        if (payment) {
+        // Debit the budget on any evidence of settlement, not only when the
+        // payments-list lookup succeeds: with a preimage in hand the payment
+        // definitely settled, and skipping finalizePayment on a list miss
+        // would let the spend escape the budget permanently (it records no
+        // activity, so pendingSpendSats never sees it either).
+        if (payment || preimage) {
             await this.finalizePayment({
                 id: request.invoice,
                 type: 'pay_invoice',
-                decoded: payment,
+                decoded:
+                    payment ||
+                    NostrConnectUtils.buildSettledPaymentFallback({
+                        invoice: request.invoice,
+                        payment_source: 'lightning',
+                        amountSats,
+                        preimage,
+                        feeSats: fees_paid,
+                        paymentHash
+                    }),
                 payment_source: 'lightning',
                 connection,
                 amountSats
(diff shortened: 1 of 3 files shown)


ajaysehwal (approved):
tACK

PR #4335, opened 2026-08-04
fix(nwc): expire unresolvable pending payments after the HTLC lifetime

kaloudis:
# Description

Stacked on #4334 (its commit is included here; only the second commit is new). Together they close out the accounting model behind NWC connection budgets.

## The bug

A pending `pay_invoice` activity whose payment never appears in the node's payments list is stuck forever:

- reconciliation (`reconcilePendingPayInvoiceActivities`) skips activities with no matching list entry
- `pendingSpendSats` derives from activity status, so the stuck pending counts against the budget indefinitely
- budget renewal resets only `totalSpendSats`, so the reservation survives every renewal period
- there is no expiry, no pruning, and no manual release

Net effect: the connection's usable budget is permanently reduced by the stuck amount. This is the fail-closed inverse of the accounting escape fixed in #4334, and was flagged during triage of the same external report (FND-004).

## The fix

A dispatched Lightning payment cannot remain genuinely in flight past its route's CLTV limit (lnd caps outgoing CLTV at 2016 blocks, roughly 14 days). Reconciliation now runs an expiry pass for pendings older than that bound:

- a definitive wide-window fetch (2000 payments, `BackendUtils.getPayments` directly so the wallet activity list and its loading flag are untouched) gets one real look at the node
- a payment that turns up is resolved through the normal reconcile path (success debits the budget; failed releases). If the node still reports it in flight (zombie records exist on lnd), the node is the authority and the reservation holds
- only a payment still missing after that look is marked failed, releasing its reservation, with an explanatory error on the activity
- a failed fetch skips the pass entirely: a reservation can never be released without evidence

On LND-family backends the payments list includes failed and in-flight entries (`include_incomplete=true`), so a wide-window miss at 14 days means the payment never left the node or failed long ago; expiring it is correct in the overwhelming case. The accepted residual: a settled payment the node's list permanently omits is released 14 days late, bounded to one payment.

Implementation notes:

- the per-activity resolution logic is extracted into `applyPendingPayInvoiceResolution` and shared by the regular loop and the expiry pass
- the threshold and age check live in `NostrConnectUtils` (`PENDING_PAY_INVOICE_EXPIRY_MS`, `isPendingPayInvoiceExpired`) with unit tests; activities of unknown age never expire
- the failure message deliberately avoids `isIgnorableError`'s substrings ("has expired" etc.) so the activity-cleanup filter does not prune the audit trail; pinned by a test
- one new `en.json` string

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Pending. Planned before ready-for-review (on top of the #4334 matrix):

- Seed a pending activity with a backdated createdAt (>14d), payment absent from the node: reconcile flips it to failed and remainingBudget increases
- Same, payment present and settled on the node: reconcile flips to success and debits (no expiry)
- Same, payment reported in flight by the node: stays pending, reservation holds
- Backend fetch failing (airplane mode): expiry pass is skipped, pending untouched

Pending, see above.

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2050,11 +2050,25 @@ export default class NostrWalletConnectStore {
                 paymentHash
             );
 
-        if (payment) {
+        // Debit the budget on any evidence of settlement, not only when the
+        // payments-list lookup succeeds: with a preimage in hand the payment
+        // definitely settled, and skipping finalizePayment on a list miss
+        // would let the spend escape the budget permanently (it records no
+        // activity, so pendingSpendSats never sees it either).
+        if (payment || preimage) {
             await this.finalizePayment({
                 id: request.invoice,
                 type: 'pay_invoice',
-                decoded: payment,
+                decoded:
+                    payment ||
+                    NostrConnectUtils.buildSettledPaymentFallback({
+                        invoice: request.invoice,
+                        payment_source: 'lightning',
+                        amountSats,
+                        preimage,
+                        feeSats: fees_paid,
+                        paymentHash
+                    }),
                 payment_source: 'lightning',
                 connection,
                 amountSats
@@ -2205,16 +2219,26 @@ export default class NostrWalletConnectStore {
 
         // CashuStore does not surface IN_FLIGHT melts yet; add recordPendingPayment
         // plus getActivities cashu reconciliation when it does.
-        if (payment) {
-            await this.finalizePayment({
-                id: request.invoice,
-                decoded: payment,
-                type: 'pay_invoice',
-                payment_source: 'cashu',
-                amountSats,
-                connection
-            });
-        }
+        //
+        // The melt succeeded (failure returned above), so debit the budget
+        // unconditionally: gating on the payments-list lookup would let the
+        // spend escape the budget when the list misses the fresh melt.
+        await this.finalizePayment({
+            id: request.invoice,
+            decoded:
+                payment ||
+                NostrConnectUtils.buildSettledPaymentFallback({
+                    invoice: request.invoice,
+                    payment_source: 'cashu',
+                    amountSats,
+                    preimage: cashuInvoice.getPreimage,
+                    feeSats: cashuInvoice.fee
+                }),
+            type: 'pay_invoice',
+            payment_source: 'cashu',
+            amountSats,
+            connection
+        });
 
         return {
             result: {
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2377,39 +2377,123 @@ export default class NostrWalletConnectStore {
         );
         let changed = false;
         for (const activity of lightningPending) {
-            const payment = payments.find(
-                (p) =>
-                    p.getPaymentRequest === activity.id ||
-                    (!!activity.paymentHash &&
-                        p.paymentHash === activity.paymentHash)
+            const payment = NostrConnectUtils.findPaymentForInvoice(
+                activity.id,
+                payments,
+                activity.paymentHash
             );
             if (!payment) continue;
 
-            let reconciled = false;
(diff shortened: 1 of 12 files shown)


stores/NostrWalletConnectStore.ts
```diff
@@ -2353,41 +2377,125 @@ export default class NostrWalletConnectStore {
...
+
+    // Definitive payments fetch for the pending-expiry pass: wide window,
+    // direct from the backend (paymentsStore.getPayments would overwrite
+    // the wallet activity list and toggle its loading flag). Returns null
+    // on failure so the caller cannot expire anything without a real look.
+    private async getPaymentsForPendingExpiryCheck(): Promise<
+        Payment[] | null
+    > {
+        try {
+            const data = await BackendUtils.getPayments({
```
ajaysehwal, line 2488: @kaloudis I think there's one edge case around pending payment expiry.

`getPaymentsForPendingExpiryCheck()` assumes the backend can return a complete payment history (including incomplete payments), but that isn't true for all backends. For example, CLN currently only returns a limited set of payments, and LndHub/LDK have similar limitations.
Because of this, an older successful payment may no longer appear in the lookup. After the pending entry expires (14 days), it could be marked as failed, `pendingSpendSats` is released, and `totalSpendSats` is never updated.
That would effectively free the reserved budget even though the payment actually succeeded, allowing the connection to spend more than its budget.
I think we should only expire pending payments when the backend can reliably provide a complete payment history. Otherwise, it may be safer to keep them pending (or use a backend-specific strategy) rather than assuming they're failed.

kaloudis: Good catch, thanks. You were right, and CLN was worse than described: our CLN getPayments is a SQL query hard-capped at 150 rows that silently ignores the maxPayments argument, so the "wide" fetch was never wide there. LndHub's /gettxs takes no filters and only returns settled payments. LDK Node is actually fine: listPayments returns the node's complete, never-auto-pruned payment store.

Pushed c531b82c4, which makes the confirmation backend-aware instead of gating the feature off wholesale:

- CLN rest and LDK Node confirm each candidate with a point query (new supportsPaymentLookupByHash / lookupPaymentByHash). CLN's listpays with a payment_hash reads the full sendpays index with no row limit; LDK filters its complete store. An empty result is proof of absence.
- LND, embedded LND, and LNC (on lnd >= 0.15.1) fetch one window with creation_date_start set a day before the oldest candidate, so the range provably contains it. A response that fills the 2000 cap may have been truncated and is treated as inconclusive (new supportsPaymentsCreationDateFilter, explicitly overridden to false in LndHub so it does not inherit LND's).
- Backends with no reliable strategy (LndHub) skip the pass entirely and stay fail-closed, exactly as before this PR.

Any fetch failure, missing/underivable payment hash, or saturated window keeps the reservation held, so nothing can be expired without a definitive look.

kaloudis:
Rebased on master (988637464), reconciling with #4435 which merged in the meantime. The helper this PR introduced (applyPendingPayInvoiceResolution) was dropped in favor of the reviewed successor that landed with #4435 (applyPayInvoiceReconcile); the expiry pass now calls that instead, so there is a single resolution path. Everything unique to this PR is retained: the 14-day HTLC-lifetime bound, the definitive backend-aware wide fetch before any reservation is expired, and the tests.

One thing worth discussing at review: master's isPayInvoiceHoldAbandoned (from #4435) already releases unresolved holds at 7 days after a node lookup, so most candidates will be abandoned before the 14-day pass fires. The remaining value of the expiry pass is the definitive unthrottled fetch, which can resolve a payment the regular throttled list misses instead of failing it. Happy to tighten the two layers into one if preferred. yarn verify green (1425 tests).

PR #4336, opened 2026-08-05
fix: stop logging sensitive payment and credential material

kaloudis:
# Description

Removes log statements that exposed secrets in release builds.

**Fixes**
- `stores/SwapStore.ts`: logged the whole persisted swap object, including `refundPrivateKey`, the ECPair private key, and the reverse-swap preimage (fund-loss grade). Now logs swap id + status only.
- `stores/LightningAddressStore.ts`: logged the raw Cashu bearer token from the ZEUS Pay redeem endpoint (spendable ecash). Now logs the quote id only.
- `views/Swaps/SwapDetails.tsx`: logged Boltz claim details including the payment preimage.
- `lndmobile/index.ts`: the keysend `sendPaymentV2` path logged the raw stream event and the decoded `lnrpc.Payment` (`payment_preimage`, full route hops) on every payment. Aligned with the non-keysend path, which never logged these.
- `android LncModule.kt`: logged the LNC pairing phrase (a bearer credential for the node) on connect, and every LNC RPC request body via `invokeRPC` (invoices, destinations, amounts, PSBTs) to logcat. Both removed. The iOS module never logged them.
- `PushNotificationManager.tsx`: device token and full notification payload logs are now gated behind `__DEV__` (payload content is server-controlled).

**Not in this PR (intentionally)**
- Build-level backstops (babel `transform-remove-console` for release bundles, ProGuard `Log.v/d/i` strip + R8 optimization switch): split out per review feedback, tracked in #4345. The ProGuard piece needs keep rules for `zeus-cashu-restore.aar` and `react-native-nitro-tor` plus a reproducible-build re-verification before it can ship.
- `CashuDevKitModule` logcat dumps (mint quote response etc.): deferred to land on top of #4211, converting them to its redacted `logToFile` channel instead of deleting the signal.
- Medium-tier object dumps (LSPStore, ContactStore, ChannelsStore, route hints, Boltz WS frames): explicit gating can follow separately; the release strip in #4345 will cover them in release builds.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

> **Draft note:** device testing pending. Required before merge: (1) LNC connect + RPC on Android to confirm the removed logs broke nothing; (2) a keysend payment on embedded LND; (3) a swap refund path hit to exercise the SwapStore change; (4) `adb logcat` spot check that the LNC pairing phrase and request bodies no longer appear.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/PushNotificationManager.tsx
+++ b/PushNotificationManager.tsx
@@ -22,7 +22,7 @@ export default class PushNotificationManager extends React.Component<any, any> {
         Notifications.events().registerRemoteNotificationsRegistered(
             (event) => {
                 const deviceToken = event.deviceToken;
-                console.log('Device Token Received', deviceToken);
+                if (__DEV__) console.log('Device Token Received', deviceToken);
                 lightningAddressStore.setDeviceToken(deviceToken);
             }
         );
@@ -52,7 +52,11 @@ export default class PushNotificationManager extends React.Component<any, any> {
     registerNotificationEvents = () => {
         Notifications.events().registerNotificationReceivedForeground(
             (notification, completion) => {
-                console.log('Notification Received - Foreground', notification);
+                if (__DEV__)
+                    console.log(
+                        'Notification Received - Foreground',
+                        notification
+                    );
                 // Don't display redeem notification if auto-redeem is on
                 if (
                     settingsStore.settings?.lightningAddress
@@ -83,7 +87,11 @@ export default class PushNotificationManager extends React.Component<any, any> {
 
         Notifications.events().registerNotificationOpened(
             (notification, completion) => {
-                console.log('Notification opened by device user', notification);
+                if (__DEV__)
+                    console.log(
+                        'Notification opened by device user',
+                        notification
+                    );
                 console.log(
                     `Notification opened with an action identifier: ${notification.identifier}`
                 );
@@ -94,7 +102,11 @@ export default class PushNotificationManager extends React.Component<any, any> {
 
         Notifications.events().registerNotificationReceivedBackground(
             (notification, completion: any) => {
-                console.log('Notification Received - Background', notification);
+                if (__DEV__)
+                    console.log(
+                        'Notification Received - Background',
+                        notification
+                    );
                 // Calling completion on iOS with `alert: true` will present the native iOS inApp notification.
                 completion({ alert: true, sound: true, badge: false });
             }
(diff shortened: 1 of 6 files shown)


shubhamkmr04 (commented):
Would it make sense to split the ProGuard change into its own PR? One thing worth knowing: R8 ignores the -optimizations line in proguard-android-optimize.txt, so switching to that file turns on full optimization rather than the safe subset the file's comments imply. That's usually fine, but it can break FFI paths quietly instead of crashing at boot — zeus-cashu-restore.aar ships no consumer proguard.txt (unlike Lndmobile.aar and cashudevkit.aar), and react-native-nitro-tor has no keep rules yet. It also changes the APK output, so the reproducible-build hashes would need re-verifying.
Since commit 1 already removes the actual leaks, the Log.d rule is more of a future safeguard. Landing them separately would let the redaction fixes merge now and keep any release regression easy to trace back.

package.json
```diff
@@ -176,6 +176,7 @@
     "@typescript-eslint/eslint-plugin": "8.65.0",
     "@typescript-eslint/parser": "8.65.0",
     "babel-jest": "29.7.0",
+    "babel-plugin-transform-remove-console": "^6.9.4",
```
shubhamkmr04: can we lock the version

kaloudis:
> Would it make sense to split the ProGuard change into its own PR? One thing worth knowing: R8 ignores the -optimizations line in proguard-android-optimize.txt, so switching to that file turns on full optimization rather than the safe subset the file's comments imply. That's usually fine, but it can break FFI paths quietly instead of crashing at boot — zeus-cashu-restore.aar ships no consumer proguard.txt (unlike Lndmobile.aar and cashudevkit.aar), and react-native-nitro-tor has no keep rules yet. It also changes the APK output, so the reproducible-build hashes would need re-verifying. Since commit 1 already removes the actual leaks, the Log.d rule is more of a future safeguard. Landing them separately would let the redaction fixes merge now and keep any release regression easy to trace back.

Not a bad point. Reeled this in and documented an issue for a follow-up in https://github.com/ZeusLN/zeus/issues/4345

shubhamkmr04 (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4340, opened 2026-08-05
fix(nwc): include routing and melt fees in connection budget

ajaysehwal:
# Description

NWC connection budgets previously tracked only the invoice amount for `pay_invoice`, so Lightning routing fees and Cashu melt fees never counted against the limit. Clients could overspend the configured budget by the fee on every payment.
This PR includes those fees in budget accounting and shows them on connection activityThis PR includes those fees in budget accounting and surfaces them in connection activity.

### Changes

- **Budget debit** — track `amountSats + feeSats` (via `NostrConnectUtils.resolveFeeSats`) on Lightning settle, Cashu melt success, and pending → success reconcile
- **Fee normalization** — `resolveFeeSats` rounds fractional sats (common after msat ÷ 1000); invalid / non-positive values become `0`
- **Stream fee fallback** — preserve stream fee when payment-list fee is unavailable (avoid `'0'` truthiness dropping the fee)
- **Activity** — store `satAmount` and `fees_paid` on pay activities
- **UI** — show amount + fee on NWC connection activity when a fee is present
- **Warnings** — sync budget limit UI after spend so Limit Exceeded stays accurate
- **Docs** — note that `validateBudgetBeforePayment` checks invoice amount only; a connection can exceed budget by up to the final payment’s fee (bounded by `PAYMENT_FEE_LIMIT_SATS`)

### Test plan

- [ ] Lightning pay with routing fee → budget includes amount + fee; activity shows both
- [ ] Cashu melt with nonzero fee → budget and activity include fee
- [ ] Pending pay that settles later → fee debited once on reconcile
- [ ] Budget near limit → fee can push over; next `pay_invoice` is rejected
- [ ] Limit Exceeded badge updates after spend crosses the budget

<table>
  <tr>
    <th>Lightning Payment Activity</th>
    <th>Cashu Payment Activity</th>
  </tr>
  <tr>
    <td align="center">
      Lightning Payment Activity
    </td>
    <td align="center">
      Cashu Payment Activity
    </td>
  </tr>
</table>

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1195,6 +1195,9 @@ export default class NostrWalletConnectStore {
         if (index !== -1) {
             this.connections[index] = connection;
         }
+        runInAction(() => {
+            connection.syncWarnings(this.maxBudgetLimit);
+        });
         if (save) {
             this.scheduleSave();
         }
@@ -2422,7 +2425,10 @@ export default class NostrWalletConnectStore {
                     const spendSats =
                         amountSats + NostrConnectUtils.resolveFeeSats(feeSats);
                     if (spendSats > 0) {
-                        connection.trackSpending(spendSats);
+                        connection.trackSpending(
+                            spendSats,
+                            this.maxBudgetLimit
+                        );
                     }
                 } else if (payment.isFailed) {
                     activity.status = 'failed';
@@ -2718,7 +2724,10 @@ export default class NostrWalletConnectStore {
         // Budget is whole sats; round fractional fees (0.026 → 0, 1.999 → 2).
         const budgetFeeSats = NostrConnectUtils.resolveFeeSats(feeSats);
         runInAction(() => {
-            connection.trackSpending(amountSats + budgetFeeSats);
+            connection.trackSpending(
+                amountSats + budgetFeeSats,
+                this.maxBudgetLimit
+            );
             connection.activity.push({
                 id,
                 type,
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2068,6 +2068,8 @@ export default class NostrWalletConnectStore {
                 paymentHash
             );
 
+        const feeSats = Number(payment?.getFee || fees_paid) || 0;
+
         // Debit the budget on any evidence of settlement, not only when the
         // payments-list lookup succeeds: with a preimage in hand the payment
         // definitely settled, and skipping finalizePayment on a list miss
@@ -2089,16 +2091,15 @@ export default class NostrWalletConnectStore {
                     }),
                 payment_source: 'lightning',
                 connection,
-                amountSats
+                amountSats,
+                feeSats
             });
         }
 
         return {
             result: {
                 preimage: preimage || payment?.getPreimage || '',
-                fees_paid: satsToMillisats(
-                    Number(fees_paid) || Number(payment?.getFee) || 0
-                )
+                fees_paid: satsToMillisats(feeSats)
             },
             error: undefined
         };
@@ -2241,6 +2242,7 @@ export default class NostrWalletConnectStore {
         // The melt succeeded (failure returned above), so debit the budget
         // unconditionally: gating on the payments-list lookup would let the
         // spend escape the budget when the list misses the fresh melt.
+        const feeSats = Number(cashuInvoice.getFee || payment?.getFee || 0);
         await this.finalizePayment({
             id: request.invoice,
             decoded:
@@ -2255,6 +2257,7 @@ export default class NostrWalletConnectStore {
             type: 'pay_invoice',
             payment_source: 'cashu',
             amountSats,
+            feeSats,
             connection
         });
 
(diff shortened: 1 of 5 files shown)


kaloudis (commented):
Concept ACK. This closes a real gap: routing and melt fees previously escaped the connection budget entirely. The mechanics look right, both `finalizePayment` call sites are covered, the reconcile path debits each pending activity exactly once inside `runInAction`, activity records now carry `satAmount`/`fees_paid`, and the list_transactions mapper already treats `fees_paid` as sats and converts to msats. UI matches the existing Activity.tsx amount + fee pattern. `resolveFeeSats` test coverage is solid.

Two things I'd like resolved before merge, both commented inline:
1. The `getFee || fees_paid` ordering on line 2071 makes the stream-fee fallback unreachable due to string `'0'` truthiness, which can drop the fee when the payments list is stale at settlement time.
2. The PR description says ceil but the code rounds; pick one.

One design note, no change requested: `validateBudgetBeforePayment` still checks the invoice amount only, so a connection can exceed its budget by up to the final payment's fee (bounded by `PAYMENT_FEE_LIMIT_SATS`) before the debit blocks subsequent payments. Probably the right tradeoff since the fee is unknown upfront, but worth a comment in the code.

Testing: checklist shows iOS + LND REST only. Before merge I'd want to see a Cashu melt with a nonzero fee, an Android pass, and a budget-boundary check (budget nearly exhausted, verify the fee pushes the connection over and the next pay_invoice is rejected).

stores/NostrWalletConnectStore.ts
```diff
@@ -2068,6 +2068,8 @@ export default class NostrWalletConnectStore {
                 paymentHash
             );
 
+        const feeSats = Number(payment?.getFee || fees_paid) || 0;
```
kaloudis: `Payment.getFee` returns the string `'0'` in its no-fee branches (including when `fee_sat` is `'0'`), and a string `'0'` is truthy, so `fees_paid` from `transactionsStore.payment_fee` is unreachable whenever a list payment was found. The payments list here was fetched immediately after `waitForPaymentCompletion`, and `findPaymentForInvoice` on line 2065 doesn't require the match to be settled, so a just-settled payment can match a stale list entry whose fee isn't final yet. In that case the real fee from the stream gets dropped: the budget misses it (the case this PR is closing) and the NIP-47 `fees_paid` response reports 0 where master reported the stream fee.

Suggest:
```ts
const feeSats = Number(payment?.getFee) || Number(fees_paid) || 0;
```
which prefers the list fee when it's nonzero and still falls back to the stream fee.

stores/NostrWalletConnectStore.ts
```diff
@@ -2241,6 +2242,7 @@ export default class NostrWalletConnectStore {
         // The melt succeeded (failure returned above), so debit the budget
         // unconditionally: gating on the payments-list lookup would let the
         // spend escape the budget when the list misses the fresh melt.
+        const feeSats = Number(cashuInvoice.getFee || payment?.getFee || 0);
```
kaloudis: Same string-`'0'` short-circuit as the lightning path: `CashuPayment.getFee` returns `'0'` when empty, which blocks the `payment?.getFee` fallback. Both would usually be 0 anyway, but for consistency with line 2071 (and to keep an NaN guard on the outside):
```ts
const feeSats = Number(cashuInvoice.getFee) || Number(payment?.getFee) || 0;
```
Also worth aligning the fallback on line 2255, which still reads `cashuInvoice.fee` while this line uses `getFee`. Same value today, but one accessor is easier to audit.

utils/NostrConnectUtils.ts
```diff
@@ -1201,6 +1201,16 @@ export default class NostrConnectUtils {
...
     }
 
+    /**
+     * Whole-sat budget debit for a fee that may be fractional (msat / 1000).
+     * Round to nearest sat; do not use for NIP-47 fees_paid.
+     */
+    static resolveFeeSats(fee: string | number | null | undefined): number {
+        const raw = Number(fee);
+        if (!Number.isFinite(raw) || raw <= 0) return 0;
+        return Math.round(raw);
```
kaloudis, line 1211: The PR description says this "ceils fractional sats ... so fees are never undercounted", but the implementation, docblock, and tests all round to nearest, so a 1.4 sat fee debits 1 sat. Either is defensible at sub-sat scale, but pick one: if the never-undercount property is the intent, make this `Math.ceil` and update the tests (`1.2` -> 2, `0.026` -> 1); otherwise fix the PR description. Right now a reader of the description gets the wrong model of the code.

stores/NostrWalletConnectStore.ts
```diff
@@ -2715,7 +2726,9 @@ export default class NostrWalletConnectStore {
                         ? new CashuPayment(decoded)
                         : new Payment(decoded),
                 status: 'success',
-                payment_source
+                payment_source,
+                satAmount: amountSats,
+                fees_paid: feeSats
```
kaloudis: Note for future readers: `ConnectionActivity.fees_paid` holds sats (possibly fractional), while NIP-47 `fees_paid` is msats. The list_transactions mapper converts correctly and the resolveFeeSats docblock warns about it, so nothing to fix here, but if this gets touched again a `feeSats`-style name on the activity field would remove the trap.

kaloudis (approved):
ACK

PR #3216, opened 2025-08-16
[feat] Nostr Wallet Connect (NWC) service

kaloudis:
# Description

Relates to issue: [ZEUS-3095](https://github.com/ZeusLN/zeus/issues/3095)

_Please enter a description and screenshots, if appropriate, of the work covered in this PR_

Screenshot 2025-08-12 at 6 05 12 AM
Screenshot 2025-08-14 at 10 01 21 PM
Screenshot 2025-08-14 at 1 14 59 PM
Screenshot 2025-08-12 at 6 08 48 AM
Screenshot 2025-08-14 at 10 11 11 PM

This pull request is categorized as a:

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

I have tested this PR on the following platforms (please specify OS version and phone model/VM):

- [ ] Android
- [x] iOS

I have tested this PR with the following types of nodes (please specify node version and API version where appropriate):

- [ ] Embedded LND
- [ ] LND (REST)
- [ ] LND (Lightning Node Connect)
- [ ] Core Lightning (CLNRest)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales
- [ ] I’ve added new locale text that requires translations
- [ ] I’m aware that new translations should be made on the ZEUS [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Third Party Dependencies and Packages

- [ ] Contributors will need to run `yarn` after this PR is merged in
- [ ] 3rd party dependencies have been modified:
    * verify that `package.json` and `yarn.lock` have been properly updated
    * verify that dependencies are installed for both iOS and Android platforms

### Other:

- [ ] Changes were made that require an update to the README
- [ ] Changes were made that require an update to onboarding

PR #4341, opened 2026-08-05
fix(nwc): rotate and persist client secret on relay change

ajaysehwal:
## Description

Fixes NWC relay-change secret rotation so the new pairing URL works, and routes edit-screen relay changes through `updateConnection` instead of delete + create.

### Problem

On relay change, `updateConnection` generated a fresh keypair but:

- Never persisted the new client key
- Never updated `connection.pubkey`
- Never revoked the old secret

The returned pairing URL was inert while the old credential kept working. The edit screen also used regenerate (delete + create), which wiped activity.

### Changes

- Persist the new client key, rebind `connection.pubkey`, delete the old key, then subscribe
- Run rotation **before** mutating the connection so failed attempts stay retryable
- If subscribe/save fails after durable rotation, still return the pairing URL (subscription recovers on restart)
- Ensure a wallet service exists for the new relay and publish service info if needed
- Drop unused relay services on delete and relay change
- Edit screen always uses `updateConnection`; show the new QR when a secret is rotated
- Preserve connection identity, budget, and activity across relay changes
- Details-screen **Regenerate Connection** stays a full re-pair (unchanged)
- Store tests for key store / delete / pubkey rebind on relay change

### Behavior note (expiry)

The old edit path went through `createConnection` with `params.id` / `params.createdAt`, which triggered `refreshExpiryForRegenerate`. Relay change now keeps the form’s expiry as-is. Intentional: relay change preserves connection identity rather than acting as a full regenerate.

### Test plan

- [ ] Relay change on a linked client → old secret dead, new QR pairs
- [ ] Name/budget-only edit → no rotation, no QR screen
- [ ] Create connection + details-screen regenerate still work
- [ ] Unit tests: relay change stores new key, deletes old, rebinds pubkey

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [x] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -701,13 +701,34 @@ export default class NostrWalletConnectStore {
                 updates.name = newName;
             }
             const oldRelayUrl = connection.relayUrl;
+            const oldPubkey = connection.pubkey;
             const newRelayUrl = updates.relayUrl;
-            const relayUrlChanged = newRelayUrl && newRelayUrl !== oldRelayUrl;
+            const relayUrlChanged = !!(
+                newRelayUrl && newRelayUrl !== oldRelayUrl
+            );
 
             if (relayUrlChanged) {
                 this.unsubscribeFromConnection(connectionId);
                 if (!this.nwcWalletServices.has(newRelayUrl)) {
-                    await this.initializeNWCWalletServices();
+                    this.nwcWalletServices.set(
+                        newRelayUrl,
+                        new NWCWalletService({
+                            relayUrls: [newRelayUrl]
+                        })
+                    );
+                }
+                if (!this.publishedRelays.has(newRelayUrl)) {
+                    const nwcWalletService =
+                        this.nwcWalletServices.get(newRelayUrl);
+                    if (
+                        nwcWalletService &&
+                        this.walletServiceKeys?.privateKey
+                    ) {
+                        await this.publishWalletServiceInfoWithRetry(
+                            nwcWalletService,
+                            newRelayUrl
+                        );
+                    }
                 }
             }
 
@@ -735,15 +756,32 @@ export default class NostrWalletConnectStore {
                 }
                 this.findAndUpdateConnection(connection);
             });
-            await this.subscribeToConnection(connection);
+
             if (relayUrlChanged) {
-                return {
-                    nostrUrl:
-                        this.generateConnectionSecret(newRelayUrl)
-                            .connectionUrl,
-                    success: true
-                };
+                const {
+                    connectionUrl,
+                    connectionPrivateKey,
+                    connectionPublicKey
+                } = this.generateConnectionSecret(newRelayUrl);
+
+                await this.storeClientKeys(
+                    connectionPublicKey,
+                    connectionPrivateKey
+                );
+                await this.deleteClientKeys(oldPubkey);
+
+                runInAction(() => {
+                    connection.pubkey = connectionPublicKey;
+                    this.findAndUpdateConnection(connection);
+                });
+
+                await this.subscribeToConnection(connection);
+                await this.saveConnections();
+
+                return { nostrUrl: connectionUrl, success: true };
             }
+
+            await this.subscribeToConnection(connection);
             return { success: true };
         } catch (error: any) {
             console.error('Failed to update NWC connection:', error);
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -632,6 +632,7 @@ export default class NostrWalletConnectStore {
                     )
                 );
             }
+            const relayUrl = connection.relayUrl;
             await this.deleteClientKeys(connection.pubkey);
             this.unsubscribeFromConnection(connectionId);
             this.lastPendingPaymentStatusFetchByConnection.delete(connectionId);
@@ -639,6 +640,7 @@ export default class NostrWalletConnectStore {
             runInAction(() => {
                 this.connections.splice(index, 1);
             });
+            this.releaseUnusedRelayService(relayUrl);
             await this.saveConnections();
         } catch (error: any) {
             runInAction(() => {
@@ -701,9 +703,10 @@ export default class NostrWalletConnectStore {
                 updates.name = newName;
             }
             const oldPubkey = connection.pubkey;
+            const oldRelayUrl = connection.relayUrl;
             const newRelayUrl = updates.relayUrl;
             const relayUrlChanged = !!(
-                newRelayUrl && newRelayUrl !== connection.relayUrl
+                newRelayUrl && newRelayUrl !== oldRelayUrl
             );
 
             // Fallible rotation work must finish before any connection mutation
@@ -790,6 +793,10 @@ export default class NostrWalletConnectStore {
                 }
                 this.findAndUpdateConnection(connection);
             });
+
+            if (relayUrlChanged) {
+                this.releaseUnusedRelayService(oldRelayUrl);
+            }
 
             if (relayUrlChanged && rotatedSecret) {
                 await this.subscribeToConnection(connection);
@@ -1493,6 +1500,35 @@ export default class NostrWalletConnectStore {
                 error
             );
         }
+    }
+
+    /**
+     * Close and drop a relay's NWCWalletService when no remaining connection
+     * references it, so unused websockets and publishedRelays entries do not
+     * linger after delete or relay change.
+     */
+    private releaseUnusedRelayService(relayUrl: string): void {
+        if (!relayUrl) return;
+        if (this.connections.some((c) => c.relayUrl === relayUrl)) {
+            return;
+        }
+
+        const service = this.nwcWalletServices.get(relayUrl);
+        if (service) {
+            try {
+                service.close();
+            } catch (error) {
+                console.warn(
+                    `NWC: Failed to close unused relay service ${relayUrl}:`,
+                    error
+                );
+            }
+        }
+
+        runInAction(() => {
+            this.nwcWalletServices.delete(relayUrl);
+            this.publishedRelays.delete(relayUrl);
+        });
     }
 
     private async unsubscribeFromAllConnections(): Promise<void> {
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -799,8 +799,18 @@ export default class NostrWalletConnectStore {
             }
 
             if (relayUrlChanged && rotatedSecret) {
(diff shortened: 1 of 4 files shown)


kaloudis:
Where did we introduce this regression @ajaysehwal?

ajaysehwal:
> Where did we introduce this regression @ajaysehwal?

It was introduced in https://github.com/ZeusLN/zeus/pull/3216/changes/0d09cdb565a715f5e869ea909ac6897490032c3f. This was one of the initial commits for the NWC service.

stores/NostrWalletConnectStore.ts
```diff
@@ -735,15 +756,32 @@ export default class NostrWalletConnectStore {
...
-                            .connectionUrl,
-                    success: true
-                };
+                const {
+                    connectionUrl,
+                    connectionPrivateKey,
+                    connectionPublicKey
+                } = this.generateConnectionSecret(newRelayUrl);
+
+                await this.storeClientKeys(
```
kaloudis: By the time we get here, `Object.assign(connection, updates)` above has already applied the new `relayUrl` and `findAndUpdateConnection` has scheduled a persist. If `generateConnectionSecret` or `storeClientKeys` throws (e.g. a storage write failure), we return `{ success: false }` with `connection.relayUrl` already flipped in memory.

On the user's retry, `relayUrlChanged` then computes as `false`, so rotation is silently skipped and the update "succeeds" without ever returning a `nostrUrl`. Since nothing reads `NWC_CLIENT_KEYS` back to re-derive a pairing URL for an existing connection, the only way out at that point is the details-screen regenerate, which drops the activity/budget history this PR is trying to preserve.

Suggest reordering so the fallible work happens before any mutation: generate the secret and `storeClientKeys` first, then apply `relayUrl` and `pubkey` together in one `runInAction`, then save and subscribe, with `deleteClientKeys(oldPubkey)` last (it's already best-effort and non-throwing). That keeps every failure point retryable.

stores/NostrWalletConnectStore.ts
```diff
@@ -701,13 +701,34 @@ export default class NostrWalletConnectStore {
...
+                    );
+                }
+                if (!this.publishedRelays.has(newRelayUrl)) {
+                    const nwcWalletService =
+                        this.nwcWalletServices.get(newRelayUrl);
+                    if (
+                        nwcWalletService &&
+                        this.walletServiceKeys?.privateKey
+                    ) {
+                        await this.publishWalletServiceInfoWithRetry(
```
kaloudis: `createConnection` wraps this same call in try/catch (warn and continue, plus a 500ms settle delay), but here a publish failure rejects the whole update. `retry()` defaults to 3 attempts and then throws, so a relay that accepts subscriptions but rejects the info event will fail the edit after the old subscription was already torn down above. The state isn't corrupted at this point so a retry works, but if the user backs out instead, the connection stays unsubscribed until restart. Suggest matching `createConnection`'s tolerant handling here.

stores/NostrWalletConnectStore.ts
```diff
@@ -701,13 +701,34 @@ export default class NostrWalletConnectStore {
...
+                            relayUrls: [newRelayUrl]
+                        })
+                    );
+                }
+                if (!this.publishedRelays.has(newRelayUrl)) {
+                    const nwcWalletService =
+                        this.nwcWalletServices.get(newRelayUrl);
+                    if (
+                        nwcWalletService &&
+                        this.walletServiceKeys?.privateKey
```
kaloudis, line 749: Minor consistency point: `createConnection` calls `await this.loadWalletServiceKeys()` when keys are missing before generating a secret. Here, missing keys silently skip the publish and then `generateConnectionSecret` throws later. Keys are loaded unconditionally at service init so this is unlikely to bite in practice, but adding the same guard keeps the two paths consistent. Could fold into the reorder from the other comment.

stores/NostrWalletConnectStore.ts
```diff
@@ -701,13 +701,34 @@ export default class NostrWalletConnectStore {
...
             const oldRelayUrl = connection.relayUrl;
+            const oldPubkey = connection.pubkey;
             const newRelayUrl = updates.relayUrl;
-            const relayUrlChanged = newRelayUrl && newRelayUrl !== oldRelayUrl;
+            const relayUrlChanged = !!(
+                newRelayUrl && newRelayUrl !== oldRelayUrl
+            );
 
             if (relayUrlChanged) {
                 this.unsubscribeFromConnection(connectionId);
```
kaloudis, line 734: Not blocking, and `deleteConnection` has the same gap, but after a relay change the old relay's `NWCWalletService` stays in `nwcWalletServices` with a live websocket, and `publishedRelays` keeps its entry, even when no remaining connection uses that relay. Worth a follow-up to drop services with zero remaining connections.

views/Settings/NostrWalletConnect/AddOrEditNWCConnection.tsx
```diff
@@ -618,7 +563,16 @@ export default class AddOrEditNWCConnection extends React.Component<
                     params
                 );
                 if (updated.success) {
-                    setTimeout(() => navigation.goBack(), 100);
+                    // Relay change rotates the client secret — show the new
```
kaloudis: nit: house style avoids em-dashes in comments and copy.

kaloudis (changes requested):
Concept ACK. The core fix is right: `subscribeToConnection` builds its keypair from `connection.pubkey`, so persisting the new key and rebinding the pubkey before subscribing is exactly what was missing, and rotation genuinely revokes the old secret. Preserving connection identity, budget, and activity across relay changes instead of delete+recreate is the right design, and keeping the details-screen regenerate as an intentional full re-pair is a sensible split.

Main ask before merge is the ordering issue flagged inline (mutation happens before the fallible rotation steps, which can leave a connection in a state where retry silently skips rotation and no new QR is ever shown). The publish error-handling inconsistency is worth aligning too; the rest are minor.

One behavior note worth confirming as intended:
- The old edit path went through `createConnection` with `params.id` and `params.createdAt` set, which triggered `refreshExpiryForRegenerate`. The new path keeps the form's expiry as-is on relay change. Probably desirable given the identity-preservation goal, but it is a behavior change and worth a line in the PR description.

Process: please fill in the PR template (category, `tsc`/`lint`/`prettier`/`test`, platforms tested). For this change I'd want at minimum: a relay change on a connection with an actually-linked client app (verify the old secret is dead and the new QR works), a name/budget-only edit (confirm no rotation and no QR screen), and create + details-screen regenerate regression checks, on at least one platform. A small store test asserting "relay change stores new key, deletes old, rebinds pubkey" would also guard against this regressing a third time.

kaloudis:
Thanks for the quick turnaround. The inline items all look good: rotation now happens before any mutation with the pubkey rebound in the same `runInAction`, the publish handling matches `createConnection`, the key-load guard is in, and `releaseUnusedRelayService` goes beyond what I asked by covering `deleteConnection` too. Four things before this is merge-ready:

1. **Store test.** Please add the store test asserting that a relay change stores the new client key, deletes the old one, and rebinds `connection.pubkey`. This behavior has regressed since #3216 and survived unnoticed until now, so I want it guarded before merge.

2. **Expiry behavior note.** The old edit path went through `createConnection` with `params.id`/`params.createdAt` set, which triggered `refreshExpiryForRegenerate`. The new path keeps the form's expiry as-is on relay change. I think that's the right call given the identity-preservation goal, but it is a behavior change: please add a line to the PR description documenting it.

3. **Template/testing.** `yarn run test` is still unchecked and the unit-tests question has no box checked. Please also confirm the manual matrix from my review: a relay change on a connection with an actually-linked client app (verify the old secret is dead and the new QR pairs), a name/budget-only edit (no rotation, no QR screen), and create + details-screen regenerate regression checks.

4. **Residual ordering edge case.** The reorder covers the storage failure modes, but if rotation succeeds and `subscribeToConnection` then throws, `relayUrl` has already been flipped. A retry computes `relayUrlChanged` as false, skips rotation, and never surfaces the pairing URL, even though the new key is persisted and the pubkey rebound (so the connection itself is fine, just unsubscribed until restart and the user never sees the QR). Much softer than the original bug, but worth closing: one option is to catch failures from the subscribe/save step and still return `{ nostrUrl, success: true }` with a warning, since the rotation itself is durable at that point and the subscription recovers on restart.

ajaysehwal:
> Thanks for the quick turnaround. The inline items all look good: rotation now happens before any mutation with the pubkey rebound in the same `runInAction`, the publish handling matches `createConnection`, the key-load guard is in, and `releaseUnusedRelayService` goes beyond what I asked by covering `deleteConnection` too. Four things before this is merge-ready:
> 
> 1. **Store test.** Please add the store test asserting that a relay change stores the new client key, deletes the old one, and rebinds `connection.pubkey`. This behavior has regressed since [[feat] Nostr Wallet Connect (NWC) service #3216](https://github.com/ZeusLN/zeus/pull/3216) and survived unnoticed until now, so I want it guarded before merge.
> 2. **Expiry behavior note.** The old edit path went through `createConnection` with `params.id`/`params.createdAt` set, which triggered `refreshExpiryForRegenerate`. The new path keeps the form's expiry as-is on relay change. I think that's the right call given the identity-preservation goal, but it is a behavior change: please add a line to the PR description documenting it.
> 3. **Template/testing.** `yarn run test` is still unchecked and the unit-tests question has no box checked. Please also confirm the manual matrix from my review: a relay change on a connection with an actually-linked client app (verify the old secret is dead and the new QR pairs), a name/budget-only edit (no rotation, no QR screen), and create + details-screen regenerate regression checks.
> 4. **Residual ordering edge case.** The reorder covers the storage failure modes, but if rotation succeeds and `subscribeToConnection` then throws, `relayUrl` has already been flipped. A retry computes `relayUrlChanged` as false, skips rotation, and never surfaces the pairing URL, even though the new key is persisted and the pubkey rebound (so the connection itself is fine, just unsubscribed until restart and the user never sees the QR). Much softer than the original bug, but worth closing: one option is to catch failures from the subscribe/save step and still return `{ nostrUrl, success: true }` with a warning, since the rotation itself is durable at that point and the subscription recovers on restart.

addressed in 8f80408e6131eb99e9838fa75b955c4a3c481e51

kaloudis (approved):
ACK

PR #4343, opened 2026-08-07
fix(nwc): safe pay_invoice activity upsert and reconcile persistence

ajaysehwal:
# Description

Hardens NWC `pay_invoice` activity writes and reconcile persistence so budget and status stay consistent when payments settle in-handler or in the background.

### Problem

- `finalizePayment` always appended a success row, so a prior **pending** row for the same invoice could remain (duplicate activity).
- Reconcile could promote pending → success and debit budget without a durable save, so `is_budget_debited` / status could be lost on process kill.
- Pending, failed, and success paths duplicated find-or-push logic.

### Changes

- Debit budget only once using `is_budget_debited` (skip if already set during reconcile)
- Centralize pending/failed / success writes in `upsertPayInvoiceActivity` (merge by id, never demote success)
- Persist the connection after reconcile mutates activity

### Flow (after)

| Path | Activity | Budget |
|------|----------|--------|
| In-flight | pending | held via `pendingSpendSats` |
| Settled in-handler | upsert → success | debit once |
| Settled later | pending → success | debit once if not already flagged |
| Failed | upsert → failed | no debit |

### Test plan

- [ ] Immediate settle → one success activity, budget debited once
- [ ] In-flight then settle → single row (pending upgraded to success), budget debited once
- [ ] Background settle + app restart → status and budget flag still correct
- [ ] Failed payment → failed activity, budget unchanged

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -248,10 +248,11 @@ export default class NWCConnection extends BaseModel {
     ): number {
         return (this.activity ?? []).findIndex(
             (a) =>
-                a.id === id ||
-                (!!paymentHash &&
-                    !!a.paymentHash &&
-                    a.paymentHash === paymentHash)
+                a.type === 'pay_invoice' &&
+                (a.id === id ||
+                    (!!paymentHash &&
+                        !!a.paymentHash &&
+                        a.paymentHash === paymentHash))
         );
     }
 
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -41,6 +41,7 @@ export interface ConnectionActivity {
     isExpired?: boolean;
     expiryLabel?: string;
     fees_paid?: number;
+    is_budget_debited?: boolean;
 }
 
 export enum BudgetRenewalType {
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -41,7 +41,7 @@ export interface ConnectionActivity {
     isExpired?: boolean;
     expiryLabel?: string;
     fees_paid?: number;
-    is_budget_debited?: boolean;
+    isBudgetDebited?: boolean;
 }
 
 export enum BudgetRenewalType {
--- a/models/NWCConnection.ts
+++ b/models/NWCConnection.ts
@@ -242,6 +242,27 @@ export default class NWCConnection extends BaseModel {
         return this.totalSpendSats + this.pendingSpendSats;
     }
 
+    public findPayInvoiceActivityIndex(
+        id: string,
+        paymentHash?: string
+    ): number {
+        return (this.activity ?? []).findIndex(
+            (a) =>
+                a.id === id ||
+                (!!paymentHash &&
+                    !!a.paymentHash &&
+                    a.paymentHash === paymentHash)
+        );
+    }
+
+    public findPayInvoiceActivity(
+        id: string,
+        paymentHash?: string
+    ): ConnectionActivity | undefined {
+        const index = this.findPayInvoiceActivityIndex(id, paymentHash);
+        return index !== -1 ? this.activity[index] : undefined;
+    }
+
     @computed public get remainingBudget(): number {
         if (!this.hasBudgetLimit) return Infinity;
         return Math.max(0, this.maxAmountSats! - this.effectiveSpendSats);
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2538,7 +2538,10 @@ export default class NostrWalletConnectStore {
             });
         }
 
-        if (changed) this.lookupPaymentsCache = null;
+        if (changed) {
+            this.lookupPaymentsCache = null;
+            this.findAndUpdateConnection(connection);
+        }
         return changed;
     }
 
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1722,7 +1722,7 @@ export default class NostrWalletConnectStore {
         }
     }
 
-    private async pushMakeInvoiceActivityToConnection({
+    private async recordMakeInvoiceActivity({
         connection,
         request,
         status,
@@ -1807,15 +1807,14 @@ export default class NostrWalletConnectStore {
                         );
                     }
 
-                    const result =
-                        await this.pushMakeInvoiceActivityToConnection({
-                            connection,
-                            request,
-                            status: 'pending',
-                            type: 'make_invoice',
-                            paymentRequest: cashuInvoice.paymentRequest,
-                            payment_source: 'cashu'
-                        });
+                    const result = await this.recordMakeInvoiceActivity({
+                        connection,
+                        request,
+                        status: 'pending',
+                        type: 'make_invoice',
+                        paymentRequest: cashuInvoice.paymentRequest,
+                        payment_source: 'cashu'
+                    });
                     return result;
                 } catch (cashuError) {
                     return NostrConnectUtils.createNip47Error(
@@ -1859,7 +1858,7 @@ export default class NostrWalletConnectStore {
 
             await this.invoicesStore.getInvoices();
 
-            const result = await this.pushMakeInvoiceActivityToConnection({
+            const result = await this.recordMakeInvoiceActivity({
                 connection,
                 request,
                 status: 'pending',
@@ -2822,28 +2821,34 @@ export default class NostrWalletConnectStore {
     }): Promise<void> {
         // Budget is whole sats; round fractional fees (0.026 → 0, 1.999 → 2).
         const budgetFeeSats = NostrConnectUtils.resolveFeeSats(feeSats);
+
         runInAction(() => {
-            connection.trackSpending(
-                amountSats + budgetFeeSats,
-                this.maxBudgetLimit
-            );
-            connection.activity.push({
+            const existing = connection.activity.find((a) => a.id === id);
+            const alreadyDebited =
+                !!existing?.is_budget_debited || existing?.status === 'success';
+
+            if (!alreadyDebited) {
+                connection.trackSpending(
+                    amountSats + budgetFeeSats,
+                    this.maxBudgetLimit
+                );
+            }
+
+            this.upsertPayInvoiceActivity(connection, {
                 id,
                 type,
-                payment:
-                    payment_source == 'cashu'
-                        ? new CashuPayment(decoded)
-                        : new Payment(decoded),
+                payment: this.toConnectionActivityPayment(
+                    payment_source,
+                    decoded
+                ),
                 status: 'success',
                 payment_source,
                 satAmount: amountSats,
-                // Stored internally in sats (may be fractional). NIP-47 `fees_paid` is
-                // msats; convert to msats only when mapping to the NIP-47 response.
                 fees_paid: feeSats,
-                is_budget_debited: true // flag to indicate that the budget was debited for this payment
+                is_budget_debited: true
             });
-            this.findAndUpdateConnection(connection);
         });
+
         this.lookupPaymentsCache = null;
         NostrConnectUtils.notifyOutgoingNwcPayment(amountSats, connection.name);
     }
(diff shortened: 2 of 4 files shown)


models/NWCConnection.ts
```diff
@@ -41,6 +41,7 @@ export interface ConnectionActivity {
     isExpired?: boolean;
     expiryLabel?: string;
     fees_paid?: number;
+    is_budget_debited?: boolean;
 }
```
kaloudis, line 45: nit: this model is a mixed of camel-cased and snake cased

ajaysehwal: Good point. But I think changing the other fields for now could impact old saved activity data. I'll only change `is_budget_debited` for now.

stores/NostrWalletConnectStore.ts
```diff
@@ -2484,29 +2486,25 @@ export default class NostrWalletConnectStore {
         return true;
     }
 
-    /**
-     * Promotes pending pay_invoice activities that have since settled or failed.
-     * Used by the activity screen and by list_transactions — without the latter,
-     * an NWC client sees a settled payment as pending until someone opens the
-     * screen. Node fetches are rate limited by the helper below.
-     */
     private async reconcilePendingPayInvoiceActivities(
```
kaloudis, line 2500: Please restore the docstring that was here. It explains why `list_transactions` triggers reconcile (an NWC client would otherwise see a settled payment as pending until someone opens the activity screen) and notes that node fetches are rate limited by the helper below. That context is load-bearing for anyone touching this next.

stores/NostrWalletConnectStore.ts
```diff
@@ -2828,27 +2828,34 @@ export default class NostrWalletConnectStore {
...
+                    payment_source,
+                    decoded
+                ),
                 status: 'success',
                 payment_source,
                 satAmount: amountSats,
-                // Stored internally in sats (may be fractional). NIP-47 `fees_paid` is
-                // msats; convert to msats only when mapping to the NIP-47 response.
-                fees_paid: feeSats
+                fees_paid: feeSats,
```
kaloudis, line 2878: Please restore the comment that was here: `fees_paid` is stored internally in sats (may be fractional), while NIP-47 `fees_paid` is msats, converted only when mapping to the NIP-47 response. That unit boundary has bitten us before and the comment was the only thing marking it.

stores/NostrWalletConnectStore.ts
```diff
@@ -2871,49 +2878,34 @@ export default class NostrWalletConnectStore {
...
+                ...(paymentHash ? { paymentHash } : {}),
+                ...(payment
+                    ? {
+                          payment: this.toConnectionActivityPayment(
+                              payment_source,
+                              payment
+                          )
+                      }
+                    : {})
+                // is_budget_debited intentionally unset — held via pendingSpendSats
```
kaloudis: nit: comment says `is_budget_debited` but the field was renamed to `isBudgetDebited` in d3e183ce. Same in the merge comment in `upsertPayInvoiceActivity` below.

stores/NostrWalletConnectStore.ts
```diff
@@ -2954,26 +2941,64 @@ export default class NostrWalletConnectStore {
...
-            } else {
-                // ➕ add new element
-                connection.activity.push(record);
-            }
-
-            this.findAndUpdateConnection(connection);
+            });
         });
+
         NostrConnectUtils.notifyOutgoingNwcPaymentFailed(
```
kaloudis: This is a behavior regression. On master, `recordFailedPayment` early-returns when the activity is already `success`, which skipped both the failed write and this notification (that guard came in with 005ae5c40). Now the demote is suppressed inside `upsertPayInvoiceActivity`, but the notification fires unconditionally.

Concrete case: a hodl/slow payment settles in the background, reconcile promotes the row to `success`, then the in-handler wait times out and calls `recordFailedPayment` with "payment timed out" (not covered by `isIgnorableError`). The activity stays correct but the user gets a payment-failed push for a payment that succeeded.

Suggestion: have `upsertPayInvoiceActivity` return a boolean (applied vs rejected demote) and gate this notification on it.

stores/NostrWalletConnectStore.ts
```diff
@@ -2954,26 +2941,64 @@ export default class NostrWalletConnectStore {
...
+    /**
+     * Insert or merge a pay_invoice activity by id.
+     * Never demotes status === 'success'.
+     */
+    @action
+    private upsertPayInvoiceActivity(
+        connection: NWCConnection,
+        record: ConnectionActivity
+    ): void {
+        const index = connection.activity.findIndex((a) => a.id === record.id);
```
kaloudis: Matching by id only still leaves the bug class this PR is targeting open for uppercase invoices. `recordPendingPayment` and `recordFailedPayment` compute `id` from `decodeInvoiceTags().paymentRequest`, and `Bolt11Utils.decode` lowercases the invoice. Both `finalizePayment` call sites pass `id: request.invoice` verbatim. A client sending an uppercase bolt11 produces a pending row and a success row with different ids: the upsert misses, the pending row keeps holding `pendingSpendSats` on top of the debit, and a later reconcile promotes the stale pending row and debits again (it has no `isBudgetDebited`).

Two-part fix: derive the id in `finalizePayment` through the same `decodeInvoiceTags` path (or normalize `request.invoice` once at handler entry), and consider a `paymentHash` fallback here, mirroring how `reconcilePendingPayInvoiceActivities` already matches payments.

stores/NostrWalletConnectStore.ts
```diff
@@ -2954,26 +2941,64 @@ export default class NostrWalletConnectStore {
...
+        const existing = index !== -1 ? connection.activity[index] : undefined;
+
+        if (existing?.status === 'success' && record.status !== 'success') {
+            return;
+        }
+
+        if (index !== -1) {
+            connection.activity[index] = {
+                ...existing,
+                ...record,
```
kaloudis, line 3004: The spread keeps existing fields that aren't present in the new record, so a failed -> retry -> success sequence leaves the old `error` message sitting on a `status: 'success'` row (and on the retry's pending row). Suggest clearing `error` when the incoming status isn't `failed`, e.g. `error: record.status === 'failed' ? record.error : undefined`.

stores/NostrWalletConnectStore.ts
```diff
@@ -3192,28 +3217,52 @@ export default class NostrWalletConnectStore {
...
+                this.maxBudgetLoadInFlight = null;
             }
-        } catch (error) {
-            runInAction(() => {
-                this.maxBudgetLimit = 0;
-            });
-            console.error('Failed to get max budget:', error);
         }
     }
+    private scheduleMaxBudgetRefresh(): void {
```
kaloudis, line 3306: Two lifecycle gaps here:

1. The timer is never cleared in `reset()`. Teardown flushes and cancels `_scheduledSave`, but a pending budget-refresh timer will still fire `loadMaxBudget()` (a forced balance fetch) mid-wallet-switch. Please clear it in the teardown block alongside the save timer.
2. The in-flight dedupe can swallow the refresh's purpose: if a `loadMaxBudget` that started before settlement is still in flight when this timer fires, the call joins the stale promise and no post-payment fetch happens. Either chain a follow-up load when one is in flight, or leave a comment that the staleness is accepted.

kaloudis (commented):
Thanks for the follow-up. The core fix is right: on master, a payment that settles via reconcile and then has its handler finalize late gets a second success row pushed and the budget debited a second time, and the `alreadyDebited` guard closes that. Treating legacy pre-flag success rows as already debited is a nice touch, and keeping the pending -> success flip and the debit inside one `runInAction` preserves the `pendingSpendSats` atomicity invariant documented on the model.

A few things inline. The two I'd like addressed before merge are the false failure notification and the id normalization, since the latter is the same bug class this PR is targeting.

Two more general notes:

- The reconcile persistence claim in the description is mostly already true on master: `getActivities`, `reconcileAllPendingPayInvoiceActivities`, and `handleListTransactions` all schedule saves after reconcile reports changes. The genuinely new durability win here is the persisted `isBudgetDebited` flag (plus the `syncWarnings` pass that `findAndUpdateConnection` adds). Worth updating the description so the history reads right.
- `upsertPayInvoiceActivity` is a great candidate for unit tests: never-demote, sticky `isBudgetDebited`, `createdAt` preservation, and the debit-once path in `finalizePayment` are all cheap to lock in. Would love to see a few cases here.

Also, the test plan checkboxes are all unchecked. The background-settle-plus-restart case is the one that most needs a device pass before this goes in.

models/NWCConnection.ts
```diff
@@ -241,6 +242,27 @@ export default class NWCConnection extends BaseModel {
         return this.totalSpendSats + this.pendingSpendSats;
     }
 
+    public findPayInvoiceActivityIndex(
+        id: string,
+        paymentHash?: string
+    ): number {
+        return (this.activity ?? []).findIndex(
+            (a) =>
+                a.id === id ||
```
kaloudis: This matcher runs across all activity types, but `make_invoice` rows are pushed into the same `connection.activity` array with `paymentHash` set, and their `id` is the same lowercase bolt11 a client would pass to `pay_invoice`.

Concrete case that needs no unusual timing: a client calls `make_invoice`, then `pay_invoice` on the invoice it just created over the same connection (a round-trip self-test some clients do). On lnd the self-payment fails, `recordFailedPayment` runs, this matcher hits the `make_invoice` row by id and hash, and the merge in `upsertPayInvoiceActivity` rewrites it to `type: 'pay_invoice', status: 'failed'`. From then on the settle pollers (they filter on `type === 'make_invoice'`) never promote it: when the invoice is later paid externally the client sees it pending forever, plus a phantom failed payment row. The cross-source variant (cashu configured, so the self-pay actually settles) silently converts the receive record the same way, and if the matched row is already `status: 'success'` the `alreadyDebited` gate in `finalizePayment` skips the debit entirely.

One-line fix: require `a.type === 'pay_invoice'` in the predicate. The method name already promises it.

stores/NostrWalletConnectStore.ts
```diff
@@ -3192,28 +3249,63 @@ export default class NostrWalletConnectStore {
...
+        })();
+
+        this.maxBudgetLoadInFlight = load;
+        try {
+            await load;
+        } finally {
+            if (this.maxBudgetLoadInFlight === load) {
+                this.maxBudgetLoadInFlight = null;
+            }
+            if (this.maxBudgetRefreshNeeded) {
```
kaloudis, line 3294: nit: `reset()` clears the timer now, but not this flag, so if a `loadMaxBudget` is in flight during a wallet switch a joiner will have set `maxBudgetRefreshNeeded` and this block fires one more forced balance fetch after teardown. Worth clearing the flag in the teardown block alongside the timer. Also note that any concurrent joiner sets the flag, so an ordinary dedup hit now costs a second fetch afterwards; if that extra fetch is intended as a freshness guarantee, fine, just leave a comment saying so.

kaloudis (commented):
Thanks, all seven points from the last round are addressed on b4d7ac9, and the new tests pin each one down:

- `finalizePayment` now derives its id through `decodeInvoiceTags` with a `paymentHash` fallback, so an uppercase bolt11 no longer produces duplicate rows or a double debit
- `recordFailedPayment` only notifies when the write was actually applied, restoring the suppression of the false "payment failed" push after a background settle
- stale `error` is cleared when a row is promoted off `failed`
- the budget refresh timer is cleared in `reset()`, and a call that joins an in-flight `loadMaxBudget` now chains a follow-up fetch
- reconcile docstring, `fees_paid` sats/msats comment, and the `isBudgetDebited` comment naming are all restored

One new issue came in with the shared matcher though: `findPayInvoiceActivityIndex` does not filter by activity type, and `make_invoice` rows live in the same `connection.activity` array with `paymentHash` set. See inline. With that one-line fix I think this is ready.

kaloudis (approved):
ACK

PR #4352, opened 2026-08-10
Require re-authentication before revealing wallet seed

kaloudis:
# Description

Once the app is unlocked, viewing the wallet mnemonic (`Seed`) and exporting extended private keys (`SeedQRExport`, reachable only from `Seed`) require no further authentication, so a briefly unattended unlocked phone allows silent seed exfiltration. Unlike sending funds, seed theft is invisible to the victim and permanent.

This PR routes every seed-reveal entry point through the existing Lockscreen re-auth mechanism:

- New `reAuthNavigation()` helper in `utils/NavigationUtils.ts`: navigates via `Lockscreen` with `pendingNavigation` whenever a login method (PIN, passphrase, or biometrics) is configured, and directly otherwise. It intentionally uses `loginMethodConfigured()` rather than `loginRequired()`, since the latter is false for the rest of the session once the app is unlocked. `protectedNavigation` (POS-only, existing callers) is untouched; new unit tests lock in its behavior alongside the helper's.
- Gated call sites: menu seed entry, BalancePane backup prompt card, both WalletConfiguration backup paths (active and inactive wallets, including the `walletSeedPhrase` param hop), the delete-wallet channel-export path, re-viewing an existing swap rescue key, and the Cashu BIP-39 seed reveal in Cashu Tools.
- Deliberately ungated: showing a swap rescue key immediately after generating it (the user just triggered creation in-session and the key exists nowhere else yet; a prompt there is friction with no confidentiality gain).
- Standalone Lockscreen bug fix: `onAttemptLogIn` checked `selectNodeOnStartup` before `pendingNavigation`, replacing the stack with the wallet picker and silently dropping the re-auth target on the PIN/passphrase path (the biometric path was already correct). This also benefits the existing POS `pendingNavigation` flow.

Behavior notes, both intended:

- Entering duress credentials at the seed prompt performs the full duress wipe. Coerced seed reveal is the canonical duress scenario, and suppressing duress for `pendingNavigation` would also strip it from the POS auth flow, which shares the param.
- Five failed attempts at the prompt trigger the standard wipe, consistent with every other Lockscreen use.

Known limitation (pre-existing Lockscreen behavior, documented rather than changed here): with biometrics configured but no PIN/passphrase, cancelling the biometric prompt falls through Lockscreen's no-credentials path and proceeds. Same behavior exists at cold start today; fixing it means reworking the pass-through logic and is left for a follow-up.

No new locale strings, no storage changes, no new dependencies.

- [x] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [x] Embedded LND

Remote
- [ ] LND (REST) - N/A, no local seed; gated entry points only render for local wallets
- [ ] LND (Lightning Node Connect) - N/A
- [ ] Core Lightning (CLNRest) - N/A
- [ ] Nostr Wallet Connect - N/A
- [ ] LndHub - N/A

Draft until the device matrix passes. Key manual cases: `selectNodeOnStartup` + PIN lands on Seed and not the wallet picker; inactive-wallet seed params survive the Lockscreen hop; iOS FaceID firing as the delete-wallet alert dismisses; duress and 5-attempt wipes verified on an emulator only.

### Locales

### Other:

### DIFF
--- a/utils/NavigationUtils.test.ts
+++ b/utils/NavigationUtils.test.ts
@@ -1,0 +1,111 @@
+jest.mock('../stores/Stores', () => ({
+    settingsStore: {
+        posStatus: 'inactive',
+        settings: {},
+        setPosStatus: jest.fn(),
+        loginMethodConfigured: jest.fn()
+    }
+}));
+
+import { settingsStore } from '../stores/Stores';
+import { protectedNavigation, reAuthNavigation } from './NavigationUtils';
+
+const mockSettingsStore = settingsStore as any;
+
+describe('NavigationUtils', () => {
+    let navigation: any;
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+        navigation = { navigate: jest.fn() };
+        mockSettingsStore.posStatus = 'inactive';
+        mockSettingsStore.settings = {};
+    });
+
+    describe('reAuthNavigation', () => {
+        it('routes via Lockscreen when a login method is configured', () => {
+            mockSettingsStore.loginMethodConfigured.mockReturnValue(true);
+
+            reAuthNavigation(navigation, 'Seed');
+
+            expect(navigation.navigate).toHaveBeenCalledTimes(1);
+            expect(navigation.navigate).toHaveBeenCalledWith('Lockscreen', {
+                pendingNavigation: { screen: 'Seed', params: undefined }
+            });
+        });
+
+        it('passes route params through the Lockscreen pendingNavigation', () => {
+            mockSettingsStore.loginMethodConfigured.mockReturnValue(true);
+            const params = {
+                walletSeedPhrase: ['abandon', 'ability', 'able'],
+                implementation: 'ldk-node',
+                isTestNet: true
+            };
+
+            reAuthNavigation(navigation, 'Seed', params);
+
+            expect(navigation.navigate).toHaveBeenCalledWith('Lockscreen', {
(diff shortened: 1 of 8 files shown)


shubhamkmr04 (commented):
@kaloudis should we also gate Cashuseeds ?

kaloudis:
Good catch @shubhamkmr04, agreed. Added in a3eb2e26f: the Cashu BIP-39 seed reveal (Tools > Cashu Tools > Seed) now goes through `reAuthNavigation` like the other seed surfaces. That was the only entry point to `CashuSeed`.

Two related notes:
- `LegacySeedRecovery` on the same screen doesn't need gating: it consumes the seed internally to recover funds but never displays it.
- The Nostr Keys tool reveals the ZEUS Pay nsec behind an in-view toggle with no re-auth. It's key material but not directly spendable like a seed, so I'd rather handle it in a follow-up than grow this PR's scope. Happy to fold it in here instead if you'd prefer.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4344, opened 2026-08-08
fix: LSPS1: update Olympus mainnet host

kaloudis:
# Description

Relates to issue: **ZEUS-0000**

The Olympus mainnet node has moved to a new host. The node URI is now:

`031b301307574bbe9b9ac7b79cbe1700e31e544513eae0b5d7497483083f99e581@45.79.207.158:9735`

The pubkey is unchanged; only the host changed (`45.79.192.236:9735` to `45.79.207.158:9735`).

Changing `DEFAULT_LSPS1_HOST_MAINNET` alone would only fix fresh installs: the `lsps1-hosts1` migration persisted the old default into every existing user's settings blob, and `getLspConfigForNetwork` prefers the persisted value. This PR therefore also ships a one-shot migration.

## Migration plan

New `MigrationUtils.migrateLsps1MainnetHost` (flag `olympus-host-2026`, EncryptedStorage):

- If `settings.lsps1HostMainnet` equals the old default `45.79.192.236:9735`, rewrite it to `45.79.207.158:9735` and persist via `settingsStore.setSettings` (awaited, object payload). Users with a custom host are left untouched.
- The flag is set only after the write succeeds, so a crash mid-migration retries on next boot.
- Runs on **both** load paths: called from `legacySettingsMigrations` (after the `lsps1-hosts1` block, which backfills missing hosts with the new default) and from the modern `zeus-settings-v2` branch of `SettingsStore.getSettings`.
- Cohorts: fresh installs get the new default directly; legacy-blob upgraders and current v2 users on the old default get rewritten once; users with custom hosts keep them.
- Idempotent: the rewrite only fires while the value equals the old default; running twice is a no-op.
- Revert story: revert-safe in code (the migration only adds a repair). If the node ever moves back to the old IP, a counter-migration with a new flag key would be needed, since migrated users have the new host persisted.

No new storage keys are introduced (only an EncryptedStorage migration flag, which is the convention for MOD_KEY flags), so no DataClearUtils / KeychainRecoveryUtils / cloud-sync registration applies.

This pull request is categorized as a:

- [ ] New feature
- [x] Bug fix
- [x] Configuration change
- [ ] Code refactor
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I've run `yarn run tsc` and made sure my code compiles correctly
- [x] I've run `yarn run lint` and made sure my code didn't contain any problematic patterns
- [x] I've run `yarn run prettier` and made sure my code is formatted correctly
[...]

PR #4356, opened 2026-08-10
config: new default LSP hosts: zeuslsp.com

kaloudis:
## Description

We're moving the default LSP hosts in our infra from lnolymp.us to zeuslsp.com. This PR updates the six default Flow 2.0 and LSPS1 REST hosts and ships a migration for users who have the old defaults persisted in their settings.

| Setting | Old default | New default |
|---|---|---|
| `lspMainnet` | `https://0conf.lnolymp.us` | `https://flow.zeuslsp.com` |
| `lspTestnet` | `https://testnet-0conf.lnolymp.us` | `https://flow.testnet.zeuslsp.com` |
| `lspMutinynet` | `https://mutinynet-flow.lnolymp.us` | `https://flow.mutinynet.zeuslsp.com` |
| `lsps1RestMainnet` | `https://lsps1.lnolymp.us` | `https://lsps1.zeuslsp.com` |
| `lsps1RestTestnet` | `https://testnet-lsps1.lnolymp.us` | `https://lsps1.testnet.zeuslsp.com` |
| `lsps1RestMutinynet` | `https://mutinynet-lsps1.lnolymp.us` | `https://lsps1.mutinynet.zeuslsp.com` |

Out of scope, unchanged: `backups.lnolymp.us` (SCB backups), neutrino peers (`btcd*.lnolymp.us`), and the LSPS1 P2P `DEFAULT_LSPS1_HOST_*` entries (mainnet P2P host is moving separately in #4344 under its own flag).

## Implementation

- The six `DEFAULT_LSP_*` / `DEFAULT_LSPS1_REST_*` constants in `stores/SettingsStore.ts` now point at the new hosts. All readers (`getLspConfigForNetwork` fallbacks, the reset buttons in LSP and LSPS1 settings views, the legacy default backfills) go through these constants, so fresh installs and resets pick up the new hosts automatically.
- New `migrateOlympusHostsToZeusLsp` in `utils/MigrationUtils.ts`, gated by the `zeuslsp-hosts-2026` flag in EncryptedStorage, modeled on `migrateRgsDefaultToZeus`.

## Migration plan

- **Cohorts:** fresh installs get new defaults from the constants (migration only sets the flag). Users on `zeus-settings-v2` and legacy blob users with an old default persisted get that value rewritten to the new host. Users with custom hosts are untouched (exact string match against the old defaults only). Unset keys stay unset; runtime falls back to the new constants.
- **Both load paths:** runs at the end of `legacySettingsMigrations` and on the modern `zeus-settings-v2` branch of `getSettings`, after `migrateInvoiceExpiryDisplay`.
- **Idempotent:** rewriting is a pure old-value to new-value map; running twice is a no-op. `setSettings` is awaited and receives the settings object (not a JSON string); the flag is only written after persistence succeeds, so a crash mid-migration retries on next boot.
- **Revert story:** the migration only rewrites old default strings to new default strings, so reverting this PR before release is safe. If it ships and needs rollback, a counter-migration under a fresh flag can map the values back the same way.

## Testing

- 8 new Jest tests in `utils/MigrationUtils.test.ts` (legacy path rewrite, custom hosts untouched, and a v2 path block covering rewrite of all six hosts, object-not-string persistence, flag no-op, unset-keys no-op, double-run idempotency); 30/30 passing
- tsc, eslint, prettier clean
- Device testing pending: fresh install, upgrade from v2 with old defaults persisted, upgrade with one custom host, on iOS and Android

This PR is consistent with the [contribution guidelines](https://github.com/ZeusLN/zeus/blob/master/CONTRIBUTING.md).

### DIFF
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1295,12 +1295,12 @@ export const LNDHUB_AUTH_MODES = [
     { key: 'Alby', value: 'Alby' }
 ];
 
-export const DEFAULT_LSP_MAINNET = 'https://0conf.lnolymp.us';
-export const DEFAULT_LSP_TESTNET = 'https://testnet-0conf.lnolymp.us';
+export const DEFAULT_LSP_MAINNET = 'https://flow.zeuslsp.com';
+export const DEFAULT_LSP_TESTNET = 'https://flow.testnet.zeuslsp.com';
 
 // LSPS1 REST
-export const DEFAULT_LSPS1_REST_MAINNET = 'https://lsps1.lnolymp.us';
-export const DEFAULT_LSPS1_REST_TESTNET = 'https://testnet-lsps1.lnolymp.us';
+export const DEFAULT_LSPS1_REST_MAINNET = 'https://lsps1.zeuslsp.com';
+export const DEFAULT_LSPS1_REST_TESTNET = 'https://lsps1.testnet.zeuslsp.com';
 
 export const DEFAULT_LSPS1_PUBKEY_MAINNET =
     '031b301307574bbe9b9ac7b79cbe1700e31e544513eae0b5d7497483083f99e581';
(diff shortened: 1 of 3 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4360, opened 2026-08-11
fix(nwc): enforce connection expiry on all subscribed methods

ajaysehwal:
# Description

Expired NWC connections were only blocked on pay_invoice (via validateBudgetBeforePayment). A connection that was already subscribed could still call get_info, get_balance, make_invoice, lookup_invoice, list_transactions, and sign_message until the next resubscribe cycle.

This PR enforces expiry at the shared request gate (withGlobalHandler), so every subscribed NWC method is rejected once connection.isExpired is true. Expired requests return NIP-47 RESTRICTED with the existing “connection expired” message, and the subscription is dropped so further events are not handled.

## Changes

- Gate all NWC handler requests in withGlobalHandler before markConnectionUsed
- Return RESTRICTED for expired connections (NIP-47-appropriate error code)
- Unsubscribe expired connections on first rejected request
- Add Nip47ErrorCode.RESTRICTED to NostrConnectUtils
- Add unit tests for expiry rejection, unsubscribe behavior, and non-expired passthrough

Test plan

- [x] Run `yarn test`
- [ ] Manual: create an NWC connection with a short expiry, wait until expired, confirm get_info / make_invoice / sign_message return RESTRICTED
- [ ] Manual: confirm non-expired connections still work normally
- [ ] Manual: confirm pay_invoice on expired connection still returns RESTRICTED (not INSUFFICIENT_BALANCE)

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1615,10 +1615,32 @@ export default class NostrWalletConnectStore {
         return successfulPublishes;
     }
 
+    /**
+     * Runs every subscribed NWC request through one gate: reject (and drop
+     * the subscription) when the connection has expired.
+     */
     private async withGlobalHandler<T>(
         connectionId: string,
         handler: () => Promise<T>
     ): Promise<T> {
+        const { connection } = this.getConnection({ connectionId });
+        if (!connection) {
+            return NostrConnectUtils.createNip47Error(
+                localeString(
+                    'stores.NostrWalletConnectStore.error.connectionNotFound'
+                ),
+                Nip47ErrorCode.NOT_FOUND
+            ) as T;
+        }
+        if (connection.isExpired) {
+            this.unsubscribeFromConnection(connectionId);
+            return NostrConnectUtils.createNip47Error(
+                localeString(
+                    'views.Settings.NostrWalletConnect.error.connectionExpired'
+                ),
+                Nip47ErrorCode.RESTRICTED
+            ) as T;
+        }
         await this.markConnectionUsed(connectionId);
         return await handler();
     }
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1617,7 +1617,7 @@ export default class NostrWalletConnectStore {
 
     /**
      * Runs every subscribed NWC request through one gate: reject (and drop
-     * the subscription) when the connection has expired.
+     * the subscription) when the connection is missing or has expired.
      */
     private async withGlobalHandler<T>(
         connectionId: string,
@@ -1625,11 +1625,12 @@ export default class NostrWalletConnectStore {
     ): Promise<T> {
         const { connection } = this.getConnection({ connectionId });
         if (!connection) {
+            this.unsubscribeFromConnection(connectionId);
             return NostrConnectUtils.createNip47Error(
                 localeString(
                     'stores.NostrWalletConnectStore.error.connectionNotFound'
                 ),
-                Nip47ErrorCode.NOT_FOUND
+                Nip47ErrorCode.UNAUTHORIZED
             ) as T;
         }
         if (connection.isExpired) {
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1618,6 +1618,8 @@ export default class NostrWalletConnectStore {
     /**
      * Runs every subscribed NWC request through one gate: reject (and drop
      * the subscription) when the connection is missing or has expired.
+     * When wallet identity is transiently unknown (e.g. during reconnect),
+     * returns INTERNAL_ERROR without dropping the subscription.
      */
     private async withGlobalHandler<T>(
         connectionId: string,
@@ -1625,6 +1627,14 @@ export default class NostrWalletConnectStore {
     ): Promise<T> {
         const { connection } = this.getConnection({ connectionId });
         if (!connection) {
+            if (!this.hasWalletContext()) {
+                return NostrConnectUtils.createNip47Error(
+                    localeString(
+                        'stores.NostrWalletConnectStore.error.walletIdentityUnavailable'
+                    ),
+                    Nip47ErrorCode.INTERNAL_ERROR
+                ) as T;
+            }
             this.unsubscribeFromConnection(connectionId);
             return NostrConnectUtils.createNip47Error(
                 localeString(
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -704,6 +704,36 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         return connection;
     }
 
+    it('rejects and unsubscribes when the connection record is missing', async () => {
+        const store = buildStore();
+        store.connections = [];
+
+        const connectionId = 'conn-missing';
+        const unsub = jest.fn();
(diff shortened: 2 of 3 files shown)


stores/NostrWalletConnectStore.ts
```diff
@@ -1609,10 +1609,32 @@ export default class NostrWalletConnectStore {
...
     ): Promise<T> {
+        const { connection } = this.getConnection({ connectionId });
+        if (!connection) {
+            return NostrConnectUtils.createNip47Error(
+                localeString(
+                    'stores.NostrWalletConnectStore.error.connectionNotFound'
+                ),
+                Nip47ErrorCode.NOT_FOUND
+            ) as T;
+        }
```
kaloudis, line 1645: Two small things on this branch:

1. Per NIP-47, `NOT_FOUND` is specified for invoice lookups ("The invoice could not be found by the given parameters"). `UNAUTHORIZED` ("This public key has no wallet connected") is the closer fit for a request from a connection we don't recognize. Since this PR is touching the enum anyway, adding `UNAUTHORIZED` and using it here would be cheap.

2. Unlike the expiry branch below, this one doesn't drop the subscription. The normal delete path unsubscribes, so this is a defensive case, but if a subscription ever outlives its connection record it would answer an error on every request forever. Mirroring the `unsubscribeFromConnection` call here would make the gate self-cleaning in both branches.

Would also be nice to cover this branch in the new test suite.

kaloudis (commented):
Concept ACK. Verified the gate covers all 7 subscribed handlers, and that the sdk's unsubscribe only closes the subscription, so the RESTRICTED reply still gets published via the pool after `unsubscribeFromConnection`. Ran the new tests and tsc on the branch, both pass. A couple of minor comments inline, plus two non-blocking observations:

- Once the first rejected request drops the subscription, subsequent requests from the expired client time out silently instead of receiving RESTRICTED. I think that's the right tradeoff, just something to keep in mind while running the manual test plan.
- The `as T` casts are technically unsound (T includes the success shape) but they compile clean and are pragmatic given the SDK's per-method response types. Fine as is.

kaloudis:
@ajaysehwal also rebase to fix conflicts

kaloudis:
ACK 906b2fedf53f4083a77587a8dbaa25e5740f9235

Both round-1 points are addressed:

1. `UNAUTHORIZED` added to the enum and used for the unknown-connection branch, `RESTRICTED` kept for expiry. Both match their NIP-47 definitions.
2. The missing-connection branch now drops the subscription, so the gate is self-cleaning in both branches, and the new test covers it.

Verified on this round:

- All seven subscribed handlers route through `withGlobalHandler`, including `pay_invoice` (the gate runs before `enqueuePayment`, so an expired connection gets `RESTRICTED` before any budget/queue logic runs).
- The expired branch returns before `markConnectionUsed`, so `lastUsed` is not bumped by rejected requests.
- Resubscribe paths (`subscribeToAllConnections`, the Android health check) all filter on `activeConnections`, and `isActive` is `!isExpired`, so a dropped expired subscription is not silently re-added.
- The error reply is published by the SDK's relay connection, not the subscription, so unsubscribe-then-reply still delivers the final error to the client (checked this in round 1).

One non-blocking observation, fine as a follow-up or to leave as-is: `getConnection` filters through `connectionBelongsToCurrentWallet`, which returns false for every connection while `NodeInfoStore.nodeInfo` is empty. `fetchData` resets `NodeInfoStore` when `connecting`, so a request that lands in that brief window hits the `!connection` branch and drops a valid subscription with `UNAUTHORIZED`. It self-heals (Android health check and iOS foreground `initializeService` resubscribe), so this is not a blocker, but if you want to harden it: in the missing branch, when `!this.hasWalletContext()` return `INTERNAL_ERROR` without unsubscribing, since the lookup failed because wallet identity is transiently unknown, not because the connection is gone.

ajaysehwal:
> ACK [906b2fe](https://github.com/ZeusLN/zeus/commit/906b2fedf53f4083a77587a8dbaa25e5740f9235)
> 
> Both round-1 points are addressed:
> 
> 1. `UNAUTHORIZED` added to the enum and used for the unknown-connection branch, `RESTRICTED` kept for expiry. Both match their NIP-47 definitions.
> 2. The missing-connection branch now drops the subscription, so the gate is self-cleaning in both branches, and the new test covers it.
> 
> Verified on this round:
> 
> * All seven subscribed handlers route through `withGlobalHandler`, including `pay_invoice` (the gate runs before `enqueuePayment`, so an expired connection gets `RESTRICTED` before any budget/queue logic runs).
> * The expired branch returns before `markConnectionUsed`, so `lastUsed` is not bumped by rejected requests.
> * Resubscribe paths (`subscribeToAllConnections`, the Android health check) all filter on `activeConnections`, and `isActive` is `!isExpired`, so a dropped expired subscription is not silently re-added.
> * The error reply is published by the SDK's relay connection, not the subscription, so unsubscribe-then-reply still delivers the final error to the client (checked this in round 1).
> 
> One non-blocking observation, fine as a follow-up or to leave as-is: `getConnection` filters through `connectionBelongsToCurrentWallet`, which returns false for every connection while `NodeInfoStore.nodeInfo` is empty. `fetchData` resets `NodeInfoStore` when `connecting`, so a request that lands in that brief window hits the `!connection` branch and drops a valid subscription with `UNAUTHORIZED`. It self-heals (Android health check and iOS foreground `initializeService` resubscribe), so this is not a blocker, but if you want to harden it: in the missing branch, when `!this.hasWalletContext()` return `INTERNAL_ERROR` without unsubscribing, since the lookup failed because wallet identity is transiently unknown, not because the connection is gone.

fixed in 39878b4fcf2afc1efb8125a63f6a6191824c383e

kaloudis:
ACK 39878b4fc

The round-3 change resolves the reconnect race I noted last round: when wallet identity is transiently unknown (e.g. `nodeInfoStore.nodeInfo` reset during reconnect), `withGlobalHandler` now returns `INTERNAL_ERROR` without dropping the subscription, instead of misclassifying a valid connection as unknown and permanently killing it with `UNAUTHORIZED`.

Verified on this round:

- `hasWalletContext()` mirrors the nullability of `connectionBelongsToCurrentWallet`: every state where `getConnection` can miss due to missing identity takes the transient branch, so a subscription is only dropped when identity is known and the connection genuinely does not belong.
- The LndHub fallback in `getCurrentWalletPubkey` still resolves from settings during reconnect, so LndHub wallets do not false-positive into `INTERNAL_ERROR`.
- Cleanup still self-heals: a connection deleted during the identity-unknown window keeps its subscription temporarily, then gets `UNAUTHORIZED` + unsubscribe on the first request after identity returns.
- Ran the store suite (32 passing) and `tsc` on 39878b4fc; CI is green on all four checks.

Non-blocking nit: the NIP-47 spec names this code `INTERNAL` rather than `INTERNAL_ERROR`, but `INTERNAL_ERROR` is the pre-existing convention in this enum, so consistency wins; not worth churn in this PR.

kaloudis (approved):

PR #4361, opened 2026-08-11
fix: enforce DNSSEC validation on BIP 353 BOLT 12 address resolution

kaloudis:
# Description

The BIP 353 BOLT 12 lookup in `utils/handleAnything.ts` (the "try BOLT 12 address first" path for `user@domain` addresses) queried Cloudflare DoH for the `_bitcoin-payment` TXT record and consumed the first answer unconditionally:

```ts
if (json.Answer && json.Answer[0]) {
    bolt12 = json.Answer[0].data;
    ...
}
```

Two problems:

1. **No DNSSEC validation.** The response was used without checking that the record was DNSSEC-signed. BIP 353 requires payment instructions to be secured by DNSSEC and states that unsigned records MUST be ignored. Cloudflare validates DNSSEC and reports the outcome in the `AD` (Authenticated Data) bit, which we were not reading.
2. **Blind answer selection.** `Answer[0]` was taken with no record-type filter and no application of BIP 353's resolution rules, so a spoofed CNAME, a decoy TXT record, or extra answers in the section could feed garbage into `processBIP21Uri`. This also broke the spec's own worked example (`matt@mattcorallo.com`), which pairs the real record with a decoy: `Answer[0]` is the decoy, which happens to contain `bitcoin:` mid-sentence, so it slipped past `processBIP21Uri`'s `.includes('bitcoin:')` and parsed no `lno`.

Because the whole block is wrapped in an empty `catch` and flows straight into the LNURL path, a failed or absent signature was silently indistinguishable from "this domain has no BIP 353 record." A forged, cache-poisoned, or DNS-hijacked TXT answer for a recipient whose zone is unsigned could substitute an attacker's BOLT 12 offer and redirect the payment.

## Fix

- Gate acceptance on `json.Status === 0 && json.AD === true` (NOERROR + DNSSEC-authenticated). Anything not proven-signed is ignored, matching BIP 353.
- Select the payment instruction per BIP 353's resolution rules: filter `Answer` to TXT records (`type === 16`), concatenate each record's character-strings in RDATA order, ignore records that don't begin (case-insensitively) with `bitcoin:`, and require exactly one `bitcoin:` record, treating multiples as invalid. This accepts the spec's worked example (real record + decoy) while rejecting spoofed non-TXT answers and injected extra `bitcoin:` records.

Transport to Cloudflare is already HTTPS, so this closes the unsigned-zone / DNS-hijack redirection vector that DNSSEC is meant to defend against. Note that the `AD` bit still trusts Cloudflare's validation; BIP 353 ultimately wants clients to validate the DNSSEC proof chain themselves, which is a larger follow-up and is marked as a `TODO` in the code.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

`utils/handleAnything.test.ts` now mocks the DoH response and covers: the AD/Status gate (unsigned or non-NOERROR replies fall through to LNURL), the BIP 353 worked example (non-`bitcoin:` decoy ignored, character-strings concatenated in RDATA order), multiple `bitcoin:` records rejected, non-TXT answers ignored even when their data looks like a `bitcoin:` URI, and empty/missing Answer sections.

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/handleAnything.test.ts
+++ b/utils/handleAnything.test.ts
@@ -904,6 +904,156 @@ describe('handleAnything', () => {
         });
     });
 
+    describe('BIP 353 BOLT 12 address resolution', () => {
+        // The worked example from BIP 353 (matt@mattcorallo.com): a decoy
+        // TXT record that must be ignored, plus the real record split
+        // across two character-strings that must be concatenated in RDATA
+        // order.
+        const DECOY_TXT =
+            '"as long as it doesn\'t start with bitcoin:, other records should be ignored"';
+        const B12_PART1 =
+            'bitcoin:bc1qztwy6xen3zdtt7z0vrgapmjtfz8acjkfp5fp7l?lno=lno1zr5qyugqgskrk70kqmuq7v3dnr2fnmhukps9n8hut48vkqpqnskt2svsqwjakp7k6pyhtkuxw7y2kqmsxlwruhzqv0zsnhh9q3t9xhx39suc6qsr07ekm5esdyum0w66mnx8vdquwvp7dp5jp7j3v5cp6aj0w329fnkqqv60q96sz5nkrc5r95qffx002q53tqdk';
+        const B12_PART2 =
+            '8x9m2tmt85jtpmcycvfnrpx3lr45h2g7na3sec7xguctfzzcm8jjqtj5ya27te60j03vpt0vq9tm2n9yxl2hngfnmygesa25s4u4zlxewqpvp94xt7rur4rhxunwkthk9vly3lm5hh0pqv4aymcqejlgssnlpzwlggykkajp7yjs5jvr2agkyypcdlj280cy46jpynsezrcj2kwa2lyr8xvd6lfkph4xrxtk2xc3lpq';
+        const REAL_TXT = `"${B12_PART1}" "${B12_PART2}"`;
+        const RECONSTRUCTED_URI = B12_PART1 + B12_PART2;
+        const ADDRESS = 'matt@mattcorallo.com';
+
+        const txt = (data: string) => ({
+            name: 'matt.user._bitcoin-payment.mattcorallo.com',
+            type: 16,
+            TTL: 3600,
+            data
+        });
+
+        const originalFetch = global.fetch;
+        let mockDnsFetch: jest.Mock;
+
+        const mockDoHResponse = (json: any) => {
+            mockDnsFetch = jest.fn().mockResolvedValue({
+                json: async () => json
+            });
+            global.fetch = mockDnsFetch as any;
+        };
+
+        const bitcoinUriCalls = () =>
+            mockProcessBIP21Uri.mock.calls.filter((call: any[]) =>
+                String(call[0]).toLowerCase().startsWith('bitcoin:')
+            );
+
+        beforeEach(() => {
+            mockIsValidLightningAddress = true;
+            mockProcessBIP21Uri.mockImplementation((input: string) => {
+                if (input === RECONSTRUCTED_URI) {
+                    return {
+                        value: 'bc1qztwy6xen3zdtt7z0vrgapmjtfz8acjkfp5fp7l',
+                        offer: 'lno1zr5qyugqgskrk70kqmuq'
+                    };
+                }
+                return { value: input };
+            });
+            // LNURL fallback succeeds so non-BOLT 12 outcomes are observable
+            mockBlobUtilFetch.mockResolvedValue({
+                info: () => ({ status: 200 }),
+                json: () => ({ callback: 'https://example.com/callback' })
+            });
+        });
+
+        afterEach(() => {
+            global.fetch = originalFetch;
+        });
+
+        it('accepts a DNSSEC-validated answer, ignoring the non-bitcoin: decoy record', async () => {
+            mockDoHResponse({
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (commented):
The DNSSEC check is a good catch. The record-count guard looks off though —
[BIP 353 §Resolution][res] says to *ignore* TXT records at the label that don't begin with `bitcoin:`, and to fail only when more than one record *does*. The BIP says as much about its own vector in [§Example proofs][proofs]: *"Note that the
second TXT record (which does not start with, case insensitively, "bitcoin:") is
ignored."*

So "exactly one TXT record" rejects valid setups — including
`matt@mattcorallo.com`, the worked example in [§Examples][ex]:

```
Status=0  AD=true
Answer[0] type=16  "as long as it doesn't start with bitcoin:, other records sh...
Answer[1] type=16  "bitcoin:bc1qztwy6xen3zdtt7z0vrgapmjtfz8acjkfp5fp7l?lno=lno1...
```

Two records, so a correctly signed offer is discarded and we fall through to LNURL, which 404s — on device, "error fetching address". (master fails here too, just differently: it takes `Answer[0]` and `processBIP21Uri` matches on
`.includes('bitcoin:')`, so the decoy slips through and parses no `lno`. The selection fix is needed — I'd key it on the prefix rather than the count. Note the decoy has `bitcoin:` mid-sentence, so it needs `startsWith` after the quotes
come off.)

```diff
-            // BIP 353 mandates exactly one TXT (type 16) record. Reject
-            // zero, multiples, or non-TXT answers (e.g. a spoofed CNAME).
-            const txtAnswers = Array.isArray(json?.Answer)
-                ? json.Answer.filter((a: any) => a?.type === 16)
-                : [];
+            // Concatenate each TXT record's character-strings in RDATA order,
+            // never across records. BIP 353 requires TXT records at the label
+            // which don't begin with 'bitcoin:' to be ignored, and treats
+            // multiple 'bitcoin:' records as invalid.
+            const bitcoinTxts: string[] = (
+                Array.isArray(json?.Answer) ? json.Answer : []
+            )
+                .filter(
+                    (a: any) => a?.type === 16 && typeof a?.data === 'string'
+                )
+                .map((a: any) => a.data.replace(/("|\\|\s+)/g, ''))
+                .filter((data: string) =>
+                    data.toLowerCase().startsWith('bitcoin:')
+                );
 
-            if (dnssecValidated && txtAnswers.length === 1) {
-                bolt12 = txtAnswers[0].data;
-                bolt12 = bolt12!.replace(/("|\\|\s+)/g, '');
-                bolt12 = bolt12!.replace(/bitcoin:b12=/, '');
+            if (dnssecValidated && bitcoinTxts.length === 1) {
+                bolt12 = bitcoinTxts[0].replace(/bitcoin:b12=/, '');
```

Keeps the type-16 filter, and stripping before the prefix test also gives us the per-record multi-`character-string` concatenation the BIP calls for. tsc, lint and prettier pass; `matt@mattcorallo.com` resolves to the BOLT 12 Send screen with it applied.

Non-blocking: BIP 353 also says clients "MUST NOT trust a remote resolver to validate DNSSEC on their behalf", so the `AD` bit isn't strictly conformant — perhaps a `TODO` so the comment doesn't read as full conformance.

[res]: https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki#resolution
[ex]: https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki#examples
[proofs]: https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki#example-proofs

kaloudis:
Thanks @shubhamkmr04, you're right on all counts. Applied in 74e13fd7d:

**Record selection**: took your diff essentially verbatim. Confirmed against the live DoH response for `matt.user._bitcoin-payment.mattcorallo.com` (Status 0, AD true, decoy plus two-part real record): the count guard rejected it, while the prefix-keyed selection accepts it and reconstructs the URI across both character-strings. Your read on master checks out too: `Answer[0]` is the decoy, which contains `bitcoin:` mid-sentence, so it slipped past `processBIP21Uri`'s `.includes('bitcoin:')` and parsed no `lno`. One tweak on top of your diff: the legacy `bitcoin:b12=` strip is now anchored and case-insensitive (`/^bitcoin:b12=/i`) so it can't fire mid-URI and agrees with the case-insensitive prefix filter.

**AD bit**: added the TODO and softened the comment so the block no longer reads as full conformance. Client-side proof-chain validation stays a follow-up.

Also added the unit tests promised in the PR body: DoH-mocked cases in `utils/handleAnything.test.ts` for the AD/Status gate, the BIP 353 worked example (decoy ignored, character-strings concatenated in RDATA order), multiple `bitcoin:` records, non-TXT answers, and empty replies. PR body updated to match.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4362, opened 2026-08-11
fix: pin reviewed invoice so a swapped payment string can't be paid

kaloudis:
# Description

`views/PaymentRequest.tsx` binds its review to the shared `InvoicesStore` singleton and re-reads the invoice from it at confirmation time. `triggerPayment` does `const { paymentRequest, pay_req } = InvoicesStore;` and dispatches `payment_request: paymentRequest`, so the invoice actually paid is whatever the store holds at the moment of the swipe/tap, not the one the user reviewed.

`InvoicesStore.getPayReq()` mutates those observables in place, and it is reachable from many entry points, including `utils/handleAnything.ts` (the universal input router for QR, clipboard, deep links, and NFC). If a second payment string reaches `getPayReq` while the review screen is open, `InvoicesStore.pay_req` / `paymentRequest` are swapped out from under the review. A confirmation the user began for invoice A can then dispatch invoice B ("review A, pay B"). The pre-existing `paymentInFlight` guard only prevents dispatching the same payment twice; it does nothing about the invoice swap.

`views/Cashu/CashuPaymentRequest.tsx` has the same shape of issue against `CashuStore`: both of its payment paths (`MultimintPayment` and `CashuSendingLightning`) read the invoice off the store at their own mount.

## Fix

- Pin the reviewed invoice (`reviewedPaymentRequest`) when the screen mounts.
- Fail closed in `triggerPayment`: if the store's `paymentRequest` no longer matches the pinned value, refuse to pay rather than paying a different invoice than the one on screen when the confirmation began, and reset the slide-to-pay knob.
- When the invoice in the store no longer matches the pin, replace the pay controls with a warning that the payment request changed plus an explicit "Review new payment request" button that re-pins to the request now on screen (the observer view is already displaying its details at that point).
- Invalidate any in-progress slide-to-pay gesture (`swipeButtonKey`) via a reaction whenever the store invoice changes, so a swipe cannot complete against a swapped-in invoice.
- The `focus` listener never advances the pin: a focus event also fires when returning from a screen pushed on top (e.g. the QR view from the header), so re-pinning there would silently adopt an invoice injected while the user was away (caught in review, thanks @shubhamkmr04). This is safe for normal flows because `SendingLightning` exits via `popTo('Wallet')`, which unmounts the review screen, so fresh reviews always re-pin at mount.
- `CashuPaymentRequest` gets the same pin, dispatch guard, reaction, and warning UI. Its existing re-decode-on-focus behavior is kept but now re-decodes the pinned invoice instead of a mount-captured value, so returning from the QR view restores the reviewed request rather than adopting an injected one.

Adds two locale strings (`views.PaymentRequest.payReqChanged`, `views.PaymentRequest.reviewNewPayReq`), shared by both views. No new dependencies.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [ ] I've run `yarn run test` and made sure all of the tests pass

## Testing

- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Manual test plan

1. Open a normal invoice so the PaymentRequest review screen shows its amount and details; confirm a normal swipe/tap still pays that invoice.
2. With the review screen open and focused, trigger a second payment string into `InvoicesStore.getPayReq` (for example a `lightning:` deep link, an NFC tap, or clipboard handling) so the store swaps to a different invoice. Confirm the slide-to-pay control resets and the pay controls are replaced by the "payment request changed" warning with a "Review new payment request" button; tapping the button and then paying pays the request shown on screen.
3. Open the QR view from the header, inject a different invoice while it is open, then navigate back. Confirm the swapped invoice is not silently paid (warning shown instead of pay controls).
4. Navigate back out and into a fresh invoice. Confirm the newly loaded invoice can be paid normally with no warning.
5. Repeat 1 through 4 on an ecash wallet (CashuPaymentRequest), including with the multi-mint toggle enabled.

### Locales

### Other:

### DIFF
--- a/utils/LinkingUtils.ts
+++ b/utils/LinkingUtils.ts
@@ -6,6 +6,15 @@ import handleAnything from './handleAnything';
 import { processSharedQRImageFast } from './ShareIntentProcessor';
 import { settingsStore } from '../stores/Stores';
 
+// Payment status screens that must never be popped out from under an
+// in-flight payment by a deep link. Mirrors App.tsx
+// SCREENS_WITH_CUSTOM_BACK_HANDLER's payment entries.
+const SENDING_SCREENS = [
+    'SendingLightning',
+    'SendingOnChain',
+    'CashuSendingLightning'
+];
+
 class LinkingUtils {
     private shareIntentProcessed = false;
     private pendingDeepLink: string | null = null;
@@ -143,7 +152,7 @@ class LinkingUtils {
                     const focusedRoute = routes[state?.index ?? 0]?.name;
                     if (
                         route === 'PaymentRequest' &&
-                        focusedRoute !== 'SendingLightning' &&
+                        !SENDING_SCREENS.includes(focusedRoute) &&
                         routes.some((r) => r.name === 'PaymentRequest')
                     ) {
                         navigation.popTo(route, props);
--- a/utils/LinkingUtils.ts
+++ b/utils/LinkingUtils.ts
@@ -126,7 +126,30 @@ class LinkingUtils {
         } else {
             handleAnything(url)
                 .then(([route, props]) => {
-                    navigation.navigate(route, props);
+                    // navigate() only reuses an existing route when it is the
+                    // currently focused one, so a payment request arriving
+                    // while another screen sits on top of PaymentRequest
+                    // (e.g. the QR view opened from its header) would push a
+                    // SECOND PaymentRequest, freshly pinned to the injected
+                    // invoice. That bypasses the payment-request-changed
+                    // warning on the buried instance and leaves stale review
+                    // screens in the stack. If a PaymentRequest already
+                    // exists, pop back to it instead: it is already showing
+                    // the new invoice alongside the warning. Never pop while
+                    // a payment status screen is focused, so an in-flight
+                    // payment's UI is not torn down.
+                    const state = navigation.getState();
+                    const routes = state?.routes ?? [];
+                    const focusedRoute = routes[state?.index ?? 0]?.name;
+                    if (
+                        route === 'PaymentRequest' &&
+                        focusedRoute !== 'SendingLightning' &&
+                        routes.some((r) => r.name === 'PaymentRequest')
+                    ) {
+                        navigation.popTo(route, props);
+                    } else {
+                        navigation.navigate(route, props);
+                    }
                 })
                 .catch((err) =>
                     console.error(
--- a/locales/en.json
+++ b/locales/en.json
@@ -850,6 +850,8 @@
     "views.PaymentRequest.thankYouForDonation": "Thank you for your generous donation!",
     "views.PaymentRequest.amountDonated": "Amount donated",
     "views.PaymentRequest.addATip": "Add a Tip with Each Payment",
+    "views.PaymentRequest.payReqChanged": "The payment request changed while this screen was open. Review the new request before paying.",
+    "views.PaymentRequest.reviewNewPayReq": "Review new payment request",
     "views.Receive.title": "Receive",
     "views.Receive.successCreate": "Successfully created invoice",
     "views.Receive.warningLndHub": "Please note that LNDHub has a fixed on-chain address",
--- a/utils/LinkingUtils.test.ts
+++ b/utils/LinkingUtils.test.ts
@@ -65,19 +65,28 @@ describe('LinkingUtils.handleDeepLink', () => {
         expect(navigation.popTo).not.toHaveBeenCalled();
     });
 
-    it('does not pop while a payment is in flight on SendingLightning', async () => {
-        const navigation = makeNavigation([
-            'Wallet',
-            'PaymentRequest',
-            'SendingLightning'
-        ]);
+    it.each(['SendingLightning', 'SendingOnChain', 'CashuSendingLightning'])(
+        'does not pop while a payment is in flight on %s',
+        async (sendingScreen) => {
+            const navigation = makeNavigation([
+                'Wallet',
+                'PaymentRequest',
+                sendingScreen
+            ]);
 
-        LinkingUtils.handleDeepLink('lightning:lnbc1...', navigation as any);
-        await flushPromises();
+            LinkingUtils.handleDeepLink(
+                'lightning:lnbc1...',
+                navigation as any
+            );
+            await flushPromises();
 
-        expect(navigation.navigate).toHaveBeenCalledWith('PaymentRequest', {});
-        expect(navigation.popTo).not.toHaveBeenCalled();
-    });
+            expect(navigation.navigate).toHaveBeenCalledWith(
+                'PaymentRequest',
+                {}
+            );
+            expect(navigation.popTo).not.toHaveBeenCalled();
+        }
+    );
 
     it('navigates normally for routes other than PaymentRequest', async () => {
         mockHandleAnythingResult = [
--- a/utils/LinkingUtils.test.ts
+++ b/utils/LinkingUtils.test.ts
@@ -1,0 +1,115 @@
+jest.mock('./LocaleUtils', () => ({
+    localeString: (s: string) => s
+}));
+jest.mock('./ShareIntentProcessor', () => ({
+    processSharedQRImageFast: jest.fn()
+}));
+
+let mockLoginRequired = false;
+jest.mock('../stores/Stores', () => ({
+    settingsStore: {
+        loginRequired: () => mockLoginRequired,
+        settings: {}
+    }
+}));
+
+let mockHandleAnythingResult: [string, any] = ['PaymentRequest', {}];
+jest.mock('./handleAnything', () => ({
+    __esModule: true,
+    default: jest.fn(() => Promise.resolve(mockHandleAnythingResult))
+}));
+
+import LinkingUtils from './LinkingUtils';
+
+const flushPromises = () => new Promise(setImmediate);
+
+const makeNavigation = (routeNames: string[], focusedIndex?: number) => ({
+    navigate: jest.fn(),
+    popTo: jest.fn(),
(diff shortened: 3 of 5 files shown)


shubhamkmr04 (commented):
https://github.com/user-attachments/assets/14d433c1-2fed-4693-8c38-c66253d88bef

we load a invoice, then injected a different invoice, tried to pay it but the swipe knob gets stuck (in case of btn, its unresponsive). 
maybe we should let the user know the invoice was changed rather than payment getting stuck without any warning

shubhamkmr04:
It does pay the second invoice when it was injected while the user was away (on the QR code view from header)

kaloudis (commented):
Self-review notes from a second pass over 20d8f91.

What holds up:

- The core race is genuinely closed. `triggerPayment` destructures `paymentRequest` once and runs synchronously through to `TransactionsStore.sendPayment`, with no `await` between the guard and dispatch, so the guard and the dispatched invoice are atomic. The guard catches exactly the window the `swipeButtonKey` remount can't: a swap landing after the gesture succeeds but before `triggerPayment` runs.
- Pin timing is sound for every current caller. `getPayReq` sets `paymentRequest` synchronously at the top (stores/InvoicesStore.ts:703) and all five navigation paths into this screen call `getPayReq` before `navigate('PaymentRequest')`, so the mount-time pin always captures the right string.
- The guard precedes the Zaplocker `broadcastAttestation()` call, so a refused payment doesn't broadcast an attestation. Both the SwipeButton and the plain pay button route through `triggerPayment`, so both are covered.

One behavioral issue found (permanently dead pay controls after an in-focus swap) and a guard-disarm edge, both inline. Smaller notes that don't need code changes:

- `getWithdrawalReq` also writes `paymentRequest` (stores/InvoicesStore.ts:729), so a background withdrawal-request decode also trips the reaction and the guard. That's fail-closed and acceptable, just noting it widens what counts as a "swap".
- Re-injecting the same string (`getPayReq(A)` while reviewing A) doesn't fire the reaction since the observable value doesn't change, so no spurious slider reset. Confirmed intentional.
- `CashuPaymentRequest` has the same class of bug (reads its store at confirmation time). Fine to keep this PR scoped, but the body should mention it as a known follow-up so reviewers know it wasn't missed.

Nits, take or leave: `payReqDisposer` could be typed `IReactionDisposer` since `mobx` is already imported; the reaction's setState could use the functional form; the guard reads `this.state.reviewedPaymentRequest` while everything else in `triggerPayment` is destructured above.

kaloudis:
Thanks for catching both of these, great find on the QR view path. Addressed in 3040bcab9:

**QR view bypass (paying the injected invoice):** the culprit was the focus listener, which re-pinned the review to whatever was in the store on every focus. Returning from a screen pushed on top of this one (the QR view from the header) fires focus too, so an invoice injected while you were away got silently re-pinned and paid. The focus listener now only resets the slide-to-pay position and never advances the pin. This has no cost to normal flows: after a payment, `SendingLightning` exits via `popTo('Wallet')`, which unmounts `PaymentRequest`, so a fresh review always remounts and pins correctly at mount.

**Stuck knob / unresponsive button:** instead of the dispatch guard silently refusing, the screen now notices the mismatch at render time (it is an observer over `InvoicesStore`, so the derived check is always current). When the request in the store no longer matches the pinned one, the pay controls are replaced with a warning that the payment request changed plus an explicit "Review new payment request" button. The screen is already displaying the new request's details at that point, so tapping the button acts as acknowledgment and re-pins to what is on screen. The dispatch-time guard is kept as defense in depth and now resets the swipe knob if it ever fires.

One behavior note: if a new payment string lands while this screen is mounted (e.g. the user themselves opens a deep link mid-review), they now see the warning and need one extra tap on "Review new payment request" before paying. I think that friction is correct given the request genuinely changed under them.

`CashuPaymentRequest` has the same shape of issue (it reads `CashuStore.paymentRequest` at swipe time); keeping that out of this PR to stay focused, will follow up separately.

kaloudis:
Follow-up to my last comment: decided to cover CashuPaymentRequest in this PR after all rather than a separate one, since it is the same class of bug and the fix reuses the same pin, guard, and warning strings. Pushed in bd3e0141a and the PR body is updated with the extended scope and test plan. One difference from the LN view: its focus listener already re-decoded the invoice on focus, so that behavior is kept but now re-decodes the pinned invoice instead of a mount-captured value, meaning a return from the QR view restores the reviewed request instead of adopting an injected one.

shubhamkmr04:
https://github.com/user-attachments/assets/dfd93cac-a37e-4bba-9c61-157c8401e5bc

shubhamkmr04:
https://github.com/user-attachments/assets/c59a8473-1129-4951-9314-57b24627d69e

when we inject a new invoice, the donation amount should also update accordingly

kaloudis:
Thanks for both videos, they exposed two distinct remaining gaps. Addressed in 0ed3f2ab2 and c34a7db72, plus a small rename in e5bc5dba9 so the two reactions in each view are named by intent (donationDisposer / swipeResetDisposer).

**Stale donation amount (second video):** the donation amount, percentage selection, and fee mode (fixed vs percent) were derived once at mount from the invoice in the store, and neither the swap reaction nor the "Review new payment request" button re-derived them. The screen live-renders whatever invoice the store holds, so after an injection you saw the new invoice's details with the old invoice's donation math, and paying would have dispatched the old donation amount. `PaymentRequest` now mirrors the reaction `CashuPaymentRequest` already had: it observes the decoded `pay_req` and re-derives donation state and fee mode the moment a different invoice lands, with the same `donationLockRequest` guard so a re-decode of the same request does not clobber a donation you customized. So in your repro, the Donate row flips to 10 sats (10%) as soon as the 100 sat invoice decodes, before you even tap the review button. One deliberate deviation from the Cashu pattern: no `getRequestAmount` guard, so no-amount invoices still recompute (with amount 0). That preserves the previous mount behavior and avoids staleness when swapping from an amount invoice to a no-amount one.

**First video (inject while on the QR view):** what you hit there is a react-navigation 7 semantics hole rather than the pin logic. `navigate()` only reuses an existing route when it is the currently focused one, so with the QR view on top, the deep link dispatch in `LinkingUtils` pushed a second `PaymentRequest` instance. That fresh instance pinned the injected invoice at mount, so no warning ever showed (you can see `checkIfLndReady` re-running in your video, which confirms the remount), and back navigation then walked through the stale QR and stale review of the old invoice. `handleDeepLink` now pops back to the existing `PaymentRequest` when one is already in the stack: that buried instance is already rendering the injected invoice alongside the changed-request warning (and, with the first commit, the correct donation amount), and the stale screens above it are discarded, so back goes straight to Wallet. Two guards on the popTo: the existence check is load-bearing because `popTo` replaces the focused route when no match exists (a cold deep link would otherwise swallow Wallet), and a deep link arriving while `SendingLightning` is focused falls back to `navigate` so an in-flight payment's status screen is never torn down. The dispatch decision is covered by a new unit test suite (`utils/LinkingUtils.test.ts`, 5 cases).

Two residual vectors I looked at and deliberately left out of this PR, since both are "review B, pay B" (the fresh instance fully displays the invoice it would pay) rather than wrong-invoice payment: `ShareIntentProcessing` dispatches via `navigation.replace`, and in ecash mode `handleAnything` routes through `ChoosePaymentMethod`, so both can still produce a duplicate review screen deeper in the stack. Happy to follow up on those separately, along with the longer-term fix of binding each `PaymentRequest` instance to its invoice via route params instead of the shared store singleton, which would eliminate this class of bug outright.

`yarn verify` green (1154 tests). Will re-run both of your repros on device.

shubhamkmr04 (approved):
tACK

utils/LinkingUtils.ts
```diff
@@ -126,7 +126,30 @@ class LinkingUtils {
...
+                    // screens in the stack. If a PaymentRequest already
+                    // exists, pop back to it instead: it is already showing
+                    // the new invoice alongside the warning. Never pop while
+                    // a payment status screen is focused, so an in-flight
+                    // payment's UI is not torn down.
+                    const state = navigation.getState();
+                    const routes = state?.routes ?? [];
+                    const focusedRoute = routes[state?.index ?? 0]?.name;
+                    if (
+                        route === 'PaymentRequest' &&
```
ajaysehwal, line 154: @kaloudis, one small consistency issue in `LinkingUtils`: only `SendingLightning` is excluded from `popTo`.

A deep link received during `CashuSendingLightning` could still navigate to `PaymentRequest` on top of the active payment flow.

I think we should mirror the existing Lightning guard:

```ts
focusedRoute !== 'SendingLightning' &&
focusedRoute !== 'CashuSendingLightning' &&

kaloudis: Good catch on the consistency gap. Interestingly, the exact Cashu scenario can't currently fire: `handleAnything` only returns `PaymentRequest` when ecash mode is off (with ecash on, a bolt11 deep link resolves to `ChoosePaymentMethod` instead), and `CashuSendingLightning` can only be focused when ecash is on, so the two conditions are mutually exclusive today.

But the same class of bug is reachable through `SendingOnChain`: PaymentRequest buried under a deep-linked `Send` flow, broadcast in progress, then a new invoice deep link pops the status screen mid-broadcast.

Addressed in 2567ec918 by excluding all three payment status screens (the same set App.tsx protects with `SCREENS_WITH_CUSTOM_BACK_HANDLER`), which also future-proofs the Cashu path against routing changes.

ajaysehwal (approved):
tACK

PR #4363, opened 2026-08-11
fix: warn before sending node credentials over cleartext HTTP

kaloudis:
# Description

When a wallet is configured with a host that uses an explicit `http://` scheme, the REST backends transmit the node's bearer credential over an unencrypted connection with no warning:

- `getURL()` in `backends/LND.ts` and `backends/CLNRest.ts` honors the scheme verbatim (`host.includes('://') ? host : 'https://'+host`), so `http://host` stays cleartext. `LndHub` inherits this via `extends LND`.
- The credential rides in the request headers regardless of scheme: LND/LndHub send the macaroon in `Grpc-Metadata-macaroon`, CLNRest sends the rune, LndHub sends its token.
- The CLNRest import parser will even force cleartext for a shared connection string via `clnrest+http://...` or the legacy `clnrest://http://...` (`utils/ConnectionFormatUtils.ts`).

A macaroon/rune is a bearer credential granting node control, so anyone on the network path (LAN, Wi-Fi, ISP) can capture it and drain funds. Nothing in the connect flow warned the user.

## Fix

`saveWalletConfiguration` now checks whether the credential-bearing host is a cleartext `http://` endpoint that is **not** loopback and **not** a `.onion` service. If so, it requires an explicit confirmation (destructive-styled) before saving.

- Exemptions: loopback (`localhost`, `127.0.0.0/8`, `::1`) never leaves the device, and `.onion` is authenticated at the Tor layer, so neither exposes the credential.
- The existing TLS certificate-verification modal is skipped for cleartext hosts, since there is no certificate to trust; the user sees the single relevant cleartext warning instead of a confusing "install your certificate" prompt.
- Applies to the credential-bearing REST backends only (`lnd`, `cln-rest` via `host`; `lndhub` via `lndhubUrl`).

This is a warn-and-confirm, not a hard block, so legitimate LAN/VPN tunneling users can still proceed deliberately.

Note: this does not change the `certVerification` default (TLS verification off), which is a separate, deliberate design decision for self-signed LND certs. Happy to discuss that separately if desired.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Manual test plan (device tests pending)

Configuring an `lnd`/`cln-rest`/`lndhub` wallet with an `http://` clearnet host now prompts before save; bare-host (defaults https), `https://`, `.onion`, and `localhost`/`127.x`/`[::1]` hosts save without the prompt.

### Locales

### Other:

### DIFF
--- a/views/Settings/WalletConfiguration.tsx
+++ b/views/Settings/WalletConfiguration.tsx
@@ -26,6 +26,7 @@ import ConnectionFormatUtils from '../../utils/ConnectionFormatUtils';
 import { localeString } from '../../utils/LocaleUtils';
 import BackendUtils from '../../utils/BackendUtils';
 import { themeColor } from '../../utils/ThemeUtils';
+import UrlUtils from '../../utils/UrlUtils';
 import ValidationUtils from '../../utils/ValidationUtils';
 import { confirmAction } from '../../utils/ActionUtils';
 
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (commented):
The lndhub branch checks lndhubUrl, but every LND request path uses host || lndhubUrl (backends/LND.ts:241, also 144, 300, 465) — host wins, and it's persisted for lndhub nodes unfiltered (WalletConfiguration.tsx:660), so a leftover http:// host keeps receiving requests while the gate stays silent. Repro: Privacy → Clipboard on; copy lndhub://user:pass; new node → LND (REST) → Server address http://192.168.1.50 → then tap Import on the clipboard banner → Save: no warning, though the same host warns if saved as LND.

views/Settings/WalletConfiguration.tsx
```diff
@@ -521,9 +521,92 @@ export default class WalletConfiguration extends React.Component<
...
+    };
+
+    private hasInsecureCleartextTransport = (): boolean => {
+        const { implementation, host, lndhubUrl } = this.state;
+        switch (implementation) {
+            case 'lnd':
+            case 'cln-rest':
+                return this.isCleartextHttpHost(host);
+            case 'lndhub':
+                return this.isCleartextHttpHost(lndhubUrl);
```
shubhamkmr04: we can update it like 
    return (
        this.isCleartextHttpHost(host) ||
        this.isCleartextHttpHost(lndhubUrl)
    );

kaloudis: The switch is intentional. State keeps values for fields the active implementation doesn't use: the save path persists all fields into the node blob, and there's an in-session path too (an `lndhub://` clipboard import flips `implementation` and sets `lndhubUrl` but doesn't clear `host`). With the blanket OR, a stale `http://` host left over from an lnd config would trigger the warning when saving an lndhub wallet even though lndhub never sends credentials to that host. False warnings teach users to tap Proceed, so I'd rather only warn when the active backend actually transmits credentials over the flagged field, which is what the per-implementation cases encode.

Separately, a follow-up push here will replace `isCleartextHttpHost` with the shared `UrlUtils.isCleartextHttpTransport` helper being hardened in #4375 (this copy has the same host-parsing bypasses found there), but the per-implementation gating will stay.

kaloudis: Round 2 pushed (f22cea073): now that #4375 is merged, the private `isCleartextHttpHost` copy is gone and the switch calls the shared, hardened `UrlUtils.isCleartextHttpTransport` (URL-parser based, covered by the UrlUtils jest suite). Branch also rebased onto master. Per-implementation gating unchanged, per the reasoning above.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4364, opened 2026-08-11
fix: bind ZEUS Pay Cashu redeem to the configured mint

kaloudis:
# Description

Security fix. The ZEUS Pay Cashu flow trusted the socket protocol as a command channel rather than a hint. A `paid` event (and the server-supplied open-payments list) carries the `mint_url` to redeem from, and the wallet passed that URL straight through `redeemCashu` -> `checkInvoicePaid` -> `checkAndMintCDK`: it added the named mint, fetched its quote state over a raw HTTP GET, and minted proofs against it, never checking the URL against the mint the lightning address was created with (`settings.lightningAddress.mintUrl`).

A malicious or compromised ZEUS Pay backend (or anyone able to make the server emit a `paid` event for a victim's pubkey) could therefore:

1. Inject a fake deposit: mint unbacked ecash at an attacker mint and fire a local "payment received: N sats" notification.
2. Silently add and trust an unapproved mint for later seed-derived operations (restore uses the seed-derived P2PK key).
3. Drive a raw GET at an attacker-named URL, including `http://` and RFC1918 / loopback / link-local hosts (SSRF-like), with the quote id interpolated unencoded.
4. Fail-open on quote state: the native `addExternalMintQuote` mapped any unrecognized NUT-04 state to `PAID`.

## Fix

Defense in depth, primary fix first:

- `redeemCashu` now refuses any redeem whose `mint_url` does not match a user-approved mint: the configured lightning address mint (`settings.lightningAddress.mintUrl`) or a mint the user has added locally (`cashuStore.mintUrls`). The locally added mints keep open quotes redeemable when the configured mint changes via the picker or is overwritten by an unrelated `lightningAddress` settings write (the settings merge is a shallow top-level spread, and `createZaplocker`/`createNWC` replace the object wholesale). This is the single chokepoint both the socket handler and `redeemAllOpenPaymentsCashu` pass through, so it closes the primary vector regardless of caller. On a match, the locally stored URL is forwarded downstream instead of the server-supplied spelling, so the server cannot influence how the mint is keyed in CDK. Normalization lowercases only the scheme and host; the path stays byte-exact. A rejected manual redeem surfaces a localized error via the existing `ErrorMessage` surface; the automatic socket and startup paths stay silent.
- `checkExternalMintQuote` (Android + iOS) refuses non-HTTPS and internal/loopback/link-local mint hosts, and percent-encodes the quote id.
- `addExternalMintQuote` (Android + iOS) fails closed on an unrecognized state instead of defaulting to `PAID`.

Adds `stores/LightningAddressStore.test.ts` covering the mint-binding guard (reject unapproved mint, accept configured mint incl. slash and scheme/host-case variants while forwarding the configured spelling, path case-sensitivity, accept locally added mints incl. when no mint is configured, reject when nothing is approved, and error surfacing on manual vs automatic rejection).

### Not included (requires server coordination)

Full end-to-end event authentication (ZEUS Pay signing `{quote_id, mint_url, amount_msat, ts}` and the client verifying against the ZEUS Pay pubkey) is a protocol change on the backend and is a recommended follow-up. The mint-binding above provides the concrete protection in the interim: even an unauthenticated forged event can now only ever reference the wallet's own configured mint.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

> Native (Kotlin/Swift) changes are pending on-device verification; the mint-binding guard is covered by unit tests. Device testing of the ZEUS Pay Cashu redeem path (legitimate deposit still redeems; forged/foreign mint is ignored) to follow.

### Locales

### Other:

### DIFF
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -7,7 +7,10 @@ import org.json.JSONArray
 import org.json.JSONObject
 import java.io.File
 import java.net.HttpURLConnection
+import java.net.Inet4Address
+import java.net.InetAddress
 import java.net.URL
+import java.net.URLEncoder
 import java.security.MessageDigest
 import java.util.concurrent.ConcurrentHashMap
 
@@ -728,6 +731,59 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
     }
 
     /**
+     * Reject loopback, private (RFC1918), link-local and other internal hosts
+     * for server-supplied mint URLs. Matches on the literal host only (no DNS
+     * resolution) to avoid a rebinding/latency side channel; this is
+     * defense-in-depth on top of the JS-side configured-mint binding.
+     */
+    private fun isDisallowedMintHost(host: String?): Boolean {
+        if (host.isNullOrBlank()) return true
+        val h = host.trim().trimStart('[').trimEnd(']').lowercase()
+        if (h == "localhost" || h.endsWith(".localhost")) return true
+        // A host containing ':' can only be an IPv6 literal (hostnames
+        // cannot contain it), so parse it numerically instead of
+        // prefix-matching one spelling: mapped (::ffff:127.0.0.1),
+        // hex-mapped (::ffff:7f00:1) and zero-expanded (0:0:0:0:0:0:0:1)
+        // forms of the same internal address all normalize to the same
+        // bytes. This still involves no DNS - the brackets force
+        // InetAddress to treat the input as a literal or throw.
+        if (h.contains(':')) return isDisallowedIpv6Literal(h)
+        return isDisallowedIpv4(h)
+    }
+
+    // IPv4 ranges: 127/8, 10/8, 169.254/16, 172.16/12, 192.168/16, 0/8
+    private fun isDisallowedIpv4(h: String): Boolean {
+        return h.startsWith("127.") ||
+            h.startsWith("10.") ||
+            h.startsWith("169.254.") ||
+            h.startsWith("192.168.") ||
+            h.startsWith("0.") ||
+            Regex("^172\\.(1[6-9]|2[0-9]|3[0-1])\\.").containsMatchIn(h)
+    }
+
+    private fun isDisallowedIpv6Literal(literal: String): Boolean {
+        val addr = try {
+            InetAddress.getByName("[$literal]")
+        } catch (e: Exception) {
+            // Not parseable as an IPv6 literal and not a valid hostname
+            // either: fail closed
+            return true
+        }
+        // IPv4-mapped literals parse to an Inet4Address in any spelling;
+        // hold the embedded address to the same rules as a dotted host
+        if (addr is Inet4Address) {
+            return isDisallowedIpv4(addr.hostAddress ?: return true)
+        }
+        if (addr.isLoopbackAddress || addr.isLinkLocalAddress ||
+            addr.isSiteLocalAddress || addr.isAnyLocalAddress
+        ) {
+            return true
+        }
+        // fc00::/7 unique-local: no InetAddress predicate covers it
+        return (addr.address[0].toInt() and 0xfe) == 0xfc
+    }
+
+    /**
      * Check mint quote status directly from the mint's HTTP API.
      * This bypasses the local database check and works for external quotes
      * (e.g., quotes created by ZeusPay server).
@@ -738,7 +794,36 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
             try {
                 // Normalize mint URL and construct the quote endpoint
                 val normalizedUrl = mintUrl.trimEnd('/')
-                val quoteUrl = "$normalizedUrl/v1/mint/quote/bolt11/$quoteId"
+
+                // Security: this issues a raw HTTP GET against a mint URL that
+                // originates from a ZEUS Pay socket event. Refuse cleartext and
+                // internal/loopback/link-local targets so a forged event cannot
+                // downgrade the fetch or drive the device into an SSRF against
+                // the local network. quoteId is percent-encoded so it cannot
+                // alter the request path.
+                val parsed = try {
+                    URL(normalizedUrl)
+                } catch (e: Exception) {
+                    withContext(Dispatchers.Main) {
+                        promise.reject("INVALID_URL", "Invalid mint URL: $mintUrl")
+                    }
+                    return@launch
+                }
+                if (!parsed.protocol.equals("https", ignoreCase = true) ||
+                    isDisallowedMintHost(parsed.host)
+                ) {
+                    Log.e(TAG, "checkExternalMintQuote: refusing unsafe mint URL $mintUrl")
+                    withContext(Dispatchers.Main) {
+                        promise.reject(
+                            "UNSAFE_MINT_URL",
+                            "Refusing to contact non-HTTPS or internal mint host: $mintUrl"
+                        )
+                    }
+                    return@launch
+                }
+
+                val encodedQuoteId = URLEncoder.encode(quoteId, "UTF-8")
+                val quoteUrl = "$normalizedUrl/v1/mint/quote/bolt11/$encodedQuoteId"
 
                 Log.d(TAG, "checkExternalMintQuote: Fetching $quoteUrl")
 
@@ -820,7 +905,16 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     "PAID" -> QuoteState.PAID
                     "PENDING" -> QuoteState.PENDING
                     "ISSUED" -> QuoteState.ISSUED
-                    else -> QuoteState.PAID // Default to PAID for external quotes
+                    // Fail closed: never assume PAID for an unrecognized state.
+                    else -> {
+                        withContext(Dispatchers.Main) {
+                            promise.reject(
+                                "INVALID_QUOTE_STATE",
+                                "Unrecognized mint quote state: $state"
+                            )
+                        }
+                        return@launch
+                    }
                 }
 
                 // Storing a key equal to cdk's seed prefix on the quote
--- a/ios/CashuDevKit/CashuDevKitModule.swift
+++ b/ios/CashuDevKit/CashuDevKitModule.swift
@@ -712,6 +712,62 @@ class CashuDevKitModule: RCTEventEmitter {
         }
     }
 
+    /// Reject loopback, private (RFC1918), link-local and other internal hosts
+    /// for server-supplied mint URLs. Matches on the literal host only (no DNS
+    /// resolution) to avoid a rebinding/latency side channel; this is
+    /// defense-in-depth on top of the JS-side configured-mint binding.
+    private static func isDisallowedMintHost(_ host: String) -> Bool {
+        let h = host.trimmingCharacters(in: CharacterSet(charactersIn: "[]"))
+            .lowercased()
+        if h.isEmpty { return true }
+        if h == "localhost" || h.hasSuffix(".localhost") { return true }
+        // A host containing ":" can only be an IPv6 literal (hostnames
+        // cannot contain it), so parse it numerically instead of
+        // prefix-matching one spelling: mapped (::ffff:127.0.0.1),
+        // hex-mapped (::ffff:7f00:1) and zero-expanded (0:0:0:0:0:0:0:1)
+        // forms of the same internal address all normalize to the same
+        // bytes. inet_pton involves no DNS.
+        if h.contains(":") { return isDisallowedIPv6Literal(h) }
+        return isDisallowedIPv4(h)
+    }
+
+    // IPv4 ranges: 127/8, 10/8, 169.254/16, 172.16/12, 192.168/16, 0/8
+    private static func isDisallowedIPv4(_ h: String) -> Bool {
+        if h.hasPrefix("127.") || h.hasPrefix("10.") ||
+            h.hasPrefix("169.254.") || h.hasPrefix("192.168.") ||
+            h.hasPrefix("0.") {
+            return true
+        }
+        return h.range(of: "^172\\.(1[6-9]|2[0-9]|3[0-1])\\.",
+                       options: .regularExpression) != nil
+    }
+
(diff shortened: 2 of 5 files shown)


android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
```diff
@@ -727,6 +728,31 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         }
     }
 
+    /**
+     * Reject loopback, private (RFC1918), link-local and other internal hosts
+     * for server-supplied mint URLs. Matches on the literal host only (no DNS
+     * resolution) to avoid a rebinding/latency side channel; this is
+     * defense-in-depth on top of the JS-side configured-mint binding.
+     */
+    private fun isDisallowedMintHost(host: String?): Boolean {
```
ajaysehwal, line 739: @kaloudis, I think there’s a small edge-case bug in `isDisallowedMintHost`.

The current check uses:

```ts
h.startsWith('fc') || h.startsWith('fd')  
```

This can produce false positives on normal hostnames such as fc-mint.example.com, since they are not IPv6 ULA addresses.

I think the fc/fd checks should only apply to IPv6 literals, for example when the host contains :. The same issue appears to exist on iOS as well.

kaloudis: Good catch, thank you. The fc/fd prefix check was meant for IPv6 unique-local literals (fc00::/7) but as written it also matched regular hostnames. Fixed in 76310c662 on both platforms: the fc/fd check now only applies when the host contains ":", which is true for every IPv6 literal and never true for a hostname (the port is split off during URL parsing). `::1` and `fe80:` are unaffected since they already require a colon. Will squash before merge.

ios/CashuDevKit/CashuDevKitModule.swift
```diff
@@ -723,7 +752,24 @@ class CashuDevKitModule: RCTEventEmitter {
...
+                guard let baseUrl = URL(string: normalizedUrl),
+                      baseUrl.scheme?.lowercased() == "https",
+                      let host = baseUrl.host,
+                      !CashuDevKitModule.isDisallowedMintHost(host)
+                else {
+                    reject("UNSAFE_MINT_URL", "Refusing to contact non-HTTPS or internal mint host: \(mintUrl)", nil)
+                    return
+                }
+
+                let encodedQuoteId = quoteId.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? quoteId
```
ajaysehwal: @kaloudis, One platform-parity gap: on iOS, `quoteId.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)` won’t escape `/`. `.urlPathAllowed` is intended for an entire path, so `/` is allowed by design, but here `quoteId` is a single path segment.

Android uses `URLEncoder.encode(quoteId, "UTF-8")`, which does escape `/`. Since `quoteId` comes from the same untrusted socket event as `mintUrl`, a `/`-containing value could still alter the request path on iOS, which undermines the guarantee described in the comment above.

Could you use `CharacterSet.urlPathAllowed.subtracting(CharacterSet(charactersIn: "/"))` (or equivalent) so both platforms actually enforce the same behavior?

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
```diff
@@ -727,6 +728,33 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
...
```
ajaysehwal, line 739: @kaloudis, this closes a real gap. One thing worth double-checking: the host filter is string-prefix-based on the literal host, which the comment says is deliberate to avoid DNS resolution.

That means it won’t catch IPv4-mapped IPv6 literals like `https://[::ffff:127.0.0.1]/...` (`host` starts with `::ffff:` so none of the existing checks match), or non-canonical loopback forms like `[0:0:0:0:0:0:0:1]`.

Could we normalize IPv4-mapped IPv6 forms specifically (e.g. strip `::ffff:` and re-check against the IPv4 rules) without doing full DNS resolution? Not asking for a fully general IP parser — just want to make sure this well-known SSRF bypass is a conscious tradeoff rather than an oversight.

ios/CashuDevKit/CashuDevKitModule.swift
kaloudis: Confirmed and fixed in e0edeac86. You are right that `.urlPathAllowed` is the wrong set here: it is scoped to a whole path, so `/` is allowed by design and survived encoding, while Android's `URLEncoder` escapes it. The gap had more bite than platform parity: the JS-side binding constrains `mint_url` but not `quote_id`, which arrives in the same socket event, so on iOS a forged quote id could steer the GET to an arbitrary path on the configured mint (`.` and `/` both passed, so `../` traversal worked). Took your suggestion: the id is now encoded with `urlPathAllowed.subtracting("/")`, and the `?? quoteId` fallback is gone, since falling back to the raw id on a nil encoding quietly contradicted the same guarantee; the call now rejects instead. `?`, `#` and `%` were already outside the set, so query injection and double-encoding were not reachable. `/` was the only gap.

Both round-2 fixes are squashed into the branch head along with the earlier fc/fd fixup, so the PR is back to a single commit.

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
kaloudis: An oversight, not a tradeoff, thanks. Confirmed all three forms passed: `::ffff:127.0.0.1` starts with `:` so none of the IPv4 prefixes matched, and it is not `::1`, `fe80`, `fc` or `fd`; same for the hex-mapped `::ffff:7f00:1` and the zero-expanded `0:0:0:0:0:0:0:1`.

Went one step past stripping `::ffff:`: a host containing `:` can only be an IPv6 literal, never a hostname, so that branch now parses the literal numerically and checks bytes instead of spellings. On Android that is `InetAddress.getByName` on a bracketed literal, which parses or throws and never resolves (verified `[fc-mint.example.com]` throws `UnknownHostException`); on iOS, `inet_pton(AF_INET6)`. Mapped forms in any spelling collapse to the embedded IPv4 and are held to the same rules as a dotted host (Java returns an `Inet4Address` for them, verified on-JVM), the v6 path rejects loopback, unspecified, link-local, site-local and `fc00::/7` by bytes, and an unparseable `:` host fails closed, which also covers zone-id forms. Verified the byte logic against a table of blocked forms (mapped-dotted, mapped-hex, expanded loopback, mapped RFC1918 and link-local, `fe80`/`fec0`/`fc00`/`fd`, junk, zone-id) and allowed forms (public v6, mapped-public `::ffff:8.8.8.8`, `64:ff9b::`), plus `:app:compileDebugKotlin`.

Two residuals stay deliberate: exotic dotted-quad spellings on the IPv4 side (Java does not parse them, and the JS-side configured-mint binding runs before any of this), and NAT64-style prefixes, which embed public IPv4 by design. Fixed in e0edeac86 alongside the quoteId encoding from your other comment, and squashed with the rest of the branch.

ajaysehwal (commented):
@kaloudis, `settings.lightningAddress.mintUrl` isn’t as stable as this guard assumes, and there’s no fallback when it drifts.

`SettingsStore.applySettingsUpdate` uses a shallow top-level spread, so `lightningAddress` gets replaced wholesale on writes. `createZaplocker` and `createNwc` both update `lightningAddress` without spreading the existing object, which can wipe `mintUrl`. The user can also change the mint from the picker, and `lightningAddress` is global while ZEUS Pay addresses are keyed per-node — so two nodes using Cashu addresses on different mints still share a single `mintUrl`.

In all of these cases, quotes already sitting in `this.paid` for the previous mint become permanently unredeemable, with only a `console.warn` as the signal.

Could we widen the trust set slightly — accept the configured mint **or** any mint the user has already added locally via `cashuStore.mintUrls`? That preserves the security property of rejecting arbitrary server-provided mints while avoiding stranding funds when the configured mint changes or gets overwritten.

stores/LightningAddressStore.ts
```diff
@@ -816,6 +819,32 @@ export default class LightningAddressStore {
...
+        // ever mint from the mint this lightning address was created with
+        // (settings.lightningAddress.mintUrl, set in createCashu). Binding
+        // here, at the single chokepoint every caller passes through, stops a
+        // malicious or compromised backend from steering the wallet to an
+        // arbitrary mint (fake-deposit injection, unapproved mint trust,
+        // raw-HTTP fetch of an attacker URL).
+        const configuredMintUrl =
+            this.settingsStore?.settings?.lightningAddress?.mintUrl;
+        if (
+            !configuredMintUrl ||
```
ajaysehwal: @kaloudis, small UX gap here: this rejection path returns `false` without setting `this.error` / `this.error_msg`, unlike the other failure branches in `redeemCashu`.

The manual Redeem button (`CashuPayment.tsx:144`) ignores the return value, so tapping Redeem on a mismatched payment gives the user no feedback — the payment just remains there.

Could we set `error` / `error_msg` with a localized message (e.g. `stores.LightningAddressStore.Cashu.mintMismatch`) so the existing `ErrorMessage` surface picks it up?

For the automatic socket path, a silent drop may be appropriate, but the manual action should give the user some feedback.

ajaysehwal (commented):
@kaloudis, worth tightening this a bit: the comparison lowercases both sides, but `CashuStore.normalizeMintUrl` only strips trailing slashes and doesn’t lowercase.

So a `mint_url` differing only in case can pass the guard and then be forwarded verbatim into `checkInvoicePaid` → `checkAndMintCDK` → `ensureMintAdded(mintUrl)`, potentially creating a second mint entry under the differently-cased URL. That’s the same unapproved-mint trust vector this PR is trying to close, just narrower.

Easy fix: once the guard passes, forward `configuredMintUrl` instead of the server-supplied `mint_url`. The latter has already served its purpose as the comparison hint and shouldn’t be used for downstream mint keying.

Also, the “trailing slash or case” test currently only asserts that `checkInvoicePaid` was called. It would be good to assert it was called with `CONFIGURED_MINT` to lock this behavior in.

stores/LightningAddressStore.ts
```diff
@@ -805,6 +801,13 @@ export default class LightningAddressStore {
         }
     };
 
+    // Normalize a mint URL for equality comparison: strip surrounding
+    // whitespace, any trailing slash, and lower-case it so cosmetic
+    // differences do not cause a false mismatch. Kept local so the security
+    // check below does not depend on another store's private helper.
+    private normalizeMintUrl = (url?: string): string =>
```
ajaysehwal: @kaloudis, this lowercases the entire URL, but only the scheme and host are case-insensitive — paths aren’t. So `https://m.com/Mint` and `https://m.com/mint` would compare equal here.

Mints with path prefixes are uncommon, so the practical risk is low, but since this is a security comparison, it might be worth normalizing only the scheme + host while keeping the path byte-exact.

stores/LightningAddressStore.ts
kaloudis: Agreed, fixed in 4dda46463. The rejection path now sets `error` / `error_msg` with a new localized string (`stores.LightningAddressStore.Cashu.mintMismatch`) when the redeem is user-initiated, so the existing `ErrorMessage` surface in the lightning address view picks it up with no view changes.

The automatic paths stay silent as you suggested. The discriminator is the existing `localNotification` flag: both automatic callers (the socket `paid` handler and the startup redeem-all) pass `true`, while both manual entry points (the per-payment Redeem button and the manual redeem-all) leave it unset. Two new tests lock this in: manual rejection sets the message, automatic rejection does not.

stores/LightningAddressStore.ts
kaloudis: Right, fixed in 4dda46463. `normalizeMintUrl` now lowercases only the `scheme://host` prefix and leaves the path byte-exact, so `https://m.com/Mint` and `https://m.com/mint` no longer compare equal. The prefix is matched with a regex rather than the URL polyfill, since the RN polyfill does not case-fold hosts. Added a test with a path-case variant of a Minibits-style URL (`/Bitcoin` vs `/bitcoin`).

Combined with the change from your other comment (forwarding the locally stored URL on a match), a case variant can now neither pass the guard through the path nor influence downstream mint keying through the host.

kaloudis:
@ajaysehwal Both of your review-level points from this round are addressed in 4dda46463, alongside the two inline ones.

**`mintUrl` drift / stranded quotes.** Confirmed on all three mechanisms you named: the settings merge is a shallow top-level spread so any `lightningAddress` write replaces the whole object, `createZaplocker` and `createNWC` both write it without spreading the existing value (wiping `mintUrl`), the mint picker legitimately changes it, and `lightningAddress` is global while addresses are keyed per node pubkey. Took your suggestion: the trust set is now the configured mint plus `cashuStore.mintUrls`. Since `mintUrls` is per-node and strictly user-added, the core property is preserved (the server can never name an arbitrary or unapproved mint), and the widened set also covers the wiped-`mintUrl` case, since the trust set stays non-empty for any Cashu user. New tests cover accepting a locally added mint both alongside and in the absence of a configured mint, and still rejecting when nothing is approved.

**Forwarding the server-supplied spelling.** Confirmed: the guard compared case-insensitively but then passed the server's `mint_url` into `checkInvoicePaid` -> `checkAndMintCDK` -> `ensureMintAdded`, and `CashuStore.normalizeMintUrl` only strips trailing slashes, so a case variant would have been keyed as a distinct mint in CDK. The guard now returns the matched locally stored URL and forwards that downstream; the server value is only ever used as a comparison hint. The slash/case test now asserts `checkInvoicePaid` was called with the configured spelling, as you suggested, and the locally-added-mint tests assert the local spelling likewise.

Branch re-squashed to a single commit. `tsc`, lint, prettier, and the full test suite (1260 tests) are green.

PR #4365, opened 2026-08-11
fix(bolt12): require offer review before paying on LDK Node

kaloudis:
# Description

**Stacked on #4353** (routing fee limit for LDK BOLT 12 payments); review the last commit only. Merging #4353 first will make this diff collapse to its own commit.

On LDK Node, tapping **Proceed** on a BOLT 12 offer paid it immediately. `LdkNode.fetchInvoiceFromOffer` called `bolt12SendUsingAmount` and then `awaitPaymentCompletion`, so the payment had fully settled before the user saw any detail of what was paid, and the slide-to-pay threshold never engaged. The same button on CLN routes through the `PaymentRequest` review screen, so the two offer-capable backends had opposite behavior behind an identical UI.

ldk-node has no fetch-invoice-only API for offers: `send`/`sendUsingAmount` dispatch the invoice_request and pay the returned invoice internally. So instead of pretending to fetch, this PR reviews what the payment will actually be bound to: the decoded offer plus the user's amount and fee limit.

## What changed

- **Native (`LdkNodeModule.kt`, `LdkNodeModule.swift`, `.m`)**: new `bolt12DecodeOffer` built on `Offer.fromStr` and the offer accessors already present in the bindings. Pure decode: no node required, nothing sent. No binary/FFI changes.
- **`backends/LdkNode.ts`**: `fetchInvoiceFromOffer` is removed and replaced by `decodeOffer` and `payOffer`, so no backend can conflate fetching an invoice with paying one. A new `supportsOffersDirectPay` capability (LDK true, CLN false) gates the two flows.
- **`views/Bolt12OfferReview.tsx`** (new): shows the decoded offer (description, issuer, destination pubkey, offer minimum, expiry, resolved fee limit), fails closed on expired offers, warns on quantity-bearing and fiat-denominated offers, applies the slide-to-pay threshold, and is the only place an offer payment can be dispatched from.
- **`stores/TransactionsStore.ts`**: `SendPaymentReq` gains an `offer` field routed to `BackendUtils.payOffer`. Offer payments now get the in-flight double-submission guard, the timeout backstop, and the store lifecycle, so `SendingLightning` renders this payment's actual outcome (this supersedes the store-side leg of #4307 for the offer path; the Proceed button is also disabled while loading).
- **`views/Send.tsx`**: on direct-pay backends, Proceed now decodes the offer and navigates to the review screen; nothing is dispatched from the Send form. The CLN flow (fetchinvoice, then `PaymentRequest`) is unchanged. Keysend sends now get the same slide-to-pay threshold gate, which they previously bypassed at any amount.

### Notes for review

- The offer amount is user-entered and ldk-node binds the invoice_request to it, so the exposure this closes is unreviewed destination/metadata plus the missing high-value gate, not arbitrary-amount extraction.
- `CLNRest.fetchInvoiceFromOffer` keeps the 4-argument signature from #4353 even though the parity rationale no longer applies; it still ignores the fee limit and applies it downstream in `PaymentRequest`.
- The review screen disables the back-swipe gesture like `PaymentRequest` does, to avoid conflicts with the swipe-to-pay control.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


stores/TransactionsStore.ts
```diff
@@ -701,13 +709,14 @@ export default class TransactionsStore {
...
         }
 
-        const payFunc =
-            (this.settingsStore.implementation === 'cln-rest' ||
-                this.settingsStore.implementation === 'embedded-lnd' ||
-                this.settingsStore.implementation === 'ldk-node') &&
-            pubkey
-                ? BackendUtils.sendKeysend
-                : BackendUtils.payLightningInvoice;
+        const payFunc = offer
```
ajaysehwal, line 727: @kaloudis  `payOffer` can return `false` on backends that don't support direct BOLT12 offers, but `sendPaymentInternal` calls `.then()` on the result whenever `offer` is set.

It's safe with the current caller because `offer` is only set when `supportsOffersDirectPay()` is true, but a future caller could pass an `offer` on CLN/LND and cause a runtime error.

Add a guard for `offer && !supportsOffersDirectPay()` and route it through `handlePaymentError` instead.

views/Send.tsx
ajaysehwal, line 1: @kaloudis  FeeLimit on the BOLT 12 Send screen is misleading on CLN , FeeLimit is now always shown for transactionType === 'BOLT 12'. On CLN, supportsCustomFeeLimit() is still false (this branch), so the control is empty or the old percent-only block, and the 4th argument to fetchInvoiceFromOffer is ignored. The real fee UI remains on PaymentRequest.
Gate that block on supportsOffersDirectPay() (or supportsCustomFeeLimit()), so CLN users do not think Send’s fee field applies to the offer fetch.

stores/TransactionsStore.ts
kaloudis: Good catch. Added the guard in c09d50adf: if `offer` is set while `supportsOffersDirectPay()` is false, `sendPaymentInternal` now fails fast through `handlePaymentError` with the existing `views.Send.payBolt12.offersNotSupported` string, instead of reaching the dispatcher where `payOffer` would resolve to `false` and crash on `.then()`. The guard runs after the in-flight bookkeeping is set up so `handlePaymentError` clears the right sequence's flag.

views/Send.tsx
kaloudis: Agreed. That block was actually introduced by #4353 (the base PR of this stack), so I fixed it there and rebased this branch on top: the FeeLimit control in the BOLT 12 section is now gated on `supportsCustomFeeLimit()`, which is true on ldk-node and false on CLN.

I went with `supportsCustomFeeLimit()` rather than `supportsOffersDirectPay()` because the latter capability is only introduced by this PR and #4353 has to stand on its own; for the backends that can reach the BOLT 12 section the two flags are equivalent. On CLN the Send screen no longer shows a fee field for offers, and the real fee UI stays on PaymentRequest where the payment is dispatched (the ignored 4th argument to `CLNRest.fetchInvoiceFromOffer` keeps its signature-parity comment). The timeout field stays visible on CLN since `fetchinvoice` does use it.

New hashes after the rebase: #4353 is now 82af24070, this branch's review commit is 6755a9bb0 plus the new guard commit c09d50adf.

ajaysehwal:
@kaloudis , Please rebase

views/Send.tsx
```diff
@@ -127,6 +131,8 @@ interface SendState {
 )
 @observer
 export default class Send extends React.Component<SendProps, SendState> {
+    listener: any;
```
ajaysehwal: @kaloudis, `listener` is declared and checked in `componentWillUnmount`, but it’s never assigned anywhere in the class.

The guard makes this a no-op today, so it’s not a runtime issue, but the leftover cleanup code is misleading and looks like a forgotten subscription. Worth removing the unused `listener` field and cleanup if it’s no longer needed.

views/Send.tsx
```diff
@@ -1308,6 +1343,29 @@ export default class Send extends React.Component<SendProps, SendState> {
                                         });
                                     }}
                                 />
+                                {/* Only shown where the fee limit applies to
+                                    this screen's offer call (ldk-node pays
+                                    inside it). On CLN only the invoice is
+                                    fetched here; its fee limit is set on
+                                    PaymentRequest, where the payment happens */}
+                                {BackendUtils.supportsCustomFeeLimit() && (
```
ajaysehwal: @kaloudis, the new FeeLimit control on the BOLT12 offer screen is gated on `supportsCustomFeeLimit()` instead of `supportsOffersDirectPay()`, so it’s shown on CLN even though the value has no effect there.

A CLN user can set a low fee limit, but `fetchInvoiceFromOffer` passes it as `_feeLimitSat` and `CLNRest.ts:708` explicitly ignores it. The user then lands on `PaymentRequest` with a separate fee limit, so the value they entered is silently discarded.

I think the FeeLimit control should only be shown when the selected backend actually supports applying it to the BOLT12 direct payment.

views/Bolt12OfferReview.tsx
```diff
@@ -0,0 +1,268 @@
...
+                    )}
+                    {decodedOffer.amountType === 'currency' && (
+                        <WarningMessage
+                            message={`${localeString(
+                                'views.Bolt12OfferReview.currencyDenominated'
+                            )}: ${decodedOffer.currencyAmount} ${
+                                decodedOffer.iso4217Code
+                            }`}
+                        />
+                    )}
```
shubhamkmr04: ldk-node rejects all three of these locally, before any `invoice_request` goes out:

- `amountType === 'currency'` → `Error::UnsupportedCurrency` (`payment/bolt12.rs:214-219`)
- `expectsQuantity` with `quantity: None`, which is what `payOffer` passes → `Bolt12SemanticError::MissingQuantity` (LDK `offers/offer.rs:970`)
- `satAmount` below the offer amount → `Error::InvalidAmount` (`payment/bolt12.rs:223-228`)

The third isn't flagged at all, and it's the one this screen already has both numbers for — the offer amount is rendered as "Offer minimum" just below, but never compared to the amount above it.

Since the screen exists to catch things before dispatch, these seem worth failing closed on like expiry rather than warning:

```ts
const offerAmountMsats =
    decodedOffer.amountType === 'bitcoin' ? decodedOffer.amountMsats : undefined;
const isUnderpaying =
    offerAmountMsats != null && Number(satAmount) * 1000 < offerAmountMsats;
const cannotPay =
    isExpired ||
    isUnderpaying ||
    decodedOffer.expectsQuantity ||
    decodedOffer.amountType === 'currency';
```

The copy would need to follow: both strings attribute the rejection to the recipient, when it's our own node refusing.

Minor, same block: `currencyAmount` is in the ISO 4217 minor unit, so a $2.50 offer renders as `250 USD`.

views/Send.tsx
kaloudis: Removed in cf32fbf86 (field and the `componentWillUnmount` check).

views/Send.tsx
kaloudis: Right, and my earlier answer here is stale: `supportsCustomFeeLimit()` was false on CLN when I wrote it, then master flipped it to true in 54941af6a.

Fixed in 50b670f49, which moves `supportsOffersDirectPay` down into #4353 and gates the block on it there, so #4353 is correct standalone too. #4365 rebased on top: cf32fbf86.

views/Bolt12OfferReview.tsx
kaloudis: All three confirmed against the fork (`src/payment/bolt12.rs:211-225`) and `lightning-0.2.x` `offers/offer.rs:970`. Fixed in cf32fbf86: single `cannotPay` (expired, fiat, expects-quantity, underpaying) drives both buttons, the two warnings are now errors, and the copy attributes the refusal to us. `currencyAmount` is rendered in the major unit via `FiatStore.symbolLookup().decimalPlaces ?? 2`.

PR #4366, opened 2026-08-11
fix(security): verify node TLS certs by default, pin connection-string certs

kaloudis:
# Description

Relates to: Bitcoin Red Team review finding `2026-08-06-zeus-default-trust-all-tls-node-credentials` (critical) — and partially addresses `2026-08-06-zeus-connection-string-cert-params-silently-dropped` (medium).

Remote LND/CLNRest connections shipped with `certVerification: false`, which makes react-native-blob-util run with `trusty: true` — all TLS certificate and hostname verification disabled. An active MITM on the phone→node path (public Wi-Fi, hostile ISP, transparent proxy) can present any self-signed certificate and silently capture the admin macaroon / rune sent in request headers: full remote control of the node.

**What this PR does:**

1. **Secure default**: `certVerification` now defaults to `true` for new configs (store default, load fallback, UI initial state, implementation switch). Nodes saved *without* the key (QR/clipboard imports dropped the `undefined` value) are normalized to an explicit `false` by the `cert-verification-default-v1` MOD_KEY migration (both load paths), so nothing breaks on upgrade — the strict default governs new configs only. Migrating the stored trust-all cohort to verified/pinned TLS is the deliberate follow-up (maintainer sign-off required).

2. **Warning modal on opt-out**: the MITM warning modal now fires when the Certificate Verification switch is turned *off* (previously only at save time), with a dedicated confirm label. The modal is suppressed when the config has pinned certs (the connection stays authenticated regardless of the toggle). Cancel leaves verification on.

3. **Certificate pinning from connection strings** (what makes the secure default usable for self-signed deployments): lndconnect `cert=` and clnrest `certs=` params — previously silently discarded — are normalized into per-node `pinnedCerts` (base64 DER; PEM bundles handled, private keys skipped) and propagated on all four import paths (handleAnything lndconnect + clnrest, importClipboard lndconnect + clnrest). A postinstall patch (`patches/patch-react-native-blob-util.mjs`, repo's existing patch system) teaches react-native-blob-util a `pinnedCerts` option: the pins become the **only trust anchors** — PKIX `TrustManagerFactory` over an in-memory KeyStore on Android; `SecTrustSetAnchorCertificates` + `SetAnchorCertificatesOnly(true)` + `SecTrustEvaluateWithError` on iOS, gated to server-trust challenges only, with the evaluation policy swapped to `SecPolicyCreateBasicX509` so iOS (like Android's permissive verifier) does not enforce hostname/SAN matching that self-signed node certs routinely fail. This covers both the self-signed-leaf case (lndconnect: the cert is its own anchor) and the CA case (clnrest: leaf validates against the pinned CA), and rejects attacker chains that merely carry the public pinned cert as an unvalidated chain member. Pins take precedence over both CA validation and the trust-all opt-out. The config UI shows a "certificate pinned" note when pins are present.

**Verification done:**

- `yarn verify` green (1118 jest tests incl. parser, store-default, migration, and import-propagation coverage; tsc; eslint + check-styles; prettier)
- Patch dry-run against pristine react-native-blob-util@0.24.10: all anchors apply, idempotent re-runs, and postinstall now *fails hard* if a future library bump breaks an anchor
- The patched Android trust manager compiled and executed on the JVM against live TLS servers:
  - self-signed node cert pinned → **accepted** (incl. non-CA self-signed certs)
  - CA-signed leaf with CA pinned → **accepted**
  - attacker leaf with the (public) pinned cert injected as a chain member → **rejected** (`unable to find valid certification path`); the earlier byte-match implementation **accepted** it — confirming both the review's chain-injection critique and its fix
  - wrong pin / strict-default controls → **rejected**

**Still needed before merge:**

- [ ] Manual on-device testing on Android + iOS (Rule 4): scan an lndconnect QR carrying `cert=` against a self-signed LND → connects; rogue-cert negative test; toggle-off modal flow; upgrade-from-previous-release with a saved self-signed remote (migration scenario 3 — blob before/after diff). The iOS challenge-handler code has not been compile-verified yet (no Xcode build run from this branch)
- [ ] Maintainer sign-off on the included `cert-verification-default-v1` migration (storage-touching gate)

Follow-ups not in scope: BTCPay `certificateThumbprint` support (same native hook); WebSocket streams (`wsReq`) have no pinning path and still require CA-valid certs; migrating the stored `certVerification: false` cohort onto verified/pinned TLS.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

_(Device testing pending — see above. Android pinning logic verified on the JVM incl. adversarial chain injection; iOS equivalent by API parity but not yet compile-verified.)_

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

_(No live-node testing yet. Only the remote LND REST and CLNRest transports are touched; embedded LND, LDK Node, LNC, and NWC do not use the blob-util REST path. LndHub never passes `trusty` and defaults to verification on. Live tests against self-signed LND REST + CLNRest are required before merge.)_

### Locales

Two new strings in `locales/en.json` only: `views.Settings.WalletConfiguration.certificateUnderstandDisable`, `views.Settings.AddEditNode.certificatePinned`.

_No dependency changes (`package.json`/`yarn.lock` untouched). A postinstall patch (`patches/patch-react-native-blob-util.mjs`) modifies the installed react-native-blob-util sources via the existing `node patches/index.mjs` step — contributors just need a normal `yarn` (or `node patches/index.mjs` on an existing install). The patch fails the install loudly if the library sources drift._

### Other:

_Worth a note in the user docs (docs.zeusln.app) once shipped: certificate verification is on by default for new node configs, and lndconnect/clnrest connection strings carrying cert material result in pinned connections. Notes for node operators: an LND TLS cert rotation invalidates a leaf pin — re-import the connection string to re-pin. Separately, anchor validation enforces the certificate validity window, so an expired node cert (lnd's autogenerated certs live ~14 months) now fails a pinned connection until the node regenerates it — expect this to surface as 'wallet stopped connecting' with no action by anyone._


kaloudis:
Review addressed in `475dc5d52` (pushed). Point-by-point:

**Critical — chain-injection bypass:** confirmed and fixed. Reproduced first on the JVM: the byte-match implementation **accepted** an injected chain `[attacker_leaf, pinned_node_cert]`. Both verifiers now treat pins as the only trust anchors — Android loads the DERs into an in-memory KeyStore and validates with the PKIX TrustManagerFactory; iOS uses SecTrustSetAnchorCertificates + SetAnchorCertificatesOnly(true) + SecTrustEvaluateWithError. JVM results on the new code: self-signed leaf pin accepted (incl. non-CA self-signed), CA-signed leaf against pinned CA accepted, injected chain rejected (`SunCertPathBuilderException`), wrong pin rejected.

**High — pins dropped on 3 import paths:** threaded through `handleAnything`'s clnrest branch and both `importClipboard` branches (lndconnect + clnrest); propagation tests added in `handleAnything.test.ts` (the WalletConfiguration clipboard paths have no view-level test harness in this repo — manually inspected).

**High — missing-key upgrade breakage:** the `cert-verification-default-v1` MOD_KEY migration normalizes nodes saved without the key to an explicit `false`, so the strict default only governs new configs and upgrades don't break self-signed remotes. Runs on both load paths (modern getSettings branch + legacySettingsMigrations), idempotent, flag-after-write, real object into setSettings; 4 tests added. This makes the PR body claim true rather than aspirational — the deliberate migration of the stored trust-all cohort remains the signed-off follow-up.

**Medium:** iOS pin branch now gated on `NSURLAuthenticationMethodServerTrust` (other challenge types fall through to default handling). clnrest client-cert pin kept but documented as an inert anchor (JS-side Basic Constraints parsing isn't worth the fragility).

**Nits:** patch script is now all-or-nothing per file and throws on any anchor miss (fails postinstall loudly); `SSLContext.getInstance("TLS")`; MITM modal suppressed when pins authenticate the config; skill docs reference PR #4366 instead of the branch name.

Still open, as flagged in the PR body: on-device testing on both platforms (Rule 4) — the iOS native code specifically has never been compiled — and sign-off on the included migration.

kaloudis:
Follow-up addressed in `efd99c0cd` (pushed):

**iOS hostname policy asymmetry:** correct — the serverTrust of a server-trust challenge is built with an SSL policy, and `SecTrustEvaluateWithError` enforces its hostname check even with custom anchors. The pin branch now swaps the policy to `SecPolicyCreateBasicX509()` before evaluation, matching Android's permissive HostnameVerifier: anchors authenticate the endpoint, hostname/SAN mismatches no longer fail pinned connections on either platform. (Ownership: the NSArray literal retains the policy; the Create-rule reference is CFReleased.) Patch re-verified against pristine 0.24.10 sources; `yarn verify` green.

**Validity-window note:** added to the PR body alongside the rotation note — an expired node cert now fails a pinned connection until the node regenerates it (lnd's autogenerated certs live ~14 months), and a leaf rotation likewise requires re-import to re-pin. Arguably correct behavior either way, now documented for node operators.

Merge gates unchanged and still acknowledged: on-device testing on both platforms (the iOS block — now including the policy swap — has only been eyeballed, never compiled), and sign-off on the `cert-verification-default-v1` migration.

views/Settings/WalletConfiguration.tsx
myxmaster: With `certVerification` now defaulting to `true`, the `CertInstallInstructions` button is always displayed on a new config, even for nodes nodes that don't support cert verification.
-> `supportsCertVerification` is needed here as well.

utils/ConnectionFormatUtils.ts
```diff
@@ -1,6 +1,44 @@
...
+        // default makes the failure loud at connect time
+        return undefined;
+    }
+    const decoded = Base64Utils.base64ToUtf8(b64);
+    if (decoded.includes('-----BEGIN CERTIFICATE-----')) {
+        return pemBundleToPinnedCerts(decoded);
+    }
+    // PEM bundles without a certificate block carry nothing pinnable;
+    // otherwise assume raw DER (lndconnect)
+    return isBase64Url ? [b64] : undefined;
```
myxmaster: If I read this right, a `cert=` value that isn't PEM gets pinned as-is, with no validation that it's actually a certificate. Previously the param was discarded entirely, so a malformed one didn't matter. Now it ends up stored as a pin on the node config. Does that surface as a connection failure later on, and if so, would the error make the cause apparent to the user?

What made me look was the test at line 54, which pins `'asfdaa=='`.

myxmaster:
`.claude/skills/zeus-diagnostics-and-tooling/SKILL.md` (below "How requests flow (verified in `backends/LND.ts` + `utils/TorUtils.ts`)") still says the patched blob-util enforces **byte-exact** cert pinning.
Same paragraph says a MITM proxy works "if you install the proxy CA in the device trust store": true for the unpinned path, but wrong for pinned connections, where anchor-only evaluation makes the device trust store irrelevant.

kaloudis:
Round addressed in `dfb6f88c7` (pushed). Point by point:

**CertInstallInstructions on unsupported backends:** confirmed. The button render was gated only on `certVerification` (now default true) and `!enableTor`, so LNC/NWC/embedded configs showed it. Now also gated on `supportsCertVerification`, matching the toggle above it.

**Non-PEM `cert=` pinned as-is:** good catch, and to answer the question directly: yes, it surfaced as a connection failure, and no, the error did not make the cause apparent. On Android a garbage pin threw a RuntimeException while building the pinned OkHttp client (`CertificateException` out of `generateCertificate`); on iOS `SecCertificateCreateWithData` returns NULL for garbage, the anchor array ends up empty, and the challenge is cancelled, which surfaces as a generic "cancelled" request error. Neither points at the imported cert param. Fixed at import time instead: `certsParamToPinnedCerts` now runs a minimal DER structural check (SEQUENCE tag, encoded length spanning the whole buffer) on raw lndconnect material and on every PEM-extracted block, and discards anything that fails. That restores the pre-PR behavior for malformed params: no pin is stored, and with strict verification on by default the failure stays loud through the normal cert-verification path, which at least has established UX (cert install instructions). The `'asfdaa=='` test you flagged now asserts no pin is stored; fixtures replaced with valid short-form and long-form (`0x82`) DER, plus new truncated-DER and non-DER-PEM-block rejection tests.

**Stale skill doc:** both claims fixed in `zeus-diagnostics-and-tooling/SKILL.md`. Pinning is now described as anchor-only trust evaluation (pins are the only trust anchors; the device trust store is never consulted), and the proxy-CA-in-device-trust-store interception route is scoped to unpinned configs, with an explicit note that pinned configs require removing the pins to intercept.

Merge gates unchanged: on-device testing on both platforms and sign-off on the `cert-verification-default-v1` migration.

ajaysehwal:
@kaloudis, Please rebase

utils/MigrationUtils.ts
```diff
@@ -736,6 +740,34 @@ class MigrationsUtils {
...
+        if (modCertVerification) return settings;
+
+        if (settings?.nodes && Array.isArray(settings.nodes)) {
+            for (const node of settings.nodes) {
+                if (node && node.certVerification === undefined) {
+                    node.certVerification = false;
+                }
+            }
+        }
+        await settingsStore.setSettings(settings);
```
ajaysehwal: @kaloudis, One inconsistency here: `migrateOlympusHostsToZeusLsp` and `migrateSwapHostsToBoltz` both track a `changed` flag and only call `setSettings` when something was actually rewritten, but `migrateCertVerificationDefault` calls it unconditionally.

The `"handles settings without nodes"` test confirms this, so we end up with an unnecessary Storage write when there are no nodes or all nodes already have an explicit `certVerification`.

Not harmful, but it might be worth using the same `changed` guard here for consistency and to avoid the redundant write.

kaloudis: Good catch, agreed on consistency. Fixed in `58b6598e6`: the migration now tracks a `changed` flag like the host migrations and only calls `setSettings` when a node was actually normalized. The MOD_KEY stamp is still written unconditionally, so the one-shot semantics are unchanged (a crash between the write and the stamp just reruns idempotently). The "handles settings without nodes" test now asserts no `setSettings` call, and the idempotency test asserts the repeat run does not persist a second time.

views/Settings/WalletConfiguration.tsx
```diff
@@ -3000,14 +3034,44 @@ export default class WalletConfiguration extends React.Component<
...
+                                            } else {
+                                                this.setState({
+                                                    certVerification:
+                                                        !certVerification,
+                                                    saved: false
+                                                });
+                                            }
+                                        }}
                                     />
+                                    {(pinnedCerts?.length ?? 0) > 0 && (
```
ajaysehwal: @kaloudis, One usability gap: once `pinnedCerts` is set, there’s no way to clear it from the UI if the node’s certificate rotates. The connection then fails closed, with no obvious recovery besides re-importing a fresh connection string or recreating the wallet.

Worth adding a “Remove pinned certificate” affordance, perhaps next to the “Node TLS certificate pinned…” text at `WalletConfiguration.tsx:3061`, or at least a clearer error pointing the user toward re-importing the connection string when a pinned connection fails.

Not a security issue — just worth addressing before this turns into “can’t connect after renewing my node’s cert” support tickets.

views/Settings/WalletConfiguration.tsx
```diff
@@ -3000,14 +3034,44 @@ export default class WalletConfiguration extends React.Component<
...
                                         value={certVerification}
-                                        onValueChange={() =>
-                                            this.setState({
-                                                certVerification:
-                                                    !certVerification,
-                                                saved: false
-                                            })
-                                        }
+                                        onValueChange={() => {
+                                            if (
```
ajaysehwal: @kaloudis, small UX gap here: once a certificate is pinned, toggling “Certificate Verification” off changes the stored value but has no effect on the actual connection — the pin still enforces verification via `trusty: !certVerification && !hasPinnedCerts` in `LND.ts` / `CLNRest.ts`.

That’s the correct security behavior, but the UI makes it look like the toggle is doing something. Might be worth disabling/greying out the switch whenever `pinnedCerts.length > 0`, or otherwise indicating that certificate verification is enforced by the pin.

views/Settings/WalletConfiguration.tsx
kaloudis: Agreed. Cert rotation would have failed closed with nothing but a generic connect error, and lnd regenerates tls.cert on expiry, so this would have hit real users eventually. Added a Remove Pinned Certificate button below the pinned-cert note in 4d049ffb2. Removal resets certVerification to true, so dropping the pin falls back to CA verification rather than any stored trust-all opt-out; turning verification off from there still goes through the warning modal. A pin-specific error message on connect failure would need error mapping in the native patch, so I'd rather track that with the other follow-ups (WebSocket pinning, clnrest client certs) than grow this PR.

views/Settings/WalletConfiguration.tsx
kaloudis: Good catch, and it was slightly worse than a cosmetic no-op: a stored false was latent, and would have become an actual trust-all the moment the pins went away. Fixed in 4d049ffb2: while a pin is present the switch is disabled and displayed as on (the effective behavior), and the pinned-cert note now states that verification is enforced by the pin. This composes with the removal button from your other comment: removing the pin re-enables the switch at true, and turning it off from there goes through the normal warning modal.

PR #4368, opened 2026-08-11
Block payment of expired lightning invoices

kaloudis:
# Description

Simulator Screenshot - iPhone 17 Pro - 2026-08-11 at 21 47 12

`Invoice.isExpired` is computed by the model but was never consulted anywhere on the user-initiated send path. An expired BOLT11 invoice loaded into `pay_req` with no error, rendered on the Payment Request screen with no warning (the screen even displays the original expiry duration), and dispatched on swipe or button press. BOLT 11 places expiry enforcement on the payer ("a reader... SHOULD NOT... pay an expired invoice"), so the outcome was delegated entirely to backend and recipient behavior: lnd and CLN payer-side checks turn this into a late, ugly failure after authorization, while backends that delegate the decision outward (LndHub, NWC client) can result in an expired invoice actually being paid if the recipient still settles it (stale invoice replay). The NWC service role already performs this exact check before paying (`NostrWalletConnectStore`), so this brings the main send path in line with it.

Changes:

- `views/PaymentRequest.tsx`: shows an expired error notice, hides the pay/swipe controls when the invoice is expired, and fail-closes in `triggerPayment` so an invoice that lapses while the user sits on the review screen is also caught at dispatch time
- `stores/InvoicesStore.ts`: `getPayReq` now threads the original bolt11 string into the `Invoice` model. This is load-bearing: the decode responses of LND (REST), embedded LND, Lightning Node Connect, and CLNRest omit the bolt11 string, which made `isExpired` silently return `false` on those backends. Without this, the view gate would be a no-op on exactly the most common backends
- `models/Invoice.ts`: adds a plain `isExpiredNow()` method that the `isExpired` computed delegates to. `Date.now()` is not observable, so a MobX computed caches its value while observed and could report a stale "not expired" when re-checked at dispatch time; the plain method always evaluates fresh
- Regression tests for the model (including the decode-response-without-bolt11 case on both sides of the threading fix, and freshness under observation with fake timers) and for `getPayReq` threading

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -810,6 +810,7 @@
     "views.Payment.writeNote": "Write your note here",
     "views.PaymentRequest.title": "Lightning Invoice",
     "views.PaymentRequest.error": "Error loading invoice",
+    "views.PaymentRequest.invoiceExpired": "This invoice has expired and can no longer be paid",
     "views.PaymentRequest.customAmt": "Custom Amount",
     "views.PaymentRequest.payDefault": "Pay default amount",
     "views.PaymentRequest.payCustom": "Pay custom amount",
--- a/models/Invoice.test.ts
+++ b/models/Invoice.test.ts
@@ -1,5 +1,7 @@
 jest.mock('../stores/Stores', () => ({}));
 
+import { autorun } from 'mobx';
+
 import Invoice from './Invoice';
 
 // decodes with timestamp 1700074718 and expiry 3600 (see Bolt11Utils.test.ts)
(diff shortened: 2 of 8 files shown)


shubhamkmr04 (commented):
Image

swapbutton is showing on expired invoices

shubhamkmr04:
we should wire this up for paying expired invoices via ecash as well

kaloudis:
Thanks for the catches @shubhamkmr04. Both addressed in 9b62324e5:

**Swap button**: now hidden in the PaymentRequest header when the invoice is expired, same condition as the pay controls.

**Ecash path**: CashuPaymentRequest gets the same treatment as PaymentRequest: expired notice banner, pay controls hidden, and a fail-closed recheck in `triggerPayment` in case the invoice lapses while the screen is open. Since MultimintPayment is only reachable through `triggerPayment` and auto-executes on mount, that gate covers the multimint branch as well. `CashuStore.payReq` is built from `Bolt11Utils.decode`, which already carries the bolt11 string, so `isExpiredNow()` works there without the threading fix the LN side needed.

Also took the opportunity to bail out of `CashuStore.getPayReq` before melt quote prep when the decoded invoice is already expired, so an unpayable invoice is never disclosed to the mint(s) in the first place.

shubhamkmr04 (approved):
tACK

PR #4370, opened 2026-08-11
Cashu: restore freshest Nostr mint backup, gate on stored timestamp floor

kaloudis:
# Description

## Problem

The Cashu mint list is backed up to Nostr relays as a NIP-44 encrypted kind 30078 replaceable event and restored on fresh installs when the local mint list is empty. The restore path had no freshness validation:

- `restoreMintsFromNostr` iterated the 8 default relays serially and returned the first decryptable result. A relay serving a stale event (an honest relay that missed the latest publish, or a malicious relay replaying an old but validly signed backup) could roll the mint list back. On a fresh install the wallet silently adopts the restored list, adds each mint to CDK, and runs proof restore against it, so a mint the user deliberately removed gets re-added and contacted.
- The persisted `nostrMintBackupTimestamp` was loaded and displayed but never consulted during restore.
- Removing the last mint never updated the backup at all (`nostrBackupMints` early-returned on an empty list), so the stale non-empty list stayed latest on every relay forever. In that case the rollback needed no misbehaving relay at all.
- The fetched event's signature and author were never checked, and the decrypted payload shape was not validated before flowing into `CashuDevKit.addMint`.

## Fix

- `restoreMintsFromNostr` now queries all relays in parallel and keeps the backup with the highest encrypted payload `timestamp`. The inner timestamp is authenticated by the NIP-44 AEAD; the outer `created_at` is not, so it is ignored. Payload format is unchanged (`{mints, timestamp}`), so cashu.me cross-compatibility is preserved.
- Restores are gated on the persisted `nostrMintBackupTimestamp` as a floor (equal passes, older is dropped) and the floor ratchets forward on every accepted restore. A replayed older backup can never be restored once a newer one has been seen.
- Removing the last mint now publishes an empty backup so the removal propagates to relays. A newer empty backup beats an older non-empty one; it advances the floor and restores nothing.
- `fetchFromRelay` verifies `event.pubkey` and the event signature and validates the decrypted payload shape (mints is a string array, timestamp is a finite number) before accepting.
- The fire-and-forget backup calls in `addMint`/`removeMint` now catch failures instead of leaving unhandled rejections. The manual Mint sync view still awaits and surfaces errors as before.
- Parallel querying also cuts the worst-case fresh-install restore from 80s (8 relays x 10s serial) to 10s.

## Behavior notes

- A device whose relays only retain events older than its stored floor now gets "No mint backup found" instead of silently restoring stale data. This is intended.
- The "Last backup" timestamp in Tools > Nostr keys > Mint sync now also updates when a restore accepts a backup, even if the user cancels adding the mints: it reflects the latest authenticated backup observed.
- Out of scope, possible follow-up: capping accepted timestamps at now plus slack, so a peer device with a fast clock cannot ratchet the floor into the future.

## Tests

`utils/NostrMintBackup.test.ts` gains a scriptable fake relay harness (partial mock of `relayInit`, real key derivation, NIP-44 encryption, and event signing) and 10 new cases: freshest-across-relays regression (fails on master, returns the stale relay's list), newer-empty-beats-older-non-empty, minTimestamp floor and equal-timestamp acceptance, wrong-pubkey rejection, tampered-signature rejection, malformed payload rejection, hung-relay timeout without losing other relays, all-relays-fail, and an empty-list backup round-trip.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/NostrMintBackup.test.ts
+++ b/utils/NostrMintBackup.test.ts
@@ -1,10 +1,120 @@
 import { sha256 } from '@noble/hashes/sha256';
-import { bytesToHex } from '@noble/hashes/utils';
-import { getPublicKey, nip19 } from 'nostr-tools';
+import { bytesToHex, hexToBytes } from '@noble/hashes/utils';
+import { schnorr } from '@noble/curves/secp256k1.js';
+import { getEventHash, getPublicKey, nip19, nip44 } from 'nostr-tools';
 import * as bip39 from '@scure/bip39';
 
-import { deriveMintBackupKeypair } from './NostrMintBackup';
+import {
+    backupMintsToNostr,
+    deriveMintBackupKeypair,
+    restoreMintsFromNostr
+} from './NostrMintBackup';
 import Base64Utils from './Base64Utils';
+
+let mockRelayFactory: ((url: string) => any) | null = null;
+
+jest.mock('nostr-tools', () => {
+    const actual = jest.requireActual('nostr-tools');
+    class MockRelay {
+        constructor(url: string) {
+            return mockRelayFactory
+                ? mockRelayFactory(url)
+                : new actual.Relay(url);
+        }
+        static async connect(url: string) {
+            const relay = mockRelayFactory
+                ? mockRelayFactory(url)
+                : new actual.Relay(url);
+            await relay.connect();
+            return relay;
+        }
+    }
+    return { ...actual, Relay: MockRelay };
+});
+
+interface RelayScript {
+    events?: any[];
+    neverRespond?: boolean;
+    failConnect?: boolean;
+}
+
+function installFakeRelays(scripts: Record<string, RelayScript>) {
+    const relays: Record<string, any> = {};
+    mockRelayFactory = (url: string) => {
+        const script = scripts[url] ?? {};
+        const relay = {
+            url,
+            connect: () =>
+                script.failConnect
+                    ? Promise.reject(new Error('connect failed'))
+                    : Promise.resolve(),
+            subscribe: (_filters: any, params: any) => {
+                const sub = { close: jest.fn() };
+                if (!script.neverRespond) {
+                    // Fires after the caller has taken the subscription
+                    // handle, under both real and fake timers
+                    queueMicrotask(() => {
+                        for (const ev of script.events ?? []) {
+                            params.onevent(ev);
+                        }
+                        params.oneose();
+                    });
+                }
+                return sub;
+            },
+            publish: jest.fn(() => Promise.resolve()),
+            close: jest.fn()
+        };
+        relays[url] = relay;
+        return relay;
+    };
+    return relays;
+}
+
+function makeBackupEvent(
+    signerPrivHex: string,
+    conversationKey: Uint8Array,
+    mints: string[],
+    timestamp: number,
+    opts: {
+        tamperSig?: boolean;
+        rawPayload?: string;
+        plainContent?: string;
+    } = {}
+) {
+    const plaintext =
+        opts.rawPayload !== undefined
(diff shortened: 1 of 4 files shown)


ajaysehwal:
@kaloudis , Please rebase

ajaysehwal (commented):
@kaloudis, one pre-existing gap I noticed: `deleteCashuData` still doesn't publish an empty Nostr backup.

This means that after deleting Cashu data and reinstalling, the wallet could still restore stale data from the relays. The comment in `CashuStore` already acknowledges this, so it's not introduced by this PR, but it would be good to address separately.

utils/NostrMintBackup.test.ts
```diff
@@ -1,11 +1,121 @@
...
+            publish: jest.fn(() => Promise.resolve()),
+            close: jest.fn()
+        };
+        relays[url] = relay;
+        return relay;
+    };
+    return relays;
+}
+
+function makeBackupEvent(
```
ajaysehwal, line 74: nit: there’s no test covering a case where the event's `created_at` differs from the payload `timestamp`.

Since the comment explicitly documents that the outer `created_at` should be ignored, adding a test for this would help lock in that behavior and prevent future regressions.

kaloudis:
Good catch @ajaysehwal. Since it's the same resurrection vector this PR closes on the restore side, I addressed it here rather than in a follow-up: 933cf0a84 publishes an empty backup in `deleteCashuData` before the seed (which derives the backup keypair) is removed.

It calls `backupMintsToNostr` directly rather than `nostrBackupMints` so nothing is persisted mid-delete, and it's raced against a 10s timer so unreachable relays can't stall deletion; failures don't abort the rest of the cleanup. Deleting while offline remains a residual gap (nothing can be published), but that's no worse than before.

utils/NostrMintBackup.test.ts
kaloudis: Added in 86d2243b2: `makeBackupEvent` now takes a `createdAt` override, with two tests where it disagrees with the payload timestamp, covering both the freshness ranking and the `minTimestamp` floor.

ajaysehwal (commented):
@kaloudis, not sure if this is related to this PR, but I found an issue with the **“Delete All Cashu Data”** button. I think we should show a loading state or disable the button while the Cashu data is being deleted to prevent multiple deletion requests.

https://github.com/user-attachments/assets/d611970e-0b99-4c72-bfdf-ddac193903b7

kaloudis:
@ajaysehwal good catch. The missing disabled state predates this PR (the button never consumed the store's loading flag), but this PR makes it much more visible: the empty-backup publish added in 933cf0a84 can hold deletion for up to 10 seconds against unreachable relays, where deletion used to be near-instant. So I've addressed it here in d89f2dbe9: the button now disables while `CashuStore.loading` is set, and `deleteCashuData` bails out if a run is already in flight.

For the record, the double-run itself was benign: every step in a concurrent second pass is idempotent or individually caught (duplicate `removeMint` failures are warned and swallowed, storage removals are idempotent, and a duplicate empty-backup publish is harmless). Worst case was stacked success alerts, so this was UX-only.

ajaysehwal (approved):
tACK

PR #4371, opened 2026-08-11
fix: never touch the network during a clipboard probe

kaloudis:
# Description

The clipboard probe (`isClipboardValue`) runs the full `handleAnything` router on every `WalletHeader` mount, every wallet-screen `focus`, and every `AppState -> active` transition (`components/WalletHeader.tsx`, `views/Send.tsx`) to decide whether to show a paste badge. It is meant to classify the clipboard string **locally**, but two input classes fell through to a network sink before the clipboard early-returns were consulted:

1. **LNURL (bech32 or `lnurl*` URL):** the branch calls `getlnurlParams(raw)`, which performs a live HTTP GET of the decoded URL, and the `isClipboardValue` guards only ran *inside* the resulting `.then()` callback, so the fetch always fired first. `js-lnurl` imposes no scheme/egress policy on bech32-decoded payloads, so a crafted `lnurl1...` can name any `http://` target, including RFC1918 (`192.168.x.x`) and link-local (`169.254.169.254`) literals.
2. **Nostr `npub`:** the branch ran `nostrProfileLookup`, opening connections to every default relay, with no clipboard guard at all.

### Impact

Any app that can write the clipboard, or a page that gets the user to copy a string, could:

- covertly beacon wallet foreground events, source IP, and timing on every app foreground while the poisoned string sits on the clipboard (Tor is bypassed on these paths);
- perform blind GETs into the local network (SSRF) from the wallet process.

No user interaction beyond opening/foregrounding the wallet is required, and `privacy.clipboard` defaults to on.

### Fix

Both branches already recognize the value locally before hitting the sink, so return early in probe mode:

- LNURL branch: `if (isClipboardValue) return true;` before `getlnurlParams`.
- npub branch: `if (isClipboardValue) return true;` before `nip19.decode` / `nostrProfileLookup`.

The actual fetch/relay lookup is deferred to when the user taps the badge or otherwise acts on the value (non-probe call), where behavior is unchanged. All other input classes (bitcoin address, BOLT11, lightning address, BTCPay, etc.) already early-returned before their network sinks and are unaffected.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Adds a `clipboard probe must not touch the network` regression suite to `utils/handleAnything.test.ts`:

- bech32 LNURL clipboard value -> `true`, `getParams` not called
- `lnurl*` URL clipboard value -> `true`, `getParams` not called
- control: non-probe call still fetches
- npub clipboard value -> `true`, no `nip19.decode` / relay init

Verified the suite fails without the source guards (3 failing) and passes with them. Full `handleAnything.test.ts` (117 tests) green; prettier, eslint, and tsc clean on the changed files.

- [ ] Android
- [ ] iOS

Device testing pending: foreground with a poisoned LNURL/npub on the clipboard; confirm no outbound request until the badge is tapped.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/handleAnything.ts
+++ b/utils/handleAnything.ts
@@ -787,6 +787,12 @@ const handleAnything = async (
         !!lnurl ||
         (value && /\/lnurl|lnurlp|lnurlw|lnurlc|lnurlauth/i.test(value))
     ) {
+        // The input is already recognized as LNURL-shaped locally (bech32
+        // decode, findlnurl, or an lnurl* URL path). That is enough to show
+        // the paste badge, so never fetch during a clipboard probe: defer
+        // getlnurlParams (a live HTTP GET of an attacker-controllable target)
+        // to when the user actually acts on the value.
+        if (isClipboardValue) return true;
         const raw: string = findlnurl(value) || lnurl || value || '';
         return getlnurlParams(raw)
             .then((params: any) => {
@@ -886,6 +892,10 @@ const handleAnything = async (
                 );
             });
     } else if (AddressUtils.isValidNpub(data)) {
+        // A valid npub is fully classified locally; do not open Nostr relay
+        // connections (nostrProfileLookup) during a clipboard probe. Defer the
+        // lookup to when the user acts on the value.
+        if (isClipboardValue) return true;
         try {
             const decoded = nip19.decode(data);
             const pubkey = decoded.data.toString();
--- a/utils/handleAnything.ts
+++ b/utils/handleAnything.ts
@@ -806,7 +806,6 @@ const handleAnything = async (
 
                 switch (params.tag) {
                     case 'withdrawRequest':
-                        if (isClipboardValue) return true;
                         if (ecash) {
                             return [
                                 'ChoosePaymentMethod',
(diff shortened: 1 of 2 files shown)


utils/handleAnything.ts
myxmaster: this is dead code now (same goes for the `isClipboardValue` guard in cases `payRequest`, `channelRequest`, `login` [both branches], `default`)

kaloudis: Good catch, thanks. Removed all six unreachable guards in 8dc670ae6 (withdrawRequest, payRequest, channelRequest, both login branches, and default). Since the probe now returns before `getlnurlParams` is called, nothing inside that `.then()` switch can ever run with `isClipboardValue` set.

myxmaster:
tACK

myxmaster (approved):

PR #4372, opened 2026-08-11
Route ancillary HTTP calls over Tor and warn on Tor-bypassing WebSocket RPCs

### LINKED ISSUE #2035: Follow tor callback address from clearnet lightning address
### Describe the feature

A follow up to: https://github.com/andrerfneves/lightning-address/issues/115

I would love to be able to point Zeus to a lightning address that has a clearnet domain, but points to a tor address via the callback.
While this works while using orbot, it would be great to remove this requirement by leveraging the built in tor.

kaloudis:
# Description

Relates to issue: [**ZEUS-2035**](https://github.com/ZeusLN/zeus/issues/2035)

With per-node **Use Tor** enabled, ZEUS only routed the LND/CLNRest **REST** path through the Tor daemon. Two classes of traffic bypassed it, leaking the user's IP (and, for several calls, the node `identity_pubkey`) to clearnet:

1. **LND WebSocket RPCs** (`wsReq`, `subscribeCustomMessages`, `openChannelStream`, `initChanAcceptor`) use the platform `WebSocket` API, which has no SOCKS wiring. They connect directly and cannot reach `.onion` nodes. This is **LND/LndHub-remote only**; `EmbeddedLND` and `LightningNodeConnect` override these methods natively/over gRPC, and CLNRest does not implement them.
2. **Every ancillary service call** (ZEUS Pay lightning-address auth, channel backups, LSP Flow/LSPS1, fiat rates, on-chain fees, block-tip sync, swaps, WIF sweeps) went direct with no `enableTor` branch. Several send the node `identity_pubkey` in the request body, directly linking IP to node identity.

Routing the platform `WebSocket` over Tor is not feasible without a SOCKS primitive in the Tor library, so that class is addressed with an **enable-time warning**. The ancillary REST calls are now routed through the existing Tor daemon.

## Changes

- **`utils/TorUtils.ts`**: add `doTorRequestRaw` (returns `{status, body}`, does not throw on non-2xx). `doTorRequest` is refactored to build on it with no behavior change for existing callers.
- **`utils/NetworkUtils.ts`** (new): `networkFetch({method, url, headers, body, enableTor})` returns a ReactNativeBlobUtil-compatible response (`.info().status` / `.json()` / `.text()` / `.data`), so call sites keep their existing status/body handling regardless of transport.
- Route **35 ancillary REST sites** through `networkFetch`, gated on `settingsStore.enableTor`: `LightningAddressStore` (12), `ChannelBackupStore` (6), `LSPStore` (6), `SwapStore` (7), `FiatStore` (2), `FeeStore` (1), `SyncStore` (1).
- **`SweepStore`**: add a `settingsStore` constructor arg (wired in `Stores.ts`) and migrate its four esplora fetch sites. **`ConnectivityStore` is intentionally left on clearnet** because Tor defeats its reachability/captive-portal probes.
- **`WalletConfiguration.tsx`**: warn when Tor is enabled on a remote **LND** node that streaming WebSocket RPCs (batch channel opens, LSP flows) bypass Tor and are unavailable for `.onion` nodes. One new `en.json` key.
- **`utils/NetworkUtils.test.ts`**: covers transport selection and the Tor shim's response shape.
- **`TransactionsStore`** (review follow-up): the mempool.space broadcast fallback (taken on backends without `publishTransaction`, e.g. CLNRest) and `broadcastRawTxToMempoolSpace` now go through `networkFetch`, so a signed transaction is no longer published from the user's real IP after Tor-routed reads.
- **`utils/ChannelMigrationUtils.ts`** (review follow-up): its eight `BACKUPS_HOST` calls now use `networkFetch`, matching `ChannelBackupStore` on transport.

## Known limitations (follow-ups)

- `LightningAddressStore`'s `socket.io` connection still emits the `pubkey`; a fetch swap cannot cover a websocket (needs socket.io SOCKS proxying).
- Nodes that reach `enableTor` via `.onion` auto-derivation (rather than the toggle) do not pass through the enable-time warning.
- The `ChannelBackupStore`, `SyncStore`, and `ChannelMigrationUtils` gates are staged but currently unreachable: those flows are embedded-LND only, and embedded LND does not get the per-node Tor toggle, so `settingsStore.enableTor` is never set for it. #4522 wires its `embeddedTor` setting into `SettingsStore.enableTor`, which activates these gates for embedded wallets once both PRs are merged.

This also subsumes the transport half of a separate BTCPay pairing-QR finding (`fetchBTCPayConfig` routes `.onion` via Tor but clearnet direct). That report's two non-transport concerns are **not** addressed here and are worth tracking separately: the config URL is fetched at scan time before confirmation, and `parseBTCPayConfig` silently prefers `adminMacaroon` over `macaroon`.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

`yarn verify` green: 1293 tests, TypeScript, Prettier, ESLint.

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

New `utils/NetworkUtils.test.ts` covers transport selection and the Tor shim's response shape.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Device testing pending, priority order: remote LND with Tor enabled (ancillary calls over Tor, WebSocket warning on enable), remote LND `.onion`, LndHub, CLNRest, plus the ancillary services themselves (ZEUS Pay, channel backups, LSP flows, swaps, fiat rates, fees, sync, sweeps) over Tor.

### Locales

### Other:

### DIFF
--- a/stores/ChannelBackupStore.ts
+++ b/stores/ChannelBackupStore.ts
@@ -1,5 +1,4 @@
 import { action, observable, runInAction } from 'mobx';
-import ReactNativeBlobUtil from 'react-native-blob-util';
 import * as CryptoJS from 'crypto-js';
 
 import NodeInfoStore from './NodeInfoStore';
@@ -16,6 +15,7 @@ import BackendUtils from '../utils/BackendUtils';
 import { LndMobileEventEmitter } from '../utils/LndMobileUtils';
 import Base64Utils from '../utils/Base64Utils';
 import { errorToUserFriendly } from '../utils/ErrorUtils';
+import { networkFetch } from '../utils/NetworkUtils';
 
 import Storage from '../storage';
 
@@ -69,14 +69,15 @@ export default class ChannelBackupStore {
         ).toString();
 
         try {
-            const authResponse = await ReactNativeBlobUtil.fetch(
-                'POST',
-                `${BACKUPS_HOST}/api/auth`,
-                { 'Content-Type': 'application/json' },
-                JSON.stringify({
+            const authResponse = await networkFetch({
+                method: 'POST',
+                url: `${BACKUPS_HOST}/api/auth`,
+                headers: { 'Content-Type': 'application/json' },
+                body: JSON.stringify({
                     pubkey: this.nodeInfoStore.nodeInfo.identity_pubkey
-                })
-            );
+                }),
+                enableTor: this.settingsStore.enableTor
+            });
 
             if (authResponse.info().status !== 200) {
                 this.logBackupStatus('ERROR');
@@ -90,17 +91,18 @@ export default class ChannelBackupStore {
             );
             const signature =
                 messageSignData.zbase || messageSignData.signature;
-            const backupResponse = await ReactNativeBlobUtil.fetch(
-                'POST',
-                `${BACKUPS_HOST}/api/backup`,
-                { 'Content-Type': 'application/json' },
-                JSON.stringify({
+            const backupResponse = await networkFetch({
+                method: 'POST',
+                url: `${BACKUPS_HOST}/api/backup`,
+                headers: { 'Content-Type': 'application/json' },
+                body: JSON.stringify({
                     pubkey: this.nodeInfoStore.nodeInfo.identity_pubkey,
                     message: verification,
                     signature,
                     backup: encryptedBackup
-                })
-            );
+                }),
+                enableTor: this.settingsStore.enableTor
+            });
             if (
                 backupResponse.info().status === 200 &&
                 backupResponse.json().success
@@ -119,14 +121,15 @@ export default class ChannelBackupStore {
     };
 
     public recoverStaticChannelBackup = async () => {
-        const authResponse = await ReactNativeBlobUtil.fetch(
-            'POST',
-            `${BACKUPS_HOST}/api/auth`,
-            { 'Content-Type': 'application/json' },
-            JSON.stringify({
+        const authResponse = await networkFetch({
+            method: 'POST',
+            url: `${BACKUPS_HOST}/api/auth`,
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({
                 pubkey: this.nodeInfoStore.nodeInfo.identity_pubkey
-            })
-        );
+            }),
+            enableTor: this.settingsStore.enableTor
+        });
         const authResponseStatus = authResponse.info().status;
         if (authResponseStatus !== 200) {
             throw new Error(`Auth response status code: ${authResponseStatus}`);
@@ -135,16 +138,17 @@ export default class ChannelBackupStore {
 
         const messageSignData = await BackendUtils.signMessage(verification);
         const signature = messageSignData.zbase || messageSignData.signature;
-        const recoverResponse = await ReactNativeBlobUtil.fetch(
-            'POST',
-            `${BACKUPS_HOST}/api/recover`,
-            { 'Content-Type': 'application/json' },
-            JSON.stringify({
+        const recoverResponse = await networkFetch({
+            method: 'POST',
+            url: `${BACKUPS_HOST}/api/recover`,
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({
                 pubkey: this.nodeInfoStore.nodeInfo.identity_pubkey,
                 message: verification,
                 signature
-            })
-        );
+            }),
+            enableTor: this.settingsStore.enableTor
+        });
         const data = recoverResponse.json();
         const { backup, created_at, success } = data;
         if (recoverResponse.info().status === 200 && success && backup) {
@@ -190,14 +194,15 @@ export default class ChannelBackupStore {
         this.error = false;
         this.backups = [];
         try {
-            const authResponse = await ReactNativeBlobUtil.fetch(
-                'POST',
-                `${BACKUPS_HOST}/api/auth`,
-                { 'Content-Type': 'application/json' },
-                JSON.stringify({
+            const authResponse = await networkFetch({
+                method: 'POST',
+                url: `${BACKUPS_HOST}/api/auth`,
+                headers: { 'Content-Type': 'application/json' },
+                body: JSON.stringify({
                     pubkey: this.nodeInfoStore.nodeInfo.identity_pubkey
-                })
-            );
+                }),
+                enableTor: this.settingsStore.enableTor
+            });
             const authResponseStatus = authResponse.info().status;
             if (authResponseStatus !== 200) {
                 throw new Error(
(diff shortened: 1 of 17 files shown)


shubhamkmr04 (commented):
Two gaps I think are worth closing before merge. One caveat on the anchors: the files that actually need the edits — `stores/TransactionsStore.ts` and `utils/ChannelMigrationUtils.ts` — aren't in this diff, so I anchored to the nearest touched code and described the target there.

stores/SweepStore.ts
```diff
@@ -211,12 +221,14 @@ export default class SweepStore {
...
-                        `${UrlUtils.getMempoolApiUrl(
+                    const res = await networkFetch({
+                        method: 'GET',
+                        url: `${UrlUtils.getMempoolApiUrl(
                             this.nodeInfoStore.nodeInfo
-                        )}/tx/${txid}`
-                    );
-                    if (!res.ok)
+                        )}/tx/${txid}`,
+                        enableTor: this.settingsStore.enableTor
```
shubhamkmr04, line 229: This read is protected now, but the flow it belongs to still ends on clearnet.

`prepareSweepInputs` (here) and `finalizeSweepTransaction` fetch UTXOs and prev-tx details over Tor. Then `views/Tools/WIFSweeper.tsx:365` hands `SweepStore.txHex` to `views/TxHex.tsx:122`, which calls `TransactionsStore.broadcast(txHex)` — a bare `ReactNativeBlobUtil.fetch('POST', .../tx, ...)` with no `enableTor` branch (`stores/TransactionsStore.ts` ~L187).

That fallback is taken whenever the backend has no `publishTransaction`, which is defined only on `LND`, `EmbeddedLND` and `LightningNodeConnect`. So CLNRest takes it — and CLNRest is inside `supportsTor`. Net effect: we hide the reads, then publish the signed transaction from the user's real IP, which is a tighter correlation than anything the reads exposed.

`TransactionsStore.broadcastRawTxToMempoolSpace` (~L875, reached from `views/RawTxHex.tsx`) is broader — no backend gate at all, so it leaks on every backend including remote LND with Tor on.

Both look like one-line `networkFetch` conversions. Happy to open a follow-up if you'd rather keep this PR scoped.

stores/ChannelBackupStore.ts
```diff
@@ -69,14 +69,15 @@ export default class ChannelBackupStore {
...
+            const authResponse = await networkFetch({
+                method: 'POST',
+                url: `${BACKUPS_HOST}/api/auth`,
+                headers: { 'Content-Type': 'application/json' },
+                body: JSON.stringify({
                     pubkey: this.nodeInfoStore.nodeInfo.identity_pubkey
-                })
-            );
+                }),
+                enableTor: this.settingsStore.enableTor
```
shubhamkmr04, line 79: I don't think this gate can ever be true — here or at the other five sites in this file.

`ChannelBackupStore` is embedded-LND-only: it imports `exportAllChannelBackups` / `restoreChannelBackups` from `lndmobile/channel` directly, and its views live under `views/Settings/EmbeddedNode/`. But `views/Settings/WalletConfiguration.tsx:1375` excludes that implementation from the Tor switch:

```ts
const isLocalImpl =
    implementation === 'embedded-lnd' || implementation === 'ldk-node';
const supportsTor =
    implementation !== 'lightning-node-connect' &&
    !isLocalImpl &&
    implementation !== 'nostr-wallet-connect';
```

So `node.enableTor` is never written for embedded LND, and `SettingsStore.enableTor` (assigned from it at L1890) stays `undefined`. `/api/auth`, `/api/backup`, `/api/recover` and `/api/recoverAdvanced` still go clearnet with `identity_pubkey` in the body.

Closing this is more a product call than a fix — the node's own traffic already has its own Tor config, so this would be a new ancillary-only control. Options as I see them: route `BACKUPS_HOST` unconditionally over Tor; give embedded LND an ancillary-only Tor setting; or leave these sites staged and note it in the PR body so the coverage claim stays accurate.

Same host and payload, not migrated: `utils/ChannelMigrationUtils.ts` POSTs `{pubkey}` to `BACKUPS_HOST` in eight places (`/api/auth` x4, `/api/status` x2, `/api/channels-backup`, `/api/restore-channels`). Worth folding into whichever option you pick so the two files don't diverge on transport.

stores/SyncStore.ts
```diff
@@ -116,20 +116,21 @@ export default class SyncStore {
...
                 // here.
-                `${UrlUtils.getMempoolApiUrl({
+                url: `${UrlUtils.getMempoolApiUrl({
                     isMutinynet: false,
                     isTestNet:
                         this.settingsStore.embeddedLndNetwork === 'Testnet'
-                })}/blocks/tip/height`
-            )
+                })}/blocks/tip/height`,
+                enableTor: this.settingsStore.enableTor
```
shubhamkmr04, line 132: Same dead gate as `ChannelBackupStore`, and the comment directly above is the tell: if SyncStore only runs for embedded LND, and embedded LND never gets the Tor switch, then `this.settingsStore.enableTor` is always `undefined` here.

stores/ChannelBackupStore.ts
kaloudis: We're gearing up to enable Tor on Embedded LND here https://github.com/ZeusLN/zeus/pull/4522

stores/SweepStore.ts
kaloudis: Good catch, and agreed that publishing the signed transaction from the real IP is the tighter correlation. Kept it in this PR rather than a follow-up: both the mempool.space fallback in `TransactionsStore.broadcast` and `broadcastRawTxToMempoolSpace` now go through `networkFetch` gated on `settingsStore.enableTor` (1bdb191a4). The store already had `settingsStore` injected and both sites already consumed `.info().status` / `.data`, so it was the one-line shape you predicted.

stores/ChannelBackupStore.ts
kaloudis: You're right that this gate is currently unreachable for the embedded-only sites. The resolution: keep the gates staged here and activate them through #4522, which adds an `embeddedTor` setting (EmbeddedNode > Advanced) routing the embedded node's own traffic through the in-app Tor proxy. As of kaloudis/zeus@b0e46a2a3 on that branch, `embeddedTor` also feeds `SettingsStore.enableTor` for the active embedded wallet, so every `networkFetch` gate in this PR goes live for embedded wallets once both merge, with a single Tor control instead of a new ancillary-only toggle. Also migrated the eight `ChannelMigrationUtils` `BACKUPS_HOST` sites in 1bdb191a4 so the two files stay on the same transport, and the staging is now noted in the PR body.

stores/SyncStore.ts
kaloudis: Same resolution as the ChannelBackupStore thread: staged until #4522 wires `embeddedTor` into `SettingsStore.enableTor` (kaloudis/zeus@b0e46a2a3), at which point this gate goes live for embedded wallets.

PR #4373, opened 2026-08-11
fix(security): save rescue key via system save dialog, purge legacy plaintext exports

kaloudis:
# Description

The swap rescue-key screen's download button writes the 12-word rescue mnemonic as plaintext JSON to a fixed, well-known path in shared storage: `Download/rescue_key.json` on Android and the Files-app-visible `Documents/rescue_key.json` on iOS (`UIFileSharingEnabled` is set). Files in the Android Downloads directory are readable by file managers and any app holding all-files access, get picked up by sync and backup tooling, and survive app uninstall. Nothing in the app ever deletes the file: the in-app "delete rescue key" flow only removes storage keys, and `clearAllData` misses it entirely, which also undercuts the duress wipe's "no recoverable key material" guarantee (#4327). The rescue key is the HD root for all swap refund/claim keys.

This is the second half of the audit finding; the first half (12-word LDK wallets reusing the wallet seed itself as the rescue key, which made this export a full wallet seed) was fixed in #4314.

**Export flow (`views/Settings/Seed.tsx`, `saveRescueKeyFile` in `utils/SwapUtils.ts`):**

* The download button is now a save button. The rescue file is staged in app-private cache (`CachesDirectoryPath`, both platforms) and handed to the system save dialog (`saveDocuments` from `@react-native-documents/picker`, already a dependency, same pattern as the wallet-config export in #4378), so the user explicitly picks the destination, including local device storage on Android (per review, the share sheet offered no save-to-device target there). Nothing is silently written to shared storage.
* Before the save dialog opens, a warning explains the file contains the rescue key in plaintext and persists wherever it is saved.
* `saveDocuments` copies the staged file to the destination before resolving, so the staging copy is always unlinked before the call returns (success, cancel, and error paths). The launch-time and wipe-path unlinks remain as belt-and-braces for crash-interrupted saves.
* A success alert confirms the export; a save failure shows an error alert; cancelling the dialog shows neither.
* The `{"mnemonic": "..."}` JSON format is unchanged, so previously exported files and files saved through the new flow both still import through the SeedRecovery upload path and remain compatible with Boltz-style rescue tooling.
* The old success alert (which pointed users to Files > Downloads) and the file-path `console.log` are removed.

**Retroactive cleanup (`purgeLegacyRescueKeyFiles` in `utils/SwapUtils.ts`):**

Best-effort delete of the legacy `rescue_key.json` from Downloads (Android) / Documents (iOS), invoked from three places:

1. A one-shot migration (`rescue-key-file-cleanup`, standard EncryptedStorage MOD_KEY pattern) that runs at the top of `SettingsStore.getSettings()`, before the settings blob is read, so it covers both legacy and modern settings paths. The flag is set regardless of unlink outcome: under Android scoped storage only the install that created the file can remove it, so a failed unlink can never succeed on retry.
2. The rescue-key delete flow in `Seed.tsx` (which previously only removed storage keys).
3. `clearAllData` in `utils/DataClearUtils.ts`, which the duress and lockout wipes route through, restoring the "no recoverable key material" claim for this artifact.

Unit tests cover the purge helper (no-op when absent, unlink when present, error swallowing) and the save flow (staging content and dialog options, staging unlink on success and cancel, dialog-reported write errors surfaced).

Note: opening the system save dialog backgrounds the app, so wallets with a PIN/password and lock-on-background may relock during the export. #4378 adds `BackgroundLockUtils` suppression for exactly this; once it lands, this flow will be wired to it in a follow-up rather than duplicating the helper here.

Heads up: this will produce a small, trivial merge conflict with #4328 in `DataClearUtils.ts` (whichever lands second rebases a one-line call site plus import).

No native rebuild required.

## Device test matrix (to be completed before undrafting)

| # | Scenario | Android | iOS |
|---|----------|---------|-----|
| 1 | Save flow: warning alert, system save dialog opens, file saves to local device storage, success alert shows, file readable at destination | [ ] | [ ] |
| 2 | No silent write: after saving, no `rescue_key.json` in Downloads (Android) / app Documents (iOS) beyond the user-chosen destination | [ ] | [ ] |
| 3 | Staging file gone from cache after save, after cancel (no alert), and after a save error (error alert) | [ ] | [ ] |
| 4 | Legacy cleanup migration: pre-planted `rescue_key.json` deleted on first launch, migration does not re-run | [ ] | [ ] |
| 5 | Rescue-key delete flow removes storage keys and file artifacts | [ ] | [ ] |
| 6 | Clear All Data / duress wipe removes legacy file | [ ] | [ ] |
| 7 | Restore round trip: file saved via the save dialog imports through SeedRecovery upload | [ ] | [ ] |

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


shubhamkmr04 (commented):
**The Android purge looks in the wrong directory, so the retroactive cleanup is a
no-op there.**

`purgeLegacyRescueKeyFiles` uses `blob-util`'s `fs.dirs.DownloadDir` = `getExternalFilesDir(DIRECTORY_DOWNLOADS)` → app-scoped `/sdcard/Android/data/app.zeusln.zeus/files/Download`. The old export wrote to `RNFS.DownloadDirectoryPath` → public `/sdcard/Download`.

Different paths, so `fs.exists` never matches and none of the three call sites deletes anything. Plant `rescue_key.json` in both and launch — only the app-scoped copy disappears.

`blob-util` has the right constant as `LegacyDownloadDir`, but using `RNFS.DownloadDirectoryPath` keeps reader and writer in sync. The tests miss itbecause they mock `fs.dirs` and only cover the iOS branch.

shubhamkmr04:
**On Android the staging file is deleted before the receiving app has read it.**

`Share.open` resolves from the chooser's `EXTRA_CHOSEN_COMPONENT` broadcast — when the user *picks* a target, not when it finishes reading. And RNShare doesn't copy the file: `ShareFile.getURI()` returns a `FileProvider` URI pointing straight at the cache file.

So the `finally` unlink kills the URI the consumer is still holding. Poll `adb shell run-as app.zeusln.zeus ls cache/rescue_key.json` and watch it vanish the moment you tap a target, with that app's dialog still open. Drive and Gmail are the likely losers.

`ChannelMigrationUtils` takes a different path on Android for the equivalent export — it copies to Downloads instead of using `Share.open`.

Deferring cleanup to the next launch would fix it; the pre-write unlink already prevents stale reuse.

shubhamkmr04:
**The cache staging file survives `clearAllData` and the duress wipe.**

`rescueKeyStagingPath` is local to `Seed.tsx`, and the purge only covers Downloads/Documents. Nothing in the wipe path touches a cache directory — `grep -ic cache utils/DataClearUtils.ts` returns 0.

Force-kill the app with the share sheet open and the `finally` never runs, leaving a plaintext 12-word mnemonic in `CachesDirectoryPath` that outlives both Clear All Data and the duress PIN wipe — the guarantee this PR cites #4327 to restore. The OS may reclaim cache eventually, but that's not a guarantee.

Moving the staging path and its unlink into `SwapUtils`, and purging both locations, closes this — and makes it safe to defer the post-share cleanup on Android.

kaloudis:
Thanks for the thorough review, all three confirmed against library source. Addressed in b8320ec9a:

**1. Android purge directory:** you were right, blob-util's `DownloadDir` is `getExternalFilesDir(DIRECTORY_DOWNLOADS)` while the legacy writer used `RNFS.DownloadDirectoryPath` (public `/sdcard/Download`). The helper now reads the same RNFS constants as the old writer on both platforms, and the tests cover the Android branch with distinct mock paths per directory so a regression here can't hide behind a shared value. Note for re-testing: if your device already ran the broken build, the one-shot `rescue-key-file-cleanup` flag is set, so you'll need to clear app data (or reinstall) for the migration to run again.

**2. Post-share unlink race:** confirmed from RNShare source, the promise resolves via the `EXTRA_CHOSEN_COMPONENT` broadcast and `compatUriFromFile` wraps the staging file directly with no copy. Went with your deferred-cleanup suggestion: the staging file now outlives `Share.open` on Android and is deleted on the next launch instead. It is still unlinked immediately on iOS (resolves after the share completes) and when the chooser is dismissed (no receiver ever got the URI). The pre-write unlink stays.

**3. Staging file vs wipes:** moved the staging path and unlink helper into `SwapUtils` as suggested. It's now cleaned up on every launch, in `clearAllData` (covers the duress wipe), and in the rescue key delete flow. One wrinkle with 'defer to next launch': the migration is one-shot by design, so the staging cleanup runs unconditionally before the one-shot gate in the renamed `purgeRescueKeyFiles`, since a new staging file can appear after any share.

`yarn verify` green locally. Will re-run the device matrix on both platforms before undrafting.

kaloudis:
Rebased on master; the round-1 commit is now 04a3b2c98 (was b8320ec9a, same content). The test CI failure was master's new `SettingsStore.test.ts` mocking `MigrationUtils` with a fixed method list that predates this PR's `getSettings` call; added `purgeRescueKeyFiles` to that mock. All four checks green.

shubhamkmr04 (changes requested):
@kaloudis can we fix this issue https://github.com/ZeusLN/zeus/pull/4378#pullrequestreview-4981072089 here also ?
also we can add a success msg for export success here as well

kaloudis:
Done in 547ee7db5, thanks @shubhamkmr04.

* The export now goes through the same system save dialog as #4378 (`saveDocuments` from `@react-native-documents/picker`), so on Android you can save the file directly to local device storage. Side benefit: `saveDocuments` copies the file to the destination before resolving, so the plaintext staging copy in app cache is now always deleted before the call returns instead of lingering until the next launch.
* Added a success alert on export, an error alert on failure, and nothing on cancel.

One known gap, same as #4378 found: the save dialog backgrounds the app, so wallets with lock-on-background may relock during the export. Rather than duplicating `BackgroundLockUtils` here while #4378 is still open, I'll wire this flow to it in a follow-up once #4378 lands.

shubhamkmr04 (commented):
Image

rescue key save is failing everytime in android.

kaloudis:
@shubhamkmr04 good catch - should be resolved in https://github.com/ZeusLN/zeus/pull/4373/commits/bf77736bbbe97bd81c498b64d6a7260f3e58ae40

shubhamkmr04 (approved):
tACK

PR #4374, opened 2026-08-11
fix(lnurl): bind paid invoice to shown metadata, amount, and safe callback

kaloudis:
# Description

## Problem

When ZEUS pays an LNURL-pay request (LUD-06) or lightning address (LUD-16), `views/LnurlPay/LnurlPay.tsx` renders the service-supplied `metadata` (payee name, text, image) on the confirmation screen but then pays an invoice whose commitment to that metadata is never checked. Specifically the flow:

1. never verifies the invoice `description_hash` equals `sha256(metadata)`, so the payee identity the user reviewed has no cryptographic relationship to the paid invoice (LUD-06 requires this check);
2. never compares the invoice amount against the amount the user requested; and
3. fetches the service-supplied `callback` verbatim: any scheme (`http://`), any host (cross-domain, RFC1918, link-local/cloud-metadata addresses).

A malicious or compromised LNURL service (or a MITM on an `http://` callback) can display benign identity metadata while the invoice the user pays commits to an unrelated description hash, invoice an arbitrary amount, leak the amount/comment in cleartext, or drive a wallet-side GET at internal infrastructure (`http://169.254.169.254/...`).

The donation path (`utils/DonationUtils.ts`, #4322) already performs exactly these LUD-06 checks because it pays silently; the interactive flow never got them.

## Fix

- New `utils/LnurlPayUtils.ts`:
  - `verifyLnurlPayInvoice(pr, metadata, expectedAmountMsat)` decodes the invoice **locally** (so the check is identical across all 7 backends, several of which omit `description_hash`/`num_msat` from their decode output) and requires both an exact amount match and `description_hash === sha256(metadata)`.
  - `isLnurlCallbackAllowed(callback)` rejects cleartext `http` to non-`.onion` hosts and any callback aimed at loopback / RFC1918 / link-local / CGNAT / `localhost` IP-literal targets.
- `views/LnurlPay/LnurlPay.tsx` runs the callback check before the fetch and the invoice-binding check after `pr` is returned, before navigating, gating **both** the Lightning (`PaymentRequest`) and ecash (`ChoosePaymentMethod`) routes. Failures show an alert and abort.
- `utils/DonationUtils.ts` refactored to reuse `verifyLnurlPayInvoice` (single source of truth; behavior unchanged, its stricter same-origin callback rule kept).

## Scope notes

- A cross-domain **https** callback is still allowed (some services front the callback on a separate CDN host). This blocks IP-literal SSRF targets only; a hostname that resolves to an internal address (DNS rebinding) is out of scope, as full protection needs connect-time resolution the RN fetch layer does not expose.
- Not addressed here (kept for focused follow-ups): client-side clamping of `comment` length and per-request `minSendable`/`maxSendable` UI enforcement.

Based on `ZeusLN/zeus:master`.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- `utils/LnurlPayUtils.test.ts` (new, 33 assertions with the donation suite): accepts a bound invoice; rejects description_hash mismatch, amount mismatch, missing metadata, and undecodable invoices; callback policy accepts https and http-onion, rejects clearnet http, 169.254.169.254, loopback, RFC1918, localhost, and IPv6 loopback.
- `yarn tsc`, `eslint`, and `prettier --check` clean on all changed files.

- [ ] Android
- [ ] iOS

Device testing (pending):

- [ ] Pay a normal lightning address (variable amount)
- [ ] Pay a fixed-amount LNURL (`minSendable === maxSendable`)
- [ ] Pay a variable-amount LNURL with a comment
- [ ] Pay a **ZEUS Pay / Zaplocker** address (the main compatibility risk: its invoices must commit `description_hash === sha256(metadata)`; if a legitimate ZEUS Pay invoice is rejected, that is a server-side non-compliance to fix rather than a reason to weaken the client)
- [ ] Pay via the ecash route from an LNURL

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


shubhamkmr04 (commented):
Image

not able to pay zeuspay addresses

shubhamkmr04 (commented):
Image

we are seeing the raw errors here

kaloudis:
Thanks @shubhamkmr04, both addressed. The branch is now rebased on upstream/master (9cf6d294a), so it picks up the cdk 0.17.4 error-parser changes; fixes are commits 5e4d3e2d7 and e2f3e688d.

**ZEUS Pay addresses blocked (5e4d3e2d7):** I pulled a live invoice from the evan@zeuspay.com callback and decoded it: the amount matches the request exactly, but the invoice carries no `description_hash` at all (just `description: "ZEUS Pay"`). That's structural for ecash-backed addresses: the invoice comes from the recipient's Cashu mint via a mint quote, so it can never commit to the LNURL metadata, and plenty of other real-world services omit the h tag too. The check now accepts an invoice with no `description_hash`, while still hard-failing when one is present and doesn't commit to the metadata shown, and the amount check remains unconditional. Verified zeuspay invoices now pass. The donation path shares the helper, so it gets the same behavior.

**Raw errors on the sending screen (e2f3e688d):** that blob is the mint's `/v1/info` document, which a CDK failure path embeds wholesale in the error string ("Unknown error response: `{...}`") and the screen displayed verbatim. Master's `cashuErrorForDisplay` (from #4424) already truncates these to 300 chars; this commit goes further inside `parseCashuDevKitError`, which it calls: when a parsed CDK error carries a sizeable JSON blob, prefer a readable `detail`/`error`/`message`/`reason` field from inside the body, otherwise drop the blob and keep the readable prefix ("Unknown error response"), falling back to a generic localized message when nothing readable remains. So instead of a truncated JSON dump the screen now shows an actual sentence. Covers the 0.17.x iOS/Android FFI shapes, the legacy shapes, and raw strings; tests added for each case.

Gates green on the rebased branch: jest, tsc, prettier, eslint (the only local failure is the known PhotoUtils path assertion that trips in any checkout not named `zeus`).

shubhamkmr04 (commented):
The error-cleanup added in e2f3e688d cuts off too much. `stripJsonBlob` looks for the first `{` or `[` anywhere in the message and throws away everything after it, so a normal error that just happens to contain a bracket loses most of its text — `"Melt failed [quote 7f3a] the mint could not complete the melt because the payment attempt timed out upstream"` becomes just `"Melt failed"`. It also runs inside `parseCashuDevKitError`, which isn't only used for display: `mapCDKError` reads that same string to decide the error type. So `Unknown error response: {"code":11000,"detail":"Quote amount not as requested",...}` gets reduced to `"Quote amount not as requested"`, the `11000` disappears, and the `MultiMintQuoteRejected` case stops matching.

Two things on the LNURL side. First, the new warning shows up for errors that have nothing to do with the invoice. The `data.status === 'ERROR'` check only catches responses in the LNURL error shape, and LNbits doesn't always use it — a deleted paylink returns `{"detail":"Pay link does not exist."}`, which slips through, leaves `pr` undefined, and the user is told "This invoice does not match the amount or recipient you approved" when the service simply isn't there any more. Second, the private-address check can be walked around — `url.parse` doesn't normalise the host, and `isPrivateIPv4` only recognises the plain `127.0.0.1` form, so `https://2130706433/`, `https://0177.0.0.1/`, `https://0x7f.0.0.1/`, `https://127.1/`, `https://127.0.0.1./` and `https://[0:0:0:0:0:0:0:1]/` all pass as allowed.

kaloudis:
Thanks @shubhamkmr04, all four points were valid; fixes are 5fe54e910, dc250f03a, and 0e2e5690f.

**stripJsonBlob over-stripping + mapCDKError (5fe54e910):** the second point was a regression this PR introduced: on master the JSON-blob shape reaches mapCDKError intact (extractMintDetail doesn't match it because of the quote before the colon), so stripping inside parseCashuDevKitError broke the 11000 / MultiMintQuoteRejected match. The stripping now lives only in cashuErrorForDisplay, so the parsed string used for classification is untouched. It also only engages when the bracketed tail actually looks like a JSON document, so "Melt failed [quote 7f3a] ..." style prose now passes through whole (the 300-char display cap still bounds oversized prose). Regression tests added for the 11000 shape surviving parse, the prose case, and detail extraction on display.

**Invoice-less service errors (dc250f03a):** a response without a `pr` is now surfaced as the service's own error (its reason/detail/error/message field when present, generic fallback otherwise) before the binding check runs, so a deleted LNbits paylink shows "Pay link does not exist." instead of the binding warning. The binding alert is reserved for the case where an invoice exists and fails verification.

**IP-literal bypass (0e2e5690f):** confirmed all the forms you listed were allowed. The check now canonicalizes before classifying: IPv4 literals are parsed the way inet_aton does (1 to 4 parts, decimal/octal/hex, last part fills the remaining bytes), IPv6 literals are expanded to their eight groups (:: compression, embedded IPv4 tails, zone IDs), trailing dots are stripped, and mapped / IPv4-compatible IPv6 classifies as the embedded IPv4 address. Every form in your list is covered by tests, plus 0x7f000001, hex 169.254.169.254, octal 10.x, short-form 192.168.1, and long-form mapped 192.168.0.1, along with negatives confirming public IPs in any spelling still pass. DNS rebinding remains out of scope, as documented.

Gates green on the branch: jest, tsc, eslint, prettier (the only local jest failure remains the known PhotoUtils path assertion in checkouts not named `zeus`).

shubhamkmr04 (approved):
tACK

PR #4375, opened 2026-08-11
Warn before adding a cleartext HTTP Cashu mint

kaloudis:
# Description

The manual Add Mint flow accepted any `http://` or `https://` URL that `UrlUtils.isValidUrl` parsed, then contacted the mint over that scheme with no warning. A cleartext `http://` mint sends invoices, ecash proof Y values, and quote/melt state over an unencrypted connection: any observer on the network path can surveil the traffic, and an active MITM can fabricate quote/melt outcomes (CDK does not verify the melt preimage against the invoice payment hash).

This gates `getMintInfo` on a cleartext-transport check that mirrors the node credential warning added for wallet configuration: if the mint host uses an explicit `http://` scheme and is neither loopback nor a Tor onion service, an explicit confirmation is required before fetching mint info. Loopback never leaves the device and onion services are authenticated at the Tor layer, so both are exempt. The fetch path is extracted into `performGetMintInfo` so it can run either directly or from the confirm callback.

### Scope

This closes the manual-entry vector only. Two related vectors are already handled separately: a ZEUS-Pay-supplied `mint_url` has a native HTTPS/internal-host guard, and a token-embedded mint is no longer polled on view. Out of scope here: the melt preimage-fabrication is an upstream CDK behavior (transport independent, since a malicious mint can already do it over HTTPS), and the app transport-security configs are intentionally permissive to support LND/LNC/LAN/onion node connections.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

Manual test plan (device tests pending):

- [ ] Add a mint with an explicit `http://` non-loopback host: confirmation dialog appears; Cancel aborts, Proceed continues to lookup.
- [ ] Add an `http://127.0.0.1` / `http://localhost` mint: no warning (loopback exempt).
- [ ] Add an `http://<host>.onion` mint: no warning (onion exempt).
- [ ] Add a normal `https://` mint: no warning, unchanged behavior.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Cashu/AddMint.tsx
+++ b/views/Cashu/AddMint.tsx
@@ -84,40 +84,6 @@ export default class AddMint extends React.Component<
         CashuStore.error = false;
     }
 
-    // Returns true when an explicit http:// scheme would carry mint traffic
-    // (invoices, ecash proofs, quote state) to a host that is neither loopback
-    // nor a Tor onion service, i.e. it would cross the network in cleartext.
-    private isCleartextHttpMint = (urlString?: string): boolean => {
-        if (!urlString) return false;
-        const trimmed = urlString.trim().toLowerCase();
-        if (!trimmed.startsWith('http://')) return false;
-
-        // Isolate the host: strip scheme, then any path/query, then a
-        // trailing :port (handling IPv6 bracket notation).
-        let hostPart = trimmed
-            .slice('http://'.length)
-            .split('/')[0]
-            .split('?')[0];
-        if (hostPart.startsWith('[')) {
-            hostPart = hostPart.slice(1).split(']')[0];
-        } else {
-            hostPart = hostPart.split(':')[0];
-        }
-
-        // Onion services are authenticated at the Tor layer and loopback
-        // never leaves the device, so neither exposes mint traffic.
-        if (hostPart.endsWith('.onion')) return false;
-        if (
-            hostPart === 'localhost' ||
-            hostPart === '::1' ||
-            hostPart.startsWith('127.')
-        ) {
-            return false;
-        }
-
-        return true;
-    };
-
     getMintInfo = async () => {
         const { mintUrl } = this.state;
         if (!mintUrl.trim()) {
@@ -154,7 +120,7 @@ export default class AddMint extends React.Component<
         // A cleartext http:// mint exposes invoices, proofs, and quote state
         // to any network observer and lets a MITM fabricate quote/melt
         // outcomes, so require an explicit confirmation before contacting it.
-        if (this.isCleartextHttpMint(mintUrl)) {
+        if (UrlUtils.isCleartextHttpTransport(mintUrl)) {
             confirmAction(
                 localeString('general.warning'),
                 localeString('views.Cashu.AddMint.cleartextHttpWarning'),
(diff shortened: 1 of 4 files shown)


views/Cashu/AddMint.tsx
```diff
@@ -83,6 +84,40 @@ export default class AddMint extends React.Component<
         CashuStore.error = false;
     }
 
+    // Returns true when an explicit http:// scheme would carry mint traffic
+    // (invoices, ecash proofs, quote state) to a host that is neither loopback
+    // nor a Tor onion service, i.e. it would cross the network in cleartext.
+    private isCleartextHttpMint = (urlString?: string): boolean => {
```
ajaysehwal: @kaloudis  , can we add some tests for isCleartextHttpMint ?

views/Cashu/AddMint.tsx
```diff
@@ -83,6 +84,40 @@ export default class AddMint extends React.Component<
...
+            hostPart = hostPart.split(':')[0];
+        }
+
+        // Onion services are authenticated at the Tor layer and loopback
+        // never leaves the device, so neither exposes mint traffic.
+        if (hostPart.endsWith('.onion')) return false;
+        if (
+            hostPart === 'localhost' ||
+            hostPart === '::1' ||
+            hostPart.startsWith('127.')
```
ajaysehwal: @kaloudis I think there’s a small issue with the localhost check.

`hostPart.startsWith('127.')` can incorrectly treat hosts like `127.0.0.1.evil.com` as loopback and skip the warning.

Since `isValidUrl` already parses the URL, I think we should use `parsed.hostname` and perform a proper `127.0.0.0/8` check instead of relying on the string prefix. This would also handle URL userinfo correctly.

kaloudis: Good catch, confirmed. The prefix check failed open on 127.0.0.1.evil.com, and while fixing it I found two more holes in the string parsing: userinfo (http://127.0.0.1@evil.com) and fragments (http://evil.com#foo.onion) could both hide the real host and pick up an exemption.

Fixed in cf009b64b: the check now parses with `new URL` (react-native-url-polyfill is already loaded in index.js and `isValidUrl` relies on it) and inspects the canonical `hostname`. Loopback is an octet-wise 127.0.0.0/8 check, IPv6 loopback compares against the bracketed `[::1]` form the parser returns, and `.onion` is checked on the real host. Unparseable http:// input fails closed and warns. Bonus: IPv4 shorthand like http://127.1 and http://0x7f.1 now canonicalizes to 127.0.0.1 and is correctly exempted.

Note: WalletConfiguration in #4363 has the same helper with the same flaw. Once this lands I'll point that PR at the shared util.

views/Cashu/AddMint.tsx
kaloudis: Done in cf009b64b. The helper moves to `UrlUtils.isCleartextHttpTransport` so it is unit-testable, and `utils/UrlUtils.test.ts` now covers it: https and scheme-less input, http to public hosts, the loopback and onion exemptions (including `[::1]` and IPv4 shorthand), and the impersonation vectors from your other comment (127.0.0.1.evil.com, userinfo, fragment).

ajaysehwal (approved):
tACK

PR #4378, opened 2026-08-11
fix: encrypt node-config exports with AES-256-GCM, drop plaintext variant

kaloudis:
# Description

Hardens the **Tools > Export/Import wallet configurations** flow, which had two credential-exposure problems:

1. **Plaintext export to shared storage.** Unchecking "Encrypt" wrote the full node record (LND `macaroonHex`, CLN `rune`, LNDHub `accessKey`, LNC `pairingPhrase`, NWC `nostrWalletConnectUrl`) as cleartext JSON to Android `DownloadDirectoryPath`, outside the app sandbox, guarded only by a non-blocking warning label.
2. **Weak KDF on the encrypted (default) path.** `CryptoJS.AES.encrypt(json, password)` is CryptoJS passphrase mode, i.e. OpenSSL `Salted__` / EVP_BytesToKey with MD5 and a single iteration and no work factor, so a captured "encrypted" backup is brute-forceable offline.

(Embedded LND / LDK Node seed material is already excluded from export by the selection UI, so this concerns remote-node credentials only.)

### What changed

- Exports are **always encrypted** now. The plaintext export variant and the checkbox are removed; a password is always required, and the export util itself refuses an empty password so no caller can produce an unencrypted file.
- Encryption uses the existing native `ZipUtils` primitive (**PBKDF2-HMAC-SHA256 + AES-256-GCM**, the same one used for channel-DB backups), not CryptoJS.
- The file is saved via the **system save dialog** (`saveDocuments` from `@react-native-documents/picker`, already a dependency: SAF `ACTION_CREATE_DOCUMENT` on Android, the Files exporter on iOS), staged from app-private cache. The user explicitly picks the destination, on-device or any document provider like Drive. Nothing is written to shared storage unprompted, and all staged files (plaintext temp, raw blob, encrypted envelope) are unlinked once the dialog settles.
- A success alert is shown once the encrypted file has been written to the chosen destination; dismissing the dialog is treated as a cancel, not an error.
- New on-disk format is `version: 2` (base64 GCM blob inside the existing JSON envelope). **Legacy `version: 1` (CryptoJS) and old plaintext files remain importable** so existing backups keep working. Import dispatches on the version field.
- The unencrypted (legacy) import path now validates that the payload actually contains a nodes array before rendering the selection modal, instead of crashing on a malformed file.
- **Background-lock fix (separate commit):** with a PIN and "Login when app returns from background" enabled, Android system dialogs (share sheet, SAF save/open pickers) run as separate activities, so presenting one fired a 'background' AppState event and the user landed on the lockscreen when the dialog closed. New `utils/BackgroundLockUtils.ts` lets such flows suppress that one lock; the lock is re-armed if the flow settles while the app is genuinely backgrounded (the user left to another app) or after a 60-second grace period, so returning still requires login in those cases. Wired up for the export and import dialogs on this screen.

### Notes / limitations

- The 100k PBKDF2 iteration count is inherited from the shipped native `ZipUtils` primitive (no native changes here). Raising it further, or moving to a memory-hard KDF, is a separable native follow-up.
- Because past exports used timestamped filenames, previously-written plaintext exports in Downloads cannot be reliably enumerated and purged (unlike a fixed-name file). The scope here is preventing future shared-storage writes.
- Old app builds will not be able to open `version: 2` files (forward-incompatible, same posture as the SCB GCM change).
- Other share-sheet call sites (invoice sharing, rescue key export, activity CSV, channel migration) have the same Android background-lock behavior; they can adopt `BackgroundLockUtils` in a follow-up PR.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

`utils/NodeConfigUtils.test.ts` coverage: v2 encrypted-envelope shape and save-dialog options, no write to shared Downloads, all staged files cleaned up after success, cancel, and save-dialog write errors, stale envelopes from earlier builds swept, empty-password export rejected, no cleartext credential in the output, v2 export/decrypt round-trip, wrong-password rejection, and a legacy v1 CryptoJS known-answer decrypt for backward compatibility. New `utils/BackgroundLockUtils.test.ts`: suppression skips exactly one lock, re-arms when the flow settles backgrounded or exceeds the grace period, no re-arm for in-process (iOS) dialogs or with the setting disabled, and state restoration on task rejection.

- [ ] Android
- [ ] iOS

_Draft: on-device testing still pending._

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


shubhamkmr04 (commented):
Image

we can't see any option to save it in the device locally in android, is this expected ?

shubhamkmr04 (commented):
https://github.com/user-attachments/assets/cdf968b7-1b75-459d-b757-8e669eba9c64

lockscreen comes up when we cancel the share sheet
happens in android only

shubhamkmr04 (commented):
we should add the `views.Tools.nodeConfigExportImport.exportSuccess` on successful export

kaloudis:
Thanks for the testing @shubhamkmr04! All three points addressed in the latest push (a272e9e94 + c6c385024):

**No option to save locally on Android:** That was a consequence of dropping the direct Downloads write: the share sheet only shows a local-save target if an app like Files by Google registers for it, which many devices lack. The export now uses `saveDocuments` from `@react-native-documents/picker` (already a dependency; it powers the import picker on this same screen) instead of the share sheet. On Android that is the system SAF save dialog (`ACTION_CREATE_DOCUMENT`), on iOS the Files exporter, so you explicitly pick where the encrypted file goes, on-device or any document provider like Drive. Nothing is written anywhere unprompted, and the staged envelope in cache is now removed as soon as the dialog settles (no lazy share-target reads to accommodate anymore).

**Lockscreen after canceling the sheet:** Root cause is in `views/Wallet/Wallet.tsx`: with a PIN and "Login when app returns from background" enabled, the Android share sheet (and the SAF dialogs, including this screen's import picker) runs as a separate activity, so presenting one fires a 'background' AppState event, the login status is cleared, and returning triggers the lockscreen. iOS presents these in-process, hence Android-only. This actually pre-exists app-wide for every share sheet with that setting enabled; it just became visible in this flow once the export started using system UI.

Fix is in its own commit (a272e9e94) since it touches the lock flow, so please give it a close look: new `utils/BackgroundLockUtils.ts` lets flows that present a system dialog suppress that one lock. To keep the security posture, the lock is re-armed if the dialog settles while the app is genuinely backgrounded (the user left to another app) or if the dialog was up for more than 60 seconds, so returning still requires login in those cases. It is wired up for the export and import dialogs on this screen; if the approach gets ACKed I can adopt it for the other share-sheet call sites (invoice sharing, rescue key export, activity CSV, channel migration) in a follow-up PR.

**exportSuccess alert:** Added. It only fires after `saveDocuments` resolves, i.e. after the encrypted file has actually been written to the chosen destination; dismissing the dialog is treated as a cancel and stays silent.

`tsc` / lint / prettier green; 22 unit tests across the two touched utils pass.

shubhamkmr04 (commented):
@kaloudis we need some kind of validation for password while exporting configs, right now even empty space ` ` passes

locales/en.json
```diff
@@ -1008,15 +1008,15 @@
     "views.Tools.developerTools.abandonChannel.confirm": "Abandon",
     "views.Tools.developerTools.updateChannelPolicy.global": "Global (all channels)",
     "views.Tools.nodeConfigExportImport.title": "Export/Import Wallet Configurations",
-    "views.Tools.nodeConfigExportImport.explainerAndroid": "Exported wallet configurations can be found in Files > Downloads.",
-    "views.Tools.nodeConfigExportImport.explaineriOS": "Exported wallet configurations can be found in the Files app under the ZEUS folder",
+    "views.Tools.nodeConfigExportImport.explainerAndroid": "Exported wallet configurations are always encrypted with your password. When exporting, you'll be prompted to choose where to save the encrypted file.",
+    "views.Tools.nodeConfigExportImport.explaineriOS": "Exported wallet configurations are always encrypted with your password. When exporting, you'll be prompted to choose where to save the encrypted file.",
```
shubhamkmr04: 2 identical locales

shubhamkmr04 (commented):
https://github.com/user-attachments/assets/20f2b7f1-4f16-4c6d-88e5-d724eb79a83c

the lockscreen still appears in the case when we hit back without saving file

kaloudis:
@shubhamkmr04 check out https://github.com/ZeusLN/zeus/pull/4378/commits/fa555af81f0b02a756dfe6fb4fde0678db0d396b - we've modified some of the re-arm logic for screen locks

locales/en.json
shubhamkmr04: @kaloudis this one is not addressed

shubhamkmr04 (changes requested):
> @kaloudis we need some kind of validation for password while exporting configs, right now even empty space ` ` passes

this is not addressed

kaloudis:
> > @kaloudis we need some kind of validation for password while exporting configs, right now even empty space ` ` passes
> 
> this is not addressed

see https://github.com/ZeusLN/zeus/pull/4378/commits/af5be53880770c169b794fcbbe7ebf8167bcc129

shubhamkmr04 (commented):
One test-only follow-up: the ZipUtils mock in NodeConfigUtils.test.ts is a reversible fake whose "ciphertext" is [0x01][passLen][password][plaintext], and the cleartext assertion checks the base64 wrapper so it passes against a payload containing the credential and password verbatim, and the suite would not notice if encryption broke entirely. Same shape of gap in the v1 test, which encrypts and decrypts with CryptoJS in-process and so can't catch a format or crypto-js change breaking existing backups

kaloudis:
@shubhamkmr04 Good catch, fixed in e7eda1385.

You're right on both. Decoding the blob showed the fake's output was literally
`hunter2!{"nodes":[...,"macaroonHex":"deadbeefcafe",...]}`; base64 was the only reason the assertion passed.
And the v1 test wasn't a known-answer test despite the comment, so a crypto-js format change would have left it
green while real pre-v2 backups stopped importing.

Changes:

- `ZipUtils` fake is now a keystream AEAD (`sha256(pass)` XOR plus an 8-byte password tag). Leaks neither
  plaintext nor password, still rejects wrong passwords.
- Cleartext assertion base64-decodes before searching.
- v1 test decrypts a fixed vector from crypto-js 4.2.0.

One correction: I mutated `exportNodeConfigs` to skip `encryptFile`, and 2 of 30 tests already failed (v2
round-trip and no-trim, both on the header parse). So the suite did notice; only the credential-leak assertion
slept through. Post-change the same mutant fails 3, including that one.

`ZipUtils` is a `NativeModules` binding, so the real AES-256-GCM primitive can't be exercised in Jest. It's
unchanged by this PR (same one shipping for channel-DB backups) and stays device-test territory.

shubhamkmr04 (approved):
tACK

PR #4379, opened 2026-08-11
fix(security): confirm before fetching a scanned BTCPay config

kaloudis:
# Description

Simulator Screenshot - iPhone 17 Pro - 2026-08-11 at 22 28 43

Scanning a BTCPay pairing QR (any string containing `config=` and `lnd.config`) made ZEUS **immediately** issue a GET to the embedded URL before the user confirmed anything (`utils/handleAnything.ts` -> `SettingsStore.fetchBTCPayConfig`). A crafted QR presented as "connect your node" could therefore:

1. contact an attacker-chosen endpoint at scan time, disclosing the user's IP, timing, and app presence (beacon), with a limited blind-SSRF surface toward internal addresses the phone can reach; and
2. propose an attacker-crafted node record (`host = uri` verbatim, any scheme; `macaroonHex = adminMacaroon || macaroon`) for the user to save.

For non-`.onion` URLs the fetch went direct (clearnet) even with Tor enabled, and `http://` was fetched in cleartext.

## Changes

- **Consent before the fetch** (`utils/handleAnything.ts`): validate the `config=` URL locally, then require an explicit confirmation dialog that shows the target host (with an added warning when the connection is cleartext `http://`) **before** any network request. Cancel or a malformed URL returns without fetching (the scanner treats the falsy result as a clean back-navigation).
- **Honor Tor** (`stores/SettingsStore.ts` `fetchBTCPayConfig`): route over Tor whenever Tor is enabled, not just for `.onion`, mirroring `fetchSponsors`, so the fetch does not leak the IP on clearnet.
- **Scheme validation** (`stores/SettingsStore.ts` `parseBTCPayConfig`): reject a non-`http(s)` `uri` before using it as the node host.
- Tests for the proceed, cancel, and malformed-URL paths.

## Out of scope (intentionally not duplicated here)

- Admin-macaroon disclosure and the save-time cleartext-host warning are `WalletConfiguration` territory (see the cleartext-HTTP-credential warning work).
- Internal-IP SSRF is not hard-blocked: a LAN/self-hosted BTCPay is a legitimate setup, so the consent dialog (which shows the host) is the mitigation rather than an outright block.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

Automated (all green): `yarn tsc`, `yarn lint`, `npx jest utils/handleAnything.test.ts` (115/115, including the 3 new BTCPay cases).

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

New tests in `utils/handleAnything.test.ts` cover the proceed, cancel, and malformed-URL paths (3 new BTCPay cases).

- [ ] Android
- [x] iOS

Device testing still pending. With a crafted QR `...?config=http://<host>/lnd.config` and an `https://` variant:
- [x] Scan shows the confirmation dialog with the host before any request; the http variant shows the cleartext warning.
- [x] Cancel closes the scanner with no network request (verify via proxy/logcat).
- [x] Proceed fetches the config and lands in Wallet Configuration as before.
- [x] With Tor enabled, the fetch routes over Tor (no direct clearnet connection to the host).
- [x] A malformed `config=` value shows an error and never fetches.
- [x] Regression: a legitimate BTCPay `lnd-rest` QR still imports the node.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


shubhamkmr04 (commented):
Concept ACK — gate's in the right spot, and I checked that fetchBTCPayConfig only has the one call site, so there's no way around it.

Two ways the dialog can still be fooled, both small fixes:

- The fetch at SettingsStore.ts:1753 follows redirects by default, so a nice-looking host can show a clean dialog and then 302 off to anywhere — LAN addresses included. Worth adding ReactNativeBlobUtil.config({ followRedirect: false }). (The Tor path already rejects 3xx, so it's only the clearnet one.)
- classifyBtcPayConfigUrl parses the host by hand and doesn't drop userinfo. https://btcpay.trusted.com:x@evil.example/lnd.config shows btcpay.trusted.com but fetches evil.example, and http://127.1/ shows up with no cleartext warning. new URL() (already used in TorUtils.ts:31) plus bailing on username/password handles both. If classify returned the cleaned URL for fetchBTCPayConfig to use, you'd also drop the second split('config=') and have one parser deciding what gets hit.

And a question, not a blocker: || this.enableTor sends clearnet over Tor, which can't reach 192.168.x.x — so the LAN BTCPay setup the description calls legit would stop working. Also this.enableTor is the currently selected node's flag, so during pairing it's really just whatever node you were on last (and undefined on a fresh install). Should private ranges skip Tor and just be called out in the dialog instead?

kaloudis:
@shubhamkmr04 all should be addressed in https://github.com/ZeusLN/zeus/pull/4379/commits/5d66e64346ac24b92e3166356936c24ddce81878 - please take a look

utils/handleAnything.ts
```diff
@@ -24,6 +24,140 @@ import wifUtils from './WIFUtils';
 const isClipboardValue = (data: string) =>
     handleAnything(data, undefined, true);
 
+// Loopback hosts resolve to the device itself; a cleartext fetch to one never
+// leaves the phone, so it needs neither an unencrypted-transport warning nor
+// Tor routing.
+const isLoopbackHost = (host: string): boolean =>
```
ajaysehwal, line 57: nit: `isLoopbackHost` currently uses `^127\.`, so a hostname like `http://127.evil.com` would also skip the cleartext warning. Low severity, but worth fixing with a proper `127.0.0.0/8` check using `URL.hostname` and octet validation.

utils/handleAnything.ts
```diff
@@ -24,6 +24,140 @@ import wifUtils from './WIFUtils';
...
+    host.endsWith('.local');
+
+// Classifies the URL embedded in a scanned BTCPay pairing QR before ZEUS
+// contacts it, using the WHATWG URL parser (react-native-url-polyfill) so the
+// host ZEUS displays is exactly the host it will fetch. `valid` is true only
+// for a well-formed, credential-free http(s) URL with a host; `url` is the
+// canonicalized URL to fetch; `isCleartext` flags an http:// host that is
+// neither a Tor onion service nor loopback, i.e. one carried over the network
+// unencrypted.
+const classifyBtcPayConfigUrl = (
```
ajaysehwal, line 84: nit: `classifyBtcPayConfigUrl` doesn't have direct unit tests right now; it's only covered indirectly through the `handleAnything` integration tests.

utils/handleAnything.ts
kaloudis: Good catch. Fixed in 2f61d158a: the octet range checks (loopback and the RFC 1918 / link-local ranges, which had the same weakness) now only apply to literal dotted-quad IPv4 addresses. The WHATWG parser already canonicalizes every real IPv4 host to that form, so anything else dotted is a DNS name and stays classified as public. Added regression tests for 127.evil.com, 10.evil.com, 192.168.evil.com, 172.16.evil.com, and 169.254.evil.com.

utils/handleAnything.ts
kaloudis: Addressed in 2f61d158a: classifyBtcPayConfigUrl is now exported and has direct table-driven tests covering loopback (including IPv4 shorthand and IPv6), private ranges, .local, onion, numeric-label DNS spoofs, credential rejection, canonicalization, and scheme validation.

utils/handleAnything.ts
```diff
@@ -736,8 +861,43 @@ const handleAnything = async (
     } else if (value.includes('config=') && value.includes('lnd.config')) {
         // BTCPay pairing QR
         if (isClipboardValue) return true;
+
+        // The QR content controls the endpoint ZEUS is about to contact, so
+        // validate it locally and require explicit consent BEFORE any network
+        // request (prevents a scan-time IP/timing beacon to an attacker URL).
+        const configRoute = value.split('config=')[1];
```
ajaysehwal, line 892: @kaloudis, I think there's a robustness issue with parsing `config=`.

This can behave unexpectedly with encoded values, multiple `config=` parameters, or additional query parameters. I think it would be safer to parse the URL/query parameters properly and then `decodeURIComponent` the `config` value. This isn't a consent bypass if `classify` still rejects invalid URLs, but it would make the parsing more reliable.

kaloudis: Fair to flag, but I'd keep the extraction as is.

The same `split('config=')[1]` is on master today, in `SettingsStore.fetchBTCPayConfig` (`stores/SettingsStore.ts:1742`). This PR only moved it earlier so the URL can be classified and consented to before any request goes out. Parsing wasn't loosened, so rewriting it here means changing a years-old wire format inside a security fix.

It also fails closed on all three of your cases, since `classifyBtcPayConfigUrl` returns `parsed.href` and that canonical string is what gets fetched verbatim:

- `config=A&config=B` and `config=A&foo=bar` both parse to A's host, which is what the dialog shows and what is contacted.
- A percent-encoded value makes `new URL` throw, so `valid: false`, error alert, no fetch.

Worst case is a 404 on a malformed QR, which matches your read that this isn't a consent bypass.

And `decodeURIComponent` would add bugs we don't have: `URLSearchParams` already decodes (so `.get()` plus `decodeURIComponent` double-decodes, `%252F` silently becomes `/`), it maps `+` to a space, and a bare `decodeURIComponent` throws `URIError` on a lone `%` outside the existing `try`. Decoding before parsing also widens the input the URL parser sees, which is backwards when the guarantee is that the displayed host equals the fetched host.

If you have a real BTCPay QR the current extraction mishandles, send it and I'll fix it as a standalone PR with that string as a test vector.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4376, opened 2026-08-11
fix(security): harden Cashu proof db at rest and on delete

kaloudis:
# Description

Ecash proofs are bearer assets: possession of a proof's `secret` + `C` lets anyone spend it from any Cashu wallet, with no seed or PIN required. Despite that, the CDK SQLite proof database (`cashu_wallet_<hash>.db`) was:

1. stored in plaintext (no encryption at rest),
2. swept into iOS iCloud/iTunes backups (the only `isExcludedFromBackup` in the tree is scoped to the LND folder, not the Cashu DB, which sits beside it in `Application Support`), and
3. left on disk by the user-facing "Delete Cashu data" action (`CashuStore.deleteCashuData` removed mints and storage keys but never deleted the DB file).

This PR closes the Zeus-side portion of that gap:

- **iOS backup exclusion + at-rest file protection.** New `secureDatabaseFile(atPath:)` in `CashuDevKitModule.swift` sets `isExcludedFromBackup` and `FileProtectionType.completeUntilFirstUserAuthentication` on the proof DB and its `-wal`/`-shm` sidecars, right after the DB is created in `initializeWallet` (mirrors the existing `LndMobileTools.excludeLndICloudBackup` pattern). `completeUntilFirstUserAuthentication` protects a powered-off/stolen device without breaking background DB access after first unlock. Android is already covered by `allowBackup="false"` plus OS file-based encryption.
- **DB deletion on "Delete Cashu data."** New native `deleteWalletDatabase()` (iOS Swift + Obj-C bridge, Android `@ReactMethod`) closes the wallet/db handles (releasing the SQLite connection) and unlinks the DB plus `-wal`/`-shm` using the tracked `currentDbPath`. Exposed through `cashu-cdk/CashuDevKit.ts` + `types.ts` and called from `CashuStore.deleteCashuData` after the mint-removal loop, wrapped so a native failure never aborts the rest of the cleanup.

**Out of scope (follow-up):** true app-keyed SQLCipher encryption is not reachable from Zeus. The prebuilt cdk-ffi (v0.14.2) opens SQLite with a keyless `WalletSqliteDatabase(filePath:)` constructor and no key crosses the RN bridge, so it requires an upstream CDK change (fork/patch `cdk-ffi` + `cdk-sqlite`, rebuild the pinned binaries, bump hashes). `NSFileProtection` is the interim at-rest control. The full-wipe wrong-filename bug is addressed separately in the orphaned-key-material PR.

This pull request is categorized as a:

[...]

PR #4380, opened 2026-08-11
fix(security): exclude LDK Node data from iOS backups

kaloudis:
# Description

On iOS, the only backup exclusion in the app covers the embedded LND directory (`excludeLndICloudBackup`, set at wallet creation). The LDK Node data directory at `Documents/ldk-node/<uuid>`, which holds on-chain wallet key material and channel state, carried no `isExcludedFromBackup` flag, so it was included in iCloud and iTunes backups by default. Backup media or Apple ID compromise yielded spendable key material.

This PR applies `NSURLIsExcludedFromBackupKey` to the shared `Documents/ldk-node` parent directory, which excludes every wallet subdirectory:

- on every node start (`startLdkNodeWallet`), so wallets created before this change are repaired idempotently
- at directory creation (`createLdkNodeDirectory`), so new wallets and seed/VSS restores are covered from the first write

No native code was needed: `react-native-fs` 2.20.0's `mkdir` accepts `NSURLIsExcludedFromBackupKey` on iOS, creates intermediate directories, and (re)applies the flag even when the directory already exists. The helper is a no-op on Android, where `allowBackup="false"` already keeps app data out of backups. Failures are logged and never block node start; the flag is simply retried on the next start.

## Behavior change to sign off on

Restoring a phone from an iTunes/iCloud backup will no longer resurrect local LDK Node state. Recovery goes through the keychain-synced wallet config plus VSS (the existing `hasLocalDb` restore-from-seed path in `startLdkNodeWallet`), which is the same model the embedded LND wallet has used since its directory was excluded.

## Non-goals

- The Cashu proof db backup exclusion is handled in #4376.
- `Documents` exposure via `UIFileSharingEnabled` (Files app / paired computers) is not affected by this flag; a companion PR converts the remaining Files-app export flows to share sheets and disables file sharing.
- `tor_data` and contact/wallet photos under Documents are out of scope (no key material).

## Notes for reviewers

- `deleteLdkNodeWallet` and the failed-init cleanup only unlink `ldk-node/<uuid>` child directories, never the parent, so the flag persists across wallet deletion.
- The exclusion helper deliberately swallows errors: worst case is one warning per start and no exclusion until the next attempt.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

New `utils/LdkNodeUtils.test.ts` covers: the flag applied to the base directory on iOS, Android no-op, non-fatal behavior when setting the flag fails, and `createLdkNodeDirectory` in both the fresh and already-exists cases.

- [ ] Android
- [ ] iOS

Device tests pending (will update):
- iOS: create LDK wallet, restart, verify the exclusion flag on `Documents/ldk-node`; upgrade path from a master build with an existing wallet
- Android: LDK wallet start smoke test (helper no-ops)
- Wallet delete flow regression

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Only LDK Node is affected; the change is iOS filesystem metadata on the LDK Node data directory.

### Locales

### Other:

### DIFF
--- a/utils/LdkNodeUtils.test.ts
+++ b/utils/LdkNodeUtils.test.ts
@@ -1,0 +1,122 @@
+const mockMkdir = jest.fn();
+const mockExists = jest.fn();
+
+jest.mock('react-native', () => ({
+    Platform: { OS: 'ios' }
+}));
+
+jest.mock('react-native-fs', () => ({
+    DocumentDirectoryPath: '/mock/documents',
+    mkdir: (...args: any[]) => mockMkdir(...args),
+    exists: (...args: any[]) => mockExists(...args),
+    unlink: jest.fn()
+}));
+
+jest.mock('../ldknode/LdkNodeInjection', () => ({
+    __esModule: true,
+    default: {}
+}));
+
+jest.mock('./LocaleUtils', () => ({
+    localeString: (key: string) => key
+}));
+
+jest.mock('./VssAuthUtils', () => ({
+    deriveVssSigningKeyFromSeed: jest.fn()
+}));
+
+import { Platform } from 'react-native';
+import {
+    createLdkNodeDirectory,
+    ensureLdkNodeBackupExclusion,
+    getLdkNodeBaseDirectory,
+    getLdkNodeStoragePath
+} from './LdkNodeUtils';
+
+describe('LdkNodeUtils', () => {
+    beforeEach(() => {
+        jest.clearAllMocks();
+        (Platform as any).OS = 'ios';
+    });
+
+    describe('getLdkNodeBaseDirectory', () => {
+        it('returns the shared ldk-node directory under Documents', () => {
+            expect(getLdkNodeBaseDirectory()).toEqual(
+                '/mock/documents/ldk-node'
+            );
+        });
+    });
+
+    describe('getLdkNodeStoragePath', () => {
+        it('returns the wallet directory under the shared base directory', () => {
+            expect(getLdkNodeStoragePath('abc-123')).toEqual(
+                '/mock/documents/ldk-node/abc-123'
+            );
+        });
+    });
+
+    describe('ensureLdkNodeBackupExclusion', () => {
+        it('applies the iOS backup exclusion flag to the base directory', async () => {
+            mockMkdir.mockResolvedValue(undefined);
+
+            await ensureLdkNodeBackupExclusion();
+
+            expect(mockMkdir).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (commented):
tACK `5d4f580` — tested on the iOS Simulator. A new LDK wallet gets the exclusion flag on `Documents/ldk-node`; stripping the attribute by hand and restarting reapplies it (the upgrade path for existing wallets); and deleting a wallet removes only its own `<uuid>` folder, leaving the parent directory and its flag intact.

**One non-blocking suggestion**

Nothing asserts that `startLdkNodeWallet` applies the exclusion — the line that repairs wallets created before this change. I removed that call locally and all 7 tests still passed, so a future refactor of `startLdkNodeWallet` could drop the repair path with CI green.

This covers it (~10ms; `shouldRetry` doesn't match `'boom'`, so there's no retry delay):

```ts
// alongside mockMkdir
const mockStart = jest.fn();

jest.mock('../ldknode/LdkNodeInjection', () => ({
    __esModule: true,
    default: { node: { start: (...a: any[]) => mockStart(...a) } }
}));

it('applies the exclusion before the node starts', async () => {
    mockMkdir.mockResolvedValue(undefined);
    mockStart.mockRejectedValue(new Error('boom'));

    await expect(
        startLdkNodeWallet({
            nodeDir: 'abc-123',
            seedMnemonic: 'x',
            network: 'mainnet',
            skipInit: true
        })
    ).rejects.toThrow('boom');

    expect(mockMkdir).toHaveBeenCalledWith('/mock/documents/ldk-node', {
        NSURLIsExcludedFromBackupKey: true
    });
});
```

Confirmed it fails when the `ensureLdkNodeBackupExclusion()` call in `startLdkNodeWallet` is removed, and it's prettier-clean.

kaloudis:
Added in 95191f91b, thanks @shubhamkmr04. Took the suggested test with two tweaks: extended the existing `LdkNodeInjection` mock to `{ node: { start } }` rather than registering a second `jest.mock` for the same module, and added an invocation-order assertion so the test also pins that the exclusion is applied before `start()`. Re-confirmed it fails when the `ensureLdkNodeBackupExclusion()` call in `startLdkNodeWallet` is removed.

PR #4381, opened 2026-08-11
fix(cashu): persist offline send token before deleting proofs

kaloudis:
# Description

## Problem

The offline ecash send path (`CashuStore.sendTokenCDK`, offline branch) bypasses CDK's atomic `prepareSend`/`confirmSend` flow and hand-rolls the send: it reads unspent proofs, selects greedily, builds the `cashuA` token in memory, then calls `CashuDevKit.removeProofs`, which hard-deletes the proofs from the native CDK SQLite database on both platforms. The sent-token record is only persisted later, in `mintToken`, after `sendTokenCDK` returns.

Two consequences:

1. **Crash window burns balance silently.** If the app is killed or crashes between `removeProofs` and the sent-tokens write, the proofs are gone from the database and the token existed only in memory: it was never displayed, never stored. The balance drops with no transaction record. The secrets were never spent at the mint, so a seed restore can technically recover them, but nothing tells the user anything went wrong.
2. **Concurrent sends can double-select proofs.** Neither `mintToken` nor `sendTokenCDK` has a re-entrancy guard, and the selection-then-removal section is not serialized. Two overlapping offline sends can read the same unspent proof set and emit two valid-looking tokens over the same proofs; the second recipient to redeem gets nothing. The online path is immune because CDK reserves proofs transactionally.

There is no bridge method to re-add proofs (both native modules hard-code `updateProofs(added: [])`), so recovery has to work without proof restoration.

## Fix

Restore the persist-before-destroy invariant with a two-phase pending record, and serialize all proof-destructive offline operations through one promise queue (same pattern as `SettingsStore.updateSettingsQueue`):

- **Persist first.** The offline branch now writes a pending record (encoded token plus the selected proofs including their `y` values) to a new `<nodeDir>-cashu-pending-offline-sends` storage key before calling `removeProofs`. The record is dropped only after `mintToken` has durably written the token to `<nodeDir>-cashu-sent-tokens`.
- **Serialize.** The whole critical section (read proofs, select, persist record, remove proofs) runs through `offlineProofOpQueue`, so a second concurrent send re-reads the database only after the first send's proofs are gone: it selects disjoint proofs or fails with insufficient funds. `mintToken` additionally refuses re-entry while a send is in flight.
- **Reconcile at startup.** `reconcilePendingOfflineSends` runs during wallet init right after CDK comes up (local database pass, works offline) and resolves any leftover record:
  - all recorded proofs still present: the crash happened before deletion, so the token was never returned or displayed. The record is dropped and funds are intact.
  - any proof missing: deletion at least partially happened, so the token may have been shown or shared. Leftover proofs are removed for consistency and the token is promoted into `sentTokens` (deduplicated by encoded token), where the user can reshare it or reclaim it via the existing Unspent Tokens flow once online.
  - proof lookup fails: the record is kept with an attempt counter and retried at the next startup. It is never dropped (that would lose the token) and never promoted without removal (the amount would be both counted in the balance and shareable).
- `deleteCashuData` clears the new key. Pending records are disk-only and namespaced by node dir, so `reset()` and wallet switches need no extra handling.

No locale, view, or native changes. A follow-up for a native atomic offline send (marshalling `updateProofs(added:)` so selection, pending state, and removal happen in one database transaction) is worth its own issue; this PR makes the existing JS path crash-safe without new native surface.

## Testing notes

New `stores/CashuStore.test.ts` (first test suite for this store) covers:

- pending record written strictly before `removeProofs` (call-order assertion), with the token and `y` values recoverable from the record
- finalize only after the sent-tokens write, leaving no pending record on the happy path
- all three reconciliation outcomes (rollback, promote, partial-deletion cleanup), dedupe against already-recorded tokens, retry-on-error, and the `cdkInitialized` gate
- two concurrent sends never remove the same proof; the second fails cleanly when the pool is exhausted
- `mintToken` re-entrancy guard

Suggested device test (pending): with the device in airplane mode and an ecash balance, start an offline send and kill the app during the spinner, then relaunch: balance should be intact with no phantom token. Repeat but kill the app after the token QR appears: after relaunch the token should be listed under Unspent Tokens and be reclaimable once back online.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/CashuStore.ts
+++ b/stores/CashuStore.ts
@@ -119,6 +119,21 @@ export interface MintReview {
     rating?: MintReviewRating;
 }
 
+// Record of an offline send persisted BEFORE its proofs are deleted from the
+// CDK database, so a crash mid-send can be reconciled at next startup instead
+// of silently burning the proofs. Proofs retain `y` so reconciliation can
+// check which ones are still present in the database.
+interface PendingOfflineSend {
+    id: string;
+    mintUrl: string;
+    encodedToken: string;
+    value: number;
+    memo?: string;
+    createdAt: number;
+    attempts?: number;
+    proofs: CDKProofWithY[];
+}
+
 /**
  * Parses a rating from review content in format [n/m] at the beginning.
  * Returns the rating object and the remaining content with the rating stripped.
@@ -293,6 +308,11 @@ export default class CashuStore {
     @observable public offlineSpentTokens: Array<CashuToken> = [];
     @observable public showOfflineSpentAlert: boolean = false;
     private isSweeping: boolean = false;
+    // Serializes proof-destructive offline operations (offline send critical
+    // section, pending-record finalization, startup reconciliation) so two
+    // operations can never select or remove the same proofs. Pattern mirrors
+    // SettingsStore.updateSettingsQueue.
+    private offlineProofOpQueue: Promise<unknown> = Promise.resolve();
 
     settingsStore: SettingsStore;
     invoicesStore: InvoicesStore;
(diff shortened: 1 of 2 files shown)


stores/CashuStore.ts
```diff
@@ -1554,6 +1542,149 @@ export default class CashuStore {
...
+        await Storage.setItem(key, records);
+    };
+
+    /**
+     * Remove the pending record for an offline send once the sent token has
+     * been durably recorded in sent-tokens storage. A record left behind by
+     * a failure here is harmless: reconcilePendingOfflineSends dedupes
+     * against sentTokens at next startup.
+     */
+    private finalizePendingOfflineSend = async (
```
ajaysehwal, line 1613: @kaloudis, nice catch pinning `pendingKey` in the offline branch of `sendTokenCDK` to guard against a wallet switch mid-queue.

One similar gap: `finalizePendingOfflineSend` recomputes the key from `this.getNodeDir()` at call time. If the active node changes between creating the pending record and `mintToken` calling `finalizePendingOfflineSend` (there’s an `await` gap for the overpay confirmation + actual send), the finalize step could target the new node’s key and leave the stale record under the original node.

This is self-healing via `reconcilePendingOfflineSends` on the next startup, so it’s not fund-unsafe. But it may be worth threading the captured `nodeDir`/`pendingKey` through to `finalizePendingOfflineSend` for consistency.

kaloudis: Good catch, and confirmed: `finalizePendingOfflineSend` recomputed `pendingOfflineSendsKey()` from `this.getNodeDir()` at call time while the send critical section pinned its key up front. Worth spelling out what I found when tracing it, because the minimal version of the fix you suggested would actually have made the failure mode worse.

There was a second recompute in the same flow: the sent-tokens write in `mintToken` also derived its key from `this.getNodeDir()` at call time. That one matters more than the finalize, because it is the write that actually lands in the wrong namespace after a mid-send wallet switch. Tracing both variants:

**Status quo (nothing pinned downstream).** After a switch, finalize loads the new node's record list, filters by the exact `encodedToken`, finds nothing, and no-ops without writing. It cannot damage the new node's legitimate records. The stale record survives under the originating node, and that node's next `reconcilePendingOfflineSends` promotes the token into its own sent history. Meanwhile `mintToken` wrote the token into the new node's sent list. Net effect: the token appears in both wallets' histories. A cosmetic duplicate, funds intact, exactly the self-healing you described.

**Pinning only the finalize key.** The originating node's record now gets correctly deleted, but the token was still recorded under the new node's key. The originating wallet ends up with its proofs gone, no pending record, and no artifact at all: reconcile has nothing left to promote. That converts a duplicate-display bug into a missing-artifact bug, so the finalize pin only makes sense together with pinning the sent-tokens write.

**What's pushed (8729484f8):** `mintToken` captures `nodeDir` once at entry, before its first await (the overpay Alert gap you pointed at), and threads it through all three touchpoints: the pending-record key inside `sendTokenCDK`, the sent-tokens write, and `finalizePendingOfflineSend`. Two extra hardening points came out of writing it down:

1. If the wallet did change mid-flight, `this.sentTokens` by that point holds the other wallet's in-memory history, so pushing to it and persisting the array would overwrite the originating node's stored list with the wrong wallet's data. The switched path instead does a read-modify-write append against the pinned key and leaves the in-memory list alone.
2. The queued critical section now aborts before proof selection when the active wallet no longer matches the pinned one. Without that, a send queued by wallet A could select and burn wallet B's proofs under a record pinned to A. The abort surfaces through the existing `mintToken` error path.

Two regression tests cover it: the queued-send abort, and a mid-send switch asserting the sent token, the finalize, and the in-memory list all stay consistent with the initiating wallet. `yarn verify` green.

Residual caveat, out of JS reach: if the switch lands inside the native calls themselves, the CDK operations resolve against whatever wallet is active at that instant. That is the same class of problem as the non-atomic `removeProofs`, tracked in #4382 for a native fix.

kaloudis:
Rebased onto current master (`1482966fb`) and fixed the CI test failure: master's Cashu key completeness guard correctly flagged that the new `pending-offline-sends` storage key was missing from `CASHU_KEY_SUFFIXES`, meaning Clear All Data would have left serialized send tokens behind. Added it to the wipe list in `8017187e4`. Full suite green locally (1589 tests, 73 suites), tsc clean.

stores/CashuStore.ts
```diff
@@ -1659,28 +1824,65 @@ export default class CashuStore {
...
+                        .toString(36)
+                        .slice(2, 10)}`,
+                    mintUrl,
+                    encodedToken: encoded,
+                    value: total,
+                    memo,
+                    createdAt: Date.now(),
+                    proofs: selected
+                });
+                await this.savePendingOfflineSends(pendingKey, records);
```
ajaysehwal, line 1867: @kaloudis, one gap I think is worth closing before the merge: in the critical section:

``` ts
await this.savePendingOfflineSends(pendingKey, records);
await CashuDevKit.removeProofs(selected.map((p) => p.y));
```

If `removeProofs` rejects, the send correctly fails, but the pending record remains on disk while the proofs are still present and selectable.

If the user retries with a different amount (or another offline send consumes a subset of those proofs), reconciliation can later see partial presence and take the promote path. That could surface a token whose proofs overlap with a different legitimately-sent token, leaving a phantom entry that fails when redeemed.

Since we’re still inside the serialized section and the proofs are untouched when `removeProofs` fails, I think it’s safe to remove the pending record in the catch before rethrowing:

```ts
try {
    await CashuDevKit.removeProofs(selected.map((p) => p.y));
} catch (e) {
    await this.savePendingOfflineSends(
        pendingKey,
        records.filter((r) => r.encodedToken !== encoded)
    );
    throw e;
}
```

Low probability, but this is the one failure path I found where the new machinery could create the kind of phantom artifact it’s meant to prevent. A focused test alongside the existing insufficient-proofs case would cover it well.

kaloudis: Good find, and confirmed. Traced the reconcile path to check severity: it is actually a bit worse than a phantom entry that fails on redeem. In the retry scenario the stale record hits the partial-presence branch, which removes the still-present non-overlapping proofs and wraps them into the promoted token. Once the real recipient redeems the second token, the overlapping proofs are spent at the mint, and since receive is all-or-nothing per swap, the non-overlapping remainder inside the phantom becomes unredeemable through that token. So the stale record can strand value, not just clutter history.

Before taking the fix I verified its load-bearing assumption, that a `removeProofs` rejection leaves the proofs untouched. Both native modules implement it as a single `db.updateProofs(added: [], removedYs:)` FFI call, and cdk-sql-common's `update_proofs` runs one `DELETE ... WHERE y IN (...)` inside a transaction that commits at the end. A rejection surfacing to JS therefore means rollback, nothing removed. (The theoretical Kotlin window where the commit succeeds but the resolve dispatch fails ends in a never-settling promise, not a rejection, so the catch cannot fire on a committed delete. That path is crash-equivalent and reconciliation handles it as before.)

Pushed in `72912a2cd`, exactly the shape you suggested: cleanup in the catch before rethrowing, still inside the serialized section so it cannot interleave with another send or with reconcile. The regression test covers the full scenario end to end: rejected removal (asserting proofs and pending storage both come back clean), an overlapping retry that succeeds, then a reconcile that promotes nothing and burns nothing. Verified the test fails against the previous commit. Full suite green (1590 tests, 73 suites), tsc and lint clean.

stores/CashuStore.ts
```diff
@@ -1734,28 +1899,81 @@ export default class CashuStore {
...
+            // Capture the storage key before queueing so a wallet switch
+            // can't redirect a queued send to another node's namespace.
+            // mintToken passes the dir it pinned before its first await;
+            // fall back to the current dir for direct callers.
+            const sendNodeDir = nodeDir ?? this.getNodeDir();
+            const pendingKey = this.pendingOfflineSendsKey(sendNodeDir);
+            // The critical section is serialized so overlapping sends can't
+            // select the same proofs: the second send re-reads the database
+            // only after the first send's proofs have been removed.
+            return this.enqueueOfflineProofOp(async () => {
```
ajaysehwal, line 1911: @kaloudis, I think there’s still a wallet-switch race in the offline critical section.

The new `getNodeDir() !== sendNodeDir` check only runs once when the queued task starts, but `CashuDevKit.getUnspentProofs()` / `removeProofs()` operate on whichever native wallet/database is loaded at call time. If the user switches from wallet A to B after that check but while one of those calls is awaiting, the native calls can operate on wallet B’s proofs while the JS records are still written under wallet A’s `nodeDir` / `pendingKey`.

That could cause wallet B’s proofs to be consumed while wallet A records the resulting token, which is a real fund/accounting issue.

The current tests don’t model this because `mockProofPool` is shared across wallets, so the wallet-switch tests only verify the JS bookkeeping, not which native wallet the proof operations actually hit.

I think the native proof operations need to be scoped/serialized with the wallet switch, or otherwise revalidated/locked for the entire critical section.

PR #4384, opened 2026-08-11
fix(security): hash app-lock verifiers and stop keychain backup migration

kaloudis:
# Description

## Problem

The app lock (PIN / passphrase and their duress variants) was a UI gate, not an encryption boundary. All four credentials were stored as plaintext fields of the `Settings` JSON blob, side by side with every wallet secret (seeds, LND password, macaroons, LNC pairing phrase, NWC secret), inside a single keychain entry. Login was a direct `===` string compare against those plaintext values.

On iOS the blob was written with react-native-keychain's default accessibility class `kSecAttrAccessibleAfterFirstUnlock`, which Apple documents as migrating to a new device through encrypted backups. So an attacker with an encrypted backup or Apple-ID restore (or a root/jailbreak keychain read) recovered the seeds and both the normal and duress credentials, and could tell exactly which credential triggers the wipe, defeating the duress feature against a forensic attacker.

## Fix

1. **Hash the verifiers.** New `utils/LockVerifierUtils.ts` derives a salted scrypt digest (`{v, kdf, n, r, p, salt, hash}`, N=32768, 16-byte random salt) and verifies in constant time, failing closed on a missing or malformed record. The four credentials now persist this record instead of the raw value. All four use the identical record shape, so a duress verifier is indistinguishable from a normal one to anyone reading the blob.

2. **One-shot migration.** `MigrationUtils.migrateLockCredentialsToVerifiers` (MOD-gated `lock-verifier-hash-mod`) hashes any existing plaintext credential into its verifier and deletes the plaintext field. It runs on both the legacy and modern (`zeus-settings-v2`) load paths.

3. **Stop backup migration.** `storage/index.ts` now writes keychain items with `accessible: AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY`. This is non-migratable, still readable at boot after first unlock, and (unlike `WhenPasscodeSet`) never requires a device passcode, so it cannot lock a user out of their own wallet. No-op on Android, which is already covered by `allowBackup="false"`.

4. **Duress stays indistinguishable.** `Lockscreen` computes the normal and duress checks at equal cost (`Promise.all`) before branching, so response time never reveals which credential matched.

5. **In-flight guard and spinner.** scrypt verification is asynchronous (~1s), so `onAttemptLogIn` guards against re-entrancy with a `verifying` flag (a double tap or second submit can no longer start a concurrent attempt and double-count the failure counter), and a full-screen spinner overlay is shown while the derivation runs. The overlay also swallows taps so a second submit cannot land.

The `SetPin` / `SetDuressPin` / `SetPassword` / `SetDuressPassword` views route their cross-checks through `verifySecret`, persist verifiers, and use presence booleans instead of prefilling saved plaintext.

## Out of scope (follow-ups)

- Hardware-backed (Secure Enclave / Keystore `accessControl`) encryption of the whole blob: the blob is read at boot, before auth (node list, wallet selection), so it cannot sit behind a biometric-gated keychain item. scrypt plus non-migratable storage is the pragmatic ceiling here. A determined attacker can still offline-brute-force a short PIN, which is inherent to any PIN scheme without hardware binding, but this raises it from trivial to costly.
- Android `securityLevel: SECURE_HARDWARE` (device-dependent failures).

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

Pending device tests (both platforms):
- [ ] Set PIN, lock, unlock with correct and incorrect entries (spinner shows during verification)
- [ ] Duress PIN triggers the full wipe
- [ ] Passphrase path
- [ ] Upgrade migration from a build carrying a plaintext PIN: unlock still works and the stored blob contains no plaintext credential
- [ ] iOS: encrypted-backup restore to a second device no longer carries the settings item
- [ ] iOS: upgrade an install with pre-existing keychain items and confirm the one-shot re-stamp runs (write-once items no longer restore to a second device)

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

App-lock and keychain-storage layer change; backend-independent.

### Locales

### Other:

### DIFF
--- a/storage/index.ts
+++ b/storage/index.ts
@@ -53,7 +53,17 @@ class Storage {
             prefixedKey,
             stringValue,
             {
-                cloudSync: false
+                cloudSync: false,
+                // iOS: keep keychain items on this device only. The default
+                // (AFTER_FIRST_UNLOCK) migrates to a new device through
+                // encrypted iCloud/Finder backups, carrying the settings blob
+                // (wallet seeds, LND password, macaroons, lock verifiers) with
+                // it. THIS_DEVICE_ONLY blocks that migration while staying
+                // readable at boot after first unlock, and (unlike
+                // WHEN_PASSCODE_SET) never requires a device passcode, so it
+                // cannot lock a user out of their own wallet. No-op on Android.
+                accessible:
+                    Keychain.ACCESSIBLE.AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY
             }
         );
         return response;
--- a/utils/LockVerifierUtils.ts
+++ b/utils/LockVerifierUtils.ts
@@ -1,0 +1,122 @@
+import { scrypt } from 'scrypt-js';
+
+// App-lock credentials (PIN / passphrase and their duress variants) must never
+// be stored in plaintext: the settings blob they live in also holds every
+// wallet secret and, on iOS, historically migrated through encrypted backups.
+// Instead we store a salted scrypt digest and compare in constant time. All
+// four verifiers use the identical record shape so a duress verifier is
+// indistinguishable from a normal one to anyone who reads the blob.
+
+export interface VerifierRecord {
+    v: 1;
+    kdf: 'scrypt';
+    n: number;
+    r: number;
+    p: number;
+    salt: string; // hex
+    hash: string; // hex
+}
+
+// scrypt work factors. N=32768 keeps a single derivation in the ~1s range on
+// device, which is a deliberate cost: it rate-limits online guessing and makes
+// an offline brute force of a short PIN expensive rather than trivial.
+const SCRYPT_N = 32768;
+const SCRYPT_R = 8;
+const SCRYPT_P = 1;
+const KEY_LEN = 32;
+const SALT_LEN = 16;
+
+const toBytes = (secret: string): Buffer => Buffer.from(secret, 'utf8');
+
+// react-native-randombytes runs a native init() at import time that throws
+// under jest, so it is loaded lazily here rather than at module scope. Keeping
+// this module side-effect free lets any store/view import it without dragging
+// the native dependency into unrelated test suites; test paths pass an explicit
+// salt and never reach this.
+const generateSalt = (length: number): Buffer => {
+    const { randomBytes } = require('react-native-randombytes');
+    return randomBytes(length);
+};
+
+/**
+ * Derive a verifier record for a secret. A fresh random salt is used per call,
+ * so the same PIN set twice yields different records. `salt` is injectable for
+ * deterministic tests only.
+ */
+export const deriveVerifier = async (
+    secret: string,
+    salt?: Buffer
+): Promise<VerifierRecord> => {
+    const saltBuf = salt ?? generateSalt(SALT_LEN);
+    const derived = await scrypt(
+        toBytes(secret),
+        saltBuf,
+        SCRYPT_N,
+        SCRYPT_R,
+        SCRYPT_P,
+        KEY_LEN
+    );
+    return {
+        v: 1,
+        kdf: 'scrypt',
+        n: SCRYPT_N,
+        r: SCRYPT_R,
+        p: SCRYPT_P,
+        salt: Buffer.from(saltBuf).toString('hex'),
+        hash: Buffer.from(derived).toString('hex')
+    };
+};
+
+/** True when `record` is a well-formed verifier (used for UI/gate presence). */
+export const hasVerifier = (record?: VerifierRecord | null): boolean =>
+    !!record &&
+    record.kdf === 'scrypt' &&
+    typeof record.salt === 'string' &&
+    record.salt.length > 0 &&
+    typeof record.hash === 'string' &&
+    record.hash.length > 0;
+
+const constantTimeEqualHex = (a: string, b: string): boolean => {
+    const bufA = Buffer.from(a, 'hex');
+    const bufB = Buffer.from(b, 'hex');
+    if (bufA.length === 0 || bufA.length !== bufB.length) {
+        return false;
+    }
+    let diff = 0;
+    for (let i = 0; i < bufA.length; i++) {
+        diff |= bufA[i] ^ bufB[i];
+    }
+    return diff === 0;
+};
+
+/**
+ * Verify a candidate secret against a stored verifier. Returns false (fail
+ * closed) for an empty secret or a missing/malformed record, and never throws.
+ */
+export const verifySecret = async (
+    secret: string,
+    record?: VerifierRecord | null
+): Promise<boolean> => {
+    if (!secret || !hasVerifier(record)) {
+        return false;
+    }
+    const rec = record as VerifierRecord;
+    try {
+        const saltBuf = Buffer.from(rec.salt, 'hex');
+        const dkLen = Buffer.from(rec.hash, 'hex').length || KEY_LEN;
+        const derived = await scrypt(
+            toBytes(secret),
+            saltBuf,
+            rec.n || SCRYPT_N,
+            rec.r || SCRYPT_R,
+            rec.p || SCRYPT_P,
+            dkLen
+        );
+        return constantTimeEqualHex(
+            Buffer.from(derived).toString('hex'),
+            rec.hash
+        );
+    } catch (e) {
+        return false;
+    }
+};
--- a/utils/LockVerifierUtils.ts
+++ b/utils/LockVerifierUtils.ts
@@ -67,6 +67,23 @@ export const deriveVerifier = async (
     };
 };
 
+/**
+ * Fixed, well-formed verifier used to equalize verification cost when a slot
+ * (e.g. duress) has no credential configured. Verifying an attempt against it
+ * runs the full scrypt derivation, so unlock latency does not reveal whether
+ * a duress credential exists, but it can never match: the stored hash is a
+ * constant, not the digest of any secret under this salt.
+ */
+export const DUMMY_VERIFIER: VerifierRecord = {
+    v: 1,
+    kdf: 'scrypt',
+    n: SCRYPT_N,
+    r: SCRYPT_R,
+    p: SCRYPT_P,
+    salt: '00'.repeat(SALT_LEN),
+    hash: '00'.repeat(KEY_LEN)
+};
+
 /** True when `record` is a well-formed verifier (used for UI/gate presence). */
 export const hasVerifier = (record?: VerifierRecord | null): boolean =>
     !!record &&
--- a/utils/MigrationUtils.ts
+++ b/utils/MigrationUtils.ts
@@ -99,6 +99,7 @@ const CASHU_MIGRATION_KEY = 'ios-keychain-cashu-fix-v1';
 
 import EncryptedStorage from 'react-native-encrypted-storage';
 import Storage from '../storage';
+import { deriveVerifier, hasVerifier } from './LockVerifierUtils';
 
 class MigrationsUtils {
     /**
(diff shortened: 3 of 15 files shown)


utils/MigrationUtils.ts
```diff
@@ -475,9 +476,58 @@ class MigrationsUtils {
...
+            if (settings[plain] !== undefined) {
+                delete settings[plain];
+                changed = true;
+            }
+        }
+
+        if (changed) {
+            await settingsStore.setSettings(settings);
+        }
+        await EncryptedStorage.setItem(MOD_KEY_LOCK_VERIFIER, 'true');
```
shubhamkmr04, line 546: The flag is set even when `setSettings` returned false — it's gated on the wipe write-latch, and the keychain write can fail on its own. Next boot then skips the migration and loads a blob that still has plaintext `pin` and no `pinVerifier`, so `loginMethodConfigured()` is false and the app opens with no lock at all. Gate this on the returned boolean.

utils/MigrationUtils.ts
```diff
@@ -475,9 +476,58 @@ class MigrationsUtils {
...
 
+    // Replace plaintext app-lock credentials with salted scrypt verifiers and
+    // strip the plaintext fields. Historically the PIN, passphrase and their
+    // duress variants were stored verbatim in the settings blob alongside every
+    // wallet secret, and login was a direct string compare - so anyone who
+    // could read the blob (root/jailbreak dump, or an iOS backup migration)
+    // recovered the credentials and could tell which one was the duress
+    // trigger. Runs once (MOD-gated) on both the legacy and modern
+    // (zeus-settings-v2) load paths.
+    public async migrateLockCredentialsToVerifiers(settings: any) {
```
shubhamkmr04, line 493: On the legacy path the stripped settings go out through `setSettings` (modern key), but the legacy `zeus-settings` EncryptedStorage blob is never rewritten or removed — `storageMigrationV2` only copies forward. The plaintext PIN/passphrase and duress variants survive there for anyone who came through that path, which is most of the population this migration is aimed at.

views/Lockscreen.tsx
```diff
@@ -295,14 +303,36 @@ export default class Lockscreen extends React.Component<
...
+        const attempt =
+            authMethod === 'passphrase' ? passphraseAttempt : pinAttempt;
+        const primaryVerifier =
+            authMethod === 'passphrase' ? passphraseVerifier : pinVerifier;
+        const duressVerifier =
+            authMethod === 'passphrase'
+                ? duressPassphraseVerifier
+                : duressPinVerifier;
+
+        const [primaryMatch, duressMatch] = await Promise.all([
```
shubhamkmr04: Not equal cost: `verifySecret` returns early when the record is missing, and scrypt-js is CPU-bound on the JS thread so `Promise.all` doesn't overlap the two anyway. One derivation (~1s) when no duress credential is configured vs two (~2s) when one is — a timing signal for whether duress exists, and doubled unlock latency for everyone who has it. Verifying against a dummy record when no duress verifier is stored would keep the cost flat.

views/Lockscreen.tsx
```diff
@@ -401,7 +430,8 @@ export default class Lockscreen extends React.Component<
                 await updateSettings({ authenticationAttempts }).then(() => {
                     this.setState({
                         error: true,
-                        pinAttempt: ''
+                        pinAttempt: '',
+                        verifying: false
```
shubhamkmr04, line 445: `verifying: false` only lands in `.then()`. `updateSettings` can reject — its queue deliberately re-throws to the caller — and a rejection here leaves the tap-swallowing overlay up with the re-entrancy guard still armed, so the user is locked out until force-quit. Needs a `catch`/`finally`.

storage/index.ts
```diff
@@ -53,7 +53,17 @@ class Storage {
...
+                // iOS: keep keychain items on this device only. The default
+                // (AFTER_FIRST_UNLOCK) migrates to a new device through
+                // encrypted iCloud/Finder backups, carrying the settings blob
+                // (wallet seeds, LND password, macaroons, lock verifiers) with
+                // it. THIS_DEVICE_ONLY blocks that migration while staying
+                // readable at boot after first unlock, and (unlike
+                // WHEN_PASSCODE_SET) never requires a device passcode, so it
+                // cannot lock a user out of their own wallet. No-op on Android.
+                accessible:
+                    Keychain.ACCESSIBLE.AFTER_FIRST_UNLOCK_THIS_DEVICE_ONLY
```
shubhamkmr04, line 66: This only applies on write, so items already in the keychain keep `AfterFirstUnlock` until something happens to rewrite them. The settings blob is rewritten often enough, but write-once keys aren't — LNC credentials, contacts, notes, cashu seeds — and those still ride an encrypted backup to a new device. A one-shot re-stamp pass (read + rewrite each key) would close it.

shubhamkmr04 (changes requested):
Blocking on the readers that weren't migrated along with it.

The migration deletes `pin` / `passphrase` (and the duress variants), and the `Set*` views never persist them again, so every remaining plaintext reader is permanently falsy — for migrated *and* freshly-set credentials. None of these files are in the diff:

- `utils/NavigationUtils.ts:12` — `protectedNavigation` gates on `settings.passphrase || settings.pin`. In POS mode, navigation to Wallet/Menu now skips the Lockscreen entirely.
- `views/Settings/Security.tsx:233` — the re-authenticate-before-changing-your-credential gate is always false, so `SetPin`/`SetPassword` open without proving the old credential.
- `views/Settings/Security.tsx:52` and `:125-126` — `buildSecurityItems` / `deriveStateFromSettings` always take the "nothing configured" branch: Change/Delete PIN, Delete Password and every duress row disappear (the Lockscreen delete flows become unreachable), and the scramblePIN / loginBackground switches hide with them.
- `views/Settings/Security.tsx:179` — biometrics can never be enabled; the "set a PIN first" modal fires unconditionally, and the `enableBiometrics` re-entry at `:168` loops straight back into it.
- `views/Settings/PointOfSale.tsx:122` — the "no authentication configured" warning is now permanent.

`hasVerifier(settings.pinVerifier)` and `settingsStore.loginMethodConfigured()` are the drop-in replacements. `NavigationUtils.test.ts` covers `reAuthNavigation` but not `protectedNavigation` — worth closing that gap here.

utils/MigrationUtils.ts
kaloudis: Good catch, and the failure mode was worse than a stale flag: with the plaintext already stripped in memory but never persisted, the next boot would have loaded a blob with plaintext fields and no verifiers, and the app would have booted with no lock at all. The flag is now only stamped after setSettings returns true, so a failed write retries on the next load. Covered by a new test (does not stamp the flag when persisting the settings fails).

utils/MigrationUtils.ts
kaloudis: Fixed. The migration now rewrites the legacy zeus-settings blob with the four plaintext fields stripped (leaving everything else intact, since KeychainRecoveryUtils still lists it as a recovery source), and it runs before the MOD flag is stamped so a failure there retries on the next load. Two new tests cover the strip and the no-credential case.

views/Lockscreen.tsx
kaloudis: Right on both counts: verifySecret returned early on the missing record, and scrypt-js is CPU-bound on the JS thread so Promise.all never overlapped the derivations. Took your dummy-record suggestion: LockVerifierUtils now exports DUMMY_VERIFIER (well-formed, constant hash that can never match) and the Lockscreen substitutes it when no duress verifier is stored, so every attempt costs exactly two derivations regardless of configuration. I also looked at deduplicating via a shared salt so both checks could ride one derivation, but that dies on PIN change: re-salting the primary would require re-deriving the duress verifier and the duress plaintext is not available at that point. If the flat ~2s is felt to be too slow, halving N is the lever, but I have left it as is for now.

views/Lockscreen.tsx
kaloudis: Fixed with a try/catch around the verification and branch logic: any rejected settings read/write now clears the verifying flag (and shows the error state) instead of leaving the overlay up. I deliberately used catch rather than finally: the success paths keep the flag armed on purpose so the overlay keeps swallowing taps during the navigation transition, and the wipe branches keep it armed so the user stays pinned to the wiping screen.

storage/index.ts
kaloudis: Added restampKeychainAccessibility, a one-shot iOS pass (MOD ios-keychain-accessible-restamp-v1) that rewrites existing items in place so they pick up THIS_DEVICE_ONLY. It walks the same inventory as keychainCloudSyncMigration: the static key list, per-note keys via the notes index, LNC credentials derived from each node's pairing phrase, and Cashu keys (including seed material) per embedded-lnd and ldk-node node, plus per-mint wallet keys via the mintUrls list. The settings blob goes through settingsStore.setSettings rather than a raw read/rewrite so it cannot race a queued settings update. The flag is only stamped when every rewrite succeeds, so partial failures retry next launch; rewrites are idempotent.

Two accepted residuals, called out in a code comment: lnurlpay:<paymentHash> and pos-<orderId> entries have no index to enumerate from JS (react-native-keychain v10 exposes no listing for internet credentials). Both hold payment metadata rather than key material. Legacy unprefixed keychain entries are already deleted by keychainCloudSyncMigration after copying, so they do not need re-stamping.

kaloudis:
Thanks for the thorough review, all six findings were real and are addressed in 6e8cb4056.

On the blocking finding: every plaintext reader you listed is migrated to verifier checks in that commit. protectedNavigation now gates on hasVerifier(passphraseVerifier) || hasVerifier(pinVerifier) (kept narrower than loginMethodConfigured, which also counts biometrics; the old gate was pin/passphrase only and the Lockscreen fallback path expects one of those). Security.tsx derives all rows, the pinExists/passphraseExists state, the re-auth gate, and the biometrics prerequisite from verifier presence, and the PointOfSale warning does the same. A repo-wide sweep for settings.pin / settings.passphrase / settings.duressPin / settings.duressPassphrase reads now comes back empty outside of test assertions.

On the test gap: NavigationUtils.test.ts did have protectedNavigation cases, but they seeded the plaintext shape ({ pin: '1234' }) so they asserted nothing about the post-migration shape. They now use verifier records (and would have caught this bug), with added cases for the passphrase verifier and the no-credential POS-active path.

Inline replies on each of the other five. yarn verify green locally; the pending device-test matrix in the PR body still stands and now includes verifying the re-stamp on an upgraded install.

utils/LockVerifierUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+    r: number;
+    p: number;
+    salt: string; // hex
+    hash: string; // hex
+}
+
+// scrypt work factors. N=32768 keeps a single derivation in the ~1s range on
+// device, which is a deliberate cost: it rate-limits online guessing and makes
+// an offline brute force of a short PIN expensive rather than trivial.
+const SCRYPT_N = 32768;
```
shubhamkmr04, line 23: .

shubhamkmr04 (commented):
Did the device testing. After migrating a PIN from master to the latest PR commit, unlocking takes at least 10-15 seconds on both iOS and Android. Details inline.

shubhamkmr04 (commented):
Did the device testing on a debug build. After migrating a PIN from the pre-PR commit, unlocking takes ~10s on Android and 10-15s on iOS, worth confirming on release, but the gap to the ~1s budget is large.

shubhamkmr04:
@kaloudis please rebase

kaloudis:
Thanks for running the device test. This is expected with the current implementation rather than a debug-only artifact: scrypt-js is pure JS, Hermes has no JIT, and since the duress-timing-parity change every attempt pays two derivations that serialize on the JS thread. For reference, one derivation at these parameters takes ~230ms under Node's JIT on an M-series laptop, so multi-second times on a phone under Hermes are in line. A release build should tighten it (optimized bytecode, no dev overhead) but it will still land in the seconds range, not the ~1s the code comment claims.

I'm taking a look at native alternatives for the derivation as part of the crypto modernization epic (#4505). The VerifierRecord format stores its KDF parameters, so swapping the backend keeps already-migrated verifiers valid with no format change. Will update here once the path is settled; holding off on asking for re-review until then.

Issue #3625, opened 2026-01-30
Multiple wallets are missing after update


LOCHER-21:
### Describe the bug

After the automatic update to v0.12.2 now 2 of my wallets was deleted again. This happened on 3 different devices on the same update that was installed automatically last night. The same wallets are missing on all of my 3 devices. It was not the first time that a few wallets are missing in Zeus after updating the app. On every update you have to pay attention and check if the update hasn't deleted some wallets again.

### Reproduce

Updating to the newest version of Zeus via App Store on iOS

### ZEUS version

v0.12.2

### Node interface

LND (REST)

### Network

Clearnet

### Device

iPhone 17 Pro

### Device operating system

iOS 26.2.1

### Log output

```shell

```

kaloudis:
We are investigating and building a wallet restore tool for the next beta - hoping to have it out later today. FWIW this appears to be an issue specific to iOS systems. Android users seem to be unaffected.

If users still can’t recover the wallets we are providing free replacements for any channels they had with the LSP.

We deeply apologize to any users affected and are committed to setting things right.

kaloudis:
In our latest beta, v0.12.3-beta1 (https://github.com/ZeusLN/zeus/releases/tag/v0.12.3-beta1) we've built out a tool that may assist in the recovery of your wallet connections.

It's located under `Menu` > `Tools` > `Export/Import Wallet Configurations` > `Restore legacy wallet configurations`

We hope you take the time to try it and report back.

ellipticKurve256k1:
This is happening in my wallet as well.
I’m using an iPhone 13 Pro on iOS 26.2.1.

I can’t even create a new embedded node, the same message keeps appearing.

![image](https://github.com/user-attachments/assets/cbe1bb5a-2f1c-4914-bc2a-12f84827a9ea)

I do have my seed backed up, so I hope I can recover the funds. However, since I’m currently unable to create a new embedded node at all, it seems the app itself is unusable.

LOCHER-21:
I successfully  recovered all the missing wallets from my own backups. Fortunately, I wasn’t traveling during the update, because then I wouldn’t have had access to my backups. That would have been a nightmare.

kaloudis:
> I successfully recovered all the missing wallets from my own backups. Fortunately, I wasn’t traveling during the update, because then I wouldn’t have had access to my backups. That would have been a nightmare.

If you don't mind, please try the recovery tool to see if your old configurations are preserved there.

[...]

PR #3307, opened 2025-10-09
fix: exclude our storage from iCloud backup

shubhamkmr04:
# Description

Relates to issue: https://github.com/ZeusLN/zeus/issues/2915

This pull request is categorized as a:

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [X] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [X] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [X] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [X] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [ ] Yes
- [X] N/A

I have tested this PR on the following platforms (please specify OS version and phone model/VM):

- [ ] Android
- [ ] iOS

I have tested this PR with the following types of nodes (please specify node version and API version where appropriate):

- [ ] Embedded LND
- [ ] LND (REST)
- [ ] LND (Lightning Node Connect)
- [ ] Core Lightning (CLNRest)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales
- [ ] I’ve added new locale text that requires translations
- [ ] I’m aware that new translations should be made on the ZEUS [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

### Third Party Dependencies and Packages

- [ ] Contributors will need to run `yarn` after this PR is merged in
- [ ] 3rd party dependencies have been modified:
    * verify that `package.json` and `yarn.lock` have been properly updated
    * verify that dependencies are installed for both iOS and Android platforms

### Other:

- [ ] Changes were made that require an update to the README
- [ ] Changes were made that require an update to onboarding

PR #4385, opened 2026-08-11
fix: stop syncing keychain data to iCloud, add consent-gated legacy purge

kaloudis:
# Description

This PR remediates KEY-004 from the 2026-08-01 security audit (wallet seeds persisting in the iCloud-synced keychain) and, in the process, fixes the root cause of the entire iCloud keychain saga: **all app keychain data, including every wallet seed, has been syncing to iCloud Keychain on iOS this whole time.**

## Root cause

react-native-keychain 10.0.0 (latest) checks the `cloudSync` option for *presence*, not value:

```objc
CFBooleanRef cloudSyncValue(NSDictionary *options)
{
  if (options && options[@"cloudSync"]) {
    return kCFBooleanTrue;
  }
  return kCFBooleanFalse;
}
```

The JS boolean `false` crosses the bridge as a non-nil `@NO` NSNumber, which is truthy as an Objective-C pointer. Every call in `storage/index.ts` passes `{ cloudSync: false }`, so every `zeus:*` write, including `zeus:zeus-settings-v2` with seeds, macaroons, and PINs, actually set `kSecAttrSynchronizable = true`. The 2025-2026 "exclude our storage from iCloud" work (#3307, #3371, the `zeus:` prefix migration) never achieved its goal, and this also explains the stale-wallet-resurrection reports on new devices: the "local-only" entries themselves were syncing.

No upstream fix exists (10.0.0 is the latest release; the presence check dates to the feature's introduction in v9.1.0). An upstream issue will be filed and linked here.

## The fix, in three ordered parts

1. **Patch the library** (`patches/patch-keychain-cloudsync.mjs`): `cloudSyncValue()` now respects the boolean. The patch is security-load-bearing, so it throws at postinstall if the pattern ever stops matching (a dependency bump cannot silently reintroduce the leak).

2. **Non-destructive desync migration** (`keychainDesyncMigration`, flag `keychain-desync-v1`): with the patch, Storage reads the device-local partition, which is empty for existing users. This one-shot iOS migration copies every synchronizable `zeus:*` item into the local partition with equality read-back verification. It deletes nothing. Ordering is load-bearing: it must run before `keychainCloudSyncMigration`, otherwise `migrateKey`'s idempotency guard misses on the empty local partition and stale pre-2026 unprefixed data would permanently shadow the fresher copy (pinned by comment and test).

3. **Automatic one-shot purge** (the KEY-004 payoff): `autoPurgeLegacyKeychain` runs from Wallet boot on iOS once the desync migration's flag proves every `zeus:*` entry has a device-local copy, plus a Tools > Keychain Cleanup screen as the manual fallback and transparency surface. Deletes all synchronizable entries (`zeus:*` and unprefixed legacy), unprefixed legacy local entries, and the legacy EncryptedStorage settings blob. On iOS candidates come from a new native enumeration module (`ios/KeychainAudit.m`, `SecItemCopyMatching` over internet passwords, servers only, no values over the bridge); on Android from a reconstructed key catalog with a source-level drift-guard test against `keychainCloudSyncMigration`'s list.

## Why this touches the no-deletion policy, and how

Deletion was disabled in `c6323fff4` after #3625: the first device to migrate deleted the shared iCloud entry and wiped the user's other devices. That policy is preserved where it matters:

- **No migration deletes anything.** `deleteFromOldKeychain` remains a no-op; the desync migration is copy-only.
- The purge **is now automatic, and that is deliberate** (maintainer decision, see the design-change comment below): the iCloud sync was an unintended side effect, never an advertised feature, and its only unique value (accidental wallet transfer between devices) never worked for node-bound wallets anyway, since embedded LND and LDK Node require seed restore on a new device regardless and Cashu proofs live in the CDK database. Consent-gating meant every user who dismissed the offer kept seeds in iCloud indefinitely. Deleting synchronizable items still propagates through iCloud Keychain, so a device that has not yet run the desync migration loses its copies, whether it is still on a pre-desync build or auto-updated in the background without being opened yet (desync only runs when the app is opened, and the purge can fire on the same boot desync completes). A grace period between desync and purge was considered and rejected: it would keep seeds in iCloud longer for every user, and since whichever device opens first purges on that boot, no warning or instruction gives users an ordering they could act on. That residual #3625-mode risk is accepted without user-facing ceremony; the release notes disclose the outcome and the recovery path (the purging device retains local copies of everything that was in the sync partition, so restore from seed or a wallet-config export). The purge only runs after the desync flag is set, and the done flag (`keychain-autopurge-v1`) is set only on a zero-failure pass, so partial passes retry on later boots. The manual Tools screen keeps a standard destructive-action confirm; its device-attestation checkbox and warning explainers were dropped for the same reason, since the automatic purge performs the identical deletion without asking. Protection in the other direction is structural, not policy: a device that has desynced reads only the device-local partition, so nothing an out-of-date device writes or deletes in the sync partition can affect it, and the only-copy guard keeps the purge from deleting entries an old device newly created.
- **Preflight before any deletion**: the live local settings blob must exist and parse (this also makes the purge a no-op after Clear All Data), and a synchronizable `zeus:*` entry is never deleted while it is the only copy (a missing local counterpart is copied and verified first). Deletion order is stale-first, `zeus:*` sync entries last, and the purge is re-runnable, so a crash mid-purge just means running it again.
- The Keychain Recovery tool still ships unchanged for anyone who never purges. `copyToLegacyLocations` (the last writer of unprefixed entries) is now internally dev-gated so release builds cannot resurrect purged data.

Relationship to #4328: the native enumeration satisfies the iOS "delete-all-in-access-group" residual from its known-residuals list, with no dependency on that PR (the preflight independently no-ops after wipes).

## Migration plan (cohorts)

| Cohort | Outcome |
|---|---|
| Fresh install (iOS) | Desync migration finds nothing, flags; all writes land device-local from day one |
| Migrated single-device iOS | `zeus:*` copied sync-to-local, verified; auto-purge can run on the same Wallet boot desync completes; post-purge iCloud is clean |
| Multi-device, another device not yet updated, or updated but never opened | The first device to open desyncs then auto-purges, possibly on that same boot; propagation deletes the other device's synced entries, so it opens to onboarding and wallets there must be restored from seed or re-entered credentials (the purging device retains local copies of everything that was in the sync partition). Accepted tradeoff, no grace period by design, disclosed in release notes |
| Multi-device, both updated and opened at least once | Purge on either device is safe for the other: opening ran its desync, so its data lives in the local partition. An out-of-date device also cannot harm an upgraded one in the reverse direction (upgraded devices no longer read the sync partition outside explicit recovery fallbacks) |
| Pre-2026 user upgrading directly to this build | No `zeus:*` sync entries exist; desync no-ops, then the legacy migration populates the local partition correctly |
| New device via iCloud Keychain sync only | First-run AppDelegate wipe clears only the non-sync partition and the EncryptedStorage flags, so the desync migration re-runs and rescues whatever is still in the sync partition to local before the auto-purge clears it; if another device already purged, the new device starts clean. iCloud auto-restore of wallets ends with this PR |
| Encrypted-backup restore | Flags restore with the data; everything skips; stable |
| Android | No partition split; patch, desync, and auto-purge are iOS-only; the manual Tools purge removes catalog-derived unprefixed legacy entries |
| Crash mid-desync / mid-purge | Flag is only set after a full pass / stale-first order plus re-runnability; both converge |
| Purge, then old build reinstalled | The old build reads the now-empty sync partition and starts fresh; data returns on re-upgrade (local partition untouched) |

## Revert story

Parts 1 and 2 are revert-safe: reverting the patch returns reads to the synchronizable partition where the original data still lives untouched (local copies go dormant), and the desync migration is purely additive. Part 3's deletions are **not revertible**; the counter-measures are that the purge runs only after the desync flag proves verified local copies exist, the preflight re-verifies the live blob and copies any missing counterpart before anything is deleted, only-copy and migrated-counterpart guards refuse unsafe deletions, and the recovery tool still ships.

## Registration points

No new persisted data keys: only two EncryptedStorage migration flags (`keychain-desync-v1`, `keychain-autopurge-v1`), consistent with flag storage convention. No additions needed to `STORAGE_KEYS`, `OTHER_KEYS`, or `migrationKeys`.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

New suites: `utils/KeychainPurgeUtils.test.ts` (19 tests: scan on both platforms, preflight refusals and catch-up copy, deletion ordering, only-copy guard, failure aggregation, auto-purge gating and retry semantics, migration-list drift guard) and a `keychainDesyncMigration` describe in `utils/MigrationUtils.test.ts` (7 tests: platform and flag gates, copy-if-missing, unprefixed skip, empty-value skip, verify-failure rethrow, flag-once).

Manual device matrix (pending, will be filled in before undrafting):

- [ ] Fresh install (iOS simulator)
- [ ] Upgrade from current release build with wallets, notes, and Cashu data: all data intact on first boot, `keychain-desync-v1` set, keychain dump shows `zeus:*` in the local partition
- [ ] Pre-2026 legacy blob upgrade (seeded via dev `copyToLegacyLocations`)
- [ ] Auto-purge on the same boot desync completes: keychain dump shows the sync partition empty, `keychain-autopurge-v1` set; recovery scanner finds nothing afterwards
- [ ] Delete Cashu data with a synchronizable seed twin seeded: the twin is cleared and the next `initializeCDK` generates a fresh seed instead of re-adopting the deleted one
- [ ] Manual Tools purge on a device seeded with safety-skipped entries
- [ ] Purge refusal after Clear All Data
- [ ] Crash mid-purge, then re-run converges
- [ ] Multi-device iCloud (real devices, same Apple ID): no purge, both fine; purge with both updated and opened, other device unaffected; destructive rehearsal (second device updated but never opened) confirming the documented outcome and recovery path
- [ ] Android: catalog purge removes seeded legacy entries
- [ ] iOS keychain attribute dump confirming `kSecAttrSynchronizable` is false for new writes with the patch

- [ ] Android
- [ ] iOS

Pending device tests (see the manual device matrix above).

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Storage-layer change; applies to all backends. Device matrix above.

## Locale additions

15 new `views.Tools.keychainCleanup.*` strings in `locales/en.json` only, per locale contribution policy.

### Locales

### Other:


shubhamkmr04 (commented):
Read the patch, the migration and the purge against react-native-keychain 10.0.0's source. The root cause checks out: `cloudSyncValue()` at `RNKeychainManager.m:138` matches the patch's expected pattern exactly, and it feeds `kSecAttrSynchronizable` on every query, insert and delete path.

Three things that could have broken this and don't:

- **Blast radius is contained.** Every other `cloudSync` call site in the app passes `true` explicitly or omits the option, and omitted was already false pre-patch. Only the three sites in `storage/index.ts` change meaning.
- **The desync copy can't destroy its source.** `setInternetCredentialsForServer` deletes before inserting, but that delete is scoped by `kSecAttrSynchronizable`, so a local write leaves the synchronizable copy intact.
- **`setRawLocalItem` is attribute-compatible with `Storage.setItem`** — same `kSecAttrAccessibleAfterFirstUnlock` default, same username-as-server convention. The native enumeration is safely scoped too: no `keychain-access-groups` entitlement, so it only sees ZEUS's own items.

Findings inline. The first two are a pair and together they're a data-loss path — I'd want those resolved before merge. The third matters before the purge reaches users; the rest are cleanups.

storage/index.ts
```diff
@@ -1,6 +1,71 @@
...
+/**
+ * Lists the servers of all keychain internet-password items in one
+ * partition (synchronizable or not) via the KeychainAudit native module.
+ * iOS only; resolves to [] elsewhere (Android has no partition split).
+ */
+export const getInternetPasswordServers = async (
+    synchronizable: boolean
+): Promise<string[]> => {
+    if (Platform.OS !== 'ios') return [];
+    return NativeModules.KeychainAudit.getInternetPasswordServers(
```
shubhamkmr04: `NativeModules.KeychainAudit` is dereferenced unguarded, and `KeychainAudit.m` rejects on any `OSStatus` other than `errSecSuccess`/`errSecItemNotFound`. Both surface as a throw inside `keychainDesyncMigration` — see the note there for why that's expensive.

utils/MigrationUtils.ts
```diff
@@ -1279,6 +1286,71 @@ class MigrationsUtils {
...
+
+            if (!settingsStore.isMigrating) settingsStore.isMigrating = true;
+
+            await setRawLocalItem(server, syncValue);
+
+            const verify = await getRawItem(server, false);
+            if (verify !== syncValue) {
+                // Rethrow so the flag stays unset and the migration retries
+                // on next boot instead of silently skipping this key forever
+                throw new Error(
```
shubhamkmr04: This is the only migration in `getSettings` that propagates errors. `keychainCloudSyncMigration` wraps its whole body in try/catch; this one throws, `getSettings` catches at `SettingsStore.ts:2008`, logs, and returns in-memory defaults **without ever reading the blob**. `Wallet.tsx:512` then falls through to `IntroSplash`.

So a user with wallets is shown onboarding. If they create a wallet from there, that write populates `zeus:zeus-settings-v2` in the local partition — and since this migration is copy-if-missing, the next boot skips the key and their real blob in the synchronizable partition is shadowed for good. The retry this throw is meant to buy has nothing left to do.

Swallowing the error alone doesn't fix it, because an empty local partition looks identical to a failed desync. The blob read needs a fallback for that boot:

```ts
let modernSettings: any = await Storage.getItem(STORAGE_KEY);
if (!modernSettings && Platform.OS === 'ios') {
    // desync did not complete; the blob may still only live in the
    // synchronizable partition. Read it, don't write it.
    modernSettings = await getRawItem(`${KEY_PREFIX}${STORAGE_KEY}`, true);
}
```

utils/KeychainRecoveryUtils.ts
```diff
@@ -410,6 +410,18 @@ class KeychainRecoveryUtils {
...
     }> {
+        // Development seeding helper only. Guarded internally (not just at
+        // the UI) because it is the last remaining writer of unprefixed
+        // keychain entries: in a release build it could resurrect exactly
+        // the data the keychain cleanup purge deletes.
+        if (!__DEV__) {
+            return {
+                success: false,
+                locations: [],
+                error: 'copyToLegacyLocations is only available in dev builds'
```
shubhamkmr04, line 459: Worth extending the source list while you're in this file. `getStorageSources()` has four readers — `current`, `unprefixed-local`, `unprefixed-cloud`, `encrypted-storage` — and `current` is `Storage.getItem`, which post-patch reads device-local only. There's no prefixed-cloud reader.

Pre-patch, `current` *was* the synchronizable partition. Post-patch, an install created after the Jan 2026 prefix migration has no unprefixed ancestors and nothing in EncryptedStorage, so if desync doesn't complete all four sources come back empty while `zeus:zeus-settings-v2` sits in the synchronizable partition holding their seeds. The tool reports nothing to recover for exactly the users it exists for — and it's the safety net for the `MigrationUtils` issue above.

```ts
private async readFromPrefixedCloud(key: string): Promise<string | null> {
    if (Platform.OS !== 'ios') return null;
    try {
        return await getRawItem(`${KEY_PREFIX}${key}`, true);
    } catch (e) {
        console.warn(`[Recovery] Error reading prefixed cloud ${key}:`, e);
        return null;
    }
}
```

utils/KeychainPurgeUtils.ts
```diff
@@ -0,0 +1,355 @@
...
+        const settings = JSON.parse(settingsJson);
+        if (!settings || typeof settings !== 'object') {
+            return { ok: false, reason: 'invalid-settings' };
+        }
+    } catch {
+        return { ok: false, reason: 'invalid-settings' };
+    }
+
+    for (const server of scan.syncServers) {
+        if (!server.startsWith(KEY_PREFIX)) continue;
```
shubhamkmr04, line 274: The only-copy guard skips everything unprefixed, but `executePurge` deletes every `legacyLocalServers` entry — on iOS, every non-`zeus:` local internet password — with no check that a migrated counterpart exists.

That assumes `keychainCloudSyncMigration` got everything, which isn't guaranteed: it covers a fixed list plus notes/LNC/embedded-lnd Cashu, and its Cashu branch is skipped outright when `hasMigrated === 'true'` from a prior launch. The UI copy promises "your active wallet data is not affected" — worth enforcing rather than assuming, in the `legacyLocalServers` loop:

```ts
if ((await getRawItem(`${KEY_PREFIX}${server}`, false)) === null) {
    // no migrated counterpart — this is the only copy
    continue;
}
```

utils/KeychainPurgeUtils.ts
```diff
@@ -0,0 +1,355 @@
...
+ * scan finds nothing to clean, silently consuming the offer.
+ */
+export const shouldOfferKeychainPurge = async (): Promise<boolean> => {
+    if (Platform.OS !== 'ios') return false;
+    try {
+        const desyncDone = await EncryptedStorage.getItem(KEYCHAIN_DESYNC_KEY);
+        if (desyncDone !== 'true') return false;
+        const offered = await EncryptedStorage.getItem(PURGE_OFFER_KEY);
+        if (offered === 'true') return false;
+        await EncryptedStorage.setItem(PURGE_OFFER_KEY, 'true');
```
shubhamkmr04: The offer flag is written before the scan runs, so if `scanPurgeCandidates` throws, the catch returns false with the one-time prompt already burned — for a user who does have candidates. Moving this below the scan still lets the nothing-found case consume it.

utils/KeychainPurgeUtils.ts
```diff
@@ -0,0 +1,355 @@
...
+    for (const server of unprefixedSync) {
+        await remove(server, true);
+    }
+
+    for (const server of prefixedSync) {
+        // Last line of defense: never delete the only copy of live data
+        const local = await getRawItem(server, false);
+        const syncValue = await getRawItem(server, true);
+        if (local === null && syncValue) {
+            failures.push(server);
```
shubhamkmr04: Two small accounting things: `resetInternetCredentials` succeeds on a missing item, so `deleted` is really an attempt count; and the only-copy refusal lands in `failures`, which renders as "finished with some failures, you can run it again" — but a re-run refuses identically, forever. A third bucket for "skipped for safety" would read truer.

utils/KeychainPurgeUtils.test.ts
```diff
@@ -0,0 +1,405 @@
...
+            expect(migrationIdentifiers.length).toBeGreaterThanOrEqual(19);
+
+            const staticCatalogMatch = purgeSource.match(
+                /const STATIC_LEGACY_KEYS = \[([\s\S]*?)\]/
+            );
+            expect(staticCatalogMatch).toBeTruthy();
+            const catalogBlock = staticCatalogMatch![1];
+
+            for (const identifier of migrationIdentifiers) {
+                expect(catalogBlock).toContain(identifier);
```
shubhamkmr04: `toContain` is a substring match, so `SWAPS_KEY` is satisfied by `REVERSE_SWAPS_KEY` — drop `SWAPS_KEY` from the catalog and this guard still passes. A word-boundary regex closes it. Also worth noting the guard only reads the `migrationKeys` array; `STORAGE_KEY` and `NOTES_KEY` are migrated outside it and aren't covered.

storage/index.ts
kaloudis: Fixed in 372b9e8a9. The deref is now guarded with a descriptive error, kept as a throw rather than an empty resolve: callers set one-shot migration flags on an empty enumeration, so a silent [] would mark the desync complete on work that never happened. The throw is now contained by the migration itself (see the other thread).

utils/MigrationUtils.ts
kaloudis: Good catch, this was the most important finding of the round. Fixed in 372b9e8a9 with both halves: keychainDesyncMigration now catches everything without setting keychain-desync-v1 (so it retries next boot), and getSettings falls back to a read-only fetch of the synchronizable blob when the local read misses on iOS, essentially your snippet. A wallet-holding user can no longer be routed to onboarding by a transient keychain error, and the shadowing sequence is closed off. Regression tests added in MigrationUtils.test.ts and SettingsStore.test.ts, including one asserting the fallback never writes the local partition.

utils/KeychainRecoveryUtils.ts
kaloudis: Added in 728e6f0ca as a prefixed-cloud source, with one refinement over the snippet: the scanner only surfaces it for a key when the current read misses. After a completed desync every zeus:* sync entry still exists until the purge runs, and listing them all would both bury the tool in noise and offer a stale copy as a restore source over fresher local data. With the suppression it lights up for exactly the failed-desync case this exists for.

utils/KeychainPurgeUtils.ts
kaloudis: Fixed in 9f28126cb. Both unprefixed loops (local and synchronizable) now require the migrated counterpart to exist before deleting, and refusals land in a new skipped bucket rather than failures. One mapping wrinkle your snippet would have hit: the legacy v1 settings blob has no zeus:zeus-settings counterpart, its migrated successor is the v2 blob, so the counterpart lookup special-cases that rename. Covered by new tests including the v1-to-v2 mapping.

utils/KeychainPurgeUtils.ts
kaloudis: Fixed in 9f28126cb: the flag is now set right after a successful scan, before the candidate-count check, so nothing-found still consumes the offer but a scan that throws leaves it unset and the next boot retries. Test added.

utils/KeychainPurgeUtils.ts
kaloudis: Both adopted in 9f28126cb: remove() now probes existence first so deleted counts real removals, and only-copy refusals go to a new skipped bucket with their own line in the cleanup screen ('entries kept because no migrated copy exists'). One small correction on the forever claim: the Tools screen re-runs verifyPurgePreflight before every purge, and for zeus:* sync entries that preflight copies the missing local counterpart, so a refusal there only recurs if the copy keeps failing. The unprefixed skips from the other thread genuinely do persist, which is exactly why they get the honest third bucket.

utils/KeychainPurgeUtils.test.ts
kaloudis: Fixed in 9f28126cb: word-boundary regex per identifier (underscore is a word character, so \bSWAPS_KEY\b no longer matches inside REVERSE_SWAPS_KEY), plus explicit assertions for STORAGE_KEY and NOTES_KEY, which the migration handles outside the migrationKeys array.

shubhamkmr04 (commented):
Re-checked the three fixes against the source: all correct. The recovery gate is per-key with `current` ordered first, the purge counterpart check fails toward keeping data, and the settings fallback handles `false`-on-miss plus the post-wipe and post-purge cases. 84/84 green across the three suites, no tsc errors in any touched file.

Two small things left:

utils/MigrationUtils.ts
```diff
@@ -1279,6 +1286,83 @@ class MigrationsUtils {
...
+
+                await setRawLocalItem(server, syncValue);
+
+                const verify = await getRawItem(server, false);
+                if (verify !== syncValue) {
+                    // Throw so the flag stays unset and the migration retries
+                    // on next boot instead of silently skipping this key
+                    // forever
+                    throw new Error(
+                        `Keychain desync migration verify failed for ${server}`
```
shubhamkmr04: The throw is inside the loop but the catch is outside it, so the first bad key abandons every remaining one — and `SecItemCopyMatching` returns servers unordered, so which keys get copied is arbitrary.

Only `STORAGE_KEY` has a read fallback. If contacts, notes or the LNC credentials land on the uncopied side, a store writing that key on the same boot shadows the real copy for good under copy-if-missing.

A per-key catch copies the maximum each pass and lets the retry converge instead of re-failing at the same arbitrary point:

```ts
let failed = false;
for (const server of syncServers) {
    try {
        // ... copy + verify ...
    } catch (error) {
        console.error(`Keychain desync failed for ${server}`, error);
        failed = true;
    }
}
if (!failed) await EncryptedStorage.setItem(KEYCHAIN_DESYNC_KEY, 'true');
```

stores/SettingsStore.ts
```diff
@@ -1939,10 +1940,39 @@ export default class SettingsStore {
...
             await MigrationsUtils.purgeRescueKeyFiles();
 
-            const modernSettings: any = await Storage.getItem(STORAGE_KEY);
+            let modernSettings: any = await Storage.getItem(STORAGE_KEY);
+
+            if (!modernSettings && Platform.OS === 'ios') {
+                // An empty local partition is indistinguishable from a failed
+                // desync migration: the blob may still live only in the
+                // synchronizable partition. Read it there for this boot, but
+                // never write it back; all writes stay device-local, and the
```
shubhamkmr04: "never write it back" doesn't quite hold — the five migrations below all call `setSettings` when `changed`, which writes the blob to the local partition on this boot. The outcome is benign (same content desync would have produced), but the comment reads as a guarantee.

utils/MigrationUtils.ts
kaloudis: Fixed in 17cc0fab8, essentially your shape: per-key catch, and keychain-desync-v1 is only set when zero keys failed, so each pass copies the maximum and the retry converges on just the failed keys. Agreed this was underweighted as a nit; it is the same shadowing class as the round-1 blocker, just for the keys without a read fallback. Tests added: a failing first key no longer blocks the keys after it (flag held), and a second pass copies only the previously failed key and then flags. The verify-mismatch test was updated to the per-key error shape.

stores/SettingsStore.ts
kaloudis: Reworded in 17cc0fab8. You're right that the retry clause overstated things: any of the five migrations below can persist the blob device-local via setSettings on this same boot, so the desync retry may find it already present. The comment now states that explicitly and why it is benign (the persisted content derives from the same synchronizable read, so copy-if-missing skipping it loses nothing), and keeps the one real guarantee: nothing is ever written back to the synchronizable partition.

kaloudis:
Pushed three additions from a self-review pass focused on what existing users hit on their first boot after updating:

- 083e305d1: `initializeCDK`'s remote-node path generated and persisted a fresh mnemonic whenever the stored cashu seed read missed. If the desync migration had failed for exactly that key, the new seed would permanently shadow the real one under copy-if-missing. It now checks the synchronizable partition before generating, mirroring the settings blob fallback in `getSettings`. Embedded and LDK nodes were never at risk here since their cashu seed re-derives deterministically from the wallet seed.
- 2fac7141c: `UnitsStore` reads its pref at store construction, which on iOS precedes the desync migration on the first post-update boot, so units showed defaults until the next restart. It now reloads alongside contacts and notes on wallet connect.
- 65dea05cc: the one-shot cleanup offer dialog read as routine housekeeping, so cautious users would dismiss it forever. The copy now names the cause (seeds kept in iCloud Keychain) and the update-all-devices precondition, and explainer2 points at the wallet configuration export as the concrete backup step before purging.

Typescript, eslint, prettier clean; the three keychain suites (86 tests) and the adjacent store suites (220 tests) pass.

kaloudis:
Design change in ed2136d3e, the one-time cleanup offer dialog on Wallet is gone, and the purge now runs automatically.

Rationale: the iCloud sync was an unintended side effect of the cloudSync bug, never an advertised feature. Its only unique value was accidental wallet transfer between devices, which never actually worked for node-bound wallets (embedded LND and LDK Node channel state lives on disk and requires seed restore on a new device regardless; Cashu proofs live in the CDK database and do not transfer either; remote connections are re-enterable credentials). Meanwhile, consent-gating the deletion meant every user who dismissed the dialog kept their seeds in iCloud indefinitely, which is the KEY-004 exposure this PR exists to end.

What changed:

- `shouldOfferKeychainPurge` and the `Alert.alert` offer are replaced by `autoPurgeLegacyKeychain` in `KeychainPurgeUtils.ts`, fired fire-and-forget from the same Wallet boot site (iOS only, after the `keychain-desync-v1` flag proves every zeus:* entry has a device-local copy).
- All safety properties from the consent-gated design are kept: preflight requires a parseable live settings blob, a synchronizable zeus:* entry is never deleted while it is the only copy, unprefixed entries are never deleted without their migrated counterpart, and the `keychain-autopurge-v1` done flag is set only on a zero-failure pass so partial passes retry on later boots. Safety-skipped entries do not block the flag.
- Tools > Keychain Cleanup stays as the manual fallback (Android legacy entries, safety-skipped items) and transparency surface.
- Test suite updated: 8 cases covering platform gating, desync ordering, flag idempotency, scan failure retry, preflight deferral, clean pass, and failure retry.

Accepted tradeoff, stated explicitly: deleting synchronizable entries propagates across the Apple ID, so a device still running a pre-desync build loses its copies, and iCloud auto-restore of wallet data on new devices ends. Both were already true of the manual purge; the difference is the user no longer attests that all their devices are updated first. The release notes will carry both behavior changes.

shubhamkmr04 (commented):
Round-2 fixes verified: per-key containment converges as described, and the reworded fallback comment is accurate. 89/89 green across the three suites, tsc clean on touched files.

I'm treating auto-purge as the settled policy decision described above. The two inline notes are about whether the implementation meets its own stated guarantees.

views/Wallet/Wallet.tsx
```diff
@@ -508,6 +509,9 @@ export default class Wallet extends React.Component<WalletProps, WalletState> {
                 if (shareIntentData) {
                     this.setState({ pendingShareIntent: shareIntentData });
                 }
+                // Fire-and-forget: one-shot removal of legacy and
+                // iCloud-synced keychain copies of wallet data
+                autoPurgeLegacyKeychain();
```
shubhamkmr04, line 514: This runs on the same boot as `keychainDesyncMigration` (`getSettings()` sets the flag a few lines up), so the cohort table's "auto-purge runs on a subsequent Wallet boot" doesn't hold.

That breaks the "Multi-device, both updated → safe for the other" row. A second device that auto-updated but hasn't been opened is still pre-desync: its only copy is the shared synchronizable entry, and this purge deletes it through iCloud before that device ever runs desync. It opens to onboarding.

That's the common two-device case, not the out-of-date one the release notes cover. Deferring the purge (say, a few days after the desync flag is first set) would give other devices a window to open. At minimum the notes should say to *open* ZEUS on every device, not just update it.

stores/CashuStore.ts
```diff
@@ -550,6 +559,61 @@ export default class CashuStore {
...
     };
 
+    /**
+     * Last-resort read of the cashu seed from the synchronizable keychain
+     * partition, for the case where keychainDesyncMigration failed for this
+     * key (oblador/react-native-keychain#800; ZeusLN/zeus#4403). Adopts and
+     * persists the recovered seed device-locally so subsequent boots read it
+     * through Storage.
+     */
+    private recoverSeedFromSyncPartition = async (): Promise<string | null> => {
```
shubhamkmr04, line 569: This can bring back a seed the user deliberately deleted. `deleteCashuData` removes `-cashu-seed-phrase` via `Storage.removeItem`, which post-patch only clears the local partition. If the auto-purge hasn't finished a clean pass, the synchronizable twin is still there, and the next `initializeCDK` on a remote node reads it back instead of generating a fresh seed.

Clearing the sync twin alongside the local removals in `deleteCashuData` closes it, matching the both-variants rule `clearKey` already follows. This path has no test yet either.

views/Wallet/Wallet.tsx
kaloudis: Confirmed: the purge can run on the very boot desync completes, and the cohort table claimed otherwise. The table is corrected, and after weighing the deferral we're staying with no grace period. It would keep seeds in iCloud longer for every user, and since whichever device opens first purges on that boot, no warning or "open every device first" instruction gives users an ordering they can actually act on. The accepted outcome is now stated plainly in the PR body: a second device that updated but was never opened loses its sync copies and opens to onboarding; the purging device retains local copies of everything that was in the sync partition, so recovery is restore from seed or a wallet-config export. The release notes will disclose the outcome and recovery path rather than prescribe an ordering.

Followed the same logic through to the manual screen in ac1329bf9: the device-attestation checkbox and warning explainers are gone, since they gated nothing the automatic path does not already do. The destructive-action confirm (propagation plus irreversibility) stays.

stores/CashuStore.ts
kaloudis: Good catch, and an interaction between two of this PR's own additions. Fixed in 90623f699: deleteCashuData now clears the synchronizable twins (seed-phrase, seed-version, and the legacy seed key) before the local removals, so a failure leaves the local seed intact and the deletion retryable rather than resurrectable, matching clearKey's both-variants rule.

On the missing test: a unit test for this path needs CashuStore's full module graph mocked, so for now the device matrix gained a line exercising the resurrection scenario end to end (seeded sync twin, delete Cashu data, next initializeCDK generates fresh). Happy to add a store-level jest suite if you'd rather see it covered there.

shubhamkmr04 (approved):
ACK https://github.com/ZeusLN/zeus/pull/4385/changes/ac1329bf9859543de10d194bed350c309e93926c Tested on iOS Simulator and Android emulator: upgrade from master, fresh install, Keychain Cleanup, Cashu data deletion and Clear All Data all pass. Multi-device iCloud not tested.

PR #4390, opened 2026-08-11
fix: migrate CLN payments from deprecated pay/keysend to xpay/xkeysend

### LINKED ISSUE #4388: CLN's `pay` is deprecated
I run my node (CLN v26.06.6) with `allow-deprecated-apis=false` to see if anything is using deprcated stuff and i noticed that Zeus is still using `pay`. `pay` is deprecated since v26.06 and will be removed in v27.03: https://github.com/ElementsProject/lightning/blob/master/doc/developers-guide/deprecated-features.md

`xpay` will be the only remaining pay command. `xpay` has been in CLN since v24.11. Please switch to `xpay`: https://docs.corelightning.org/reference/xpay

kaloudis:
# Description

Relates to issue: #4388

CLN deprecated the `pay` command in v26.06 and will remove it in v27.03. As #4388 reports, ZEUS's CLNRest backend still pays through `POST /v1/pay`, which already fails on nodes running with `allow-deprecated-apis=false`. While auditing every clnrest endpoint ZEUS calls against CLN's [deprecated features list](https://github.com/ElementsProject/lightning/blob/master/doc/developers-guide/deprecated-features.md), I found `keysend` is on the same deprecation clock (v26.06 deprecated, v27.03 removed, replaced by `xkeysend`). No other endpoint ZEUS uses is deprecated.

This PR migrates both send paths, version-gated so older nodes keep working:

- `payLightningInvoice` uses `POST /v1/xpay` when the node is v24.11+ (when `xpay` was introduced), otherwise falls back to the existing `pay` request unchanged
- `sendKeysend` uses `POST /v1/xkeysend` when the node is v26.06+ (when `xkeysend` was introduced), otherwise falls back to the existing `keysend` request unchanged

The gates read node version through the existing `CLNRest.supports()` helper. If a payment is somehow dispatched before `getinfo` resolves, the gate fails toward the legacy path, which remains functional until v27.03.

The migration is not a URL swap; two API differences are handled:

**1. `xpay`/`xkeysend` take an absolute `maxfee` (msat), not `maxfeepercent`.** When the user picks a percent cap, the new `getMaxFeeMsat` helper converts: percent x amount, using the caller-supplied amount when present (zero-amount invoices, keysend) and otherwise decoding the invoice via `/v1/decode` to learn it (works for both bolt11 and bolt12 invoices). Fixed caps and silent sends (donations) pass `fee_limit_sat` instead, which maps directly. If no cap can be derived, `maxfee` is omitted and xpay's default applies (greater of 5000 msat or 1%).

**2. The response omits `status` and `payment_hash`** (xpay returns only `payment_preimage`, `failed_parts`, `successful_parts`, `amount_msat`, `amount_sent_msat`). `formatXpayResult` synthesizes both: `status: 'complete'` is safe because xpay only returns success (failures reject through the HTTP error path into `handlePaymentError`, same as `pay` today), and `payment_hash` is computed as SHA256 of the returned preimage. Without the hash, payment notes saved at send time would key on the preimage while the payments list (which reads `sendpays` via `/v1/sql`) keys on the hash, so notes would silently detach. Fee display is unaffected: `Payment.getFee` derives the fee from `amount_sent_msat - amount_msat`, which xpay provides.

A side benefit: `xpay` accepts bolt12 invoices natively via `invstring`, which is the path CLN offer payments already flow through (`fetchinvoice` then `sendPayment`).

**Update (second commit): fixed fee limits in sats for CLN.** Since `xpay`/`xkeysend` take an absolute `maxfee`, CLN no longer has to be percent-only. `supportsCustomFeeLimit()` is now enabled for CLNRest, so CLN gets the same dual fixed/percent fee limit UI as LND (both on the payment screens and for the default in Settings > Payments), and the CLN-only percent branches in `FeeLimit`, `PaymentRequest`, and `PaymentsSettings` are removed. Because CLN rejects requests carrying both `maxfee` and `maxfeepercent`, the views now only send `max_fee_percent` when the percent option is selected (via a new `onFeeOptionChange` callback on `FeeLimit`); otherwise the fixed `fee_limit_sat` maps to `maxfee`. On the legacy fallback paths: `pay` has supported `maxfee` since before clnrest existed so it is used unconditionally, while `keysend` only gained `maxfee` in v24.11, so on older nodes the fixed cap is converted to the equivalent percent of the keysend amount (capped at CLN's 100% parameter limit).

- [x] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Planned test matrix (draft until complete), on a CLN v26.06 node with `allow-deprecated-apis=false`:
- fixed-amount bolt11 pay (fee cap honors max fee percent)
- fixed-amount bolt11 pay with a fixed sat fee cap (new dual UI, `maxfee` mapping)
- zero-amount bolt11 pay with user-supplied amount
- bolt12 offer pay
- keysend (both percent and fixed sat caps)
- donation flow (silent send, `fee_limit_sat` mapping)
- payment note persists from send to payments list (synthesized `payment_hash`)
- default fee limit (fixed and percent) set from Settings > Payments carries into the payment screen

Plus one pre-v24.11 node to exercise the legacy fallback leg, including a fixed sat cap through legacy `pay` (`maxfee`) and legacy `keysend` (percent conversion).

### Locales

### Other:

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -410,8 +410,9 @@ export default class CLNRest {
     // xpay and xkeysend only accept an absolute fee cap (maxfee), not
     // maxfeepercent: derive one from the user's max fee percent, decoding
     // the invoice server-side when the caller didn't supply the amount,
-    // and fall back to fee_limit_sat (silent sends). Returning undefined
-    // leaves the cap to xpay's default of max(5000msat, 1%).
+    // and fall back to fee_limit_sat (fixed fee limit mode, silent sends).
+    // Returning undefined leaves the cap to xpay's default of
+    // max(5000msat, 1%).
     private getMaxFeeMsat = async (data: any, invstring?: string) => {
         if (data.max_fee_percent) {
             let amountMsat = data.amt ? Number(data.amt) * 1000 : undefined;
@@ -462,14 +463,21 @@ export default class CLNRest {
         // pay is deprecated in CLN v26.06 and removed in v27.03; use xpay
         // (available since v24.11) when the node has it
         if (!this.supports('v24.11')) {
+            const legacyRequest: any = {
+                bolt11: data.payment_request,
+                amount_msat: Number(data.amt && data.amt * 1000),
+                retry_for: data.timeout_seconds
+            };
+            // pay takes either a percent cap or an absolute maxfee, not
+            // both. maxfee predates clnrest itself, so no version gate.
+            if (data.max_fee_percent) {
+                legacyRequest.maxfeepercent = data.max_fee_percent;
+            } else if (data.fee_limit_sat) {
+                legacyRequest.maxfee = Number(data.fee_limit_sat) * 1000;
+            }
             return this.postRequest(
                 '/v1/pay',
-                {
-                    bolt11: data.payment_request,
-                    amount_msat: Number(data.amt && data.amt * 1000),
-                    maxfeepercent: data.max_fee_percent,
-                    retry_for: data.timeout_seconds
-                },
+                legacyRequest,
                 data.timeout_seconds * 1000
             );
         }
@@ -493,14 +501,32 @@ export default class CLNRest {
         // xkeysend (introduced alongside the deprecation in v26.06) when
         // the node has it
         if (!this.supports('v26.06')) {
+            const legacyRequest: any = {
+                destination: data.pubkey,
+                amount_msat: Number(data.amt && data.amt * 1000),
+                retry_for: data.timeout_seconds
+            };
+            // keysend takes either a percent cap or an absolute maxfee,
+            // not both, and only gained maxfee in v24.11: on older nodes
+            // express the fixed cap as the equivalent percent of the
+            // keysend amount (capped at CLN's 100% param limit)
+            if (data.max_fee_percent) {
+                legacyRequest.maxfeepercent = data.max_fee_percent;
+            } else if (data.fee_limit_sat) {
+                if (this.supports('v24.11')) {
+                    legacyRequest.maxfee = Number(data.fee_limit_sat) * 1000;
+                } else if (data.amt) {
+                    legacyRequest.maxfeepercent = BigNumber.min(
+                        new BigNumber(data.fee_limit_sat)
+                            .times(100)
+                            .dividedBy(data.amt),
+                        100
+                    ).toNumber();
+                }
+            }
             return this.postRequest(
                 '/v1/keysend',
-                {
-                    destination: data.pubkey,
-                    amount_msat: Number(data.amt && data.amt * 1000),
-                    maxfeepercent: data.max_fee_percent,
-                    retry_for: data.timeout_seconds
-                },
+                legacyRequest,
                 data.timeout_seconds * 1000
             );
         }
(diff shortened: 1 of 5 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (commented):
[`FeeLimit`](https://github.com/ZeusLN/zeus/blob/54941af6a7e4efb53acd0c84d318716f9893786c/components/FeeLimit.tsx#L58-L59) reads the `feeOption` prop only in `componentDidMount`, before its `await`, so it captures `'fixed'` — [`PaymentRequest`](https://github.com/ZeusLN/zeus/blob/54941af6a7e4efb53acd0c84d318716f9893786c/views/PaymentRequest.tsx#L170-L176) doesn't flip its own copy to `'percent'` until its own `componentDidMount`, which runs after the child mounts.

On CLN that means a bolt11 over 1000 sats with an untouched fee UI highlights the sat input and emits `fee_limit_sat: '100'`, while the payload still carries `max_fee_percent: '5.0'` — and [`getMaxFeeMsat`](https://github.com/ZeusLN/zeus/blob/54941af6a7e4efb53acd0c84d318716f9893786c/backends/CLNRest.ts#L416) prefers the percent. A 1,000,000 sat invoice shows a 100 sat cap and applies 50,000.

Not a wire regression (CLN was percent-only before), but the new fixed-cap option is inert unless the user taps the field.

Simplest fix: let `fee_limit_sat` be the source of truth, since `FeeLimit` already emits `percentAmount` as `fee_limit_sat` in percent mode:

```suggestion
    private getMaxFeeMsat = (data: any) =>
        data.fee_limit_sat ? Number(data.fee_limit_sat) * 1000 : undefined;
```

The cap then always matches what's on screen, and the `/v1/decode` round trip goes away with it — `max_fee_percent` stays only for the pre-v24.11 legacy `keysend` leg.

kaloudis:
Good catch, confirmed. The child captures `feeOption` before its `await` and never re-syncs it from props (`componentDidUpdate` only watches `satAmount`), while the parent's percent flip runs in its own `componentDidMount` after the child's, so the desync is permanent until the user taps a field. And since the accordion mounts collapsed children eagerly, it happens even with the fee UI never opened.

Took your suggestion in 9d41e9d94: `getMaxFeeMsat` now derives `maxfee` solely from `fee_limit_sat` (sync now, `/v1/decode` round trip gone), and I flipped the preference in the legacy `pay` leg too, since the same mismatch would have persisted on pre-v24.11 nodes. `max_fee_percent` is now only consulted in the legacy `keysend` leg, as you suggested.

One adjacent issue this surfaced: for no-amount invoices, `FeeLimit` received `customAmount`, which is the display-unit string, so with a fiat unit selected the percent-derived `percentAmount` was computed from the fiat number. LND already uses that value as its cap today, and this change would have extended that to CLN percent mode. Both `FeeLimit` sites in `PaymentRequest` now pass the converted `this.state.satAmount` (`Send` already did).

Side note: the over-1000-sats percent-mode default in `PaymentRequest` is now fully inert (it already was on LND, for the same capture reason). Left it in place for now; if we want that default back, the root fix is making `FeeLimit` honor `feeOption` prop updates, which stays consistent with this change since the wire now follows whatever the component emits.

tsc, eslint, prettier, and jest (1108) all green.

shubhamkmr04 (approved):
tACK

PR #4392, opened 2026-08-12
fix(nwc): scope list_transactions and lookup_invoice to connection activity

ajaysehwal:
# Description

Fixes a medium-severity NWC data-scoping vulnerability where read-only connections could export the full wallet history and look up arbitrary invoices by hash.

- Scope `list_transactions` to each connection’s own `activity` log for all permission tiers (removes the inverted branch that returned wallet-wide Lightning, on-chain, and Cashu history for read-only connections).
- Scope `lookup_invoice` to invoices/payments recorded in the calling connection’s activity (no more node-wide `BackendUtils.lookupInvoice` or payment-store scans).
- Remove the wallet-wide on-chain conversion path from `list_transactions` (read-only connections no longer receive on-chain history at all).
- Deduplicate Cashu `make_invoice` status refresh into `reconcilePendingCashuMakeInvoice`, shared by activity refresh and `lookup_invoice` polling.
- Restore preimage/description_hash on activity-scoped `lookup_invoice` via shape-tolerant invoice coercion and persisted `activity.preimage` after reconcile.

Connection expiry and missing-connection rejection (`withGlobalHandler` → `RESTRICTED` / `UNAUTHORIZED`, orphaned subscription cleanup) landed on master in #4360; this branch inherits that behavior and does not duplicate it.

## Security impact

**Before:** A read-only NWC connection string (`get_info`, `get_balance`, `lookup_invoice`, `list_transactions`) could:
- Receive the wallet’s complete Lightning payment/invoice history, all on-chain transactions (including destination addresses and raw tx hex), and full Cashu history via `list_transactions`.
- Resolve any payment hash or invoice created outside NWC via `lookup_invoice`.

**After:** Each connection only sees and can look up records it owns (its `activity` log). Read-only connections with no activity get an empty history and `NOT_FOUND` on foreign hashes.

## Test plan

- [x] `yarn test`
- [ ] Read-only `list_transactions` returns only connection activity and does not call wallet stores
- [ ] `lookup_invoice` returns `NOT_FOUND` for wallet-only hashes
- [ ] `lookup_invoice` succeeds for hashes in connection activity
- [ ] Expired/missing connections are rejected and unsubscribed
- [ ] Manual: connect a read-only NWC client and confirm it cannot enumerate wallet history or probe arbitrary invoice hashes
- [ ] Manual: confirm paying/invoice-creating connections still see their own activity via `list_transactions` and `lookup_invoice`

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -93,8 +93,6 @@ const SAVE_CONNECTIONS_DEBOUNCE_MS = 500;
 const PENDING_PAYMENT_STATUS_MIN_AGE_MS = 5_000;
 /** Per-connection debounce for pending pay_invoice status refresh in getActivities */
 const PENDING_PAYMENT_STATUS_REFRESH_MS = 30_000;
-/** Debounce for lookup_invoice outgoing-payment fetches (NWC client polling) */
-const LOOKUP_PAYMENTS_CACHE_MS = 5_000;
 
 export const DEFAULT_NOSTR_RELAYS = [
     'wss://relay.getalby.com/v1',
@@ -179,8 +177,6 @@ export default class NostrWalletConnectStore {
         string,
         number
     >();
-    private lookupPaymentsCache: Payment[] | null = null;
-    private lookupPaymentsFetchedAt = 0;
 
     settingsStore: SettingsStore;
     balanceStore: BalanceStore;
@@ -1115,52 +1111,9 @@ export default class NostrWalletConnectStore {
             )
         ) {
             await Promise.all(
-                pendingCashuInvoiceActivities.map(async (activity) => {
-                    const cashuInv = activity.invoice as
-                        | CashuInvoice
-                        | undefined;
-                    const quote = cashuInv?.quote;
-                    if (!quote || !cashuInv) return;
-
-                    try {
-                        const result = await this.cashuStore.checkInvoicePaid(
-                            quote,
-                            cashuInv.mintUrl,
-                            undefined,
-                            true
-                        );
-                        runInAction(() => {
-                            if (result.isPaid) {
-                                activity.status = 'success';
-                                const updated =
-                                    result.updatedInvoice ||
-                                    this.cashuStore.invoices?.find(
-                                        (inv) => inv.quote === quote
-                                    );
-                                if (updated) {
-                                    activity.invoice = updated;
-                                }
-                                const paidSats = Math.floor(
-                                    Number(result.amtSat) || 0
-                                );
-                                if (
-                                    paidSats > 0 &&
-                                    (!activity.satAmount ||
-                                        activity.satAmount <= 0)
-                                ) {
-                                    activity.satAmount = paidSats;
-                                }
-                            } else if (cashuInv.isExpired) {
-                                activity.status = 'expired';
-                            }
-                        });
-                    } catch (err) {
-                        console.error(
-                            'NWC: failed to refresh cashu make_invoice status:',
-                            err
-                        );
-                    }
-                })
+                pendingCashuInvoiceActivities.map((activity) =>
+                    this.reconcilePendingCashuMakeInvoice(activity)
+                )
             );
             this.lastPendingInvoiceStatusFetchByConnection.set(
                 connectionId,
@@ -1408,7 +1361,7 @@ export default class NostrWalletConnectStore {
             if (connection.hasPermission('lookup_invoice')) {
                 handler.lookupInvoice = (request: Nip47LookupInvoiceRequest) =>
                     this.withGlobalHandler(connection.id, () =>
-                        this.handleLookupInvoice(request)
+                        this.handleLookupInvoice(connection, request)
                     );
             }
 
@@ -1922,20 +1875,16 @@ export default class NostrWalletConnectStore {
     }
 
     private async handleLookupInvoice(
+        connection: NWCConnection,
         request: Nip47LookupInvoiceRequest
     ): NWCWalletServiceResponsePromise<Nip47Transaction> {
         try {
-            const result = await NostrConnectUtils.lookupInvoiceTransaction({
-                request,
-                isCashu: this.isCashuConfigured,
-                getCashuInvoices: () =>
-                    Promise.resolve(this.cashuStore.invoices || []),
-                getCashuPayments: () =>
-                    Promise.resolve(this.cashuStore.payments || []),
-                getLightningPayments: () => this.getPaymentsForLookup()
-            });
-
-            if (!result) {
+            const activity =
+                await NostrConnectUtils.findConnectionActivityByLookupRequest(
+                    connection.activity,
+                    request
+                );
+            if (!activity) {
                 return NostrConnectUtils.createNip47Error(
                     localeString(
                         'stores.NostrWalletConnectStore.error.invoiceNotFound'
@@ -1943,6 +1892,13 @@ export default class NostrWalletConnectStore {
                     Nip47ErrorCode.NOT_FOUND
                 );
             }
+
+            await this.refreshActivityForLookup(connection, activity);
+
+            const result =
+                NostrConnectUtils.convertConnectionActivityToNip47Transaction(
+                    activity
+                );
 
             return { result, error: undefined };
         } catch (error) {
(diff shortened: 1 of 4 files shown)


stores/NostrWalletConnectStore.ts
```diff
@@ -1881,99 +1843,138 @@ export default class NostrWalletConnectStore {
...
+            if (reconciled) {
+                this.scheduleSave();
             }
 
+            const nip47Transactions = connection.activity
+                .map((activity) =>
+                    convertConnectionActivityToNip47Transaction(activity)
+                )
+                .filter((tx) => tx.amount > 0)
+                .sort((a, b) => b.created_at - a.created_at);
```
kaloudis, line 2020: Two behavior changes land here and in `handleLookupInvoice`, and they deserve an explicit maintainer product decision rather than only a vulnerability framing:

1. Read-only connections drop from full wallet history (Lightning + on-chain + Cashu) to their own, usually empty, activity log. The wallet-wide branch was added deliberately in bb70b1333, so "read-only = companion viewer" (accounting or monitoring apps) may have been the intended use case. Scoping is defensible given what a leaked read-only string exposed, but existing read-only integrations will silently start seeing empty history, so this should be a conscious call and a release-note item. If the viewer use case matters, the Alby Hub precedent (per-app isolated vs non-isolated toggle) would let both coexist later.

2. `lookup_invoice` is now activity-scoped for every permission tier, including full access. Any client that resolves invoices created outside the connection (for example an invoice generated in the ZEUS UI or by ZEUS Pay, checked by an external tool over NWC) now gets `NOT_FOUND`. Worth confirming no shipped integrations rely on node-wide lookup before this merges.

stores/NostrWalletConnectStore.ts
```diff
@@ -1881,99 +1843,138 @@ export default class NostrWalletConnectStore {
...
                         'stores.NostrWalletConnectStore.error.invoiceNotFound'
                     ),
                     Nip47ErrorCode.NOT_FOUND
                 );
             }
 
+            await this.refreshActivityForLookup(connection, activity);
+
+            const result =
+                convertConnectionActivityToNip47Transaction(activity);
```
kaloudis: Regression: settled incoming invoices lose their preimage (and `description_hash`) in `lookup_invoice` responses.

The old path returned `lightningInvoiceToNip47Transaction(new Invoice(rawInvoice))`, which sets `preimage: invoice.getRPreimage` when paid and always sets `description_hash`. The new path goes through `convertConnectionActivityToNip47Transaction`, whose preimage resolution is `activity.preimage || activity.payment?.getPreimage || ''`. A `make_invoice` activity has neither, so a paid incoming invoice now returns an empty preimage even though `reconcilePendingLightningMakeInvoice` has just stored the refreshed `Invoice` (with `getRPreimage`) on `activity.invoice`.

NIP-47 clients use the `lookup_invoice` preimage as proof of settlement (zap receipts, order fulfillment), so this breaks real flows. Suggest extending `convertConnectionActivityToNip47Transaction` to fall back to `activity.invoice?.getRPreimage` for settled incoming activities (and to carry `description_hash` through), plus a test asserting the preimage is present after a settled lookup.

utils/NostrConnectUtils.ts
```diff
@@ -1673,9 +1647,7 @@ export default class NostrConnectUtils {
                 metadata: {
                     block_height: tx.block_height,
                     block_hash: tx.block_hash,
-                    num_confirmations: tx.num_confirmations,
-                    dest_addresses: tx.dest_addresses,
-                    raw_tx_hex: tx.raw_tx_hex
+                    num_confirmations: tx.num_confirmations
```
kaloudis: Stripping `raw_tx_hex`/`dest_addresses` is the right instinct, but after this PR `convertOnChainTransactionsToNip47Transactions` has no production callers: its only caller was the wallet-wide `handleListTransactions` branch this PR removes. Same for `convertLightningDataToNip47Transactions`, `convertCashuDataToNip47Transactions`, and the lookup helpers that only they and the deleted `lookupInvoiceTransaction` path reached (`resolveLookupPaymentHash`, `findPaymentByLookupRequest`, `findCashuInvoiceByLookupRequest`, `cashuInvoiceToNip47Transaction`, `cashuPaymentToNip47Transaction`, `lightningPaymentToNip47Transaction`; `lightningInvoiceToNip47Transaction` survives only via the dead converters).

So the "strip sensitive on-chain fields" bullet in the PR body is currently a no-op in shipped code. Either delete the now-dead chain (my preference, it is a lot of unreachable surface in a security-sensitive file), or keep it deliberately if maintainers opt to retain a wallet-wide tier per the other thread, in which case the stripping becomes live again and this change is correct as is.

stores/NostrWalletConnectStore.ts
```diff
@@ -128,6 +126,17 @@ export interface CreateConnectionParams {
...
+const {
+    buildSettledPaymentFallback,
+    convertConnectionActivityToNip47Transaction,
+    createNip47Error,
+    getFullAccessPermissions,
+    getNotifications,
+    getPayInvoiceAmountSats,
+    isIgnorableError,
+    resolveFeeSats
+} = NostrConnectUtils;
```
kaloudis: Nit: this module-level destructure plus the ~40 `NostrConnectUtils.x` to `x` call-site rewrites is unrelated churn inside a security fix. It works (these statics reference the class by name internally, not `this`), but it doubles the diff to review and pollutes blame on lines that security follow-ups will need to archaeology later. Suggest dropping it from this PR and doing it as a standalone refactor if it is wanted.

kaloudis (commented):
Concept ACK on the scoping direction. A leaked "read-only" connection string exporting the full wallet history (Lightning, Cashu, and on-chain including `dest_addresses` and `raw_tx_hex`) plus node-wide `lookup_invoice` probing is a real data-exposure problem, and the new store-level scoping tests are a welcome addition. The `getPaymentsForLookup`/`lookupPaymentsCache` removal is also complete (field, fetch helper, and all three invalidation sites).

A few things need resolving, though:

1. **The PR body describes a change that is not in the diff.** "Enforce connection expiry and missing-connection rejection in `withGlobalHandler`, dropping orphaned subscriptions with NIP-47 `RESTRICTED`/`UNAUTHORIZED` errors" and the matching test-plan item have no corresponding code: on this branch `withGlobalHandler` still only calls `markConnectionUsed`, and nothing in the diff touches it or references those error codes. Did a commit not get pushed? Note the gap is real: connection expiry is currently only enforced in `validateBudgetBeforePayment`, so an expired connection can keep calling `get_balance`/`list_transactions`/`lookup_invoice` indefinitely. Either push the missing commit or trim the body and test plan to match.

2. **`lookup_invoice` loses the preimage for settled incoming invoices** (inline comment).

3. **The behavior change needs an explicit product ACK from maintainers, not just a bug-fix framing** (inline comment).

Non-blocking:

- `refreshActivityForLookup` reconciles pending Lightning `pay_invoice` and pending `make_invoice`, but not pending Cashu `pay_invoice`. The old path consulted the Cashu payments store directly, so a settled Cashu payment could now report `pending` on lookup until another path reconciles it. Minor, worth a follow-up.
- Newly dead conversion chain in `NostrConnectUtils` (inline comment).
- The module-level destructuring refactor adds a lot of unrelated churn to a security fix (inline comment).
- Test asks: once the preimage fix lands, add an assertion that a settled `make_invoice` lookup includes it; and if the `withGlobalHandler` enforcement commit shows up, a test for the expired-connection rejection.

kaloudis:
@ajaysehwal needs a rebase now

stores/NostrWalletConnectStore.ts
kaloudis: Thanks, 32b542e1d addresses this for the live-session case and the two new tests cover it well.

One residual gap: the fallback guards with `activity.invoice instanceof Invoice`, but persisted connections rehydrate through `NWCConnection.normalizeDates()`, which only coerces dates. After an app restart `activity.invoice` is a plain JSON object, `instanceof` fails, and a settled invoice's preimage and `description_hash` are empty again. The raw fields survive persistence (the reconciled `Invoice` with `r_preimage` is what got serialized), they are just unwrapped. `resolveMakeInvoiceActivitySats` in `NWCConnection` already handles exactly this shape, so the converter can mirror it:

```ts
const lightningInvoice =
    !activity.invoice || activity.payment_source === 'cashu'
        ? undefined
        : activity.invoice instanceof Invoice
        ? activity.invoice
        : new Invoice(activity.invoice);
```

Belt and braces would be to also set `activity.preimage = refreshed.getRPreimage` in `reconcilePendingLightningMakeInvoice` when the refreshed invoice is paid, so the settlement proof persists as a scalar. A test seeding `activity.invoice = JSON.parse(JSON.stringify(settledInvoice))` would pin the rehydrated shape.

utils/NostrConnectUtils.ts
kaloudis: Resolved in f88e1cf6a, thanks. Deleting the chain outright was my preference too. Note the commit subject still says "reject expired connections", but post-rebase (with #4360 on master) it only removes the converters; worth a reword if you touch the branch again.

stores/NostrWalletConnectStore.ts
kaloudis: Resolved, thanks for dropping it. The diff is much easier to audit now.

kaloudis (commented):
Round 2 review of f88e1cf6a. The rebase resolved the body/diff mismatch the right way: #4360 landed the `withGlobalHandler` expiry and UNAUTHORIZED enforcement on master, and this branch now inherits it, so lookup and list are expiry-gated without duplicating logic here. The dead converter chain is gone, the destructure churn was dropped, and the preimage restoration comes with tests at both the utils and store level. CI is green. Nice round.

Remaining items:

1. **Main ask: the preimage fallback goes dead after an app restart** (detail in the existing thread). `activity.invoice instanceof Invoice` fails for rehydrated plain objects, so settled lookups lose their preimage and `description_hash` again after a restart. Small fix, shape-tolerant wrap plus optionally persisting `activity.preimage` at reconcile time.

2. **Suggestion (non-blocking): invoice-only lookups lost hash-derived matching.** The deleted `resolveLookupPaymentHash` decoded `request.invoice` into a payment hash when `payment_hash` was absent. `connectionActivityMatchesLookupRequest` now only string-compares `request.invoice` against `activity.id` and the stored payment request (`paymentRequestMatchesLookupRequest` decodes the activity side, but only compares against a hash derived from `request.payment_hash`). BOLT11 is case-insensitive, so an invoice-only lookup that arrives with different casing misses. Cheap restore: when `payment_hash` is absent, decode `request.invoice` once at the top of `findConnectionActivityByLookupRequest` and match on that hash.

3. **Nit:** `refreshActivityForLookup` can flip an activity to success or expired, but `handleLookupInvoice` never calls `scheduleSave()`, unlike the list path, so the flip only persists on the next unrelated save.

4. **Doc nits:** the PR body's expiry bullet now describes what #4360 landed on master rather than this diff, and the "strip sensitive on-chain fields" bullet became "delete the on-chain conversion outright". Worth a touch-up so the body matches the diff that merges.

The round-1 product thread (read-only viewer removal plus all-tier lookup scoping) is on me to resolve and stays open for now. The round-1 non-blocking note about pending cashu pay_invoice activities returning stale status on lookup also still stands, fine to leave as is.

kaloudis (approved):
Round 3 review of b2ff23148. All three remaining items from round 2 are addressed, and addressed well:

1. **Restart preimage gap (main ask): fixed.** `coerceLightningInvoice` wraps rehydrated plain-object invoices at every consumer (converter, expiry extraction, pending filter, reconcile, backfill), and both reconcile and backfill now persist `activity.preimage`, so settled lookups keep their preimage and `description_hash` without depending on re-coercion at read time. The store-level test seeds the activity through `new NWCConnection(...)` with a raw JSON invoice, which reproduces the persisted shape exactly, and the utils test covers the `r_preimage` base64-to-hex path on a plain object.

2. **Invoice-only lookups: hash-derived matching restored.** `resolveLookupPaymentHash` is back, decoding `request.invoice` once at the top of `findConnectionActivityByLookupRequest`, plus case-insensitive BOLT11 comparison on the string paths. Reusing the BOLT11 spec vector in the test is a nice touch.

3. **Lookup-path persistence: fixed.** `refreshActivityForLookup` reports changes and `handleLookupInvoice` schedules a save. The status-flip proxy is faithful, since every mutating path in both reconcilers also flips status, and the pay_invoice leg returns the reconciler's own `changed` flag.

The PR body now reads accurately against the diff. CI is green and I re-ran both suites locally (361 passing).

**Closing out the round-1 product thread: the scoping stands.** Read-only connections were never documented as a wallet-wide viewer tier, and a leaked read-only string exporting full Lightning, on-chain, and Cashu history is the worse failure mode. Two follow-ups that do not block this PR:

- Release notes should call out that read-only NWC connections now see only their own activity, since any existing companion-app integrations will silently start receiving empty history.
- If the viewer use case resurfaces, the right shape is an explicit opt-in permission per connection (Alby Hub's isolated vs. full-access split is the precedent), done as a new PR rather than by reviving the deleted converter chain.

Non-blocking notes, fine as follow-ups or not at all:

- `resolveLookupPaymentHash`'s second `paymentRequest` parameter is unused by its only caller.
- The cashu leg has the same rehydration blind spot this PR just fixed for Lightning: `reconcilePendingCashuMakeInvoice` reads `cashuInv.isExpired`, a prototype getter, so after a restart a pending cashu invoice that expired never flips to `expired` via the lookup path (paid detection still works, since `quote` is a plain field). This is pre-existing on master (the dedupe moved it, did not introduce it), so out of scope here.
- `extractExpiresAtFromActivity` now routes rehydrated cashu invoices through the Lightning `Invoice` wrapper, because the `instanceof CashuInvoice` guard in `coerceLightningInvoice` cannot recognize plain objects. Today the math degenerates to the same absolute expiry the old fallback returned (creation date resolves to 0, so `0 + expiry`), but a `payment_source === 'cashu'` guard like the converter's would make that correctness not depend on luck.
- f88e1cf6a's subject still says "reject expired connections". Not worth a force-push on its own; only if another push happens anyway.

utACK b2ff231485e69b8fbb0bb913c29da9436796ef70

Code review plus local test runs only, no device pass on my end. The author's iOS + LDK Node testing and the store-level scoping tests cover the risk surface well; a quick manual check that a settled `lookup_invoice` still returns its preimage after an app restart would be a good final confirmation before merge.

kaloudis (approved):

PR #4393, opened 2026-08-12
fix(nwc): rate-limit make_invoice requests per connection

ajaysehwal:
# Description

In this PR, we protect NWC `make_invoice` from remote abuse.

### Changes

**1. Per-connection rate limiting for `make_invoice`**
- Max **5** `make_invoice` requests per **60s** window per NWC connection
- Max **10** concurrent pending `make_invoice` activities per connection
- Returns NIP-47 `RATE_LIMITED` error with a clear message when either limit is hit
- Prunes connection activity history to a max of 100 entries to avoid unbounded growth

### Why
Remote NWC clients can spam `make_invoice`. Rate limiting gives basic abuse protection without impacting legitimate usage.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [x] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


stores/NostrWalletConnectStore.ts
```diff
@@ -1779,24 +1785,78 @@ export default class NostrWalletConnectStore {
...
+        }
+    }
+
+    private checkMakeInvoiceLimits(
+        connection: NWCConnection
+    ): ReturnType<typeof NostrConnectUtils.createNip47Error> | null {
+        const pending = connection.activity.filter(
+            (a) => a.type === 'make_invoice' && a.status === 'pending'
+        ).length;
+        if (pending >= MAX_PENDING_MAKE_INVOICES_PER_CONNECTION) {
```
kaloudis: A `make_invoice` activity only transitions out of `pending` inside `getActivities` (the reconcile around L1029-1091), and that is only called from `NWCConnectionDetails` and `NWCConnectionActivity`, both user-driven screens. `handleLookupInvoice` and `handleListTransactions` only reconcile `pay_invoice` entries (`reconcilePendingPayInvoiceActivities`).

So for a background client (the main NWC receive use case), every invoice ever created on the connection stays `pending` forever, whether it was paid or expired. Once 10 have accumulated, this returns `RATE_LIMITED` on every `make_invoice` until the user opens that connection's activity screen. That effectively bricks long-running receive connections.

Minimum fix: exclude entries whose `expiresAt` has passed (it is stored on the activity, so this stays a cheap in-memory check):

```ts
const now = Date.now();
const pending = connection.activity.filter(
    (a) =>
        a.type === 'make_invoice' &&
        a.status === 'pending' &&
        (!a.expiresAt || a.expiresAt.getTime() > now)
).length;
```

Ideally also skip entries whose invoice is already paid (e.g. via the `invoicesStore`/`cashuStore` caches, like `findLightningInvoiceInStore` does) so a busy receiver with more than 10 paid invoices inside their expiry window is not throttled either.

stores/NostrWalletConnectStore.ts
```diff
@@ -1779,24 +1785,78 @@ export default class NostrWalletConnectStore {
...
+                const createdAt =
+                    connection.activity[i]?.createdAt?.getTime() ?? 0;
+                if (createdAt < oldestTime) {
+                    oldestTime = createdAt;
+                    oldestIdx = i;
+                }
             }
+            connection.activity.splice(oldestIdx, 1);
+        }
+    }
```
kaloudis, line 1828: This prunes oldest-first regardless of type/status, which can evict a pending `pay_invoice` entry, and budget enforcement depends on those surviving until reconcile:

- `NWCConnection.pendingSpendSats` sums pending `pay_invoice` amounts into `effectiveSpendSats`, so pruning one frees budget while the payment is still in flight.
- `reconcilePendingPayInvoiceActivities` finds the settled payment via the activity entry and only then calls `trackSpending`. If the entry was pruned, the debit never happens and the spend permanently escapes the budget.

This is reachable at this PR's own rate limits: 5 `make_invoice`/min fills 100 entries in about 20 minutes, so a connection holding a slow-settling (e.g. hodl) payment can push its pending record out of the log and reuse the budget.

Suggest never pruning `status === 'pending'` entries, or only pruning settled/expired/failed `make_invoice` entries (which is also where the growth this PR targets comes from).

Two side effects worth noting either way: `totalReceivedActivitySats` is computed from this log too, so pruning successful `make_invoice` entries quietly shrinks it; and pruning silently deletes the user's audit history for the connection.

kaloudis (commented):
Concept ACK on both goals: `make_invoice` needs abuse protection, and the per-invoice OS notification was noise. The notification removal looks clean (sole caller of `buildMakeInvoiceSuccessPayload` updated, remaining `numberWithCommas`/`millisatsToSats` uses intact, non-English locales left to the Transifex sync).

However, I think both new guards interact badly with how the connection activity log actually behaves, and one of them undermines the budget enforcement machinery. Details inline; short version:

1. **The pending-invoice cap counts a status that never updates in the background.** `make_invoice` activities only leave `pending` inside `getActivities`, which is only called from the two connection settings screens. For a background NWC client, every invoice ever created stays `pending` (paid or expired alike), so after 10 invoices the connection returns `RATE_LIMITED` on every `make_invoice` until the user happens to open that connection's activity view.

2. **The activity prune can evict pending `pay_invoice` entries that budget enforcement depends on.** `pendingSpendSats` holds in-flight amounts against the budget and reconcile debits `totalSpendSats` when it finds the pending entry. Pruning oldest-first regardless of status can delete that entry, freeing budget immediately and permanently skipping the debit when the payment settles. Reachable at this PR's own limits (5/min fills 100 entries in ~20 minutes).

Non-blocking notes:

- 5/min and 10 outstanding may be tight for a legitimately busy receiver. `make_invoice` is far less dangerous than `pay_invoice` (cost is invoice DB bloat, not funds), so more generous defaults seem fine. Maintainer call.
- The window timestamp is recorded before the invoice attempt, so 5 backend failures lock the client out for the rest of the minute. Reasonable for abuse protection if intentional.
- `makeInvoiceTimestampsByConnection` entries are never removed when a connection is deleted. Bounded at ~5 timestamps per connection, so just a nit.
- The limiter is in-memory and resets on app restart; worth a one-line comment on the constants.
- Dropping the invoice-created notification is an unconditional behavior removal (no setting), and remote-activity notifications have some value as a "your connection is being used" signal, so flagging it for an explicit product ACK from the maintainers.
- Both blocking issues are logic bugs a unit test would have caught (a pending-but-expired activity; a pending `pay_invoice` at prune time). Would be great to add store tests alongside the fix.

stores/NostrWalletConnectStore.ts
kaloudis: Resolved by 0ee2ddb37, thanks. `isActivePendingMakeInvoice` excludes expired entries (and `expiresAt` is always set: `decodeInvoiceTagsForMakeInvoice` returns an absolute timestamp for both the lightning and cashu paths, with a now+3600 fallback) plus already-paid invoices via the store lookups. The lightning paid-check stays reasonably fresh because `handleMakeInvoice` calls `getInvoices()` on every create, and the shape-tolerant `new Invoice(activity.invoice)` wrap for rehydrated activities is a nice touch (same pattern I asked for on #4392).

One residual, non-blocking: for a cashu background-only receiver, `isPaid` on `cashuStore.invoices` only flips via view-driven `checkInvoicePaid` today, so paid-but-unexpired quotes can briefly count toward the cap. The expiry backstop turns that into a temporary throttle rather than a permanent brick, and #4392's lookup-driven reconcile closes it fully once that merges. Fine as is.

stores/NostrWalletConnectStore.ts
kaloudis: Resolved by e5ec4e5a9, thanks. Skipping every pending entry keeps `pendingSpendSats` holds alive until reconcile can debit them, and the two tests pin exactly the scenarios I was worried about, including the all-pending stop condition.

kaloudis (commented):
Round 2 review of e5ec4e5a9. Both blocking asks are addressed with store-level tests (which also covers my round-1 ask for unit tests), and CI is green. utACK e5ec4e5a9 on the code side.

Remaining items, none blocking on the author:

1. The unconditional removal of the invoice-created OS notification is a product call I still owe this PR (it loses the "this connection is being used" signal for invoice-only connections). Will resolve before merge.
2. The limits themselves (5 per 60s, 10 outstanding, 100 activity entries) read as sensible abuse-protection defaults; they are easy to tune later if busy receivers hit them.
3. Heads-up on merge order: this branch and #4392 both touch the constants and private-field block around `lookupPaymentsCache` (deleted over there) and the tail of `NostrWalletConnectStore.test.ts`, so whichever merges second needs a small rebase.
4. Round-1 nits stand as non-blocking: failed create attempts still consume window slots, and `makeInvoiceTimestampsByConnection` entries are not removed on connection delete (bounded at 5 timestamps each, so cosmetic).

ajaysehwal:
> Round 2 review of [e5ec4e5](https://github.com/ZeusLN/zeus/commit/e5ec4e5a9021e5fd9fbfdb38eb07340e9b0dfe59). Both blocking asks are addressed with store-level tests (which also covers my round-1 ask for unit tests), and CI is green. utACK [e5ec4e5](https://github.com/ZeusLN/zeus/commit/e5ec4e5a9021e5fd9fbfdb38eb07340e9b0dfe59) on the code side.
> 
> Remaining items, none blocking on the author:
> 
> 1. The unconditional removal of the invoice-created OS notification is a product call I still owe this PR (it loses the "this connection is being used" signal for invoice-only connections). Will resolve before merge.
> 2. The limits themselves (5 per 60s, 10 outstanding, 100 activity entries) read as sensible abuse-protection defaults; they are easy to tune later if busy receivers hit them.
> 3. Heads-up on merge order: this branch and #4392 both touch the constants and private-field block around `lookupPaymentsCache` (deleted over there) and the tail of `NostrWalletConnectStore.test.ts`, so whichever merges second needs a small rebase.
> 4. Round-1 nits stand as non-blocking: failed create attempts still consume window slots, and `makeInvoiceTimestampsByConnection` entries are not removed on connection delete (bounded at 5 timestamps each, so cosmetic).

fixed in d490ace348813ac2d1787035c563e9737c284739

kaloudis:
Round 3 review of d490ace34.

Thanks for restoring the invoice-created notification. That resolves the product question I still owed this PR from round 2, and since the restore is byte-identical to master the diff is now purely the rate limiting work. The rate-limit bookkeeping changes look right: `checkMakeInvoiceLimits` is now side-effect free, timestamps are recorded only after a successful invoice, the map is cleaned up on connection deletion, and the new tests cover all of it. CI is green.

Two asks before this is merge-ready, one in code and one editorial:

**1. (blocking) Expired pending `make_invoice` entries are never prunable, so the activity cap can still be defeated by the client this PR defends against.**

`pruneConnectionActivity` skips every `status === 'pending'` entry, but pending `make_invoice` activities are only reconciled to `expired`/`success` from user-driven paths (`getActivities`, `refreshActivityForLookup`). A client that only calls `make_invoice` never triggers those. Since `request.expiry` is honored, a client can request `expiry: 1`; after a second the entry no longer counts toward the 10-outstanding cap (`isActivePendingMakeInvoice` correctly excludes it), so `checkMakeInvoiceLimits` keeps returning null and the client sustains 5 invoices/min indefinitely. Every entry stays `pending`, the prune loop hits `oldestIdx === -1` and gives up, and `connection.activity` grows ~7,200 entries/day, each with a serialized invoice, persisted on every debounced save. That defeats `MAX_CONNECTION_ACTIVITY_ENTRIES` for exactly the adversary in this PR's threat model.

Repro test (fails on d490ace34):

```ts
it('prunes expired pending make_invoice entries past the cap', () => {
    const store = buildMakeInvoiceTestStore();
    const connection = seedMakeInvoiceConnection(store, {
        activity: Array.from({ length: 150 }, (_, i) =>
            makePendingMakeInvoiceActivity(
                `lnbc-expired-${i}`,
                new Date('2020-01-01T00:00:00Z'),
                { createdAt: new Date(2024, 0, 1, 0, 0, i) }
            )
        )
    });
    // the outstanding cap never trips, so the client keeps creating:
    expect((store as any).checkMakeInvoiceLimits(connection)).toBeNull();
    (store as any).pruneConnectionActivity(connection);
    expect(connection.activity.length).toBeLessThanOrEqual(100); // fails: 150
});
```

Suggested fix, which I verified passes your two prune tests plus the repro above, and also brings the code in line with its own comment (the comment only claims to protect pending pay_invoice, the code currently protects everything pending):

```ts
if (
    activity.status === 'pending' &&
    (activity.type !== 'make_invoice' ||
        this.isActivePendingMakeInvoice(activity))
) {
    continue;
}
```

This keeps pending `pay_invoice` (budget held via `pendingSpendSats`) and still-active pending `make_invoice` (outstanding-cap integrity), while letting expired or already-paid pending `make_invoice` entries age out normally. If you would rather preserve paid-but-unreconciled entries for history, an expiry-only check works too; either way expired entries must become prunable.

**2. (blocking, editorial) The PR title and body describe a notification removal that is no longer in the diff.** After d490ace34 the notification code is unchanged from master. Please retitle (e.g. "fix(nwc): rate-limit make_invoice requests per connection") and drop section 2 of the description. Optional but nice on the same push: rebase out the remove/restore pair (4d004da2a plus the restore hunks of d490ace34 cancel to zero), so history does not contain a self-revert.

Non-blocking notes, fine as follow-ups or to fold in if you are already touching the code:

- `pruneConnectionActivity` is only invoked from the make_invoice path. `upsertPayInvoiceActivity` and the failed-payment recorders append without pruning, so a connection that never calls `make_invoice` can still grow past 100 via failed pay attempts. Calling the prune from `upsertPayInvoiceActivity` as well would close that.
- The window-rejection branch (5 recent timestamps -> RATE_LIMITED) has no direct test; the rejection test you have covers the pending cap. One more seed-5-and-expect-error case would complete the matrix.
- Two make_invoice events processed concurrently can both pass the check before either records a timestamp, so a burst can slightly exceed 5/window. Fine for a best-effort in-memory limiter, just noting it is understood.
- The new `makeInvoiceRateLimited` locale string needs the translations checklist box ticked (Transifex).

kaloudis:
Round 4 review of 0ee191fd2.

Both blocking items from round 3 are resolved:

1. **Prune blocker fixed.** `pruneConnectionActivity` now only protects pending entries that are pending pay_invoice or *active* pending make_invoice (`activity.type !== 'make_invoice' || this.isActivePendingMakeInvoice(activity)`). Expired and already-paid pending make_invoice entries are now prunable, so the `expiry=1` spam vector no longer grows activity without bound. The new test (`prunes expired pending make_invoice entries past the cap`) reproduces exactly the scenario I flagged: the outstanding cap stays null while entries accumulate, and prune now brings the log back under the cap. I re-ran the suite locally at 0ee191fd2: 32/32 passing.

2. **Title/body updated.** The notification-removal section is gone and the title now describes only the rate-limiting work.

Bonus: adding the `pruneConnectionActivity` call to `upsertPayInvoiceActivity` also closes the non-blocking gap I mentioned, where failed-pay spam could grow the activity log on connections that never call make_invoice. Both write paths are now bounded.

Remaining nits, none blocking:
- The Locales checklist should have "I've added new locale text that requires translations" ticked, since this PR adds `makeInvoiceRateLimited` to `en.json`.
- The remove/restore commit pair (4d004da2a + the restore hunks of d490ace34) is net-zero in the aggregate diff; rebasing it out would tidy history, but I won't hold the PR on it.

utACK 0ee191fd2. CI is green (tsc, lint, test, prettier).

kaloudis (approved):

PR #4395, opened 2026-08-12
fix(ldk): real UTXO confirmations in coin control, anchor reserve no longer shown as pending

### LINKED ISSUE #4394: LDK Node UTXOs remain at 1 confirmation and anchor reserve appears pending
### Describe the bug

When using the embedded LDK Node wallet, confirmed on-chain UTXOs are displayed with exactly **1** confirmation in the Coins screen, even when the transactions have hundreds of confirmations.

The Activity screen reports the correct confirmation count for the same transactions.

 Additionally, the wallet’s main on-chain balance can continuously show the pending clock indicator even after all relevant transactions are fully confirmed. The amount classified as pending may actually be the confirmed anchor-channel reserve, which is unavailable for ordinary spending but is not **unconfirmed**.

### Reproduce

1. Configure ZEUS to use the embedded LDK Node wallet 
2. Receive an on-chain transaction. 
3. Wait until the transaction has multiple confirmations. 
4. Open the main wallet screen. 
5. Observe that the on-chain balance may still display the pending clock indicator. 
6. Swipe the on-chain balance row and select Coins. 7. Open the received UTXO. 
8. Observe that it reports exactly 1 confirmation. 9.
9. Open Activity and inspect the same transaction.
10. Observe that Activity reports the correct, much higher confirmation count.

The UTXO confirmation count should be calculated from its confirmation height and the current best block height. A confirmed anchor-channel reserve should not be classified as unconfirmed.

### ZEUS version

v13.1.3

### Node interface

LDK Node

### Network

Clearnet

### Device

Samsung Galaxy S25 Ultra

### Device operating system

Android 16 / OneUI 8.5

### Log output

```shell

```

kaloudis:
# Description

Relates to issue: **#4394**

Fixes both LDK Node display bugs reported in #4394:

**1. UTXOs always showed 1 confirmation in the Coins screen**

`getUTXOs` hardcoded `confirmations: u.is_spent ? '0' : '1'`. The underlying `WalletUtxo` type carries no confirmation height, so the real depth is now derived by joining each UTXO's txid against the on-chain payment that created it (every wallet UTXO originates from a tracked inbound receive or outbound change) and computing `bestBlockHeight - confirmationHeight + 1`. This is the same data source `formatOnchainTransaction` already uses, which is why the Activity screen was correct all along. UTXOs with no confirmed payment record show 0 confirmations, matching LND behavior for unconfirmed outputs. Spent outputs are now filtered out of coin control instead of being listed with "0 confirmations".

**2. Confirmed anchor channel reserve classified as pending forever**

`getBlockchainBalance` computed `unconfirmed = total - spendable`, but ldk-node's `spendableOnchainBalanceSats` excludes `totalAnchorChannelsReserveSats`. Once an anchor channel is open, the fully confirmed reserve was permanently reported as unconfirmed, so the wallet balance showed the pending clock indefinitely. The reserve (already forwarded by both native bridges, it was just missing from the TS interface) is now subtracted when deriving unconfirmed and folded into confirmed. This matches LND semantics, where `confirmed_balance` includes the anchor reserve. If the reserve is temporarily backed by unconfirmed funds, unconfirmed is clamped at 0 rather than going negative.

No native code changes; both fixes are in the JS mapping layer.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Device testing pending; will update the matrix after verifying on an LDK Node wallet with an open anchor channel and aged UTXOs.

### Locales

### Other:

### DIFF
--- a/backends/LdkNode.ts
+++ b/backends/LdkNode.ts
@@ -259,12 +259,20 @@ export default class LdkNode {
      */
     getBlockchainBalance = async (): Promise<any> => {
         const balances = await LdkNodeInjection.node.listBalances();
+        // spendableOnchainBalanceSats excludes the anchor channel reserve,
+        // which is confirmed but held back - it must not be counted as
+        // unconfirmed or the balance shows as pending forever
+        const anchorReserve = balances.totalAnchorChannelsReserveSats || 0;
+        let unconfirmed = new BigNumber(balances.totalOnchainBalanceSats)
+            .minus(balances.spendableOnchainBalanceSats)
+            .minus(anchorReserve);
+        if (unconfirmed.lt(0)) unconfirmed = new BigNumber(0);
         return {
             total_balance: balances.totalOnchainBalanceSats.toString(),
-            confirmed_balance: balances.spendableOnchainBalanceSats.toString(),
-            unconfirmed_balance: new BigNumber(balances.totalOnchainBalanceSats)
-                .minus(balances.spendableOnchainBalanceSats)
-                .toString()
+            confirmed_balance: new BigNumber(balances.totalOnchainBalanceSats)
+                .minus(unconfirmed)
+                .toString(),
+            unconfirmed_balance: unconfirmed.toString()
         };
     };
 
--- a/backends/LdkNode.ts
+++ b/backends/LdkNode.ts
@@ -1090,17 +1090,45 @@ export default class LdkNode {
      * Get UTXOs for coin control
      */
     getUTXOs = async (): Promise<any> => {
-        const utxos = await LdkNodeInjection.onchain.listUtxos();
+        const [utxos, payments, status] = await Promise.all([
+            LdkNodeInjection.onchain.listUtxos(),
+            LdkNodeInjection.payments.listPayments(),
+            LdkNodeInjection.node.status()
+        ]);
+        const bestBlockHeight = status.currentBestBlock_height;
+
+        // WalletUtxo carries no confirmation height, so look it up from
+        // the on-chain payment that created each UTXO
+        const confirmationHeights: { [txid: string]: number } = {};
+        for (const payment of payments) {
+            if (
+                payment.kind.type === 'onchain' &&
+                payment.kind.txid &&
+                payment.kind.confirmationHeight
+            ) {
+                confirmationHeights[payment.kind.txid] =
+                    payment.kind.confirmationHeight;
+            }
+        }
+
         return {
-            utxos: utxos.map((u) => ({
-                outpoint: {
-                    txid_str: u.txid,
-                    output_index: u.vout
-                },
-                address: u.address,
-                amount_sat: u.value_sats,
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (commented):
deriving confirmed_balance by subtracting unconfirmed breaks when confirmed on-chain balance falls below the anchor reserve. spendableOnchainBalanceSats is a saturating_sub, so it pins at 0 there and confirmed collapses to exactly the reserve — the wallet reports a flat 25,000 sats regardless of what's actually on-chain. The lt(0) clamp only catches total < reserve. It's reachable right after open_channel_fund_max, and it matters beyond the balance label since views/OpenChannel.tsx:329 uses confirmedBlockchainBalance as the displayed fund-max channel size. Also a regression in the risky direction — the old code showed 0 here, understated but safe. Could we compute confirmed directly (S + min(R, total - S)) instead of round-tripping through unconfirmed?

kaloudis:
Good catch, the failure mode is real and I could confirm it with concrete numbers. Once the trusted balance falls below the reserve, `spendableOnchainBalanceSats` saturates at 0 and confirmed pins at exactly the reserve, e.g. right after `open_channel_fund_max` the wallet reported 25,000 sats confirmed when the true confirmed balance is 0, and that flowed into the fund-max amount on the OpenChannel screen.

One note on the suggested formula: `S + min(R, total - S)` is algebraically the same function as what was shipped. The old code computed `confirmed = total - max(0, total - S - R)`, and both simplify to `min(total, S + R)`, so it returns the same flat 25,000 in the saturated case. The information is genuinely destroyed by the saturating_sub; no arithmetic over `(total, spendable, reserve)` can recover confirmed once spendable hits 0.

Pushed a fix that splits the two regimes instead:

- `spendable > 0`: nothing was clipped, so `spendable + reserve` is the confirmed balance exactly. Same result as before in this regime, no extra native calls.
- `spendable == 0` with funds present: sum confirmed UTXOs directly, reusing the confirmation-height join that the coin control commit added. UTXOs whose creating tx has no confirmation height yet count as unconfirmed, so this path can only understate, matching the safe direction of the old behavior.

The expensive path (`listUtxos` + `listPayments` + `status`) only runs in the rare saturated state, so the hot balance-refresh path is unchanged for normal wallets. Will re-verify the fund-max flow on device before marking ready.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

fonzdm:
Can confirm the 25000 sats behavior

PR #4396, opened 2026-08-12
LSPS1/LSPS7 order screens: fix cleared amount fields re-rendering as 0 and doubled background on gradient themes

kaloudis:
# Description

Two UI fixes for the LSPS1/LSPS7 order screens:

**1. Cleared amount fields no longer re-render as `0`**

Deleting the value in the LSP balance, client balance, or channel expiry blocks fields on the LSPS1 order screen (and the channel extension blocks field on the LSPS7 order screen) immediately re-rendered the field as `0`, forcing users to fight the stray digit when entering an amount manually. The cause is `numberWithCommas`, which falls back to `'0'` for empty input. The empty state is now guarded at the input bindings, so a cleared field stays empty and shows its placeholder. `numberWithCommas` itself is unchanged, since its `'0'` fallback is relied on for display elsewhere.

**2. Order content no longer paints a second background over the gradient**

`LSPS1OrderResponse` and `LSPS7OrderResponse` wrapped their content in `<Screen>` (the full-screen theme background wrapper) plus their own `<ScrollView>`, but all four call sites (`views/LSPS1/index.tsx`, `views/LSPS1/Order.tsx`, `views/LSPS7/index.tsx`, `views/LSPS7/Order.tsx`) already render them inside a host `Screen` and `ScrollView`. The nested `Screen` painted a second copy of the theme background sized to the content pane. On solid themes this is invisible (same color over same color), but on gradient themes the gradient restarts mid-screen and reads as a second background behind the order details. The components now render their content `View` directly; the diff is large only because removing the two wrapper levels re-indents the whole render body. This also removes a same-direction nested `ScrollView`.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/LSPS1/OrderResponse.tsx
+++ b/views/LSPS1/OrderResponse.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react';
 import { inject, observer } from 'mobx-react';
-import { ScrollView, View } from 'react-native';
+import { View } from 'react-native';
 import moment from 'moment';
 import BigNumber from 'bignumber.js';
 
-import Screen from '../../components/Screen';
 import KeyValue from '../../components/KeyValue';
 import Amount from '../../components/Amount';
 import Button from '../../components/Button';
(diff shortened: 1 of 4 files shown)


ajaysehwal (commented):
https://github.com/user-attachments/assets/106f1f93-1d7a-4ba5-8a7c-9fce57e841bb

@kaloudis  Maybe this is a pre-existing bug.

kaloudis:
@ajaysehwal good catch, and you're right that it's pre-existing. The same sequence reproduces on master, so it isn't introduced by this PR.

What's happening in your video: "Purchase inbound channel" in LSPS1 Settings uses `navigation.navigate('LSPS1')`, which under react-navigation 7 pushes a second LSPS1 instance when Settings was opened from the LSPS1 screen itself. Both instances observe the singleton `LSPStore`, and the second instance's `componentWillUnmount` calls `resetLSPS1Data()`, wiping `getInfoData` out from under the original instance, which never refetches. That's why the Service info accordion, slider, and Advanced settings disappear, and the slider min/max labels fall back to `0` / `0` via `numberWithCommas(undefined)`.

I've documented the full root cause and fix options in #4417 so this PR can stay scoped to the two UI fixes here.

ajaysehwal (approved):
tACK

PR #4397, opened 2026-08-12
deps: Bump Cashu Dev Kit to 0.17.3: fix ecash operations against cdk-mintd 0.17 mints (Minibits)

kaloudis:
# Description

Fixes ecash operations against Minibits and any other mint running an onchain-enabled cdk-mintd 0.17.x, by bumping the bundled Cashu Dev Kit from 0.14.2 to 0.17.3.

## Root cause

Minibits migrated its mint to cdk-mintd 0.17.2 (`motd: "migrated"`), which advertises NUT-30 onchain support in `/v1/info`. cdk 0.14.2 cannot deserialize that document. Two fields fail (verified with a Rust test parsing the live Minibits JSON against the pinned `cashu` crate versions):

- NUT-17: the closed `WsCommand` enum rejects `onchain_mint_quote` / `onchain_melt_quote`
- NUT-4: the `MintMethodOptions` untagged enum rejects the onchain entry's `options: {confirmations: 1}`

`Wallet::mint_quote` loads mint info as its first step, so every mint quote and melt against such a mint fails, and cdk's `ErrorResponse::from_value` fallback stuffs the entire raw info JSON into the error message, producing an unreadable full-screen banner. The tolerant parsing landed upstream in cdk 0.15.0 with no 0.14.x backport, and 0.15.0 is also the release that removed the `MultiMintWallet` API, so there is no fix version without this migration. cashu 0.15.0 and 0.17.3 both parse the Minibits document cleanly.

## What changed

**Artifact plumbing** (distribution changed upstream at 0.17.x):
- Android moved to Maven Central and is split in two: `cdk-android` (per-ABI `libcdk_ffi.so` only) plus `cdk-jvm` (compiled Kotlin bindings). Both are vendored with pinned sha256 hashes; JNA was already a gradle dependency. Both new hashes are wired into the fail-closed hash format check in `fetch-libraries.sh`.
- iOS framework zip renamed to `CashuDevKitFFI.xcframework.zip` (sha256 matches the binaryTarget checksum in cdk-swift's own `Package.swift`); Swift bindings moved to `Sources/Cdk/`. The upstream xcframework ships a macOS slice as a dynamic library while the iOS slices are static, which CocoaPods refuses; `fetch-libraries.sh` strips the macOS slice after extraction.

**Native module ports** (iOS and Android in lockstep). The JS bridge contract is unchanged: method names, parameters, resolved JSON keys, omit-when-nil behavior, and state string casing are all preserved, so `CashuStore.ts` logic is untouched. Internally:
- `MultiMintWallet` becomes a `WalletRepository` over the same `WalletSqliteDatabase` handle plus per-mint `Wallet` objects created on demand
- melts use the new two-phase `prepareMelt` / `prepareMeltProofs` + `confirm()` flow inline within a single bridge call; `FinalizedMelt` is encoded into the legacy `Melted` JSON shape
- the upstream `payment_preimage` to `payment_proof` rename stays `payment_preimage` in bridge JSON
- `receive` / `restoreFromSeed` add the token's mint on demand, preserving the previous hardcoded `allowUntrusted: true` behavior; decoding a foreign token still does not add or contact its mint
- `checkProofsState` maps `checkProofsSpent` booleans to the same `Spent` / `Unspent` strings, which also removes an old iOS/Android divergence (`Unknown` vs `Reserved` for exotic states; JS treats anything not `Unspent` as spent)
- `restore` resolves `Restored.unspent` (the recovered spendable amount)
- the FFI error enum collapsed from 19 variants to `Cdk(code, message)` / `Internal(message)`; both `mapFfiError` implementations derive the legacy bridge error code strings from protocol codes and pass the message through for JS-side classification
- `meltMpp` is removed (zero app call sites; its global mint selection has no 0.17 equivalent)
- iOS `meltPartial` gains the `FfiError` catch arm it was missing (previously the main source of raw FFI dumps in the UI)
- `addExternalMintQuote` (ZEUS Pay) stores the quote without its P2PK secret key and writes cdk's NpubCash quote-key marker instead. The ZEUS Pay key is the wallet's seed prefix (seed[0..32]), byte-identical to what cdk's npubcash feature treats as a legacy NpubCash key; storing it made the mint saga scrub the quote mid-flight (an optimistic-lock version bump) and the saga's post-mint write then failed with "Concurrent update detected" after the mint had already issued signatures, stranding the funds until restore. With the marker, cdk re-derives the identical key at signing time and no mid-saga write occurs. Verified deterministic on every lightning address redemption before the fix. This knowingly couples to cdk's npubcash kv layout; to be removed when upstream fixes the scrub/saga interaction (filed as cashubtc/cdk#2335). A related follow-up worth considering: calling `recoverIncompleteSagas` at wallet init to self-heal any interrupted operation.

**Error display hardening:**
- `parseCashuDevKitError` handles the 0.17.x error shapes on both platforms (including the Kotlin package move from `uniffi.cdk_ffi` to `org.cashudevkit`), with test vectors
- new `cashuErrorForDisplay` truncates parsed error text at 300 chars at display time, applied at every `CashuStore` `error_msg` / `paymentErrorMsg` assignment fed by raw exception text, so a raw JSON body can never fill the screen again; `mapCDKError` classification still sees the full message
- `ensureMintAdded` no longer swallows `addMint` failures: it logs, records `errorConnecting`, and rethrows so failures surface at the first broken step

## Wallet database migration

The cdk wallet sqlite migrations from 0.14.2 to 0.17.3 are additive and auto-applied on first open; no manual steps. **The upgrade is one-way**: a `payment_preimage` column rename means a wallet db opened by 0.17.3 cannot roll back to a 0.14.2 build.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [x] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS (simulator; on-device passes in progress, see device checklist below)

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

    * verify that `package.json` and `yarn.lock` have been properly updated (unchanged: the CDK is vendored via `fetch-libraries.sh`, pinned in `fetch-libraries-versions.json`)

### Other:

## Device test checklist

- [x] iOS: fresh ecash wallet init
- [x] iOS: create ecash invoice from Minibits (the original bug): mint quote now succeeds
- [x] iOS: pay that invoice and confirm payment detection + balance credit
- [x] Android: fresh ecash wallet init
- [x] Existing 0.14.2 wallet db auto-migration (throwaway wallet first; migration is one-way)
- [x] Mint / melt / send / receive round-trip on a second mint
- [x] Multimint payment (melt planner) across two mints
- [x] ZEUS Pay external quote redemption
- [x] Restore from seed (v1 legacy and v2)
- [x] Offline send (prepareSend / confirmSend) and reclaim
- [x] Token from a not-yet-added mint still auto-adds on receive
- [x] Forced mint error shows a humanized, truncated message (no raw JSON banner) — tested on iOS against a mock mint serving a 1.9 KB unparseable info document (same failure class as the original incident); banner truncates at ~300 chars with ellipsis
- [ ] v1-wallet ZEUS Pay redemption (v1 seed, key persisted on quote rather than marker)
- [ ] Retry after a failed external mint: a quote left claimed by a dead saga mints on the next poll (quote version CAS fix)
- [ ] Init timing: `initializeWallet` with 4+ mints, one unreachable host, timed against a 0.14.2 build (see #4413)
- [ ] Remove a multi-unit mint (e.g. testnut, advertises usd), restart, confirm it stays removed (see review video)
- [ ] Remove a mint that still has a pending unexpired invoice, restart, confirm it stays removed (pending-poll re-add fix)
- [ ] Remove the LAST remaining mint, restart, confirm it stays removed (Nostr fresh-recovery discriminator fix)

### DIFF
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -835,7 +835,7 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                 // Create the MintQuote object
                 // For external quotes that are PAID, we set amountPaid = amount
                 val zeroAmount = Amount(0UL)
-                val quote = MintQuote(
+                fun makeQuote(version: UInt) = MintQuote(
                     id = quoteId,
                     mintUrl = url,
                     amount = amt,
@@ -849,13 +849,25 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     paymentMethod = PaymentMethod.Bolt11,
                     secretKey = if (useSeedPrefixMarker) null else storedSecretKey,
                     usedByOperation = null,
-                    version = 0u
+                    version = version
                 )
 
                 Log.d(TAG, "addExternalMintQuote: Adding quote $quoteId to database")
 
-                // Add to database
-                db!!.addMintQuote(quote)
+                // addMintQuote is a CAS upsert: the update only applies while
+                // the stored row still has the version we write, and a failed
+                // mint attempt leaves the row bumped by the saga's claim.
+                // Reuse the stored version (0 for a fresh insert) so retries
+                // do not die with ConcurrentUpdate before minting
+                val storedVersion = db!!.getMintQuote(quoteId)?.version ?: 0u
+                try {
+                    db!!.addMintQuote(makeQuote(storedVersion))
+                } catch (e: FfiException) {
+                    // Lost the CAS race between read and write; re-read and
+                    // retry once
+                    val retryVersion = db!!.getMintQuote(quoteId)?.version ?: 0u
+                    db!!.addMintQuote(makeQuote(retryVersion))
+                }
 
                 Log.d(TAG, "addExternalMintQuote: Successfully added quote $quoteId")
 
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -445,10 +445,19 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
             try {
                 val normalized = normalizeMintUrl(mintUrl)
                 val url = MintUrl(normalized)
-                // removeWallet only drops the in-memory wallet; tolerate a
-                // mint the repository does not know (parity with the
-                // non-throwing 0.14.x removeMint)
-                runCatching { repo.removeWallet(url, walletUnit) }
+                // load_wallets creates one wallet per supported unit, so
+                // drop every unit's wallet for this mint, not just the
+                // configured one: a leftover handle keeps the mint in
+                // getMintUrls, and the persisted list then re-adds the
+                // mint at the next boot's reconcile. Compare
+                // case-insensitively since cdk canonicalizes scheme and
+                // host casing. removeWallet failures are tolerated (parity
+                // with the non-throwing 0.14.x removeMint)
+                repo.getWallets().forEach { wallet ->
+                    val wUrl = runCatching { wallet.mintUrl() }.getOrNull() ?: return@forEach
+                    if (!wUrl.url.equals(normalized, ignoreCase = true)) return@forEach
+                    runCatching { repo.removeWallet(wUrl, wallet.unit()) }
+                }
                 wallets.remove(normalized)
                 db!!.removeMint(url)
 
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -793,7 +793,7 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
 
         scope.launch {
             try {
-                val url = MintUrl(mintUrl)
+                val url = MintUrl(normalizeMintUrl(mintUrl))
                 val amt = Amount(amount.toLong().toULong())
 
                 // Map state string to QuoteState enum
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -151,9 +151,11 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         }
 
         val url = MintUrl(normalized)
-        if (!currentRepo.hasMint(url)) {
-            currentRepo.createWallet(url, walletUnit, null)
-        }
+        // hasMint matches any unit for the URL while getWallet is keyed on
+        // (URL, unit), so a mint loaded only under a non-sat unit would skip
+        // creation here and then miss. createWallet is a no-network map
+        // insert, so create unconditionally instead of guarding
+        currentRepo.createWallet(url, walletUnit, null)
         val wallet = currentRepo.getWallet(url, walletUnit)
         wallets[normalized] = wallet
         return wallet
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -151,12 +151,19 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         }
 
         val url = MintUrl(normalized)
-        // hasMint matches any unit for the URL while getWallet is keyed on
-        // (URL, unit), so a mint loaded only under a non-sat unit would skip
-        // creation here and then miss. createWallet is a no-network map
-        // insert, so create unconditionally instead of guarding
-        currentRepo.createWallet(url, walletUnit, null)
-        val wallet = currentRepo.getWallet(url, walletUnit)
+        // Try the (URL, unit)-keyed lookup first and only create on a miss:
+        // creating unconditionally would overwrite a wallet configured by
+        // addMint (e.g. a custom targetProofCount) with a default-config
+        // handle. The FFI does not expose the unit-keyed hasWallet, so a
+        // failed get is the miss signal; createWallet is a no-network map
+        // insert, and a concurrent double-create is a benign same-config
+        // overwrite
+        val wallet = try {
+            currentRepo.getWallet(url, walletUnit)
+        } catch (e: FfiException) {
+            currentRepo.createWallet(url, walletUnit, null)
+            currentRepo.getWallet(url, walletUnit)
+        }
         wallets[normalized] = wallet
         return wallet
     }
@@ -407,7 +414,10 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         scope.launch {
             try {
                 val url = MintUrl(normalizeMintUrl(mintUrl))
-                repo.createWallet(url, walletUnit, targetProofCount?.toUInt())
+                // 0 is the JS sentinel for "use default"; match iOS, which
+                // maps it to null rather than a target of zero proofs
+                val count = targetProofCount?.takeIf { it > 0 }?.toUInt()
+                repo.createWallet(url, walletUnit, count)
 
                 withContext(Dispatchers.Main) {
                     promise.resolve(null)
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -777,6 +777,7 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         state: String,
         expiry: Double,
         secretKey: String?,
+        useSeedPrefixMarker: Boolean,
         promise: Promise
     ) {
         if (!isInitialized || db == null) {
@@ -798,18 +799,25 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     else -> QuoteState.PAID // Default to PAID for external quotes
                 }
 
-                // ZEUS Pay locks quotes to the wallet's seed-prefix key
-                // (seed[0..32]), which is byte-identical to cdk's "legacy
-                // NpubCash" key. Storing that key on the quote makes cdk's
-                // mint saga scrub it mid-flight (a version bump), and the
-                // saga's post-mint write then dies with ConcurrentUpdate
-                // AFTER the mint has issued the signatures. Instead, store
-                // the quote with no key and write cdk's NpubCash quote-key
-                // marker: at signing time cdk re-derives the identical
-                // seed-prefix key from the marker, with no mid-saga write.
+                // For v2-bip39 wallets, ZEUS Pay locks quotes to the seed-
+                // prefix key (seed[0..32]), which is byte-identical to cdk's
+                // "legacy NpubCash" key. Storing that key on the quote makes
+                // cdk's mint saga scrub it mid-flight (a version bump), and
+                // the saga's post-mint write then dies with ConcurrentUpdate
+                // AFTER the mint has issued the signatures. For those quotes
+                // (useSeedPrefixMarker, decided by JS via byte comparison),
+                // store the quote with no key and write cdk's NpubCash
+                // quote-key marker: at signing time cdk re-derives the
+                // identical seed-prefix key from the marker, with no mid-saga
+                // write. v1 wallets sign with a different key (LND seed
+                // bytes [32:64]), so the marker would derive the wrong key
+                // for them; their key is stored on the quote instead, which
+                // is safe because the scrub only triggers on byte equality
+                // with the seed prefix.
                 // Upstream bug: https://github.com/cashubtc/cdk/issues/2335
                 // Remove when fixed: https://github.com/ZeusLN/zeus/issues/4402
-                if (!secretKey.isNullOrEmpty()) {
+                val storedSecretKey = secretKey?.takeIf { it.isNotEmpty() }
+                if (storedSecretKey != null && useSeedPrefixMarker) {
                     db!!.kvWrite(
                         primaryNamespace = "npubcash",
                         secondaryNamespace = "quotes",
@@ -833,7 +841,7 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     amountIssued = if (quoteState == QuoteState.ISSUED) amt else zeroAmount,
                     estimatedBlocks = null,
                     paymentMethod = PaymentMethod.Bolt11,
-                    secretKey = null,
+                    secretKey = if (useSeedPrefixMarker) null else storedSecretKey,
                     usedByOperation = null,
                     version = 0u
                 )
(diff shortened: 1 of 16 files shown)


ajaysehwal (commented):
@kaloudis I think this introduces a regression for v1 wallets. When JS provides a `secretKey`, the native code now ignores it and stores the `"legacy-seed-prefix"` marker, causing CDK to derive the signing key from `seed[..32]`.

That matches the v2 BIP39 flow, but not v1. For v1 wallets, the P2PK key comes from the LND seed bytes `[32:64]`, so the key is different from CDK's `seed[..32]`.

This means a v1 wallet can have the quote marked `PAID` by the mint, but then fail `mintExternal` because the NUT-20 signature is made with the wrong key.

I think we should only use the `"legacy-seed-prefix"` marker when the supplied key actually matches CDK's seed prefix. Otherwise, we should persist the JS-provided key for the v1 path.

kaloudis:
Good catch @ajaysehwal, confirmed. For v1 wallets `deriveCashuSecretKey()` returns the key from LND seed bytes `[32:64]` (via `originalSeedVersion === 'v1'` in `ensureSeed`), while the CDK wallet is initialized from a 12-word mnemonic derived from LND seed bytes `[48:64]`, so cdk's `derive_legacy_npubcash_secret_key` (its `seed[..32]`) produces a different key. With the unconditional marker, a v1 ZEUS Pay redemption would fail NUT-20 signature verification at the mint on every attempt. (The quote stays `PAID` rather than being stranded, since the mint verifies the witness before issuing signatures, but the redemption deterministically fails.)

Fixed in f8a90506b along the lines you suggested:

- JS now byte-compares the derived key against the CDK wallet's seed prefix (`mnemonicToSeedSync(cdkMnemonic)[0..32]`) and passes a `useSeedPrefixMarker` flag over the bridge.
- Match (v2-bip39 wallets): quote is stored keyless with the NpubCash marker, as before, to dodge the cdk#2335 scrub.
- No match (v1 wallets): the JS-provided key is persisted on the quote. This cannot trip the scrub because `is_legacy_npubcash_secret_key` is byte equality with the seed prefix, which is exactly the comparison we just performed.

The comparison mirrors cdk's own heuristic, so any odd seed shape that fails the comparison falls through to the store-the-key path, which is safe by construction. Both native modules (Kotlin + Swift) carry the same branch, and the workaround comments now document the v1/v2 split.

shubhamkmr04 (commented):
cdk 0.14's load_wallets created one wallet per mint for the configured unit. 0.17's creates one per supported unit — it calls info.supported_units() and loops. So repo.getBalances() returns [WalletKey: Amount] keyed by (url, unit), and both modules fold that into a per-URL total with result[key.mintUrl.url, default: 0] += amount.value. A mint advertising usd or auth alongside sat will have its non-sat balance added into the sat figure. Suggest filtering on walletUnit in both getBalances and getTotalBalance, e.g. for (key, amount) in balances where key.unit == unit.

shubhamkmr04:
has_mint is "any wallet for this URL, any unit", but get_wallet is keyed on (url, unit) and returns Error::UnknownWallet on a miss. So a mint whose repo entry exists only under a non-sat unit skips creation and then throws. There's also a benign TOCTOU race between the check and the create. Since create_wallet does no network — it just builds a Wallet and inserts into a BTreeMap — the guard buys nothing; dropping it and always calling createWallet removes both problems. Worth keeping the hasMint check in encodeToken though, that one is correct and load-bearing.

shubhamkmr04:
add_mint_quote in cdk-sql-common is a CAS upsert — ON CONFLICT(id) DO UPDATE SET … WHERE mint_quote.version = :expected_version, returning database::Error::ConcurrentUpdate when rows_affected == 0. Both modules hardcode version: 0. The first write inserts fine, but if an earlier attempt left the row at version ≥ 1 (the mint saga bumps it when it claims the quote via used_by_operation), a retry fails with "Concurrent update detected" — the same error class this PR is fixing, reached by a different path. Reachable when the mint marks a quote PAID, mintExternal fails after the saga claims the quote but before signatures are issued, and the next poll re-enters the PAID branch. Reading the stored quote and reusing its version, or retrying once on ConcurrentUpdate, would close it. Happy for this to be a follow-up rather than in-scope here.

kaloudis:
Good catch @shubhamkmr04, confirmed against the cdk v0.17.3 source. 0.14's `MultiMintWallet` was explicitly single-unit ("manages multiple mints but supports only one currency unit"), while 0.17's `WalletRepository::load_wallets` runs on every init, fetches each stored mint's info, and creates one wallet per `supported_units()` entry (falling back to sat-only when the info fetch fails). `get_balances()` then returns the map keyed by (mint URL, unit), and our folds summed every unit into the sat figure.

Worth noting the bug is latent rather than actively corrupting balances today: ZEUS only ever creates and uses sat wallet handles, `Wallet::receive` of a foreign-unit token fails at the mint's swap (cross-unit swap is rejected), and restore is unit-filtered inside `get_mint_keysets` (keysets with `unit != self.unit` are skipped). So the auto-created usd/auth wallets sit at zero and the fold currently adds zeros. But it is wrong by construction and would silently mis-report the moment any non-sat proofs land in the db.

Fixed in dea04aef3 as you suggested: all four fold sites (Swift + Kotlin, `getBalances` + `getTotalBalance`) now filter on the wallet's configured unit before summing. This also matches cdk's own `total_balance()`, which groups by unit rather than summing across units.

kaloudis:
Both confirmed @shubhamkmr04, thanks. Fixed in 5acb67bce and a68a777fd.

**getWallet guard:** verified against the cdk source that `has_mint` matches any wallet for the URL while `get_wallet` is keyed on (URL, unit) and returns `UnknownWallet` on a miss. The trap arms after a restart: within the first session `addMint` explicitly creates the sat wallet, but `load_wallets` recreates only the advertised units, so a mint advertising no sat unit comes back as (URL, usd) only, passes the `hasMint` check, and then misses on `get_wallet`. Since `create_wallet` is a no-network build plus map insert with overwrite semantics (and no call site passes a custom `targetProofCount` that could get clobbered), both modules now create unconditionally, which also removes the TOCTOU. The `hasMint` gate in `decodeToken` is untouched; as you said, that one is load-bearing (it keeps foreign tokens from adding or contacting their mint).

**Quote version:** verified the CAS upsert in cdk-sql-common (`ON CONFLICT(id) DO UPDATE ... WHERE mint_quote.version = :expected_version`, `ConcurrentUpdate` on zero rows) and the re-entry path: `checkAndMintCDK` re-enters the PAID branch on every poll and calls `addExternalMintQuote` before `mintExternal`, so a row left at version >= 1 by a dead saga claim would fail the CAS on every retry, before minting even starts. Took it in-scope rather than as a follow-up since it is the same wedge class this PR exists to fix: both modules now read the stored quote via `getMintQuote` and reuse its version (0 for a fresh insert), with a single re-read retry if the write loses a race with a concurrent bump. A side benefit of the re-add is that it writes `used_by_operation = NULL`, clearing the dead saga's claim, which nothing else currently recovers.

shubhamkmr04:
WalletRepository(mnemonic:store:) calls load_wallets(), which loops every mint in the db and awaits fetch_mint_info — a real GET /v1/info on a fresh HTTP client — before the constructor returns. 0.14's load_wallets was db-only, so initializeWallet was effectively local; it's now N sequential round trips, and one dead or slow mint costs a full timeout before the cashu balance renders. Failures are tolerated (fallback to [Sat]) so this isn't breakage, but it scales linearly with mint count and hits multi-mint users hardest. Probably a follow-up issue (parallelize, or defer load_wallets off the init path), but worth timing against a 0.14.2 build with 4+ mints and one unreachable host before merge.

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
ajaysehwal, line 1: nit: addExternalMintQuote still builds MintUrl without normalizeMintUrl (addMint/getWallet do).

kaloudis: Done in c0e255bb2. Functionally it turned out to be moot: the FFI boundary converts through cdk's `MintUrl::from_str`, whose `format_url` trims trailing slashes and lowercases the scheme and host, which is strictly stronger than our `normalizeMintUrl`, so the stored quote's mint URL was canonical either way. Taken for consistency with addMint/getWallet all the same.

kaloudis:
Confirmed @shubhamkmr04, and it is a bit worse than described. Two additions from digging through the 0.17.3 source:

- There is no request timeout anywhere in the connector layer: `fetch_mint_info` builds a fresh `HttpClient` per mint and nothing in `mint_connector/transport` configures one (reqwest's default is none). So an unreachable host that black-holes packets costs the OS TCP connect timeout (roughly 1 to 2 minutes on mobile), and a reachable but hung server can stall the constructor indefinitely, not just for one timeout period.
- The refetch is also redundant: `load_wallets` destructures `for (mint_url, _mint_info) in mints` and discards the mint info already stored in the db row before fetching it again over the network.

Agreed on follow-up scope, so nothing changes in this PR. Filed both sides:

- ZeusLN/zeus#4413 for the ZEUS-side mitigation (move `initializeCDK` off the boot-critical path or defer to first cashu use).
- cashubtc/cdk#2351 upstream (use the stored mint info with lazy refresh, parallelize the fetches, add a transport timeout, or a builder flag to skip discovery).

Also added your timing measurement to the device-test checklist in the PR body: `initializeWallet` timed with 4+ mints, one of them an unreachable host, against a 0.14.2 build.

ajaysehwal:
@kaloudis  I think there’s still an issue with `targetProofCount` handling on Android. 

JS sends `targetProofCount ?? 0`, but Android treats `0` as `Some(0)`, while iOS converts `0` to `nil`. Also, `addMint` calls `getWallet()` first and then `createWallet(..., null)`, which can overwrite the in-memory wallet with the default value (`3`). This means Android may appear to self-heal on first use, but `addMint` itself is still incorrect, and a real non-zero `targetProofCount` could also get overwritten.

I think we should preserve the actual `targetProofCount` value through the `getWallet`/`createWallet` flow and handle `0` consistently across Android and iOS.

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
kaloudis: https://github.com/ZeusLN/zeus/pull/4397/commits/c0e255bb23b740b376f030f8df7b0a4b198f7d5d

kaloudis:
Confirmed both halves @ajaysehwal, fixed in 1c4b4092e. One small correction to the mechanics: `addMint` does not call `getWallet` first, it is the reverse. `addMint` creates the wallet with its count, then the first operation goes through `getWallet`, which since 5acb67bce created unconditionally and overwrote the entry with a default-config handle. Same conclusion though: Android's `Some(0)` wallet was masked by that overwrite, and a real non-zero `targetProofCount` would have been silently clobbered on both platforms.

- `getWallet` now tries the (URL, unit)-keyed lookup first and only creates on a miss. The FFI does not export the unit-keyed `hasWallet`, so a failed get is the miss probe. This preserves whatever config `addMint` established, keeps the round-2b fix (the get is keyed on exactly (URL, unit), so the original wrong-key `hasMint` trap stays dead), and the remaining race is a benign same-config double create.
- Android `addMint` now maps the JS `0` sentinel to null like iOS (`targetProofCount?.takeIf { it > 0 }?.toUInt()`), so wallets no longer get created with a target of zero proofs.

ajaysehwal:
https://github.com/user-attachments/assets/42bf41d1-fcdf-4316-a030-063d43ebcb01

shubhamkmr04:
kind of similar behaviour in android, first mint is not deleting 

https://github.com/user-attachments/assets/2abc64f9-760b-4ff9-8049-8737bc33f656

kaloudis:
Thanks for the video @ajaysehwal, reproduced the chain in code and fixed in 74f5df759. It is another member of the per-unit `load_wallets` family:

1. 0.17's `load_wallets` creates one wallet per advertised unit, so a mint advertising usd (or auth) alongside sat sits in the in-memory repository as multiple (URL, unit) entries.
2. `removeMint` removed only the configured unit's wallet (`removeWallet(url, sat)`) before deleting the db mint row, leaving the other units' handles in the map.
3. The JS side then refreshes its mint list from `getMintUrls`, which is derived from `repo.getWallets()`, so the leftover handle kept the removed mint in the list, and that list is persisted to storage.
4. On restart the db row is gone and CDK correctly reports the mint absent, but the boot reconcile walks the persisted list, sees the mint "missing from CDK", and re-adds it.

A sat-only mint removes cleanly, which is why this survived the earlier passes. The fix is in `removeMint` on both platforms: iterate `repo.getWallets()` and remove every wallet whose mint URL matches (case-insensitive, since cdk canonicalizes scheme and host casing), each with its own `unit()`. With all units gone, `getMintUrls` drops the mint, the persisted list stays correct, and the reconcile has nothing to re-add. No JS change needed.

shubhamkmr04:
> kind of similar behaviour in android, first mint is not deleting
> 
>  Screen.Recording.2026-08-13.at.11.57.59.PM.mov

issues is fixed in android

kaloudis:
Found the remaining reappear path and fixed it in 7ca7237a8. It was not actually iOS-specific, it is state-specific, which is why the two platforms behaved differently:

`checkPendingItems` runs on wallet boot and polls pending invoices (`checkInvoicePaid` -> `checkAndMintCDK`) and unspent sent tokens (`checkTokenSpent`). Both take their mint URL from the persisted invoice/token record, not the live mint list, and both go through `ensureMintAdded`, whose in-memory cache is empty after a restart. So a mint removed while it still had a pending unexpired invoice or an unspent sent token was re-added by `CashuDevKit.addMint` on the next boot. The iOS test wallet had pending Minibits invoices from earlier device testing; the Android wallet did not. The same repro on Android with a pending invoice would resurrect the mint there too.

Fix, all JS this time:

- `checkPendingItems` and `checkSentTokensSpentStatus` now skip records whose mint is no longer in the configured mint list, so background polls cannot resurrect a removed mint.
- `removeMint` now also evicts the mint from `addedMintsCache`, closing the in-session variant where a stale cache hit made flows treat the removed mint as still present.

The native multi-unit `removeWallet` fix from 74f5df759 stands; it closed the persisted-list chain, this closes the pending-records chain.

kaloudis:
Third and final resurrection chain found and fixed in a36bd1e5a. This one only affects the LAST mint, which matches the remaining repro:

1. Removing the last mint persists `-cashu-mintUrls` as an empty list, then fires the Nostr mint backup, but `nostrBackupMints` deliberately never publishes an empty list, so the relays keep the stale backup that still contains the mint.
2. On the next boot, `initializeWallets` saw zero local mints, treated it as a fresh recovery, and called `nostrRestoreMints`, which returned the stale backup and re-added the mint.

Any non-last mint leaves a non-empty list, so the restore never triggers, which is why the earlier fixes held for every mint except the final one.

The fix corrects the fresh-state discriminator: the Nostr restore (and the onboarding `initialMintUrls` add, which had the same zero-mints trigger) now only runs when the mint list key has never been stored at all, not when it holds a deliberately emptied list.

One known cross-device caveat to note, not a regression from this PR: a genuine fresh recovery on a different device will still pick up the stale relay backup, because empty lists are intentionally never published (an accidental empty publish would clobber a good backup). Propagating deletions across devices safely is the freshness problem tracked in #4370's territory.

So three chains total, all now closed:
- persisted-list chain: native removeMint left non-sat wallet handles alive (74f5df759)
- pending-records chain: boot polls re-added mints referenced by pending invoices and unspent sent tokens (7ca7237a8)
- last-mint chain: emptied list mistaken for fresh recovery, restored from stale Nostr backup (a36bd1e5a)

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4399, opened 2026-08-12
Swaps: retire ZEUS swap server, default to Boltz, fetch rates only in Swaps view

kaloudis:
# Description

Relates to issue: **N/A**

Three changes to the swaps stack:

**1. Retire the ZEUS swap server, make Boltz the default**

- Removes `swaps.zeuslsp.com` / `testnet-swaps.zeuslsp.com` from the mainnet and testnet swap host lists.
- `DEFAULT_SWAP_HOST_MAINNET` / `DEFAULT_SWAP_HOST_TESTNET` now point at Boltz (`https://api.boltz.exchange/v2`, `https://api.testnet.boltz.exchange/v2`).
- New one-shot `swap-hosts-boltz` migration moves users whose persisted settings still point at the retired hosts to the Boltz defaults. It runs on both the legacy and modern (`zeus-settings-v2`) settings paths, only rewrites hosts that exactly match the retired defaults (SwapMarket, Eldamar, and custom hosts are untouched), and only persists when something actually changed.

**2. Only fetch swap rates when the user enters the Swaps view**

Previously swap rates were fetched on startup: once from `Wallet.tsx` after connect, and again via a `SwapStore` reaction that fired whenever the resolved swap host changed (settings load, node info resolving testnet vs mainnet). Both are removed. `getSwapFees` is now only called from the Swaps view.

- The store tracks which host rates were last fetched from (`fetchedRatesHost`); the Swaps view re-fetches on focus if the provider changed since the last fetch (e.g. returning from Swap Settings).
- Swap status polling for existing swaps (`fetchAndUpdateSwaps`) is unchanged.

**3. Remove the swap shortcuts from Send and PaymentRequest**

The swap shortcut icons in Send (on-chain reverse swap) and PaymentRequest (submarine swap) were gated on cached rate data that no longer loads at startup, so they would have appeared only after visiting the Swaps view. The shortcuts and their supporting fee/limit math are removed; swaps remain available from the Swaps view.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [x] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

New unit tests cover the `swap-hosts-boltz` migration (ZEUS hosts rewritten, other hosts untouched, no-op on flag, no rewrite when nothing changed).

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Device tests pending: fresh install (Boltz default), upgrade with persisted ZEUS host (migration), swaps view rate fetch, provider change in Swap Settings followed by return to the Swaps view, confirming no requests to the swap host on startup, and confirming Send / PaymentRequest headers render correctly without the swap shortcut.

### Locales

### Other:

### DIFF
--- a/stores/SwapStore.ts
+++ b/stores/SwapStore.ts
@@ -269,11 +269,12 @@ export default class SwapStore {
     };
 
     @action
-    public getLockupTransaction = async (id: string) => {
+    public getLockupTransaction = async (id: string, endpoint?: string) => {
         try {
+            const host = endpoint || this.getHost;
             const response = await ReactNativeBlobUtil.fetch(
                 'GET',
-                `${this.getHost}/swap/submarine/${id}/transaction`,
+                `${host}/swap/submarine/${id}/transaction`,
                 this.getHeaders
             );
 
@@ -606,10 +607,15 @@ export default class SwapStore {
 
                 if (shouldSkip) continue;
 
+                // poll the server the swap was created on; the currently
+                // selected provider may differ (e.g. after the ZEUS ->
+                // Boltz host migration) and won't know this swap's ID
+                const host = swap.endpoint || this.getHost;
+
                 try {
                     const response = await ReactNativeBlobUtil.fetch(
                         'GET',
-                        `${this.getHost}/swap/${swap.id}`,
+                        `${host}/swap/${swap.id}`,
                         this.getHeaders
                     );
 
--- a/stores/SwapStore.ts
+++ b/stores/SwapStore.ts
@@ -1,4 +1,4 @@
-import { action, observable, computed, runInAction } from 'mobx';
+import { action, observable, computed, runInAction, reaction } from 'mobx';
 import ReactNativeBlobUtil from 'react-native-blob-util';
 import { ECPairAPI, ECPairFactory } from 'ecpair';
 import ecc from '@bitcoinerlab/secp256k1';
@@ -75,6 +75,19 @@ export default class SwapStore {
         this.settingsStore = settingsStore;
         initEccLib(ecc);
         this.ECPair = ECPairFactory(ecc);
+
+        reaction(
+            () => this.getHost,
+            (host) => {
+                // only refetch when rates were already fetched this
+                // session; getHost also changes on startup as settings
+                // load and node info resolves, and rates must not be
+                // fetched on startup
+                if (this.fetchedRatesHost && this.fetchedRatesHost !== host) {
+                    this.getSwapFees();
+                }
+            }
+        );
     }
 
     @computed get claimMinerFee(): number {
@@ -186,12 +199,12 @@ export default class SwapStore {
     public getSwapFees = async () => {
         this.loading = true;
         this.apiError = '';
-        this.fetchedRatesHost = this.getHost;
-        console.log(`Fetching fees from: ${this.getHost}`);
+        const host = this.getHost;
+        console.log(`Fetching fees from: ${host}`);
         try {
             const response = await ReactNativeBlobUtil.fetch(
                 'GET',
-                `${this.getHost}/swap/submarine`,
+                `${host}/swap/submarine`,
                 this.getHeaders
             );
             const status = response.info().status;
@@ -222,7 +235,7 @@ export default class SwapStore {
         try {
             const response = await ReactNativeBlobUtil.fetch(
                 'GET',
-                `${this.getHost}/swap/reverse`,
+                `${host}/swap/reverse`,
                 this.getHeaders
             );
             const status = response.info().status;
@@ -249,6 +262,9 @@ export default class SwapStore {
             console.error('Error fetching reverse swap fees:', e);
             this.apiError = localeString('views.Swaps.fetchFeesFailed');
         }
+        // only record the host once both fetches succeed, so a failed
+        // fetch is retried the next time the Swaps view gains focus
+        if (!this.apiError) this.fetchedRatesHost = host;
         this.loading = false;
     };
 
--- a/stores/SwapStore.ts
+++ b/stores/SwapStore.ts
@@ -1,4 +1,4 @@
-import { action, observable, computed, runInAction, reaction } from 'mobx';
+import { action, observable, computed, runInAction } from 'mobx';
 import ReactNativeBlobUtil from 'react-native-blob-util';
 import { ECPairAPI, ECPairFactory } from 'ecpair';
 import ecc from '@bitcoinerlab/secp256k1';
@@ -63,6 +63,9 @@ export default class SwapStore {
     @observable public swapsLoading = false;
     @observable public DERIVATION_PATH = 'm/44/0/0/0';
     @observable ECPair: ECPairAPI;
+    // host the swap rates were last fetched from; rates are only
+    // fetched from the Swaps view, never on startup
+    @observable public fetchedRatesHost: string | undefined;
 
     nodeInfoStore: NodeInfoStore;
     settingsStore: SettingsStore;
@@ -72,11 +75,6 @@ export default class SwapStore {
         this.settingsStore = settingsStore;
         initEccLib(ecc);
         this.ECPair = ECPairFactory(ecc);
-
-        reaction(
-            () => this.getHost,
-            () => this.getSwapFees()
-        );
     }
 
     @computed get claimMinerFee(): number {
@@ -188,6 +186,7 @@ export default class SwapStore {
     public getSwapFees = async () => {
         this.loading = true;
         this.apiError = '';
+        this.fetchedRatesHost = this.getHost;
         console.log(`Fetching fees from: ${this.getHost}`);
         try {
             const response = await ReactNativeBlobUtil.fetch(
--- a/views/Swaps/SwapDetails.tsx
+++ b/views/Swaps/SwapDetails.tsx
@@ -373,7 +373,8 @@ export default class SwapDetails extends React.Component<
                         !this.state.swapData?.lockupTransaction?.hex
                     ) {
                         const lockupTx = await SwapStore?.getLockupTransaction(
-                            createdResponse.id
+                            createdResponse.id,
+                            endpoint
                         );
                         this.setState((prevState) => ({
                             swapData: new Swap({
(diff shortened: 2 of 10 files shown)


shubhamkmr04 (commented):
2 issues: 
The first is that swap rates can get stuck at zero for a whole session. In `getSwapFees`, `fetchedRatesHost` is set at the top of the function, before the two network calls actually run, and it never gets cleared if they fail. So a failed fetch gets recorded as if it succeeded. If the user is offline (or just has a flaky moment) when they open Swaps, the fetch fails, and after that the new focus check `fetchedRatesHost !== getHost` is always false, so it never retries — the view sits at zero limits even once the network is back. You only recover by leaving Swaps entirely and coming back, since the mount path calls `getSwapFees` unconditionally. Setting `fetchedRatesHost` only after a successful fetch fixes it. Worth noting the same variable causes a second, smaller issue: settings are written asynchronously (`setSettings` only assigns `this.settings` after the storage write resolves), so if someone changes the provider in Swap Settings and immediately hits back, `getHost` may still be returning the old value when focus fires, so no re-fetch happens and the view ends up showing one provider's rates while swaps get created at another. The `reaction(() => this.getHost, ...)` that was removed from the constructor covered exactly that — bringing it back would handle both cases, and since it only fires when the host actually changes, it won't re-introduce a fetch on startup.

shubhamkmr04:
The second is that reverse swap claims can now be built with a miner fee of zero. SwapDetails uses SwapStore.claimMinerFee, which comes from reverseInfo, but after this change nothing populates reverseInfo unless the
user opens the Swaps view — getSwapFees is only called from views/Swaps/index.tsx now. So if someone force-quits the app with a reverse swap in flight, reopens, and taps that swap straight from Activity, componentDidMount subscribes to the websocket and auto-builds the claim transaction with minerFee: 0. The easiest way to confirm this one is by
commit: with the migration commit applied but this one reverted, users are already moved to a live Boltz host and Wallet.tsx still fetches at connect, so claimMinerFee is correct — reverting just this commit fixes it while keeping the migration. A lazy fetch (call getSwapFees from SwapDetails when reverseInfo is empty) would keep the "don't hit the swap server on boot" goal intact. Same root cause makes the claim miner fee display wrong in Activity and SwapsPane for older swaps that don't have claimMinerFee stored.

kaloudis:
Thanks for the thorough review, both catches are real. Pushed fixes for both.

**1. Rates stuck at zero**

Confirmed. `fetchedRatesHost` was recorded before the fetches ran, so a failed fetch was treated as a successful one and the focus check never retried. Fixed: `getSwapFees` now captures the host at the start and only assigns `fetchedRatesHost` after both fetches complete without an API error. A failed fetch leaves it unset, so the focus listener retries as soon as the view regains focus.

The provider-switch race is also confirmed: `setSettings` only assigns `this.settings` after the storage write resolves, so a fast back-tap can beat it. I brought the reaction back, but guarded: it only refetches when `fetchedRatesHost` is already set and differs from the new host. The unguarded version would have re-introduced the startup fetch for some users: at boot the store is constructed before settings load and before `nodeInfo.isTestNet` resolves, so `getHost` transitions from the mainnet Boltz default to the persisted host (custom host, SwapMarket, or the testnet default) and the reaction fires. That is exactly why the old reaction caused fetches at startup. With the guard, the reaction only fires once rates have actually been fetched in the session, which covers the settings race without any boot traffic. Also added a `loading` guard to the focus check so the reaction and the focus listener cannot double-fetch.

**2. Reverse claim miner fee**

Confirmed the mechanism: nothing populated `reverseInfo` on the Activity reopen path, and the claim used the store's computed `claimMinerFee`. One correction on severity though: the claim would not actually have been built with a zero miner fee. The mobile binding falls back to a fee rate when `minerFee <= 0` (ZeusLN/lnd `mobile/swaps.go` on `lnd-v0.21.2-beta-zeus`, the version pinned in `fetch-libraries-versions.json`):

```go
if minerFee > 0 {
    sats := uint64(minerFee)
    fee = boltz.Fee{Sats: &sats}
} else {
    satPerVbyte := float64(feeRate)
    fee = boltz.Fee{SatsPerVbyte: &satPerVbyte}
}
```

and `feeRate` is `Number(fee || 2)`, so on the reopen-from-Activity path (no `fee` route param) the claim would have gone out at 2 sat/vB instead of the quoted absolute fee. Still a regression worth fixing, since 2 sat/vB can be too low to confirm before the swap timeout in a high-fee environment, just not an unclaimable zero-fee transaction.

For the fix I went with your lazy-fetch direction rather than reverting the commit, since a revert would bring back the boot fetch this PR removes:

- The claim now prefers `swapData.claimMinerFee`, which `createReverseSwap` already persists on the swap record at creation time, so the normal reopen-from-Activity path works fully offline with the fee that was quoted when the swap was created.
- For legacy swaps that predate the stored field, `SwapDetails` lazily calls `getSwapFees()` on mount before subscribing to updates, so the store value is populated by claim time.

The receive-amount display in Activity/SwapsPane already prefers the stored per-swap fee; legacy swaps without it will show the correct amount once rates have been fetched in the session (visiting the Swaps view or opening the swap details).

shubhamkmr04 (approved):
tACK

stores/SwapStore.ts
ajaysehwal, line 1: @kaloudis  I think this breaks status polling for existing ZEUS swaps. Each saved swap already has its own `endpoint`, and the details/claim/refund flows use that stored endpoint. But `fetchAndUpdateSwaps()` always polls `this.getHost`.

Since this PR changes `this.getHost` to the Boltz API, existing ZEUS swaps will now be queried against the wrong server and can remain stuck with a stale status.

This is especially risky for time-sensitive submarine/reverse swaps, where users could miss a claim or refund window.

I think we should poll the swap's stored endpoint and only fall back to `this.getHost` for old records without an endpoint:

```ts
const host = swap.endpoint || this.getHost;

const response = await ReactNativeBlobUtil.fetch(
    'GET',
    `${host}/swap/${swap.id}`,
    this.getHeaders
);

File: views/Swaps/SwapDetails.tsx [hunk 1]
```diff
@@ -142,6 +142,16 @@ export default class SwapDetails extends React.Component<
...
             }
 
+            // legacy swaps predate claimMinerFee being stored at
+            // creation; fetch rates so the claim fee does not fall
+            // back to zero
+            if (
+                swapData.claimMinerFee == null &&
+                !this.props.SwapStore?.claimMinerFee
+            ) {
+                this.props.SwapStore?.getSwapFees();
```
ajaysehwal, line 152: @kaloudis  I think there’s still an endpoint issue for existing reverse swaps. If `claimMinerFee` isn’t stored, `SwapDetails` calls `SwapStore.getSwapFees()`, which always uses the global `getHost` (now Boltz). For a ZEUS-created swap, this can fetch Boltz’s current fee instead of the fee expected by the ZEUS swap server.

That could cause the claim to underpay and get stuck, or overpay and lose extra sats.

I think `getSwapFees()` should use the swap’s stored `endpoint` (or `route.params.endpoint`) instead of the global host. Also, the claim flow shouldn’t rely on the fire-and-forget fee request completing in time.

stores/SwapStore.ts
kaloudis: Good catch, thank you. Fixed in 0edd333. Confirmed: `fetchAndUpdateSwaps` polled `this.getHost` while every saved swap persists its own `endpoint`, and the details/claim/refund flows already use the stored endpoint. The polling behavior predates this PR (it was only reachable before by manually switching providers with a swap in flight), but the host migration here would have made it bite every legacy pending ZEUS swap on first launch, so it needs to be fixed in this PR. Applied your suggestion as written: `const host = swap.endpoint || this.getHost;` with the `getHost` fallback covering any record predating the `endpoint` field.

While verifying this I found a second instance of the same class: `getLockupTransaction` also queried `this.getHost`, and it is called from SwapDetails and from the Refund view's lockup prefetch. For a legacy submarine swap whose lockup transaction was never persisted, that would have queried the wrong server and broken the refund flow outright, the exact time-sensitive scenario you flagged. It now takes an optional `endpoint` param (falling back to `this.getHost`), and both call sites pass the swap's stored endpoint.

One pre-existing oddity I noticed but left alone as out of scope: `getLockupTransaction` writes the HTTP status code (`200`) into `swap.status` when persisting the lockup transaction. Happy to open a follow-up issue for that.

views/Swaps/SwapDetails.tsx
kaloudis: Thanks for taking another close look. This one is half right, and I have addressed the half that holds up in 8696233.

On the endpoint question I do not think there is an issue. `claimMinerFee` is not a value the swap server expects or enforces: it is the on-chain fee of the client-built claim transaction, subtracted from the locked amount when the native code builds the sweep. The right value depends only on current Bitcoin mempool feerates, which are the same no matter which swap server created the swap, so Boltz's fresh estimate is perfectly applicable to a ZEUS-created swap (and being current, it is arguably better than whatever the ZEUS server quoted at creation time). Pointing `getSwapFees()` at the swap's stored endpoint would also cut the other way twice: the retired server will eventually go dark, killing rates fetches exactly when legacy users need to claim, and `getSwapFees()` populates store-wide observables (`subInfo`, `reverseInfo`, `fetchedRatesHost`) that the Swaps view consumes for the currently selected provider, so fetching from a per-swap legacy endpoint would poison that state and break the host-change refetch logic from the last round.

The fire-and-forget point is a fair catch though. The websocket can deliver `transaction.mempool` before the mount-time prefetch resolves, and the miner fee is read when the claim is built, so a fast update could still produce a 2 sat/vB sweep for a legacy swap, which is exactly what the earlier fix was meant to prevent. `createReverseClaimTransaction` now resolves the miner fee before building: if neither the swap nor the store has one, it awaits a rates fetch and re-reads. The mount-time prefetch stays as a warm-up so the await is usually instant, and if the fetch fails (for example claiming offline) the existing `feeRate` fallback still applies, unchanged from current behavior.

ajaysehwal (approved):
tACK

PR #4406, opened 2026-08-13
ui: KeyValue: add copy icon and make tap copy values

kaloudis:
# Description

|-|-|-|
|-|-|-|
|Simulator Screenshot - iPhone 17 Pro - 2026-08-12 at 22 02 53|Simulator Screenshot - iPhone 17 Pro - 2026-08-12 at 22 02 45|Simulator Screenshot - iPhone 17 Pro - 2026-08-12 at 22 02 40|

Copying a value from a detail screen (payment hash, preimage, address, pubkey, etc.) currently requires a long press on the value, which is an invisible gesture most users never discover, and any string value is silently copyable. This PR makes copyability explicit and visible:

- `KeyValue` gets a `showCopyIcon` prop that renders the existing `Copy.svg` next to the value (16px, `secondaryText` stroke), with its own touch target, `hitSlop`, and an accessibility label reusing the existing `components.CopyButton.copy` locale key
- The icon is now the copyability contract: rows with `showCopyIcon` copy on a regular press (existing "Copied!" toast + vibration feedback); rows without it are not copyable at all, and long press is removed
- Rows with a `mempoolLink` keep press = open block explorer; the copy icon is the copy affordance there, and long-press copy is retained as a fallback
- `showCopyIcon` is applied across the detail views to identifier-type fields: payment hashes, preimages, payment requests, tx ids, block hashes, addresses, pubkeys, node URIs, channel ids/points, order/swap IDs, nostr keys, and mint/relay/watchtower URLs. Amounts, fees, dates, statuses, and other display-only fields are no longer copyable
- The masked nsec/privkey rows in the NostrKeys views pass `disableCopy` while hidden, so the icon only appears (and copy only works) once revealed
- `flexShrink` added to the value text so long values do not push the icon off-screen

Lurker mode behavior is unchanged: no copying, and the icon does not render.

Screenshots to follow with device testing.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Device and backend testing pending; will check off before marking ready for review.

### Locales

No new locale strings; the icon's accessibility label reuses `components.CopyButton.copy`.

### Other:

### DIFF
--- a/components/FeeBreakdown.tsx
+++ b/components/FeeBreakdown.tsx
@@ -454,6 +454,7 @@ export default class FeeBreakdown extends React.Component<
                                 value={channelId}
                                 color={themeColor('chain')}
                                 sensitive
+                                showCopyIcon
                                 mempoolLink={() =>
                                     UrlUtils.goToBlockExplorerChannelId(
                                         channelId,
@@ -470,6 +471,7 @@ export default class FeeBreakdown extends React.Component<
                                 value={channelPoint}
                                 color={themeColor('highlight')}
                                 sensitive
+                                showCopyIcon
                                 mempoolLink={() =>
                                     UrlUtils.goToBlockExplorerTXID(
                                         channelPoint,
--- a/components/KeyValue.tsx
+++ b/components/KeyValue.tsx
@@ -90,17 +90,23 @@ export default class KeyValue extends React.Component<KeyValueProps> {
         const lurkerMode: boolean =
             SettingsStore?.settings?.privacy?.lurkerMode || false;
 
-        // lurker mode only redacts sensitive values; copying or opening a
-        // link on a redacted value would expose the real value
+        // lurker mode only redacts sensitive values; copying a redacted
+        // value would put the real value on the clipboard, so copy actions
+        // are disabled for redacted values. mempoolLink stays active:
+        // navigation is an explicit act and remains available elsewhere in
+        // lurker mode (e.g. the note edit button, custom explorer links)
         const valueHidden = lurkerMode && sensitive;
 
         {
             /* TODO: rig up RTL */
         }
+        // long press copies any copyable value; the icon and tap-to-copy
+        // are opt-in per call site
         const isCopyable =
             !disableCopy &&
-            !!showCopyIcon &&
+            !valueHidden &&
             (typeof value === 'string' || typeof value === 'number');
+        const showIcon = isCopyable && !!showCopyIcon;
         const rtl = false;
         const KeyBase = (
             <Body>
@@ -183,7 +189,7 @@ export default class KeyValue extends React.Component<KeyValueProps> {
                         ? PrivacyUtils.sensitiveValue({ input: value })
                         : value}
                 </Text>
-                {isCopyable && !valueHidden && (
+                {showIcon && (
                     <Pressable
                         onPress={() => this.copyText()}
                         onPressIn={() => this.setValueOpacity(0.2)}
@@ -209,15 +215,17 @@ export default class KeyValue extends React.Component<KeyValueProps> {
         );
 
         let Value: any;
-        if (!valueHidden && (isCopyable || mempoolLink)) {
+        if (isCopyable || mempoolLink) {
             Value = (
                 <Pressable
-                    onPress={mempoolLink ? mempoolLink : () => this.copyText()}
-                    onLongPress={
-                        mempoolLink && isCopyable
+                    onPress={
+                        mempoolLink
+                            ? mempoolLink
+                            : showIcon
                             ? () => this.copyText()
                             : undefined
                     }
+                    onLongPress={isCopyable ? () => this.copyText() : undefined}
                     onPressIn={() => this.setValueOpacity(0.2)}
                     onPressOut={() => this.setValueOpacity(1)}
                 >
--- a/components/KeyValue.tsx
+++ b/components/KeyValue.tsx
@@ -90,6 +90,10 @@ export default class KeyValue extends React.Component<KeyValueProps> {
         const lurkerMode: boolean =
             SettingsStore?.settings?.privacy?.lurkerMode || false;
 
+        // lurker mode only redacts sensitive values; copying or opening a
+        // link on a redacted value would expose the real value
+        const valueHidden = lurkerMode && sensitive;
+
         {
             /* TODO: rig up RTL */
         }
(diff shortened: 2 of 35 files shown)


myxmaster:
Very nice idea!

I noticed that a long press on the note text (on the transaction details screen) now navigates to "Update Note". Before, a long press (on the actual note text, not the label "Note"!) copied the note text. This is probably not the only example where behaviour is changed.

And then this PR revealed a pre existing bug:
With lurker mode enabled, the copy icon is gone even for content that is not protected, for example "Channel ID" in channel details or "Block Hash" on Node Info screen (before this PR, the copy icon was obviously never there, but copy functionality for long press was also gone with lurker mode enabled).

kaloudis:
> Very nice idea!
> 
> I noticed that a long press on the note text (on the transaction details screen) now navigates to "Update Note". Before, a long press (on the actual note text, not the label "Note"!) copied the note text. This is probably not the only example where behaviour is changed.
> 
> And then this PR revealed a pre existing bug: With lurker mode enabled, the copy icon is gone even for content that is not protected, for example "Channel ID" in channel details or "Block Hash" on Node Info screen (before this PR, the copy icon was obviously never there, but copy functionality for long press was also gone with lurker mode enabled).

1. Note long press: confirmed. The transaction details note and label were wrapped in their own `TouchableOpacity`, causing the inner press handler to disappear and and long presses to fall through to the navigation handler.

Both are migrated same pattern `Payment` and `Invoice` use. Tap opens `Update Note`, long press or the copy icon copies the text. Bonus: tapping directly on the note text now also opens `Update Note`. Previously, taps on the text itself did nothing and only the label area worked.

2. Lurker mode: good catch, and confirmed pre-existing: the old code gated all copying on lurker mode regardless of the sensitive flag. Now links are only disabled when the value is actually redacted. Non-sensitive fields like Channel ID and Block Hash keep their copy icon in lurker mode.

myxmaster:
re 1:
With lurker mode enabled, tapping the note row now does nothing at all (the pencil icon in the header still navigates to "Update Note"). Not sure if that is wanted behaviour.
There is a bigger/structural issue though: `isCopyable` still depends on `showCopyIcon`, so copying remains opt-in, so lots of examples that used to be copyable by long press no longer are.
AI says:
"it would be easy to keep long press working everywhere and only gate the icon and the tap-to-copy on the prop:
``` 
    const isCopyable =
        !disableCopy &&
        (typeof value === 'string' || typeof value === 'number');
    const showIcon = isCopyable && !!showCopyIcon;
```
"

re 2:
One case the new `valueHidden` gate misses: the three rows under the "Aliases" accordion in channel details (Alias SCIDs, Peer Alias SCID, 0-conf Confirmed SCID) don't use the `sensitive` prop, they redact inline by passing `PrivacyUtils.sensitiveValue({...})` as the value. So `valueHidden` is false, the copy icon stays visible in lurker mode, and tapping it puts the ancient Greek placeholder into the clipboard.
AI says (I didn't test/look into it further):
"Same pattern (inline `PrivacyUtils` without the prop) also exists in
Mint.tsx, Transaction.tsx and TxHex.tsx. Those have no `showCopyIcon`
today, so they're fine for now, but the trap is armed."

kaloudis:
Both addressed, thank you.

re 1: took the structural suggestion. `isCopyable` no longer depends on `showCopyIcon`: long press copies any string or number value again, matching the old behavior everywhere, and only the icon and tap-to-copy are opt-in per call site.

For the note row in lurker mode: tapping now navigates to Update Note again. I made `mempoolLink` actions stay active even when the value is redacted, because lurker mode never consistently blocked navigation: the header pencil still opens Update Note, and custom explorer-link rows like the tx hash on the transaction details screen always kept working in lurker mode. Copy actions (icon, tap, long press) remain disabled for redacted values, since those would silently put the real value on the clipboard.

re 2: the three alias rows in channel details now pass the raw SCIDs with the `sensitive` prop instead of redacting inline, so the display is identical but KeyValue's lurker-mode gate applies: no copy icon, and nothing copies the ancient Greek placeholder.

On the armed trap in `Mint.tsx`, `Transaction.tsx` and `TxHex.tsx`: I checked all of them, plus `PaymentPath.tsx` and `Invoice.tsx`. Those inline redactions all feed either plain `Text` elements or JSX passed as the KeyValue `value` (their own `TouchableOpacity` with their own handlers). KeyValue's copy path only engages for string and number values, so even with universal long press restored they can't copy a placeholder. The three channel rows were the only string-valued inline redactions reaching KeyValue.

ajaysehwal (commented):
Image

@kaloudis , can we add copy button for bolt12 address as well?

ajaysehwal:
Simulator Screenshot - iPhone 17 Pro - 2026-08-18 at 19 08 23


nit: use copy icon here also

myxmaster:
Fixes look good, tACK https://github.com/ZeusLN/zeus/commit/261bf1ae54ed799d5722b8ec65997b12ae7bf9f1

And I agree with @ajaysehwal, copy icon would be useful there.

kaloudis:
Good call. Added in 06a96f082: the Source/Destination Channel ID rows now show the copy icon, matching the single channel view.

Two details worth noting:
- When the channel is unknown (your screenshot's case) the row now falls back to the raw channel ID rather than the alias map entry, so what you copy always matches the label.
- When the channel is still open, the value is the tappable alias link that navigates to the Channel view, and KeyValue only copies string values, so no icon renders there. The ID is one tap away on that screen with its own copy icon. Letting element-valued rows copy an underlying string (a `copyValue` prop) is on the follow-up list along with the Transaction view's tx/block hash rows.

kaloudis:
@ajaysehwal  Re: the BOLT 12 address screen (review above): good idea, but that screen deserves more than a bare copy button and this PR is scoped to KeyValue rows, so I've split it out to #4444. The plan there is to match the ZEUS Pay Lightning address QR view: expanded QR with copy/share, NFC on Android, and Address/Offer tabs.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4207, opened 2026-07-01
fix: SyncStore: fallback block height APIs with timeout

ajaysehwal:
## Summary

Fixes embedded LND sync UI getting stuck showing `isSyncing: true` even when the node is already caught up and internet is available.

Sync progress depends on fetching the chain tip (`bestBlockHeight`) from an external Esplora-compatible API. Previously this only hit `mempool.space`, which could hang on rate limits or fail entirely — leaving `bestBlockHeight` stale and `numBlocksUntilSynced > 0` indefinitely.

## Problem

- Sync state used a single `mempool.space` endpoint for chain tip height
- Rate-limited or slow mempool responses could block sync progress indefinitely
- `isSyncing` stayed true, disabling wallet actions and showing a misleading sync banner

## Solution

- Add fallback block height APIs with ordered retry
- Add a 5s timeout per request (same as `ConnectivityStore`) so slow endpoints don't block fallbacks
- 
## Changes

**`stores/SyncStore.ts`**

| Network | Fallback order |
|---|---|
| Mainnet | mempool.space → blockstream.info → mempool.emzy.de |
| Testnet | mempool.space → blockstream.info |
| Testnet4 | mempool.space |

- `getBestBlockHeight()` tries each URL in order until one succeeds
- On total failure: sets `this.error = true` and returns (no throw)
- `setSyncInfo` unchanged: if `this.error`, sets `bestBlockHeight = currentBlockHeight` so sync can complete
- Initial sync loop retries on `!this.bestBlockHeight` instead of try/catch

## Test plan

- [ ] Start embedded LND wallet on mainnet — sync banner clears when node is caught up
- [ ] Verify sync banner clears when node reports `synced_to_chain.`
- [ ] Repeat on testnet
- [ ] Confirm wallet actions (send/receive) are not blocked after sync completes
- [ ] Confirm sync UI still shows progress while node is genuinely behind chain tip

This pull request is categorized as a:

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

[...]

PR #4407, opened 2026-08-13
SyncStore: estimate chain tip from node headers when mempool instance is unreachable

kaloudis:
# Description

Previously stacked on #4227; now that it has merged, this PR is rebased onto master and contains a single commit.

Fixes the embedded LND Sync view wedging at 0% (height 0, tip 0, "1 block until synced") when the configured mempool instance is unreachable, even while the node itself syncs fine. Found while device-testing #4401 on a network where mempool.space is blocked: `startSyncing` retried the tip fetch forever, never consulted GetInfo, and `isSyncing` never cleared even after the node reached `synced_to_chain`.

Follow-up to the discussion in #4207, which was closed in favor of #4227. #4227 fixes which instance is queried; this fixes the loop shape so no single unreachable endpoint can hold the sync UI hostage. No third-party fallback endpoints are added: the fallback is the node itself.

Changes in `stores/SyncStore.ts` + new `utils/SyncUtils.ts`:

- Initial tip fetch is bounded to 3 attempts (3s apart), then the sync loop proceeds.
- With no external tip, the tip is estimated from the node's own `best_header_timestamp`, assuming one block per ten minutes of header lag. During initial header sync the estimate tracks the real tip closely; once headers catch up it converges on the node's height.
- While on an estimated tip, only `synced_to_chain` can declare the sync complete, since the estimate can undershoot. With a real tip, completion behaves exactly as before.
- The periodic 30s tip query now recovers from estimated mode: on success it swaps the real tip back in and clears the error state.
- Hardening: a 200 response with a non-numeric body (captive portals) is treated as a failure, and `setSyncInfo` no longer throws out of the sync loop when the tip query rejects.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


stores/SyncStore.ts
```diff
@@ -85,16 +89,28 @@ export default class SyncStore {
...
             this.updateProgress();
         }
 
-        if (nodeInfo?.synced_to_chain || this.numBlocksUntilSynced <= 0) {
+        // An estimated tip can undershoot the real one, so while the mempool
+        // instance is unreachable only the node's synced_to_chain may declare
+        // the sync complete
+        if (
+            nodeInfo?.synced_to_chain ||
+            (!this.error && this.numBlocksUntilSynced <= 0)
```
ajaysehwal, line 112: @kaloudis I think we should keep `synced_to_chain` as the only completion signal while `error` is set.

The `numBlocksUntilSynced <= 0` estimator can be inaccurate, so using it during an error state could mark the node as synced too early while Neutrino is still catching up.

With `error === true`, I think only LND's `synced_to_chain` should clear `isSyncing`. Once the mempool connection recovers and `error` is cleared, the existing height-based completion can resume.

kaloudis: The condition already works exactly this way. The `!this.error` guard on this line disables the `numBlocksUntilSynced <= 0` clause whenever the mempool instance is unreachable, so `synced_to_chain` is the only completion signal while `error` is set. Once a mempool fetch succeeds, `getBestBlockHeight` clears `error` and replaces the estimated tip with the real one in the same handler, so height-based completion only ever resumes against a real tip. The comment block above the condition documents this rationale.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

Issue #3642, opened 2026-02-01
LNC: `hasPerms` call returns false for all permissions after upgrade to Embedded LND v0.20.0-beta-zeus


kaloudis:
All calls in this block will return `false`, regardless of the permissions granted to your LNC session:

```
    checkPerms = async () => {
        this.permOpenChannel = await this.lnc.hasPerms(
            'lnrpc.Lightning.OpenChannel'
        );
        this.permSendCoins = await this.lnc.hasPerms(
            'lnrpc.Lightning.SendCoins'
        );
        this.permSendLN = await this.lnc.hasPerms(
            'routerrpc.Router.SendPaymentV2'
        );
        this.permNewAddress = await this.lnc.hasPerms(
            'lnrpc.Lightning.NewAddress'
        );
        this.permImportAccount = await this.lnc.hasPerms(
            'walletrpc.WalletKit.ImportAccount'
        );
        this.permForwardingHistory = await this.lnc.hasPerms(
            'lnrpc.Lightning.ForwardingHistory'
        );
        this.permSignMessage = await this.lnc.hasPerms(
            'signrpc.Signer.SignMessage'
        );
    };
```

https://github.com/ZeusLN/zeus/blob/master/backends/LightningNodeConnect.ts#L68

Upon testing, it appears that these calls has the proper permissions according to their LNC session.

This has resulted in users in seeing `Read-only wallet` warnings in their wallet headers and causing confusion.

Until remedied, we will mark all permissions as `true`, degrading our permission checking functionality, but ensuring LNC users continue to have an uninterrupted experience.

kaloudis:
Workaround pushed in https://github.com/ZeusLN/zeus/commit/3366b07199d8e8f7868935c2312718ed3f599e14

PR #4411, opened 2026-08-13
fix(nwc): keep sign_message off full-access by default

ajaysehwal:
# Description

NWC `sign_message` signed arbitrary client text with the Lightning node identity key. That method was included in the default full-access grant, so a normal pairing string became a remote signing oracle (LNURL-auth-style logins, gossip, anything that trusts this node’s signatures).
This PR makes `sign_message` opt-in:
- Drop it from the Full Access preset. New connections do not get it unless the user enables it.
- Still advertise it on the wallet-info event so clients can request it when granted.
- Show a warning and confirm before enabling the permission.
- Reject missing/empty/non-string `message` with `UNKNOWN`; do not call `SignMessage`.
- Existing connections that already have `sign_message` keep it.
Connection expiry was already enforced on every subscribed method (including `sign_message`) via `withGlobalHandler`. This change does not add per-request signing prompts; NWC is meant to run unattended once granted.

Simulator Screenshot - iPhone 17 Pro - 2026-08-13 at 20 27 18
Simulator Screenshot - iPhone 17 Pro - 2026-08-13 at 20 27 07

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [x] LDK Node
- [x] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


kaloudis:
utACK e130062bb

Reviewed the branch and verified locally at head `e130062bb`: both touched Jest suites pass (134 tests) and `yarn tsc` is clean. Have not device-tested.

The approach is the right call: the default pairing no longer hands out a node-identity signing oracle, opt-in requires an explicit warning acknowledgment, and the empty-message validation closes the malformed-request path. I also checked the edge behavior and it holds up:

- Preset round-trips preserve an existing `sign_message` grant (`withPreservedSignMessage`), and `determinePermissionType` strips it before matching, so old full-access connections still classify correctly.
- `checkForChanges` compares sorted arrays, so the append-at-end ordering can't cause phantom "has changes"; regenerate passes `connection.permissions` through, so grants survive secret rotation.
- When the handler isn't registered (granted but backend can't sign), the SDK replies with spec-compliant `NOT_IMPLEMENTED`, so requests fail cleanly.

A few asks before merge:

1. **`Nip47ErrorCode.UNKNOWN` is not a NIP-47 error code** (`utils/NostrConnectUtils.ts`, used in `handleSignMessage`). The spec's bucket for an invalid request payload is `OTHER`; a conformant client won't know what `UNKNOWN` means. The enum already deviates in legacy spots, but let's not add a new nonstandard code.
2. **`handleGetInfo` advertises unfiltered permissions** (`methods: connection.permissions`). A connection holding `sign_message` on a backend that can't sign advertises a method whose handler was never registered. Filtering with the same `supportsMessageSigning()` gate would keep `get_info` consistent with the handler set. Low priority since connections are node-bound, but the LNC read-only-pairing case can hit it.
3. **LNC timing question**: `LightningNodeConnect.supportsMessageSigning()` returns `this.permSignMessage`, which is populated asynchronously after LNC connects. Can the NWC service subscribe before the permission check completes? If so `sign_message` would be silently dropped until the next resubscribe. Worth a quick confirmation on an LNC pairing.
4. Nit: in `getInvoiceDetails` the `catch (decodeError) { throw decodeError; }` now does nothing after the `console.error` removal; the try/catch can go entirely.
5. Nit, copy: "read my connection transaction history" reads awkwardly; suggest "read this connection's transaction history". (The `fullAccess` -> `walletAccess` key rename is the right pattern, stale Transifex translations will fall back to English.)

Process items:

- Please tick the Locales checklist items; this PR adds new `en.json` strings that need to go to Transifex.
- The Testing section should have "Yes" ticked (unit tests were added), and an Android pass is still owed before merge.
- Please squash the four progressive commits into one before merge.

One deliberate tradeoff worth keeping in mind (stated in the PR body): already-paired apps keep their existing `sign_message` grant, so this doesn't revoke access from a previously paired client. Since the connection details view now labels the permission, users can see and remove it, which seems like the right balance against silently breaking working setups.

ajaysehwal:
> utACK [e130062](https://github.com/ZeusLN/zeus/commit/e130062bb86e55da7687bca3a42a7d9946c6a032)
> 
> Reviewed the branch and verified locally at head `e130062bb`: both touched Jest suites pass (134 tests) and `yarn tsc` is clean. Have not device-tested.
> 
> The approach is the right call: the default pairing no longer hands out a node-identity signing oracle, opt-in requires an explicit warning acknowledgment, and the empty-message validation closes the malformed-request path. I also checked the edge behavior and it holds up:
> 
> * Preset round-trips preserve an existing `sign_message` grant (`withPreservedSignMessage`), and `determinePermissionType` strips it before matching, so old full-access connections still classify correctly.
> * `checkForChanges` compares sorted arrays, so the append-at-end ordering can't cause phantom "has changes"; regenerate passes `connection.permissions` through, so grants survive secret rotation.
> * When the handler isn't registered (granted but backend can't sign), the SDK replies with spec-compliant `NOT_IMPLEMENTED`, so requests fail cleanly.
> 
> A few asks before merge:
> 
> 1. **`Nip47ErrorCode.UNKNOWN` is not a NIP-47 error code** (`utils/NostrConnectUtils.ts`, used in `handleSignMessage`). The spec's bucket for an invalid request payload is `OTHER`; a conformant client won't know what `UNKNOWN` means. The enum already deviates in legacy spots, but let's not add a new nonstandard code.
> 2. **`handleGetInfo` advertises unfiltered permissions** (`methods: connection.permissions`). A connection holding `sign_message` on a backend that can't sign advertises a method whose handler was never registered. Filtering with the same `supportsMessageSigning()` gate would keep `get_info` consistent with the handler set. Low priority since connections are node-bound, but the LNC read-only-pairing case can hit it.
> 3. **LNC timing question**: `LightningNodeConnect.supportsMessageSigning()` returns `this.permSignMessage`, which is populated asynchronously after LNC connects. Can the NWC service subscribe before the permission check completes? If so `sign_message` would be silently dropped until the next resubscribe. Worth a quick confirmation on an LNC pairing.
> 4. Nit: in `getInvoiceDetails` the `catch (decodeError) { throw decodeError; }` now does nothing after the `console.error` removal; the try/catch can go entirely.
> 5. Nit, copy: "read my connection transaction history" reads awkwardly; suggest "read this connection's transaction history". (The `fullAccess` -> `walletAccess` key rename is the right pattern, stale Transifex translations will fall back to English.)
> 
> Process items:
> 
> * Please tick the Locales checklist items; this PR adds new `en.json` strings that need to go to Transifex.
> * The Testing section should have "Yes" ticked (unit tests were added), and an Android pass is still owed before merge.
> * Please squash the four progressive commits into one before merge.
> 
> One deliberate tradeoff worth keeping in mind (stated in the PR body): already-paired apps keep their existing `sign_message` grant, so this doesn't revoke access from a previously paired client. Since the connection details view now labels the permission, users can see and remove it, which seems like the right balance against silently breaking working setups.

fixed in 2ac2525c2ce0692f71389b9b6efc567a83e7881b

kaloudis:
utACK 2ac2525c2

Re-reviewed the new round: all five asks from my last pass are addressed. Verified locally at head `2ac2525c2`: both touched Jest suites pass and `yarn tsc` is clean.

- `UNKNOWN` is renamed to the spec-compliant `OTHER`, with no stray references left behind.
- `get_info` now advertises through `advertisedConnectionMethods()`, which drops `sign_message` when the backend can't sign. That makes all three surfaces consistent: handler registration, the 13194 info event, and per-connection `get_info`. The new tests cover it.
- The LNC answer is the honest one: `permSignMessage` defaults to `true`, which matches what `checkPerms` hardcodes today under the ZEUS-3642 workaround (#3642), so the change only closes the undefined window before `checkPerms` completes and is behavior-preserving in steady state. The comment correctly flags that `hasPerms` must be restored there before the flag becomes meaningful.
- Dead try/catch removed; copy fixed. Checklist items are ticked.

One optional nit, non-blocking: the new "omits sign_message from get_info" test restores the `supportsMessageSigning` mock inline at the end, so if the assertion throws, the mock stays `false` for any test added after it. A `try/finally` or an `afterEach` restore would be sturdier.

Remaining before merge:
- Android device pass (checklist item still unticked).
- Squash the five commits into one.

ajaysehwal:
> utACK [2ac2525](https://github.com/ZeusLN/zeus/commit/2ac2525c2ce0692f71389b9b6efc567a83e7881b)
> 
> Re-reviewed the new round: all five asks from my last pass are addressed. Verified locally at head `2ac2525c2`: both touched Jest suites pass and `yarn tsc` is clean.
> 
> * `UNKNOWN` is renamed to the spec-compliant `OTHER`, with no stray references left behind.
> * `get_info` now advertises through `advertisedConnectionMethods()`, which drops `sign_message` when the backend can't sign. That makes all three surfaces consistent: handler registration, the 13194 info event, and per-connection `get_info`. The new tests cover it.
> * The LNC answer is the honest one: `permSignMessage` defaults to `true`, which matches what `checkPerms` hardcodes today under the ZEUS-3642 workaround ([LNC: `hasPerms` call returns false for all permissions after upgrade to Embedded LND v0.20.0-beta-zeus #3642](https://github.com/ZeusLN/zeus/issues/3642)), so the change only closes the undefined window before `checkPerms` completes and is behavior-preserving in steady state. The comment correctly flags that `hasPerms` must be restored there before the flag becomes meaningful.
> * Dead try/catch removed; copy fixed. Checklist items are ticked.
> 
> One optional nit, non-blocking: the new "omits sign_message from get_info" test restores the `supportsMessageSigning` mock inline at the end, so if the assertion throws, the mock stays `false` for any test added after it. A `try/finally` or an `afterEach` restore would be sturdier.
> 
> Remaining before merge:
> 
> * Android device pass (checklist item still unticked).
> * Squash the five commits into one.

fixed in dcd50ce1ceff2fd8a41d5bffca00593d6b5a9e8c

kaloudis:
re-utACK dcd50ce1c

Compared the squashed commit against my last ACK at `2ac2525c2` (diff-of-diffs, same merge base `6c55fc3b8`): production code is byte-identical. The only change is in `stores/NostrWalletConnectStore.test.ts`, where the `supportsMessageSigning` mock reset moved from the tail of a test into an `afterEach` hook. That's a small hygiene improvement, since the mock now resets even if the backend-cannot-sign test fails midway.

Verified locally at `dcd50ce1c`: both touched Jest suites pass (135 tests) and `yarn tsc` is clean.

kaloudis (approved):

PR #4414, opened 2026-08-14
ui: Sweepremoteclosed: fix fee fetch, loading and button states

kaloudis:
# Description

Fixes three issues on the sweep remote closed view (Settings > Embedded node > Chantools > Sweep remote closed), most noticeable when running it against LDK Node:

1. **Recommended fee rate is now fetched by default.** The view initialized the fee field with a hardcoded `21` sat/vB. `OnchainFeeInput` skips the mempool recommended-rate fetch whenever it is handed an already-valid fee, so the fetch never ran. The fee state now starts empty, which triggers the fetch of the user's preferred mempool rate on mount. With mempool rates disabled, the plain numeric input behaves as before.
2. **Loading indicator moved to the top right of the header.** Replaces the indicator that rendered below the header, using the same `rightComponent` pattern as `NeutrinoPeers`.
3. **Start button is disabled while a sweep is processing.** `disabled` is now `!sweepAddr || loading`. While wiring this up, found that `loading` was never reset on the success path (only in the catch), so after a successful sweep the spinner (and now the disabled button) would persist when navigating back from the TxHex view. Both success branches (LDK Node and chantools) now reset `loading` before navigating.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Settings/EmbeddedNode/Chantools/Sweepremoteclosed.tsx
+++ b/views/Settings/EmbeddedNode/Chantools/Sweepremoteclosed.tsx
@@ -90,7 +90,7 @@ export default class Sweepremoteclosed extends React.Component<
         sweepAddr: '',
         apiUrl: 'https://api.node-recovery.com',
         customApiUrl: '',
-        feeRate: '21',
+        feeRate: '',
         recoveryWindow: '200',
         sleepSeconds: '0',
         loading: false,
@@ -135,10 +135,16 @@ export default class Sweepremoteclosed extends React.Component<
                                     fontFamily: 'PPNeueMontreal-Book'
                                 }
                             }}
+                            rightComponent={
+                                loading ? (
+                                    <View>
+                                        <LoadingIndicator size={30} />
+                                    </View>
+                                ) : undefined
+                            }
                             navigation={navigation}
                         />
                         {error && <ErrorMessage message={error} dismissable />}
-                        {loading && <LoadingIndicator />}
                         <ScrollView style={{ margin: 10 }}>
                             {isLdkNode && (
                                 <Text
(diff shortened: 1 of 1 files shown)


shubhamkmr04 (changes requested):
If the mempool fetch fails, OnchainFeeInput sets the fee to the string '0' on error. In JavaScript '0' counts as "truthy", so feeRate || 2 does not fall back to 2, it gives 0 sat/vB. 

Neither could happen before, because the hardcoded '21' meant the field was never empty and the fetch never mattered.

Fix: use Number(feeRate) || 2 instead of Number(feeRate || 2), and ideally also block the Start button when the fee is invalid via OnchainFeeInput's onFeeError.

kaloudis:
Good catch, thanks. Confirmed: OnchainFeeInput's catch handler calls onChangeFee('0'), and since the string '0' is truthy, Number(feeRate || 2) resolved to 0 sat/vB. This path was unreachable before this PR because the hardcoded '21' made the component skip the fetch entirely.

Pushed 9ae8e31d3 with both suggestions:

1. Both sweep calls now use Number(feeRate) || 2, so '0' and empty string fall back to 2.
2. Wired up onFeeError into a feeLoadingError state field, following the existing WIFSweeper pattern (set on fetch error, cleared in onChangeFee so recovering via EditFee re-enables the button).

For the Start button I went slightly broader than the error flag alone: disabled is now !sweepAddr || loading || feeLoadingError || !(Number(feeRate) > 0). The validity check also covers a manually typed 0 or an empty field in the plain-input mode (which would otherwise silently sweep at 2 sat/vB while the placeholder shows 10), and it keeps the button disabled while the mempool fetch is still in flight, since feeRate is empty until the rate arrives. Given this is sweeping force-close funds, refusing to build a 0 sat/vB transaction seemed like the right call across the board.

shubhamkmr04 (approved):
tACK

PR #4424, opened 2026-08-14
chore: bump CDK to 0.17.4; drop npubcash quote-key marker workaround

### LINKED ISSUE #2335: wallet: mint saga aborts with ConcurrentUpdate after the mint has issued signatures when a quote's secret_key equals seed[..32] (npubcash legacy scrub invalidates the saga's quote snapshot)
## Summary

With the `npubcash` feature enabled (as in the released cdk-ffi binaries), `Wallet::mint` deterministically fails with `Error::Cdk(50000, "Concurrent update detected")` for any mint quote whose stored `secret_key` is byte-identical to the first 32 bytes of the wallet seed. The failure occurs **after** the mint has already issued the blind signatures, so the quote is consumed on the mint side (state `ISSUED`) while no proofs, transaction, or balance are recorded locally. The funds are only recoverable through a NUT-13 restore scan.

The trigger is not limited to npub.cash quotes. Any application that derives an external quote's NUT-20 key as `seed[0..32]` collides with `is_legacy_npubcash_secret_key`, which treats byte equality with the seed prefix as proof that the key came from cdk's legacy npub.cash scheme. ZEUS hit this on every ZEUS Pay lightning address redemption after upgrading from cdk-ffi 0.14.2 to 0.17.3: ZEUS Pay has locked quotes to the wallet's seed prefix since before cdk had npubcash support, so every redemption stranded the funds until restore.

Genuine legacy npub.cash quotes that carry the stored key should hit the identical crash, since the scrub runs inside the mint saga itself.

## Versions

- cdk 0.17.3 (via cdk-kotlin 0.17.3 and cdk-swift 0.17.3, both of which build cdk-ffi with `npubcash` enabled)
- Reproduced on Android (JVM bindings) and applicable to iOS (same Rust path)
- Wallet store: `WalletSqliteDatabase` (passed as `WalletStore::Custom`; the built-in sqlite store takes the same code path)

## Reproduction

No npub.cash account is needed; only a quote whose `secret_key` equals the seed prefix.

1. Create a `WalletRepository` (or `Wallet`) from a mnemonic. Let `seed` be the BIP39 seed (empty passphrase); let `k = SecretKey::from_slice(&seed[..32])`.
2. At any real mint, create a bolt11 mint quote locked to `k.public_key()` (NUT-20 `pubkey`), pay its invoice out of band, so the quote is `PAID`.
3. Insert the quote into the wallet database the way an external-quote consumer does:

```rust
db.add_mint_quote(MintQuote {
    id: quote_id,
    state: MintQuoteState::Paid,
    amount: Some(amount),
    amount_paid: amount,
    amount_issued: Amount::ZERO,
    secret_key: Some(k),        // == seed[..32]
    payment_method: PaymentMethod::Bolt11,
    version: 0,
    used_by_operation: None,
    // ...
}).await?;
```

4. Call `wallet.mint(&quote_id, SplitTarget::None, None).await`.

Observed (every time, not a race):

```
Err(Cdk { code: 50000, error_message: "Concurrent update detected" })
```

and afterwards `GET /v1/mint/quote/bolt11/{quote_id}` at the mint reports `"state": "ISSUED", "amount_issued": <amount>`, while the local database contains no proofs for the quote and the quote row's `version` has advanced past the value the saga expected.

## Root cause

Two defects compose:

**1. The legacy-key heuristic is unsound.** `is_legacy_npubcash_secret_key` claims a quote key as cdk's own based purely on byte equality with the seed prefix:

https://github.com/cashubtc/cdk/blob/v0.17.3/crates/cdk/src/wallet/npubcash.rs#L378-L380

`seed[..32]` is the most obvious trivial derivation available to any wallet built over the same seed, so byte equality cannot establish provenance. Applications that chose the same derivation for their own external quotes (ZEUS Pay did, predating cdk's npubcash support) are misclassified.

**2. The scrub invalidates the mint saga's own snapshot.** `mint_quote_signing_key` is called from `prepare_after_reserve` inside the mint saga. In the legacy branch it calls `scrub_legacy_npubcash_quote`, which rewrites the quote row (removing the key and writing the kv marker), bumping the row's optimistic-lock `version`:

https://github.com/cashubtc/cdk/blob/v0.17.3/crates/cdk/src/wallet/issue/mod.rs#L75-L102
https://github.com/cashubtc/cdk/blob/v0.17.3/crates/cdk/src/wallet/npubcash.rs#L392-L408

The saga, however, keeps using the `quote_info` it loaded before reserving the quote (`prepare`, saga/mod.rs#L311) and, after the mint HTTP call has succeeded, persists the updated state through the versioned upsert with the stale `version`:

https://github.com/cashubtc/cdk/blob/v0.17.3/crates/cdk/src/wallet/issue/saga/mod.rs#L735-L737
https://github.com/cashubtc/cdk/blob/v0.17.3/crates/cdk-sql-common/src/wallet/mod.rs#L1174-L1230

`UPDATE ... WHERE mint_quote.version = :expected_version` matches zero rows and returns `ConcurrentUpdate`. The saga aborts and compensates, after the external side effect (signature issuance) has already happened, so the constructed proofs are dropped.

## Impact

- Funds appear lost to the user: the mint marks the quote issued, the wallet records nothing. Recovery requires a NUT-13 restore.
- Deterministic on every affected mint operation, not a concurrency edge case.
- Affects any cdk-ffi consumer implementing server-created (LNURL-style) quote redemption with seed-prefix keys, and plausibly legacy npub.cash quotes themselves when the key is stored on the quote.

## Suggested fixes

In increasing order of rigor; the first alone resolves the fund-stranding:

1. Reload `quote_info` from the localstore after `mint_quote_signing_key` returns (or whenever the scrub ran), so the saga's post-mint write uses the current row version.
2. Alternatively, have `scrub_legacy_npubcash_quote` return the refreshed record (or perform a version-preserving write) so a hidden write inside key resolution cannot invalidate callers' snapshots.
3. Fix the heuristic: only scrub quotes whose npub.cash provenance cdk can actually establish (e.g. its own kv metadata or an import marker), never on secret-key byte equality alone. As a general principle, a key-resolution helper mutating quote rows mid-saga is surprising even for true positives.

## Workaround we ship in the meantime

ZeusLN/zeus#4397 stores external quotes with `secret_key: None` and pre-writes cdk's own kv marker (`npubcash` / `quotes` / `{quote_id}` = `legacy-seed-prefix`) so `mint_quote_signing_key` resolves the identical key through `derive_legacy_npubcash_secret_key` without triggering the scrub:

https://github.com/ZeusLN/zeus/pull/4397 (commit 65cead73f)

This knowingly couples to `pub(crate)` kv constants; we would like to remove it once this is fixed. Happy to test a patch.

### LINKED ISSUE #4402: Remove npubcash kv-marker workaround in addExternalMintQuote once cashubtc/cdk#2335 ships a fix
## Context

PR #4397 (CDK 0.17.3 bump) contains a deliberate workaround in `addExternalMintQuote` on both platforms:

- `ios/CashuDevKit/CashuDevKitModule.swift`
- `android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt`

ZEUS Pay locks external mint quotes to the wallet's seed-prefix key (`seed[0..32]`). cdk's `npubcash` feature treats any stored quote key equal to `seed[0..32]` as a legacy NpubCash key and scrubs it from the quote record mid-mint-saga. The scrub bumps the quote row's optimistic-lock version, so the saga's post-mint write fails with `Cdk(50000, "Concurrent update detected")` after the mint has already issued the blind signatures. Locally no proofs or transaction are recorded and the funds are only recoverable via NUT-13 restore. This was deterministic on every lightning address redemption.

Upstream bug: https://github.com/cashubtc/cdk/issues/2335

## The workaround

Instead of storing the P2PK secret key on the quote, `addExternalMintQuote` stores the quote with no key and pre-writes cdk's own NpubCash quote-key marker:

```
kvWrite(primaryNamespace: "npubcash", secondaryNamespace: "quotes", key: quoteId, value: "legacy-seed-prefix")
```

At signing time cdk resolves the marker and re-derives the identical seed-prefix key (`derive_legacy_npubcash_secret_key` is `SecretKey::from_slice(&seed[..32])`), so the NUT-20 signature is unchanged and no mid-saga quote write occurs.

## Why this needs tracking

The workaround couples ZEUS to cdk internals that are `pub(crate)`: the `npubcash`/`quotes` kv namespaces and the `legacy-seed-prefix` marker bytes. If a future cdk release renames the namespace, changes the marker, or reworks the NpubCash quote-key resolution, ZEUS Pay redemptions will break again, likely silently and with the same stranded-funds failure mode.

## Exit criteria

- [ ] cashubtc/cdk#2335 is fixed in a released cdk version (saga reloads the quote after key resolution, or the scrub stops claiming keys cdk did not create)
- [ ] ZEUS bumps to that CDK release
- [ ] Revert the workaround: store `secretKey` on the quote again in `addExternalMintQuote` and drop the `kvWrite` marker call on both platforms
- [ ] Re-test ZEUS Pay lightning address redemption on both platforms

## Related hardening (optional, can split out)

cdk 0.15+ provides `Wallet::recoverIncompleteSagas()` to complete or roll back interrupted operations (crash mid-mint/melt leaves proofs reserved or unpersisted). ZEUS never calls it. Calling it non-fatally at wallet initialization would self-heal states like the one this bug produced.

### LINKED ISSUE #4413: CDK 0.17: wallet init blocks on sequential /v1/info fetches for every stored mint
### Background

#4397 bumps CDK to 0.17.3. In cdk 0.14, `MultiMintWallet::load_wallets` was db-only: it read the stored mints and built wallet handles locally. In 0.17, the `WalletRepository` constructor awaits `load_wallets()`, which loops every mint stored in the wallet db and awaits `fetch_mint_info` for each, a real `GET /v1/info` on a freshly constructed HTTP client, before the constructor returns ([wallet_repository.rs#L629-L667](https://github.com/cashubtc/cdk/blob/v0.17.3/crates/cdk/src/wallet/wallet_repository.rs#L629-L667)). Failures are tolerated (fallback to a sat-only wallet), so nothing breaks, but init latency now scales linearly with mint count.

Raised by @shubhamkmr04 in https://github.com/ZeusLN/zeus/pull/4397#issuecomment-5284252972.

### Impact

- `CashuStore.initializeCDK` awaits `CashuDevKit.initializeWallet`, and `initializeWallets` awaits that on the Wallet view boot path, so the cashu balance render blocks on N sequential round trips.
- The cdk mint connector configures no request timeout (reqwest default is none), so an unreachable host that black-holes packets costs the OS TCP connect timeout (roughly 1 to 2 minutes on mobile), and a reachable but hung server can stall init indefinitely.
- Multi-mint users are hit hardest; one dead mint delays every mint behind it in the loop.

### Options

- Upstream (filed as cashubtc/cdk#2351): have `load_wallets` use the mint info already stored in the db row (the loop currently destructures it as `_mint_info` and discards it), refresh lazily, parallelize the fetches, and/or add a transport timeout.
- ZEUS-side: move `initializeCDK` off the boot-critical path (render cached balances first, finish init in the background), or defer the blocking init until first cashu use.

### Pre-merge measurement for #4397

Time `initializeWallet` on a build with 4+ mints configured, one of them an unreachable host, and compare against a 0.14.2 build. Tracked in the #4397 device-test checklist.

kaloudis:
# Description

Relates to issue: #4413, #4402

Follow-up to #4397. cdk shipped [v0.17.4](https://github.com/cashubtc/cdk/releases/tag/v0.17.4) today in response to the two issues we filed during that review, so this bumps to it and removes the workaround the old behavior forced on us.

**Commit 1: bump Cashu Dev Kit to 0.17.4.** `WalletRepository` construction is offline-first again ([cashubtc/cdk#2351](https://github.com/cashubtc/cdk/issues/2351)): `load_wallets` restores wallet units from the mint info already persisted in the wallet db instead of issuing a sequential `GET /v1/info` per stored mint (on a client with no request timeout) before the constructor returns. Wallet init no longer scales with mint count or blocks on unreachable mints, which fixes #4413 with no ZEUS-side changes. Upstream added regression coverage asserting zero network connections during startup.

Artifact integrity: the iOS `CashuDevKitFFI.xcframework.zip` SHA256 pinned here matches the checksum in cdk-swift v0.17.4's own `Package.swift`. The uniffi contract version is unchanged at 30, and the bindings diff over 0.17.3 is doc comments plus the two `WalletRepository` constructor checksums. The 0.17.4 xcframework still ships the mixed macOS slice, so the existing strip step in `fetch-libraries.sh` stays.

**Commit 2: drop the npubcash marker workaround in `addExternalMintQuote`.** cdk 0.17.4 also fixes [cashubtc/cdk#2335](https://github.com/cashubtc/cdk/issues/2335): signing-key resolution now consults the kv marker only and otherwise returns the quote's stored key without touching wallet state, the legacy scrub only runs on quotes that provably came from an NpubCash account, and the mint saga reloads quote snapshots after key resolution. Upstream's `external_seed_prefix_key_is_not_claimed_by_npubcash` test covers our exact case. So the P2PK secret key is stored on external quotes again for both v1 and v2 wallets, and the keyless-quote-plus-marker path, the `useSeedPrefixMarker` bridge parameter, and the JS seed-prefix byte comparison are removed. Closes #4402.

Notes:
- In-flight quotes stored keyless with the marker by earlier builds still sign correctly; 0.17.4 retains the marker resolution path.
- The stored-version CAS handling in `addExternalMintQuote` stays; it addresses a separate failure mode (retrying after a dead saga claim bumped the row version).
- The db migration from 0.17.3 is additive (`legacy_quotes_migrated_v2` kv key); no ZEUS data migration is needed.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [x] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

Device test checklist (both platforms):
- [ ] ZEUS Pay lightning address redemption on a v2-bip39 wallet (key now stored on the quote)
- [ ] ZEUS Pay lightning address redemption on a v1 wallet, if a v1 seed is available
- [ ] ZEUS Pay redemption of a quote created by a pre-0.17.4 build (keyless with marker) after upgrading
- [ ] Fresh wallet init with 4+ mints configured, one of them unreachable: cashu balances render without delay
- [ ] Retry a redemption after a failed mint attempt (dead saga claim, CAS version fix from #4397 still applies)
- [ ] Minibits invoice create/pay, multimint melt, restore, offline send (regression sweep from the #4397 checklist)

Build verification so far: `yarn verify` green (1149 tests), Android `compileDebugKotlin` clean against the 0.17.4 jar, iOS simulator `xcodebuild` BUILD SUCCEEDED against the 0.17.4 xcframework.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

_`package.json`/`yarn.lock` are untouched; the dependency change is the native CDK artifact pins in `fetch-libraries-versions.json` (plus the podspec version and `Podfile.lock`). Re-running `yarn` covers it: postinstall runs `fetch-libraries` and `pod-install`._

### Other:

### DIFF
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -802,7 +802,6 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         state: String,
         expiry: Double,
         secretKey: String?,
-        useSeedPrefixMarker: Boolean,
         promise: Promise
     ) {
         if (!isInitialized || db == null) {
@@ -824,32 +823,11 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     else -> QuoteState.PAID // Default to PAID for external quotes
                 }
 
-                // For v2-bip39 wallets, ZEUS Pay locks quotes to the seed-
-                // prefix key (seed[0..32]), which is byte-identical to cdk's
-                // "legacy NpubCash" key. Storing that key on the quote makes
-                // cdk's mint saga scrub it mid-flight (a version bump), and
-                // the saga's post-mint write then dies with ConcurrentUpdate
-                // AFTER the mint has issued the signatures. For those quotes
-                // (useSeedPrefixMarker, decided by JS via byte comparison),
-                // store the quote with no key and write cdk's NpubCash
-                // quote-key marker: at signing time cdk re-derives the
-                // identical seed-prefix key from the marker, with no mid-saga
-                // write. v1 wallets sign with a different key (LND seed
-                // bytes [32:64]), so the marker would derive the wrong key
-                // for them; their key is stored on the quote instead, which
-                // is safe because the scrub only triggers on byte equality
-                // with the seed prefix.
-                // Upstream bug: https://github.com/cashubtc/cdk/issues/2335
-                // Remove when fixed: https://github.com/ZeusLN/zeus/issues/4402
+                // Storing a key equal to cdk's seed prefix on the quote
+                // requires cdk >= 0.17.4: earlier versions claim it as a
+                // legacy NpubCash key and scrub it mid-mint-saga, stranding
+                // the minted funds (cashubtc/cdk#2335).
                 val storedSecretKey = secretKey?.takeIf { it.isNotEmpty() }
-                if (storedSecretKey != null && useSeedPrefixMarker) {
-                    db!!.kvWrite(
-                        primaryNamespace = "npubcash",
-                        secondaryNamespace = "quotes",
-                        key = quoteId,
-                        value = "legacy-seed-prefix".toByteArray(Charsets.UTF_8)
-                    )
-                }
 
                 // Create the MintQuote object
                 // For external quotes that are PAID, we set amountPaid = amount
@@ -866,7 +844,7 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     amountIssued = if (quoteState == QuoteState.ISSUED) amt else zeroAmount,
                     estimatedBlocks = null,
                     paymentMethod = PaymentMethod.Bolt11,
-                    secretKey = if (useSeedPrefixMarker) null else storedSecretKey,
+                    secretKey = storedSecretKey,
                     usedByOperation = null,
                     version = version
                 )
--- a/cashu-cdk/CashuDevKit.ts
+++ b/cashu-cdk/CashuDevKit.ts
@@ -316,12 +316,6 @@ class CashuDevKit {
      * @param state - The quote state (UNPAID, PAID, PENDING, ISSUED)
      * @param expiry - The quote expiry timestamp
      * @param secretKey - Optional hex-encoded secret key for P2PK-locked quotes
-     * @param useSeedPrefixMarker - When the secret key equals CDK's seed
-     * prefix (seed[0..32]), store the quote keyless and write cdk's NpubCash
-     * quote-key marker instead, to avoid cdk's legacy npubcash scrub
-     * (https://github.com/cashubtc/cdk/issues/2335). Must be false when the
-     * key differs from the seed prefix (v1 wallets), or cdk would sign with
-     * the wrong key.
      */
     async addExternalMintQuote(
         mintUrl: string,
@@ -330,8 +324,7 @@ class CashuDevKit {
         request: string,
         state: string,
         expiry: number,
-        secretKey?: string,
-        useSeedPrefixMarker: boolean = false
+        secretKey?: string
     ): Promise<boolean> {
         try {
             return await CashuDevKitModule.addExternalMintQuote(
@@ -341,8 +334,7 @@ class CashuDevKit {
                 request,
                 state,
                 expiry,
-                secretKey,
-                useSeedPrefixMarker
+                secretKey
             );
         } catch (error) {
             throw mapCDKError(error);
--- a/cashu-cdk/types.ts
+++ b/cashu-cdk/types.ts
@@ -259,8 +259,7 @@ export interface CashuDevKitNativeModule {
         request: string,
         state: string,
         expiry: number,
-        secretKey?: string,
-        useSeedPrefixMarker?: boolean
+        secretKey?: string
     ): Promise<boolean>;
     mint(
         mintUrl: string,
--- a/fetch-libraries-versions.json
+++ b/fetch-libraries-versions.json
@@ -10,11 +10,11 @@
         "iosSha256": "fe50c9796f91f121ed58bd2aa3970dc2c3048358b750ff5b8fafa3e6a8b50fce"
     },
     "cdk": {
-        "version": "0.17.3",
-        "androidSha256": "632e4d593a7ed3abfc142df9cfc9b8ef68cd29db8cd12de260eb568f9828f39a",
-        "androidBindingsSha256": "56d990510fbbfed182e21549168742b007fc736df5184ce6cadf729b47eccc93",
-        "iosSha256": "02a93d7e649069dd8c75b9f14f74606ca23bf081725be57786524c60839395de",
-        "iosBindingsSha256": "2bca7a65eeee1e1ebc45f33287293e31bb473ed3662cb74a3d699444e3c831e8"
+        "version": "0.17.4",
+        "androidSha256": "9717f92f62ae9c9e044b5f40f5831a2bd0d9fc3bb9dd63739468500bfe70f4cd",
+        "androidBindingsSha256": "520bdd3ff49d7f0b5376b4313720665d7208634a1c463620d3959437484d8260",
+        "iosSha256": "0908c519ee72a0d80713dbb96d787fd5a3bd8ee1fb38074cc2e562b417911c96",
+        "iosBindingsSha256": "7ebcce2363c346daaae4c01c090d34a6d73bf168ddef0395a5eae9aab36bf53f"
     },
     "zeus-cashu-restore": {
         "version": "0.1.0",
--- a/ios/CashuDevKit.podspec
+++ b/ios/CashuDevKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'CashuDevKit'
-  s.version          = '0.17.3'
+  s.version          = '0.17.4'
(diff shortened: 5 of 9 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4427, opened 2026-08-15
LSPS1: refetch service info on focus after stacked instance wipes store

### LINKED ISSUE #4417: LSPS1: create-order screen loses service info after returning from a stacked LSPS1 instance
### Description

The LSPS1 create-order screen loses its service info after the user navigates through LSPS1 Settings back into a second copy of the create-order screen and then returns. On the original screen, the Service info accordion, the LSP balance slider, and the Advanced settings accordion disappear, and the slider min/max labels render as `0` / `0`.

Demo by @ajaysehwal (from https://github.com/ZeusLN/zeus/pull/4396#pullrequestreview-4936081084):

https://github.com/user-attachments/assets/106f1f93-1d7a-4ba5-8a7c-9fce57e841bb

### Steps to reproduce

1. Open the LSPS1 create-order screen (`views/LSPS1/index.tsx`). Everything renders correctly.
2. Tap the gear icon to open LSPS1 Settings (`views/LSPS1/Settings.tsx`).
3. Tap "Purchase inbound channel". This lands on a second copy of the create-order screen.
4. Create an order (optional, makes it easier to see in the video) and press back twice, through Settings, to the original create-order screen.
5. The original screen is now degraded: no Service info accordion, no slider, no Advanced settings, and the slider min/max labels show `0` / `0`.

### Root cause

1. **Duplicate screen instances.** `views/LSPS1/Settings.tsx` uses `navigation.navigate('LSPS1')` for "Purchase inbound channel". Under react-navigation 7, `navigate` no longer pops back to an existing screen in the stack, it pushes. When Settings was reached from the LSPS1 screen itself, the stack becomes LSPS1 -> Settings -> LSPS1 with two live instances.
2. **Shared singleton store, reset on unmount.** Both instances observe the same `LSPStore`. `componentWillUnmount` in `views/LSPS1/index.tsx` calls `LSPStore.resetLSPS1Data()`, which wipes `getInfoData` and `createOrderResponse`. When the second instance is popped, it wipes the store out from under the first instance, which is still mounted below and never refetches (`getInfo` only runs in `componentDidMount`; the `focus` listener only refreshes the discount token).
3. **The `0` / `0` labels** are `numberWithCommas(info?.min_initial_lsp_balance_sat)` receiving `undefined` from the emptied store and hitting the `'0'` fallback in `numberWithCommas`. Everything gated on `Object.keys(getInfoData).length > 0` or `info?.min_initial_lsp_balance_sat` (both accordions, the slider) simply disappears.

This predates #4396, which only touches the TextInput value bindings for user-entered state and the OrderResponse wrappers. The same sequence reproduces on master.

### Possible fixes

- **Refetch on focus**: in the LSPS1 screen's `focus` listener, re-run the `getInfo` flow when `getInfoData` is empty. This also heals any other path that clears the store.
- **Don't reset from the unmounting duplicate**: only reset in `componentDidMount` (which already does it), or guard the unmount reset. Riskier, since the unmount reset also stops free-order status polling.
- A `popTo` in Settings is not viable on its own: `LSPS1Settings` is also reachable from `views/Settings/LSPServicesList.tsx` and from the LSPS7 screen, where LSPS1 is not in the stack, so "Purchase inbound channel" legitimately needs to push in those cases.

### Notes

LSPS7 (`views/LSPS7/index.tsx`) has the same reset-on-unmount pattern and shares the same settings screen, so it likely has the same defect and should be checked alongside any fix.

kaloudis:
# Description

Fixes #4417

The LSPS1 create-order screen loses its service info after navigating LSPS1 -> Settings -> "Purchase inbound channel" (a second LSPS1 copy under react-navigation 7's push-only `navigate`) and returning. The popped duplicate's `componentWillUnmount` calls `LSPStore.resetLSPS1Data()`, wiping `getInfoData` and `createOrderResponse` out from under the original instance, which only fetched in `componentDidMount`. The Service info accordion, LSP balance slider, and Advanced settings disappear and the slider min/max labels render as `0` / `0`. Demo in the issue (found by @ajaysehwal during review of #4396; reproduces on master, predates that PR).

This implements the "refetch on focus" fix recommended in the issue:

- `views/LSPS1/index.tsx`: the getInfo dispatch (native / REST / custom message) is extracted into `fetchGetInfo()`. The existing `focus` listener now re-runs it whenever `getInfoData` is empty and no fetch is in flight. This heals the stacked-duplicate sequence and any other path that clears the store while the screen is mounted below. The custom-message transport setup (`connectPeer` + `subscribeToCustomMessages`) still happens only on mount, under the exact same conditions as before; on the heal path the original instance's peer connection and subscription are still live, so only the getInfo request needs to be re-sent.
- The unmount reset also stops the shared free-order status polling (`stopFreeOrderStatusPolling` is one slot shared by LSPS1 and LSPS7, and `resetLSPS1Data` / `clearLSPS7Order` each kill it). Per the issue's note to check LSPS7 alongside: both screens now resume polling on focus when the store holds a pending (non-terminal) free order. LSPS7 itself cannot be duplicated via the settings screen (Settings only links to LSPS1) and renders from route params, so the polling kill is its only exposure; `startFreeOrderStatusPolling` is a no-op if polling for that order is already active.

The mount-time reset and unmount reset are intentionally left in place: the unmount reset is what stops polling and clears state when the user actually leaves the flow.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [ ] I've run `yarn run tsc` and made sure my code compiles correctly
- [ ] I've run `yarn run lint` and made sure my code didn't contain any problematic patterns
- [ ] I've run `yarn run prettier` and made sure my code is formatted correctly
- [ ] I've run `yarn run test` and made sure all of the tests pass

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/LSPS1/index.tsx
+++ b/views/LSPS1/index.tsx
@@ -33,6 +33,7 @@ import BackendUtils from '../../utils/BackendUtils';
 import {
     buildPaymentAwaitParams,
     isOrderFree,
+    LSPOrderState,
     LSPService
 } from '../../models/LSP';
 import { themeColor } from '../../utils/ThemeUtils';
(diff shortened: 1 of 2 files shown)


ajaysehwal (approved):
tACK

Issue #1752, opened 2023-10-09
`accessibilityState` of swipeable rows should be dynamically updated


sha-265:
### Describe the bug

The `accessibilityState` attribute of the wallet's swipeable rows (part of the [`LayerBalances`](https://github.com/ZeusLN/zeus/blob/ebf5d0674dd996b0bd9e8c2553188bf8e5783de4/components/LayerBalances/index.tsx#L62) component) is not updated dynamically. 

For better accessibility the `accessibilityState` value should be update dynamically when swiping the rows, from `{ expanded: false }` to `{ expanded: true }` and vice versa.

As I understand maybe we need to use [direct manipulation](https://reactnative.dev/docs/0.70/direct-manipulation), or as @kaloudis [suggested](https://github.com/ZeusLN/zeus/pull/1590#issuecomment-1739452837) to track the state of the rows somehow.

### Reproduce

Just swipe the rows with screen reader enable, and listen to `accessibilityState` state.

### Zeus version

v0.8.0

### Node interface

Multiple (specified in comments)

### Network

Clearnet

### Device

Any

### Device operating system

Any

kaloudis:
addressed in https://github.com/ZeusLN/zeus/pull/4433

kaloudis:
ZEUS v13.2.1-rc1 which includes a fix for this, is now available for testing: https://github.com/ZeusLN/zeus/releases/tag/v13.2.1-rc1

It will be submitted to app stores shortly. Please reopen if you hit any problems.

Thank you for reporting.

kaloudis:
@sha-265 ZEUS v13.2.1, which includes a fix, is now available on app stores https://github.com/ZeusLN/zeus/releases/tag/v13.2.1

Please reopen the issue and report back if it recurs. Thanks again for reporting.

PR #4433, opened 2026-08-18
components: sync accessibilityState of wallet rows

TheSeydiCharyyev:
# Description

Relates to issue: **#1752**

The swipeable wallet rows in `LayerBalances` did not expose `accessibilityState`. A screen reader could not tell if a row was open or closed.

Each row component now keeps an `expanded` flag in component state and passes it to the row as `accessibilityState={{ expanded }}`, together with `accessibilityRole="button"`. `Accordion` already uses the same pair, so this follows a pattern that is in the repo.

The flag is driven by `onSwipeableWillOpen` and `onSwipeableWillClose`, not by `onSwipeableOpen` and `onSwipeableClose`. In `react-native-gesture-handler` 2.31.2 the second pair runs inside the spring animation callback, and only when the animation reports `finished`. If the user interrupts a swipe, that callback does not run and the state stays wrong. The `Will` callbacks always run, and they also cover the programmatic `open()` and `close()` calls used by row taps and by the action buttons.

The rows also declare `accessibilityActions` for `expand` and `collapse` and handle them. Without that, setting `expanded` makes `ReactAccessibilityDelegate` add `ACTION_EXPAND` or `ACTION_COLLAPSE` to the node, so TalkBack offers the action, but performing it only flips a view tag and dispatches nothing to JS, because `accessibilityActionsMap` is filled from declared actions only. The announced action did nothing, and the flipped tag disagreed with component state until the next swipe. This matters most for rows with a payment value, where a double tap starts the pay flow instead of opening the row.

`PaymentMethodList` passes `locked={true}` to every row, so those rows take the early-return branches and are not covered by the `expanded` work. Those branches are still tappable — they open an invoice, send to an address, or show the sync info modal — so they now carry `accessibilityRole="button"`. No `expanded` state is added there, because a locked row does not open. Branches that only return `children` or a plain `View` are left alone: they have no press handler, so a button role would be wrong.

### How this was checked

On the emulator the row is one node with `class="android.widget.Button"` and the label `Lightning, 250,000, sats`. Opening by swipe and by tap gives `expanded: true`, closing gives `expanded: false` — confirmed with a temporary log, removed before the commit. Swipe-to-open, swipe-to-close and tap-to-open still behave the same after the action wiring.

I could not perform the Expand action through TalkBack on the emulator. TalkBack was active (`dumpsys accessibility` shows the bound service, `touchExplorationEnabled=true`, `TouchExplorer` and `KeyboardInterceptor` enabled), but events injected with `adb shell input` go in below touch exploration — an injected swipe reached the app and opened the row instead of being consumed for navigation, and TalkBack's keyboard shortcuts did not open its menu either. So the inert-action claim above is from the `ReactAccessibilityDelegate` source, not from a device run.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [x] LndHub — local stub server, only to bring the wallet screen up; the change itself does not touch any backend

### Locales

### Other:

### DIFF
--- a/components/LayerBalances/EcashSwipeableRow.tsx
+++ b/components/LayerBalances/EcashSwipeableRow.tsx
@@ -43,8 +43,10 @@ interface EcashSwipeableRowProps {
 @observer
 export default class EcashSwipeableRow extends Component<
     EcashSwipeableRowProps,
-    {}
+    { expanded: boolean }
 > {
+    state = { expanded: false };
+
     private renderAction = (
         text: string,
         x: number,
@@ -249,6 +251,7 @@ export default class EcashSwipeableRow extends Component<
             needsConfig,
             navigation
         } = this.props;
+        const { expanded } = this.state;
 
         if (locked && lightning) {
             return (
@@ -277,6 +280,8 @@ export default class EcashSwipeableRow extends Component<
                 rightThreshold={40}
                 renderLeftActions={this.renderActions}
                 containerStyle={{ width: '100%' }}
+                onSwipeableWillOpen={() => this.setState({ expanded: true })}
+                onSwipeableWillClose={() => this.setState({ expanded: false })}
             >
                 <TouchableOpacity
                     onPress={() =>
@@ -288,6 +293,8 @@ export default class EcashSwipeableRow extends Component<
                     }
                     activeOpacity={1}
                     style={{ opacity: needsConfig ? 0.4 : 1 }}
+                    accessibilityRole="button"
+                    accessibilityState={{ expanded }}
                 >
                     {children}
                 </TouchableOpacity>
--- a/components/LayerBalances/EcashSwipeableRow.tsx
+++ b/components/LayerBalances/EcashSwipeableRow.tsx
@@ -259,6 +259,7 @@ export default class EcashSwipeableRow extends Component<
                     onPress={() => (disabled ? null : this.fetchLnInvoice())}
                     activeOpacity={1}
                     style={{ width: '100%' }}
+                    accessibilityRole="button"
                 >
                     {children}
                 </TouchableOpacity>
--- a/components/LayerBalances/EcashSwipeableRow.tsx
+++ b/components/LayerBalances/EcashSwipeableRow.tsx
@@ -296,6 +296,16 @@ export default class EcashSwipeableRow extends Component<
                     style={{ opacity: needsConfig ? 0.4 : 1 }}
                     accessibilityRole="button"
                     accessibilityState={{ expanded }}
+                    accessibilityActions={[
+                        { name: 'expand' },
+                        { name: 'collapse' }
+                    ]}
+                    onAccessibilityAction={({
+                        nativeEvent: { actionName }
+                    }) => {
+                        if (actionName === 'expand') this.open();
+                        else if (actionName === 'collapse') this.close();
+                    }}
                 >
                     {children}
                 </TouchableOpacity>
--- a/components/LayerBalances/LightningSwipeableRow.tsx
+++ b/components/LayerBalances/LightningSwipeableRow.tsx
@@ -378,6 +378,7 @@ export default class LightningSwipeableRow extends Component<
                             text: localeString('views.Wallet.waitForSync')
                         })
                     }
+                    accessibilityRole="button"
                 >
                     <View style={{ opacity: 0.25 }}>{children}</View>
                 </TouchableOpacity>
@@ -395,6 +396,7 @@ export default class LightningSwipeableRow extends Component<
                 <TouchableOpacity
                     onPress={() => (disabled ? null : this.fetchLnInvoice())}
                     activeOpacity={1}
+                    accessibilityRole="button"
                 >
                     {children}
                 </TouchableOpacity>
--- a/components/LayerBalances/LightningSwipeableRow.tsx
+++ b/components/LayerBalances/LightningSwipeableRow.tsx
@@ -49,8 +49,10 @@ interface LightningSwipeableRowProps {
 @observer
 export default class LightningSwipeableRow extends Component<
     LightningSwipeableRowProps,
-    {}
+    { expanded: boolean }
 > {
+    state = { expanded: false };
+
     private renderAction = (
         text: string,
         x: number,
@@ -366,6 +368,7 @@ export default class LightningSwipeableRow extends Component<
             lnurlParams,
             SyncStore
         } = this.props;
+        const { expanded } = this.state;
         const { isSyncing } = SyncStore!;
         if (isSyncing) {
             return (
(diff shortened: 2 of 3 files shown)


kaloudis (commented):
Concept ACK

Verified the mechanism against the vendored `react-native-gesture-handler` 2.31.2 source: the `Will*` callbacks fire synchronously in `animateRow` outside the spring's `finished` gate, while `onSwipeableOpen`/`onSwipeableClose` only fire when the animation completes, so the choice of `onSwipeableWillOpen`/`onSwipeableWillClose` is correct and covers gesture releases, tap-to-open (`openLeft()`), and action-button `close()` alike. A below-threshold swipe fires `WillClose` on a row that never opened, but the resulting `setState` is idempotent, so state can't drift. Matches the `Accordion` precedent.

Two asks before this goes in:

1. **PaymentMethodList does not pick this change up.** It passes `locked={true}` to every row, so those rows always hit the early-return locked branches, which this PR doesn't touch. Please correct the PR description, or better, extend `accessibilityRole="button"` to the locked-but-tappable branches (`LightningSwipeableRow`, `OnchainSwipeableRow`, `EcashSwipeableRow`), since those are activatable buttons with no role today. No `expanded` state needed there as they aren't expandable.

2. **The announced Expand/Collapse action is likely inert.** Setting `expanded` in `accessibilityState` makes `ReactAccessibilityDelegate` add `ACTION_EXPAND`/`ACTION_COLLAPSE`, so TalkBack offers the action in its menu, but without `accessibilityActions` + `onAccessibilityAction` wiring, performing it does nothing. This matters most for rows with a payment value, where double-tap triggers the pay flow rather than `open()`, leaving TalkBack users no working way to expand. Suggested addition to each row:

```tsx
accessibilityActions={[{ name: 'expand' }, { name: 'collapse' }]}
onAccessibilityAction={({ nativeEvent }) =>
    nativeEvent.actionName === 'expand' ? this.open() : this.close()
}
```

Happy to take (2) as a follow-up if you'd rather keep this PR scoped to state sync, but please confirm on-device whether TalkBack's Expand action currently does anything; announcing an action that's dead can be worse than not announcing it.

Nits (optional): the `isSyncing` info-modal branches could take `accessibilityRole="button"` too.

TheSeydiCharyyev:
Both asks are in, as two commits on top.

**1. PaymentMethodList.** You're right and my description was wrong — it passes `locked={true}` to every row, so those rows take the early-return branches and never reached the `expanded` work. I took the second option: those branches now carry `accessibilityRole="button"`, since they are activatable (open an invoice, send to an address, or show the sync info modal). No `expanded` there, as they don't open. I left alone the branches that only return `children` or a plain `View` — no press handler, so a button role would be a lie. The `isSyncing` modal branches got the role too, per your nit. Description is updated.

**2. Expand/collapse actions.** Wired in all three rows:

```tsx
accessibilityActions={[{ name: 'expand' }, { name: 'collapse' }]}
onAccessibilityAction={({ nativeEvent: { actionName } }) => {
    if (actionName === 'expand') this.open();
    else if (actionName === 'collapse') this.close();
}}
```

One deviation from your snippet: explicit branches instead of a ternary, so an action name that is neither does not fall through to `close()`. Today only the declared names reach the handler, but that relies on current RN behaviour rather than on a contract.

On your question of whether the action does anything today — I could not answer it on-device, and I want to be exact about why. TalkBack was active on the emulator (`dumpsys accessibility` shows the bound service, `touchExplorationEnabled=true`, `TouchExplorer` and `KeyboardInterceptor` in the enabled features), but events injected with `adb shell input` land below touch exploration: an injected swipe reached the app and opened the row instead of being consumed for navigation, and TalkBack's keyboard shortcuts did not open its menu either. So I have no device run to show for it.

From the source it is conclusive. `performAccessibilityAction` flips the view tag for `ACTION_EXPAND`/`ACTION_COLLAPSE` and only dispatches to JS `if (accessibilityActionsMap.containsKey(action))`, and that map is filled from declared `accessibilityActions` only. With none declared it is empty, so the action was inert and the flipped tag disagreed with component state until the next swipe. After this change `expand` and `collapse` register the same action ids the `expanded` state adds — they are in RN's standard action-name map — so `containsKey` matches and the handler runs. If you can perform the action on a physical device it would confirm the second half in seconds.

I did re-check on the emulator that normal interaction is unchanged: tap-to-open, swipe-to-open and swipe-to-close all still work.

One thing I deliberately did not do: the three row components are now near-identical in their `Swipeable` config, state and accessibility props, and this is the third time the same block gets copied into all three. Extracting a shared container would remove that, but it is a refactor rather than a fix, so I kept this PR scoped. Happy to open it separately if you want it.

kaloudis:
utACK c9d2adc

Round 2 addresses both asks; verified each against the sources.

**1. Branch coverage is now right.** Every early-return branch with a press handler carries `accessibilityRole="button"` (Lightning locked and isSyncing, Onchain locked-with-value and isSyncing, Ecash locked-with-lightning), and the bare `children` / plain `View` branches are correctly left alone since a button role there would be a lie, as you said.

**2. The action wiring is sound on both platforms.** I checked the halves you couldn't reach from the emulator:

- Android: `expand` and `collapse` are in `ReactAccessibilityDelegate`'s standard action name map (`ACTION_EXPAND` / `ACTION_COLLAPSE`), so the declared actions register under the same ids the `expanded` state advertises, `accessibilityActionsMap.containsKey` now matches, and the handler runs. Your source-level reading checks out.
- iOS (Fabric): in `RCTViewComponentView.mm`, label-less custom actions fall back to the name for both the rotor entry and the JS dispatch, so VoiceOver users get a working expand path too, and `expanded: true` is announced through a localized "expanded" value. The one cosmetic gap is that the rotor shows raw lowercase "expand" / "collapse" in English; localizing needs `label`s plus Transifex strings, fine as a follow-up if ever.
- RN's `TouchableOpacity` forwards both `accessibilityActions` and `onAccessibilityAction`, and since `open()` / `close()` route through the `Will*` callbacks, state stays in sync when the row is driven from the a11y menu. The explicit if/else over my ternary is a good call.

One optional nit, take it or leave it: the locked branches gate their handlers on `disabled` (`onPress={() => (disabled ? null : ...)}`), so with `disabled` set they now announce as a button that silently does nothing on double-tap. `accessibilityState={{ disabled }}` on those branches would let the screen reader tell the truth. Happy for that to ride the shared-container refactor, which I agree belongs in a separate PR; please do open it.

Approved the CI runs; all four checks are green. I'll give this a TalkBack pass on a physical device before merge.

kaloudis (approved):

PR #4435, opened 2026-08-18
fix(nwc): hold NWC budget for late-settling pays

ajaysehwal:
# Description

Fixes NWC `pay_invoice` budget accounting when a payment settles after the store's ~125s wait.

A payment that was still unknown at timeout was recorded as `failed`. Since `failed` is terminal for the budget and `reconcile` only looked at `pending`, `totalSpendSats` stayed at `0`.

After a restart, a replayed request could then pass the budget check and pay the same invoice again on backends that re-pay the same invoice, resulting in bounded overspend (~2× in the reproduced case).

The concurrent burst / check-then-act race is **not** this bug. That path is already closed by the payment queue, `pendingSpendSats`, and the preimage fallback.

## This PR

- Treats timeout / still-loading payments with no listed outcome as **in-flight**, so `pendingSpendSats` continues to hold the budget.
- If the payments list already shows success, **debits** the spend instead of classifying the payment as failed.
- Reconciles **timeout-failed** activity that later settles or is still in-flight, covering already-persisted bad records.
- Runs reconciliation **before** the next budget check so a restart + replay cannot spend twice.

## Scope

This PR does **not** add a processed-event ledger. That belongs to the separate NWC replay finding.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2382,6 +2382,9 @@ export default class NostrWalletConnectStore {
                 paymentHash
             );
 
+        // Deliberate: debit when the node shows this invoice settled even if
+        // Zeus did not initiate the pay via this NWC connection. Without that,
+        // a wallet-paid invoice retried over NWC would escape the budget cap.
         const settledEvidence =
             !!preimage ||
             (!!payment && NostrConnectUtils.isSettledPayment(payment));
@@ -2730,26 +2733,48 @@ export default class NostrWalletConnectStore {
     private async reconcilePendingPayInvoiceActivities(
         connection: NWCConnection
     ): Promise<boolean> {
+        let changed = false;
+
         const pending = connection.activity.filter((a) =>
             connection.isUnresolvedPayInvoiceActivity(a)
         );
         if (pending.length === 0) return false;
 
+        for (const activity of pending) {
+            if (this.abandonStaleUnresolvedPayInvoiceHold(activity)) {
+                changed = true;
+            }
+        }
+
+        const stillUnresolved = connection.activity.filter((a) =>
+            connection.isUnresolvedPayInvoiceActivity(a)
+        );
+        if (stillUnresolved.length === 0) {
+            if (changed) {
+                this.scheduleMaxBudgetRefresh();
+                this.findAndUpdateConnection(connection);
+            }
+            return changed;
+        }
+
         const payments = await this.getPaymentsForPendingPayInvoiceRefresh(
             connection.id,
-            pending
+            stillUnresolved
         );
 
-        let changed = false;
-
-        for (const activity of pending) {
+        for (const activity of stillUnresolved) {
             const payment = payments.find(
                 (p) =>
                     p.getPaymentRequest === activity.id ||
                     (!!activity.paymentHash &&
                         p.paymentHash === activity.paymentHash)
             );
-            if (!payment) continue;
+            if (!payment) {
+                if (this.abandonStaleUnresolvedPayInvoiceHold(activity)) {
+                    changed = true;
+                }
+                continue;
+            }
 
             runInAction(() => {
                 if (
@@ -2765,6 +2790,25 @@ export default class NostrWalletConnectStore {
             this.findAndUpdateConnection(connection);
         }
         return changed;
+    }
+
+    /**
+     * Release a hold the node will never list. Non-timeout error exits reconcile
+     * and prune protection so budget and replay recover without deleting the
+     * connection.
+     */
+    private abandonStaleUnresolvedPayInvoiceHold(
+        activity: ConnectionActivity
+    ): boolean {
+        if (!NostrConnectUtils.isPayInvoiceHoldAbandoned(activity)) {
+            return false;
+        }
+
+        runInAction(() => {
+            activity.status = 'failed';
+            activity.error = 'error.paymentFailed';
+        });
+        return true;
     }
 
     private applyPayInvoiceReconcile(
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1806,16 +1806,20 @@ export default class NostrWalletConnectStore {
             let oldestTime = Infinity;
             for (let i = 0; i < connection.activity.length; i++) {
                 const activity = connection.activity[i];
-                // Pending pay_invoice holds budget via pendingSpendSats until
-                // reconcilePendingPayInvoiceActivities settles it. Active pending
-                // make_invoice must stay for the outstanding cap. Expired or
-                // already-paid pending make_invoice is prunable: those entries are
-                // only reconciled from user-driven paths, so a make_invoice-only
-                // client would otherwise grow activity without bound.
+                // Unresolved pay_invoice (pending or timeout-failed) must survive
+                // until reconcilePendingPayInvoiceActivities can debit. Active
+                // pending make_invoice must stay for the outstanding cap.
+                if (
+                    activity.type === 'pay_invoice' &&
+                    activity.payment_source !== 'cashu' &&
+                    connection.isUnresolvedPayInvoiceActivity(activity)
+                ) {
+                    continue;
+                }
                 if (
                     activity.status === 'pending' &&
-                    (activity.type !== 'make_invoice' ||
-                        this.isActivePendingMakeInvoice(activity))
+                    activity.type === 'make_invoice' &&
+                    this.isActivePendingMakeInvoice(activity)
                 ) {
                     continue;
                 }
@@ -2066,8 +2070,8 @@ export default class NostrWalletConnectStore {
     ): Promise<boolean> {
         if (
             activity.type === 'pay_invoice' &&
-            activity.status === 'pending' &&
-            activity.payment_source !== 'cashu'
+            activity.payment_source !== 'cashu' &&
+            connection.isUnresolvedPayInvoiceActivity(activity)
         ) {
             return this.reconcilePendingPayInvoiceActivities(connection);
         }
@@ -2380,9 +2384,7 @@ export default class NostrWalletConnectStore {
 
         const settledEvidence =
             !!preimage ||
-            (!!payment &&
-                NostrConnectUtils.isSettledPayment(payment) &&
-                !!paymentHash);
+            (!!payment && NostrConnectUtils.isSettledPayment(payment));
 
         if (!settledEvidence) {
             const paymentError = this.checkPaymentErrors(
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2270,7 +2270,11 @@ export default class NostrWalletConnectStore {
         // Late-settling pays can be recorded failed when the 125s wait
         // expires; debit them before enforcing budget so a replay cannot
         // spend the same sats twice.
-        await this.reconcilePendingPayInvoiceActivities(connection);
+        try {
+            await this.reconcilePendingPayInvoiceActivities(connection);
+        } catch (error) {
+            console.error('NWC: pay_invoice reconcile failed:', error);
+        }
 
         const { paymentRequest, paymentHash: decodedPaymentHash } =
             await NostrConnectUtils.decodeInvoiceTags(request.invoice);
@@ -2279,12 +2283,16 @@ export default class NostrWalletConnectStore {
             decodedPaymentHash
         );
         if (existing && connection.isUnresolvedPayInvoiceActivity(existing)) {
-            return NostrConnectUtils.createNip47Error(
-                localeString(
-                    'stores.NostrWalletConnectStore.error.invoicePaymentInProgress'
-                ),
-                Nip47ErrorCode.FAILED_TO_PAY_INVOICE
-            );
+            const fees_paid =
+                existing.fees_paid || existing.payment?.getFee || 0;
+
+            return {
+                result: {
+                    preimage: '',
+                    fees_paid: satsToMillisats(Number(fees_paid) || 0)
+                },
+                error: undefined
+            };
         }
 
         const budgetCheck = this.validateBudgetBeforePayment(
@@ -2372,7 +2380,9 @@ export default class NostrWalletConnectStore {
 
         const settledEvidence =
             !!preimage ||
-            (!!payment && NostrConnectUtils.isSettledPayment(payment));
+            (!!payment &&
+                NostrConnectUtils.isSettledPayment(payment) &&
+                !!paymentHash);
 
         if (!settledEvidence) {
             const paymentError = this.checkPaymentErrors(
@@ -2778,20 +2788,15 @@ export default class NostrWalletConnectStore {
             paymentStatus: activity.payment?.status
         };
 
-        if (raw.isFailed) {
-            // CLN listpays does not populate failure_reason/htlcs, so
-            // Payment.isFailed may stay false for genuine CLN failures;
-            // those rows remain in the reconcile filter until expiry.
+        if (NostrConnectUtils.isListedPaymentFailed(raw)) {
             activity.status = 'failed';
-            activity.error =
-                listed.failure_reason &&
-                String(listed.failure_reason) !== 'FAILURE_REASON_NONE'
-                    ? String(listed.failure_reason)
-                    : localeString('error.paymentFailed');
-        } else if (raw.isIncomplete) {
+            activity.error = NostrConnectUtils.paymentFailureReasonLocaleKey(
+                listed.failure_reason
+            );
+        } else if (NostrConnectUtils.isListedPaymentInTransit(raw)) {
             activity.status = 'pending';
             activity.error = undefined;
-        } else {
+        } else if (!raw.isIncomplete) {
             activity.status = 'success';
             activity.error = undefined;
             const amountSats =
@@ -2805,6 +2810,8 @@ export default class NostrWalletConnectStore {
                 connection.trackSpending(spendSats, this.maxBudgetLimit);
                 activity.isBudgetDebited = true;
             }
+        } else {
+            return false;
         }
 
         activity.payment = listed;
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1809,6 +1809,8 @@ export default class NostrWalletConnectStore {
                 // Unresolved pay_invoice (pending or timeout-failed) must survive
                 // until reconcilePendingPayInvoiceActivities can debit. Active
                 // pending make_invoice must stay for the outstanding cap.
+                // Cashu pay_invoice is excluded: melt never records pending today;
+                // extend this guard if that changes (pendingSpendSats counts it too).
                 if (
                     activity.type === 'pay_invoice' &&
                     activity.payment_source !== 'cashu' &&
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2267,6 +2267,11 @@ export default class NostrWalletConnectStore {
             }
         }
 
+        // Late-settling pays can be recorded failed when the 125s wait
+        // expires; debit them before enforcing budget so a replay cannot
+        // spend the same sats twice.
+        await this.reconcilePendingPayInvoiceActivities(connection);
+
         const budgetCheck = this.validateBudgetBeforePayment(
             connection,
             amountSats,
@@ -2339,28 +2344,6 @@ export default class NostrWalletConnectStore {
             };
         }
 
-        const paymentError = this.checkPaymentErrors(
-            Nip47ErrorCode.FAILED_TO_PAY_INVOICE,
-            localeString('views.SendingLightning.paymentTimedOut'),
-            localeString(
-                'stores.NostrWalletConnectStore.error.noPreimageReceived'
-            )
-        );
-        if (
-            paymentError &&
-            paymentError.error &&
-            paymentError.error.code === Nip47ErrorCode.FAILED_TO_PAY_INVOICE
-        ) {
-            await this.recordFailedPayment({
-                rawInvoice: request.invoice,
-                connection,
-                amountSats,
-                payment_source: 'lightning',
-                errorMessage: paymentError.error?.message
-            });
-            return paymentError;
-        }
-
         const preimage = this.transactionsStore.payment_preimage;
         const fees_paid = this.transactionsStore.payment_fee;
 
@@ -2371,6 +2354,34 @@ export default class NostrWalletConnectStore {
                 payments,
                 paymentHash
             );
+
+        const settledEvidence =
+            !!preimage ||
+            (!!payment && NostrConnectUtils.isSettledPayment(payment));
+
+        if (!settledEvidence) {
+            const paymentError = this.checkPaymentErrors(
+                Nip47ErrorCode.FAILED_TO_PAY_INVOICE,
+                localeString('views.SendingLightning.paymentTimedOut'),
+                localeString(
+                    'stores.NostrWalletConnectStore.error.noPreimageReceived'
+                )
+            );
+            if (
+                paymentError &&
+                paymentError.error &&
+                paymentError.error.code === Nip47ErrorCode.FAILED_TO_PAY_INVOICE
+            ) {
+                await this.recordFailedPayment({
+                    rawInvoice: request.invoice,
+                    connection,
+                    amountSats,
+                    payment_source: 'lightning',
+                    errorMessage: paymentError.error?.message
+                });
+                return paymentError;
+            }
+        }
 
         const feeSats = Number(payment?.getFee) || Number(fees_paid) || 0;
 
@@ -2683,10 +2694,11 @@ export default class NostrWalletConnectStore {
     }
 
     /**
-     * Promotes pending pay_invoice activities that have since settled or failed.
-     * Used by the activity screen and by list_transactions — without the latter,
-     * an NWC client sees a settled payment as pending until someone opens the
-     * screen. Node fetches are rate limited by the helper below.
+     * Promotes pending (and timeout-failed) pay_invoice activities that have
+     * since settled or failed. Used by the activity screen, list_transactions,
+     * and pay_invoice budget checks. A payment recorded failed after the store
+     * wait can still settle; without this, that spend never hits the budget.
+     * Node fetches are rate limited by the helper below.
      */
     private async reconcilePendingPayInvoiceActivities(
         connection: NWCConnection
@@ -2694,8 +2706,11 @@ export default class NostrWalletConnectStore {
         const pending = connection.activity.filter(
             (a) =>
                 a.type === 'pay_invoice' &&
-                a.status === 'pending' &&
-                a.payment_source !== 'cashu'
+                a.payment_source !== 'cashu' &&
+                !a.isBudgetDebited &&
+                (a.status === 'pending' ||
+                    (a.status === 'failed' &&
+                        NostrConnectUtils.isPaymentTimedOutMessage(a.error)))
         );
         if (pending.length === 0) return false;
 
@@ -2716,7 +2731,13 @@ export default class NostrWalletConnectStore {
             if (!payment) continue;
 
             runInAction(() => {
-                if (activity.status !== 'pending') return;
+                if (activity.isBudgetDebited) return;
+                if (
+                    activity.status !== 'pending' &&
+                    activity.status !== 'failed'
+                ) {
+                    return;
+                }
 
                 activity.payment = new Payment(payment);
                 changed = true;
@@ -2725,9 +2746,14 @@ export default class NostrWalletConnectStore {
                     activity.status = 'failed';
                     return;
                 }
-                if (payment.isIncomplete) return;
+                if (payment.isIncomplete) {
+                    activity.status = 'pending';
+                    activity.error = undefined;
+                    return;
+                }
 
                 activity.status = 'success';
+                activity.error = undefined;
                 const amountSats =
                     Math.floor(Number(activity.satAmount)) ||
                     Math.floor(Number(payment.getAmount) || 0);
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2742,29 +2742,12 @@ export default class NostrWalletConnectStore {
         );
         if (pending.length === 0) return false;
 
-        for (const activity of pending) {
-            if (this.abandonStaleUnresolvedPayInvoiceHold(activity)) {
-                changed = true;
-            }
-        }
-
-        const stillUnresolved = connection.activity.filter((a) =>
-            connection.isUnresolvedPayInvoiceActivity(a)
-        );
-        if (stillUnresolved.length === 0) {
-            if (changed) {
-                this.scheduleMaxBudgetRefresh();
-                this.findAndUpdateConnection(connection);
-            }
-            return changed;
-        }
-
         const payments = await this.getPaymentsForPendingPayInvoiceRefresh(
             connection.id,
-            stillUnresolved
+            pending
         );
 
-        for (const activity of stillUnresolved) {
+        for (const activity of pending) {
             const payment = payments.find(
                 (p) =>
                     p.getPaymentRequest === activity.id ||
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -3201,7 +3201,9 @@ export default class NostrWalletConnectStore {
                 satAmount: amountSats,
                 status: 'failed',
                 payment_source,
-                error: errorMessage,
+                error: NostrConnectUtils.isPaymentTimedOutMessage(errorMessage)
+                    ? 'views.SendingLightning.paymentTimedOut'
+                    : errorMessage,
                 createdAt: new Date(),
                 ...(paymentHash ? { paymentHash } : {})
             });
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -2272,6 +2272,21 @@ export default class NostrWalletConnectStore {
         // spend the same sats twice.
         await this.reconcilePendingPayInvoiceActivities(connection);
 
+        const { paymentRequest, paymentHash: decodedPaymentHash } =
+            await NostrConnectUtils.decodeInvoiceTags(request.invoice);
+        const existing = connection.findPayInvoiceActivity(
+            paymentRequest || request.invoice,
+            decodedPaymentHash
+        );
+        if (existing && connection.isUnresolvedPayInvoiceActivity(existing)) {
+            return NostrConnectUtils.createNip47Error(
+                localeString(
+                    'stores.NostrWalletConnectStore.error.invoicePaymentInProgress'
+                ),
+                Nip47ErrorCode.FAILED_TO_PAY_INVOICE
+            );
+        }
+
         const budgetCheck = this.validateBudgetBeforePayment(
             connection,
             amountSats,
@@ -2703,14 +2718,8 @@ export default class NostrWalletConnectStore {
     private async reconcilePendingPayInvoiceActivities(
         connection: NWCConnection
     ): Promise<boolean> {
-        const pending = connection.activity.filter(
-            (a) =>
-                a.type === 'pay_invoice' &&
-                a.payment_source !== 'cashu' &&
-                !a.isBudgetDebited &&
-                (a.status === 'pending' ||
-                    (a.status === 'failed' &&
-                        NostrConnectUtils.isPaymentTimedOutMessage(a.error)))
+        const pending = connection.activity.filter((a) =>
+            connection.isUnresolvedPayInvoiceActivity(a)
         );
         if (pending.length === 0) return false;
 
@@ -2731,40 +2740,10 @@ export default class NostrWalletConnectStore {
             if (!payment) continue;
 
             runInAction(() => {
-                if (activity.isBudgetDebited) return;
                 if (
-                    activity.status !== 'pending' &&
-                    activity.status !== 'failed'
+                    this.applyPayInvoiceReconcile(activity, connection, payment)
                 ) {
-                    return;
-                }
-
-                activity.payment = new Payment(payment);
-                changed = true;
-
-                if (payment.isFailed) {
-                    activity.status = 'failed';
-                    return;
-                }
-                if (payment.isIncomplete) {
-                    activity.status = 'pending';
-                    activity.error = undefined;
-                    return;
-                }
-
-                activity.status = 'success';
-                activity.error = undefined;
-                const amountSats =
-                    Math.floor(Number(activity.satAmount)) ||
-                    Math.floor(Number(payment.getAmount) || 0);
-                const feeSats = Number(payment.getFee) || 0;
-                activity.fees_paid = feeSats;
-
-                const spendSats =
-                    amountSats + NostrConnectUtils.resolveFeeSats(feeSats);
-                if (spendSats > 0 && !activity.isBudgetDebited) {
-                    connection.trackSpending(spendSats, this.maxBudgetLimit);
-                    activity.isBudgetDebited = true;
+                    changed = true;
                 }
             });
         }
@@ -2774,6 +2753,70 @@ export default class NostrWalletConnectStore {
             this.findAndUpdateConnection(connection);
         }
         return changed;
+    }
+
+    private applyPayInvoiceReconcile(
+        activity: ConnectionActivity,
+        connection: NWCConnection,
+        raw: Payment
+    ): boolean {
+        if (activity.isBudgetDebited) return false;
+        if (activity.status !== 'pending' && activity.status !== 'failed') {
+            return false;
+        }
+
+        const listed = new Payment(raw);
+        const before = {
+            status: activity.status,
+            error: activity.error,
+            fees_paid: activity.fees_paid,
+            isBudgetDebited: !!activity.isBudgetDebited,
+            paymentHash:
+                activity.payment instanceof Payment
+                    ? activity.payment.paymentHash
+                    : undefined,
+            paymentStatus: activity.payment?.status
+        };
+
+        if (raw.isFailed) {
+            // CLN listpays does not populate failure_reason/htlcs, so
+            // Payment.isFailed may stay false for genuine CLN failures;
+            // those rows remain in the reconcile filter until expiry.
+            activity.status = 'failed';
+            activity.error =
+                listed.failure_reason &&
+                String(listed.failure_reason) !== 'FAILURE_REASON_NONE'
+                    ? String(listed.failure_reason)
+                    : localeString('error.paymentFailed');
+        } else if (raw.isIncomplete) {
+            activity.status = 'pending';
+            activity.error = undefined;
+        } else {
+            activity.status = 'success';
+            activity.error = undefined;
+            const amountSats =
+                Math.floor(Number(activity.satAmount)) ||
+                Math.floor(Number(raw.getAmount) || 0);
+            const feeSats = Number(raw.getFee) || 0;
+            activity.fees_paid = feeSats;
+            const spendSats =
+                amountSats + NostrConnectUtils.resolveFeeSats(feeSats);
+            if (spendSats > 0) {
+                connection.trackSpending(spendSats, this.maxBudgetLimit);
+                activity.isBudgetDebited = true;
+            }
+        }
+
+        activity.payment = listed;
+
+        return (
+            before.status !== activity.status ||
+            before.error !== activity.error ||
+            before.fees_paid !== activity.fees_paid ||
+            before.isBudgetDebited !== !!activity.isBudgetDebited ||
+            before.paymentHash !== listed.paymentHash ||
+            before.paymentStatus !== listed.status
+        );
     }
 
     private async getPaymentsForPendingPayInvoiceRefresh(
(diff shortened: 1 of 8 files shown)


stores/NostrWalletConnectStore.ts
```diff
@@ -2372,6 +2355,34 @@ export default class NostrWalletConnectStore {
...
+                localeString(
+                    'stores.NostrWalletConnectStore.error.noPreimageReceived'
+                )
+            );
+            if (
+                paymentError &&
+                paymentError.error &&
+                paymentError.error.code === Nip47ErrorCode.FAILED_TO_PAY_INVOICE
+            ) {
+                await this.recordFailedPayment({
```
kaloudis, line 2407: **Blocking:** a replay of the same invoice can still release the hold and lose the debit, and it's exactly the scenario this PR targets (restart + replay of a timed-out pay), whenever there's budget headroom.

1. The timed-out pay is now recorded `pending`, held via `pendingSpendSats`. Good.
2. The replay arrives. Reconcile keeps the row pending (node shows in-flight). If remaining budget >= amount, the budget check passes and `sendPayment` fires again.
3. lnd rejects the duplicate with `payment is in transition` (`routing.ErrPaymentInFlight`). That message is neither in `isIgnorableError` nor matched by `isPaymentTimedOutMessage`, so `recordFailedPayment` runs here, and `upsertPayInvoiceActivity` (which only refuses to demote `success` rows) overwrites the pending row to `failed` with a non-timeout error. The `pendingSpendSats` hold is released while the HTLC is still out.
4. When it settles, the reconcile filter (`failed` + timeout message) no longer matches the row, so the spend never hits `totalSpendSats`. Same escape class this PR is fixing.

Repro (passes on this head; drop into `NostrWalletConnectStore.test.ts`):

<details>
<summary>failing scenario as a test</summary>

```ts
it('releases pendingSpendSats and loses the debit when a replay records "payment is in transition"', async () => {
    const store = buildPayInvoiceTestStore();
    const invoice = 'lnbc1replay-demote';
    const paymentHash = hex64('a');
    jest.spyOn(NostrConnectUtils, 'decodeInvoiceTags').mockResolvedValue({
        paymentRequest: invoice,
        paymentHash,
        descriptionHash: '',
        description: '',
        amount: 10000,
        expiryTime: 0,
        createdAt: 0,
        isExpired: false,
        network: 'bitcoin'
    } as any);
    const connection = seedPayInvoiceConnection(store, {
        maxAmountSats: 100000,
        totalSpendSats: 0,
        activity: [
            {
                id: invoice,
                type: 'pay_invoice',
                payment_source: 'lightning',
                status: 'pending',
                satAmount: 10000,
                paymentHash,
                createdAt: new Date(Date.now() - 60_000)
            }
        ]
    });
    expect(connection.pendingSpendSats).toBe(10000);

    // Replay of same invoice: lnd rejects the duplicate in-flight payment,
    // handler hits recordFailedPayment with the duplicate error.
    await (store as any).recordFailedPayment({
        rawInvoice: invoice,
        connection,
        amountSats: 10000,
        payment_source: 'lightning',
        errorMessage: 'payment is in transition'
    });

    expect(connection.activity[0].status).toBe('failed');
    expect(connection.pendingSpendSats).toBe(0); // hold released, HTLC still out

    // Original HTLC later settles; reconcile filter no longer matches
    // (failed + non-timeout error), so the spend is never debited.
    const settled = new Payment({
        payment_request: invoice,
        payment_hash: paymentHash,
        status: 'SUCCEEDED',
        value_sat: 10000,
        fee_sat: '50',
        payment_preimage: hex64('9')
    });
    (store as any).paymentsStore = {
        payments: [settled],
        getPayments: jest.fn(async () => [settled])
    };
    await (store as any).reconcilePendingPayInvoiceActivities(connection);

    expect(connection.totalSpendSats).toBe(0); // escaped the budget
    expect(connection.canSpend(100000)).toBe(true);
});
```
</details>

Suggested fix: before paying, refuse to re-pay an invoice that already has an unresolved activity row for the same id/paymentHash (status `pending`, or `failed` with a timeout error) and return an error instead. That closes this path at the root and also covers the fetch-failure case where reconcile couldn't heal. Alternatively (or additionally), make `upsertPayInvoiceActivity` refuse to demote `pending` rows from the handler's error path, reserving pending -> failed for node-confirmed failure in reconcile.

stores/NostrWalletConnectStore.ts
```diff
@@ -2725,9 +2746,14 @@ export default class NostrWalletConnectStore {
                     activity.status = 'failed';
```
kaloudis: The `payment.isFailed` branch keeps `status = 'failed'` but leaves the timeout `error` in place, so the row matches the new filter on every subsequent pass: the payment is re-attached, `changed` is set, and `scheduleMaxBudgetRefresh()` + `findAndUpdateConnection()` re-persist the connection on every pay_invoice, list_transactions, and activity-screen open, for as long as the row survives prune.

Note the new flow keeps generating such rows: when a timeout resolves to a listed failed payment, the failure is recorded with the *timeout* message from `checkPaymentErrors`, not the node's failure reason.

Repro (passes on this head): three consecutive reconcile calls all report changed and persist.

<details>
<summary>churn repro test</summary>

```ts
it('keeps matching the filter and re-persisting after the node confirms failure', async () => {
    const store = buildPayInvoiceTestStore();
    const invoice = 'lnbc1churn';
    const paymentHash = hex64('b');
    const failedPayment = new Payment({
        payment_request: invoice,
        payment_hash: paymentHash,
        status: 'FAILED',
        failure_reason: 'FAILURE_REASON_NO_ROUTE',
        value_sat: 10000
    });
    (store as any).paymentsStore = {
        payments: [failedPayment],
        getPayments: jest.fn(async () => [failedPayment])
    };
    const connection = seedPayInvoiceConnection(store, {
        maxAmountSats: 10000,
        activity: [
            {
                id: invoice,
                type: 'pay_invoice',
                payment_source: 'lightning',
                status: 'failed',
                satAmount: 10000,
                paymentHash,
                error: 'views.SendingLightning.paymentTimedOut',
                createdAt: new Date(Date.now() - 600_000)
            }
        ]
    });
    const persistSpy = (store as any).findAndUpdateConnection;

    const first = await (store as any).reconcilePendingPayInvoiceActivities(connection);
    const second = await (store as any).reconcilePendingPayInvoiceActivities(connection);
    const third = await (store as any).reconcilePendingPayInvoiceActivities(connection);

    expect([first, second, third]).toEqual([true, true, true]);
    expect(persistSpy.mock.calls.length).toBe(3);
});
```
</details>

Suggested: when the node confirms failure, overwrite `activity.error` with the node's failure reason (or otherwise mark the row) so it exits the filter, and only set `changed` when something actually changed.

utils/NostrConnectUtils.ts
```diff
@@ -1170,13 +1170,25 @@ export default class NostrConnectUtils {
...
+            }
+            if (listed && listed.isFailed) {
+                return { inTransit: false, payment: listed };
+            }
+            // Timeout / still-loading with no listed outcome: the HTLC may
+            // still settle. Treat as in-flight so budget is held via
+            // pendingSpendSats instead of recording a terminal failure.
+            if (timedOutOrStillLoading) {
+                return { inTransit: true, payment: listed };
+            }
```
kaloudis, line 1285: This branch changes the client-facing semantics: a timeout with *nothing* listed on the node now returns a NIP-47 success result with an empty preimage instead of `FAILED_TO_PAY_INVOICE`.

`isPaymentTimedOutMessage` matches any "timed out" substring, which includes transport-level REST/Tor timeouts to an unreachable node, where the request may never have left the device. In that case the client sees a success-shaped response and the budget is held indefinitely; the release valve for that hold is #4335's expiry pass, which is still a draft.

Fail-closed on budget seems right to me, but flagging for an explicit decision since the empty-preimage-success pattern is already NIP-47-questionable and this widens it to cases where no payment may exist at all.

stores/NostrWalletConnectStore.ts
```diff
@@ -2683,19 +2694,23 @@ export default class NostrWalletConnectStore {
...
         const pending = connection.activity.filter(
             (a) =>
                 a.type === 'pay_invoice' &&
-                a.status === 'pending' &&
-                a.payment_source !== 'cashu'
+                a.payment_source !== 'cashu' &&
+                !a.isBudgetDebited &&
+                (a.status === 'pending' ||
+                    (a.status === 'failed' &&
+                        NostrConnectUtils.isPaymentTimedOutMessage(a.error)))
```
kaloudis: Minor: `activity.error` is persisted as the localized string at record time, but `isPaymentTimedOutMessage` compares against the *current* locale plus an English substring. Rows recorded under a non-English locale stop matching after a locale switch, so those persisted bad records won't be healed. Probably acceptable, but worth knowing the healing path is locale-sensitive.

kaloudis (commented):
Concept ACK on the direction. Treating a timed-out pay with no listed outcome as in-flight (fail-closed hold) is the right call for the budget, and running reconcile before the budget check is the right placement. CI is green and both changed suites pass locally at 905eb90.

Two issues found, both confirmed with repro tests written against this head. The first reopens the escape this PR is closing, so I'd consider it blocking; details and repros inline.

1. A replay of the same invoice can demote the pending hold via lnd's "payment is in transition" error, releasing `pendingSpendSats` while the HTLC is still out; when it later settles, the reconcile filter no longer matches and the spend never hits `totalSpendSats` (inline at `recordFailedPayment`).
2. Confirmed-failed timeout rows re-match the reconcile filter forever, re-persisting the connection and scheduling a budget refresh on every pay_invoice / list_transactions / activity-screen open (inline in the reconcile loop).

Also inline: the NIP-47 semantics change on the new timeout branch, and a locale note on the filter.

Non-inline notes:

- **CLN predicate gap (pre-existing, widened here):** `Payment.isFailed` reads `failure_reason`/htlcs, which CLN `listpays` does not set, and `isIncomplete` is preimage-based. A genuinely failed CLN payment therefore reads as incomplete-but-not-failed, and under this PR becomes a permanent pending hold that reconcile can never flip to failed. Fail-closed, so no overspend, but worth a code comment; backend-aware lookups (as in #4335 round 2) are the structural fix.
- **Coordination with #4335:** that draft restructures `reconcilePendingPayInvoiceActivities` (resolution helper + expiry pass). Whichever lands second has real conflicts to resolve, and #4335's expiry failure message must not match `isPaymentTimedOutMessage` or expired rows re-enter this PR's filter and churn.
- **Test gap:** there's no handler-level test for the new `settledEvidence` branch (store error set, list shows settled, expect finalize + debit instead of recordFailed), nor for the replay path in the blocking comment.
- **PR body claim:** "a restart + replay cannot spend twice" is best effort. If the reconcile's `getPayments` fetch fails (Tor down, node unreachable), the cached/empty list yields no debit and the replay still passes the budget check. The same-invoice guard suggested inline covers that case too.

stores/NostrWalletConnectStore.ts
ajaysehwal: fixed in 1b40bdce12e31cedea50f5e54b16a77dae1be04c

stores/NostrWalletConnectStore.ts
ajaysehwal: fixed in 2f5cc27cd7e115a959976f7b2d06dd26260547c1

stores/NostrWalletConnectStore.ts
ajaysehwal: fixed in https://github.com/ZeusLN/zeus/commit/1b40bdce12e31cedea50f5e54b16a77dae1be04c

utils/NostrConnectUtils.ts
ajaysehwal: fixed in https://github.com/ZeusLN/zeus/commit/1b40bdce12e31cedea50f5e54b16a77dae1be04c

kaloudis:
Applied this onto its base (`2d65e2c`) in a scratch worktree and verified locally: `tsc` clean, prettier/eslint clean, the three NWC suites green (162 tests), full suite green.

The core diagnosis is right, and two of the changes are clear wins: moving `checkPaymentErrors` behind `settledEvidence` closes a real escape (pre-PR, a payment with a preimage in hand but `payment_error` set was recorded failed and never debited), and normalizing timeout errors to locale keys + `localeString(item.error) || item.error` in the activity modal is the correct shape.

But the "hold the budget" half introduces holds that can never be released. Most severe first.

---

### 1. CLN-shaped failures get promoted `failed` → `pending`, and that hold is permanent

`applyPayInvoiceReconcile` (`stores/NostrWalletConnectStore.ts:2791`) treats "not `isFailed`" as "still in flight":

```js
} else if (raw.isIncomplete) {
    activity.status = 'pending';
```

`Payment.isFailed` (`models/Payment.ts:169-191`) needs `htlcs[].status === FAILED` or a `failure_reason`. CLNRest's `getPayments()` SQL projection (`backends/CLNRest.ts:340-364`) returns neither — only `status`, `preimage`, amounts. So a genuinely failed CLN `sendpay` has `isFailed === false`, `isIncomplete === true`, and lands in the `pending` branch.

The comment at `:2782-2784` acknowledges the CLN gap but says those rows "remain in the reconcile filter until expiry." There is no expiry for a pending row:

- `pruneConnectionActivity` (`:1815-1821`) explicitly skips pending `pay_invoice`, so the row is never evicted at the 100-entry cap. Failed rows *are* prunable — the promotion is what makes it permanent.
- `pendingSpendSats` (`models/NWCConnection.ts:225-233`) is derived from activity, and `resetBudget()` (`:437`) only zeroes `totalSpendSats`. **The hold survives every budget renewal, forever.**
- The new replay guard (`:2281`) then returns "payment in progress" for that invoice forever.
- No UI path clears a connection's activity log — the only recovery is deleting the connection and its budget.

Reproduced with a CLN-shaped payment row:

```
isFailed: false  isIncomplete: true
status after reconcile:  pending
pendingSpendSats: 10000      canSpend(1): false
after resetBudget: totalSpend 0, pending 10000, canSpend(1) false
```

<details>
<summary>repro test</summary>

```ts
describe('CLN-shaped failed payment reconcile', () => {
    it('promotes a timeout-failed row to permanently pending on CLN', async () => {
        const store = buildPayInvoiceTestStore();
        const invoice = 'lnbc1cln-failed';
        const paymentHash = hex64('7');
        // CLNRest getPayments() row shape: no htlcs, no failure_reason,
        // preimage null for a genuinely failed sendpay.
        const clnFailed = new Payment({
            payment_hash: paymentHash,
            status: 'failed',
            bolt11: invoice,
            payment_request: invoice,
            amount_msat: 0,
            amount_sent_msat: 0,
            preimage: null,
            created_at: 1
        } as any);
        (store as any).paymentsStore = {
            payments: [clnFailed],
            getPayments: jest.fn(async () => [clnFailed])
        };
        const connection = seedPayInvoiceConnection(store, {
            maxAmountSats: 10000,
            totalSpendSats: 0,
            budgetRenewal: 'daily',
            activity: [
                {
                    id: invoice,
                    type: 'pay_invoice',
                    payment_source: 'lightning',
                    status: 'failed',
                    satAmount: 10000,
                    paymentHash,
                    error: 'views.SendingLightning.paymentTimedOut',
                    createdAt: new Date(Date.now() - 60_000)
                }
            ]
        });

        await (store as any).reconcilePendingPayInvoiceActivities(connection);

        expect(connection.activity[0].status).toBe('pending');
        expect(connection.pendingSpendSats).toBe(10000);
        expect(connection.canSpend(1)).toBe(false);

        // budget renewal does not release the hold
        connection.resetBudget();
        expect(connection.canSpend(1)).toBe(false);
    });
});
```

</details>

The same shape reaches the handler directly, not just reconcile: `resolveLightningPaymentInTransit` (`utils/NostrConnectUtils.ts:1189-1191`) returns `{inTransit: true}` on timeout when the listed payment is neither settled nor `isFailed`.

Suggest gating the promotion on *positive* evidence of in-flight (`isInTransit`, or `status` in {`IN_FLIGHT`, `pending`}) rather than the absence of failure evidence, and treating a listed payment with a terminal non-success status as failed. LndHub is a related case — `/gettxs` (`backends/LndHub.ts:33`) never lists failed payments, so those rows just never resolve.

### 2. Both escape hatches for a stuck pending row are removed in the same PR

Pre-PR, a stale pending row could be cleared by the next request for that invoice: `upsertPayInvoiceActivity` overwrote pending → failed. This PR closes that twice over — the replay guard short-circuits before any write, and the new demote guard (`:3236-3242`) refuses pending → failed (the PR's own test asserts this).

Independent of #1, that means any hold the node never resolves is unrecoverable. A TTL on the hold — release once the invoice's own expiry has passed, or after N hours — would make the guard safe to merge. Right now the failure mode is silent and permanent.

### 3. The replay guard reports `FAILED_TO_PAY_INVOICE` for a payment that may still settle

`:2281-2288`. This is inconsistent with the same handler's in-transit response: the *first* request that times out returns success with `preimage: ''` (`:2353-2359`), but a retry of that same live payment returns a terminal payment failure. A client that surfaces "payment failed" invites the user to pay again by another route while the HTLC is still out — the exact double-spend this PR is trying to prevent, moved up a layer.

Either return the same `{preimage: ''}` in-transit shape, or `RATE_LIMITED` (already used for the stale-queue case at `:1358`).

### 4. Reconciled failures write a raw enum — a bare number on embedded LND and LNC

`:2786-2790` stores `String(listed.failure_reason)`. Embedded LND returns the decoded protobuf message (`lndmobile/index.ts:632`) and LNC `snakeize`s without enum→string conversion (`backends/LightningNodeConnect.ts:211`), so `failure_reason` is numeric — which is why `Payment.isFailed` compares against `lnrpc.PaymentFailureReason.FAILURE_REASON_NONE` at all. Verified: `activity.error` becomes `"2"`, and the failure modal renders `2`.

`utils/ErrorUtils.ts:14-19` already maps `FAILURE_REASON_*` → locale keys. Using it fixes the display and keeps this path locale-stable — unlike the `localeString('error.paymentFailed')` fallback on the next line, which persists localized text into the activity log and partly undoes what 2f5cc27 set out to fix.

### 5. `settledEvidence` debits the budget for an invoice settled outside NWC

`:2373-2399`. If the user paid invoice X from the Zeus UI and an NWC client then requests it, `sendPayment` errors ("invoice is already paid"), `findPaymentForInvoice` returns the settled payment, `checkPaymentErrors` is skipped, and `finalizePayment` debits the connection's budget and returns the preimage as success. Bounded to once per invoice and fail-closed, so not urgent — but a client can shave budget without the wallet spending anything. Worth being a deliberate decision rather than a side effect.

### 6. Reconcile can now throw out of the pay path

`PaymentsStore.getPayments()` rethrows (`stores/PaymentsStore.ts:46-49`), and the new call at `:2273` isn't wrapped — unlike the `list_transactions` and background callers (`:1479`, `:2144`). A Tor blip now rejects the whole `pay_invoice` handler before the budget check even runs.

### 7. Nit

`locales/en.json` gains `...error.invoicePaymentInProgress`, but the checklist has "added new locale text" unticked.

---

#1 and #2 need addressing before merge — as written, a CLN user (or anyone whose node stops listing a payment) can end up with a connection whose budget is permanently consumed and one invoice permanently unpayable, with no in-app recovery. #3 is a small change with real safety value. #4–#6 can ride along or be follow-ups.

ajaysehwal:
> Applied this onto its base (`2d65e2c`) in a scratch worktree and verified locally: `tsc` clean, prettier/eslint clean, the three NWC suites green (162 tests), full suite green.
> 
> The core diagnosis is right, and two of the changes are clear wins: moving `checkPaymentErrors` behind `settledEvidence` closes a real escape (pre-PR, a payment with a preimage in hand but `payment_error` set was recorded failed and never debited), and normalizing timeout errors to locale keys + `localeString(item.error) || item.error` in the activity modal is the correct shape.
> 
> But the "hold the budget" half introduces holds that can never be released. Most severe first.
> 
> ### 1. CLN-shaped failures get promoted `failed` → `pending`, and that hold is permanent
> `applyPayInvoiceReconcile` (`stores/NostrWalletConnectStore.ts:2791`) treats "not `isFailed`" as "still in flight":
> 
> ```js
> } else if (raw.isIncomplete) {
>     activity.status = 'pending';
> ```
> 
> `Payment.isFailed` (`models/Payment.ts:169-191`) needs `htlcs[].status === FAILED` or a `failure_reason`. CLNRest's `getPayments()` SQL projection (`backends/CLNRest.ts:340-364`) returns neither — only `status`, `preimage`, amounts. So a genuinely failed CLN `sendpay` has `isFailed === false`, `isIncomplete === true`, and lands in the `pending` branch.
> 
> The comment at `:2782-2784` acknowledges the CLN gap but says those rows "remain in the reconcile filter until expiry." There is no expiry for a pending row:
> 
> * `pruneConnectionActivity` (`:1815-1821`) explicitly skips pending `pay_invoice`, so the row is never evicted at the 100-entry cap. Failed rows _are_ prunable — the promotion is what makes it permanent.
> * `pendingSpendSats` (`models/NWCConnection.ts:225-233`) is derived from activity, and `resetBudget()` (`:437`) only zeroes `totalSpendSats`. **The hold survives every budget renewal, forever.**
> * The new replay guard (`:2281`) then returns "payment in progress" for that invoice forever.
> * No UI path clears a connection's activity log — the only recovery is deleting the connection and its budget.
> 
> Reproduced with a CLN-shaped payment row:
> 
> ```
> isFailed: false  isIncomplete: true
> status after reconcile:  pending
> pendingSpendSats: 10000      canSpend(1): false
> after resetBudget: totalSpend 0, pending 10000, canSpend(1) false
> ```
> 
> repro test
> ```ts
> describe('CLN-shaped failed payment reconcile', () => {
>     it('promotes a timeout-failed row to permanently pending on CLN', async () => {
>         const store = buildPayInvoiceTestStore();
>         const invoice = 'lnbc1cln-failed';
>         const paymentHash = hex64('7');
>         // CLNRest getPayments() row shape: no htlcs, no failure_reason,
>         // preimage null for a genuinely failed sendpay.
>         const clnFailed = new Payment({
>             payment_hash: paymentHash,
>             status: 'failed',
>             bolt11: invoice,
>             payment_request: invoice,
>             amount_msat: 0,
>             amount_sent_msat: 0,
>             preimage: null,
>             created_at: 1
>         } as any);
>         (store as any).paymentsStore = {
>             payments: [clnFailed],
>             getPayments: jest.fn(async () => [clnFailed])
>         };
>         const connection = seedPayInvoiceConnection(store, {
>             maxAmountSats: 10000,
>             totalSpendSats: 0,
>             budgetRenewal: 'daily',
>             activity: [
>                 {
>                     id: invoice,
>                     type: 'pay_invoice',
>                     payment_source: 'lightning',
>                     status: 'failed',
>                     satAmount: 10000,
>                     paymentHash,
>                     error: 'views.SendingLightning.paymentTimedOut',
>                     createdAt: new Date(Date.now() - 60_000)
>                 }
>             ]
>         });
> 
>         await (store as any).reconcilePendingPayInvoiceActivities(connection);
> 
>         expect(connection.activity[0].status).toBe('pending');
>         expect(connection.pendingSpendSats).toBe(10000);
>         expect(connection.canSpend(1)).toBe(false);
> 
>         // budget renewal does not release the hold
>         connection.resetBudget();
>         expect(connection.canSpend(1)).toBe(false);
>     });
> });
> ```
> 
> The same shape reaches the handler directly, not just reconcile: `resolveLightningPaymentInTransit` (`utils/NostrConnectUtils.ts:1189-1191`) returns `{inTransit: true}` on timeout when the listed payment is neither settled nor `isFailed`.
> 
> Suggest gating the promotion on _positive_ evidence of in-flight (`isInTransit`, or `status` in {`IN_FLIGHT`, `pending`}) rather than the absence of failure evidence, and treating a listed payment with a terminal non-success status as failed. LndHub is a related case — `/gettxs` (`backends/LndHub.ts:33`) never lists failed payments, so those rows just never resolve.
> 
> ### 2. Both escape hatches for a stuck pending row are removed in the same PR
> Pre-PR, a stale pending row could be cleared by the next request for that invoice: `upsertPayInvoiceActivity` overwrote pending → failed. This PR closes that twice over — the replay guard short-circuits before any write, and the new demote guard (`:3236-3242`) refuses pending → failed (the PR's own test asserts this).
> 
> Independent of #1, that means any hold the node never resolves is unrecoverable. A TTL on the hold — release once the invoice's own expiry has passed, or after N hours — would make the guard safe to merge. Right now the failure mode is silent and permanent.
> 
> ### 3. The replay guard reports `FAILED_TO_PAY_INVOICE` for a payment that may still settle
> `:2281-2288`. This is inconsistent with the same handler's in-transit response: the _first_ request that times out returns success with `preimage: ''` (`:2353-2359`), but a retry of that same live payment returns a terminal payment failure. A client that surfaces "payment failed" invites the user to pay again by another route while the HTLC is still out — the exact double-spend this PR is trying to prevent, moved up a layer.
> 
> Either return the same `{preimage: ''}` in-transit shape, or `RATE_LIMITED` (already used for the stale-queue case at `:1358`).
> 
> ### 4. Reconciled failures write a raw enum — a bare number on embedded LND and LNC
> `:2786-2790` stores `String(listed.failure_reason)`. Embedded LND returns the decoded protobuf message (`lndmobile/index.ts:632`) and LNC `snakeize`s without enum→string conversion (`backends/LightningNodeConnect.ts:211`), so `failure_reason` is numeric — which is why `Payment.isFailed` compares against `lnrpc.PaymentFailureReason.FAILURE_REASON_NONE` at all. Verified: `activity.error` becomes `"2"`, and the failure modal renders `2`.
> 
> `utils/ErrorUtils.ts:14-19` already maps `FAILURE_REASON_*` → locale keys. Using it fixes the display and keeps this path locale-stable — unlike the `localeString('error.paymentFailed')` fallback on the next line, which persists localized text into the activity log and partly undoes what [2f5cc27](https://github.com/ZeusLN/zeus/commit/2f5cc27cd7e115a959976f7b2d06dd26260547c1) set out to fix.
> 
> ### 5. `settledEvidence` debits the budget for an invoice settled outside NWC
> `:2373-2399`. If the user paid invoice X from the Zeus UI and an NWC client then requests it, `sendPayment` errors ("invoice is already paid"), `findPaymentForInvoice` returns the settled payment, `checkPaymentErrors` is skipped, and `finalizePayment` debits the connection's budget and returns the preimage as success. Bounded to once per invoice and fail-closed, so not urgent — but a client can shave budget without the wallet spending anything. Worth being a deliberate decision rather than a side effect.
> 
> ### 6. Reconcile can now throw out of the pay path
> `PaymentsStore.getPayments()` rethrows (`stores/PaymentsStore.ts:46-49`), and the new call at `:2273` isn't wrapped — unlike the `list_transactions` and background callers (`:1479`, `:2144`). A Tor blip now rejects the whole `pay_invoice` handler before the budget check even runs.
> 
> ### 7. Nit
> `locales/en.json` gains `...error.invoicePaymentInProgress`, but the checklist has "added new locale text" unticked.
> 
> #1 and #2 need addressing before merge — as written, a CLN user (or anyone whose node stops listing a payment) can end up with a connection whose budget is permanently consumed and one invoice permanently unpayable, with no in-app recovery. #3 is a small change with real safety value. #4–#6 can ride along or be follow-ups.

fixed in e5cc326ecdc98657b3eaaa4b24dca17865d85450

kaloudis:
Re-reviewed at fa2f764. CI is green and the three NWC suites (165 tests) pass locally at this head. Disposition of my last review, item by item:

### 1. CLN failed-to-pending promotion: fixed

`isListedPaymentFailed` catches CLN's htlc-less `status: 'failed'` rows in both reconcile and `resolveLightningPaymentInTransit`, and the pending promotion now requires positive in-flight evidence via `isListedPaymentInTransit` instead of the absence of failure evidence. The new tests in both suites cover the shape I reproduced, and the reconciled row exits the timeout filter (error becomes the `error.paymentFailed` locale key), so it does not re-enter the churn path either.

### 2. No escape hatch for stuck holds: not addressed, still blocking

Neither commit adds a TTL, and fa2f764 widens prune protection to timeout-failed rows. Item 1 shrinks the set of ways to get here, but the failure mode is unchanged wherever the node never lists the payment: LndHub (`/gettxs` never lists failed pays), payment history that ages out or paginates away, or a restored node. At this head:

- reconcile is a no-op for the row (`if (!payment) continue`), so it stays `pending`
- `resetBudget()` only zeroes `totalSpendSats`, so the hold survives every budget renewal
- prune never evicts the row
- the replay guard now returns empty-preimage success for that invoice forever, so the client sees an eternally in-flight payment

Deleting the connection remains the only recovery. Repro (passes at fa2f764, drop into `NostrWalletConnectStore.test.ts`):

<details>
<summary>permanent-hold repro test</summary>

```ts
describe('unresolved hold permanence', () => {
    it('hold survives reconcile, budget renewal, and prune when the node never lists the payment', async () => {
        const store = buildPayInvoiceTestStore();
        const invoice = 'lnbc1lndhub-lost';
        const paymentHash = hex64('c');
        (store as any).paymentsStore = {
            payments: [],
            getPayments: jest.fn(async () => []) // LndHub-like: failed pays never listed
        };
        const connection = seedPayInvoiceConnection(store, {
            maxAmountSats: 10000,
            totalSpendSats: 0,
            budgetRenewal: 'daily',
            activity: [
                {
                    id: invoice,
                    type: 'pay_invoice',
                    payment_source: 'lightning',
                    status: 'pending',
                    satAmount: 10000,
                    paymentHash,
                    createdAt: new Date(Date.now() - 30 * 24 * 3600 * 1000)
                }
            ]
        });

        const changed = await (
            store as any
        ).reconcilePendingPayInvoiceActivities(connection);
        expect(changed).toBe(false);
        expect(connection.activity[0].status).toBe('pending');
        expect(connection.pendingSpendSats).toBe(10000);
        expect(connection.canSpend(1)).toBe(false);

        // budget renewal does not release the hold
        connection.resetBudget();
        expect(connection.canSpend(1)).toBe(false);

        // prune never evicts it
        for (let i = 0; i < 120; i++) {
            connection.activity.push({
                id: `fill-${i}`,
                type: 'make_invoice',
                payment_source: 'lightning',
                status: 'success',
                createdAt: new Date()
            } as any);
        }
        (store as any).pruneConnectionActivity(connection);
        const row = connection.activity.find((a: any) => a.id === invoice);
        expect(row).toBeDefined();

        // replay guard treats it as unresolved forever -> empty-preimage success
        expect(connection.isUnresolvedPayInvoiceActivity(row!)).toBe(true);
    });
});
```
</details>

Suggested shape: release the hold once the invoice's own expiry plus a grace period has passed, or after N days without the node listing the payment. Flip the row to `failed` with a non-timeout error so it exits the reconcile filter and becomes prunable. Either bound makes the hold safe to merge; without one, a single unlisted payment permanently consumes budget and permanently fake-succeeds that invoice, with connection deletion as the only way out.

### 3. Replay guard response: fixed

Returns the same `{preimage: '', fees_paid}` in-transit shape as the first request, and the locale key is gone. Note the interaction with item 2: a permanently unresolved invoice now reads as in-flight forever rather than failing loudly, which raises the value of a TTL.

### 4. Raw failure_reason enum: fixed

`paymentFailureReasonLocaleKey` handles both string names and numeric enums (the protobufjs reverse lookup works via the `Object.create(valuesById)` chain), all the `error.failureReason*` keys exist in `en.json`, and the stored value is a locale key rather than localized text. Nit: this duplicates the `FAILURE_REASON_*` map in `utils/ErrorUtils.ts`; fine to leave, but a shared map would keep the two from drifting.

### 5. settledEvidence debiting external settles: reverted, not fixed

e5cc326 added a `!!paymentHash` guard here, then fa2f764 removed it again, I assume because `payment_hash` is not populated on all backends and requiring it would reopen the missed-debit escape. That is the right call between the two, but the net behavior is identical to what I flagged: an invoice settled outside NWC still debits the connection budget once. I am fine accepting that as deliberate; please add a short code comment saying so (or spin a follow-up issue) so it does not read as an oversight later.

### 6. Reconcile throw in the pay path: fixed

try/catch in place, and the replay guard below it does not depend on reconcile succeeding, so the fetch-failure plus replay case stays closed.

### 7. Locale checklist: moot

The key was deleted, so the PR now has a net-zero `locales/en.json` diff. The PR body still ticks "added new locale text"; worth unticking.

---

One new minor note from this pass: the rewritten prune predicate no longer protects pending cashu `pay_invoice` rows. Unreachable today (the melt path never records pending, per the comment at its call site), but `pendingSpendSats` has no cashu exclusion, so if pending recording is ever added for cashu, prune would silently release counted holds. A one-line comment on the predicate would prevent that trap.

From my side, item 2 is the one thing left between this and merge, plus the small comment ask on item 5.

ajaysehwal:
> Re-reviewed at [fa2f764](https://github.com/ZeusLN/zeus/commit/fa2f7641f91877efd5d6aba60cc21144fb6ab757). CI is green and the three NWC suites (165 tests) pass locally at this head. Disposition of my last review, item by item:
> 
> ### 1. CLN failed-to-pending promotion: fixed
> `isListedPaymentFailed` catches CLN's htlc-less `status: 'failed'` rows in both reconcile and `resolveLightningPaymentInTransit`, and the pending promotion now requires positive in-flight evidence via `isListedPaymentInTransit` instead of the absence of failure evidence. The new tests in both suites cover the shape I reproduced, and the reconciled row exits the timeout filter (error becomes the `error.paymentFailed` locale key), so it does not re-enter the churn path either.
> 
> ### 2. No escape hatch for stuck holds: not addressed, still blocking
> Neither commit adds a TTL, and [fa2f764](https://github.com/ZeusLN/zeus/commit/fa2f7641f91877efd5d6aba60cc21144fb6ab757) widens prune protection to timeout-failed rows. Item 1 shrinks the set of ways to get here, but the failure mode is unchanged wherever the node never lists the payment: LndHub (`/gettxs` never lists failed pays), payment history that ages out or paginates away, or a restored node. At this head:
> 
> * reconcile is a no-op for the row (`if (!payment) continue`), so it stays `pending`
> * `resetBudget()` only zeroes `totalSpendSats`, so the hold survives every budget renewal
> * prune never evicts the row
> * the replay guard now returns empty-preimage success for that invoice forever, so the client sees an eternally in-flight payment
> 
> Deleting the connection remains the only recovery. Repro (passes at [fa2f764](https://github.com/ZeusLN/zeus/commit/fa2f7641f91877efd5d6aba60cc21144fb6ab757), drop into `NostrWalletConnectStore.test.ts`):
> 
> permanent-hold repro test
> ```ts
> describe('unresolved hold permanence', () => {
>     it('hold survives reconcile, budget renewal, and prune when the node never lists the payment', async () => {
>         const store = buildPayInvoiceTestStore();
>         const invoice = 'lnbc1lndhub-lost';
>         const paymentHash = hex64('c');
>         (store as any).paymentsStore = {
>             payments: [],
>             getPayments: jest.fn(async () => []) // LndHub-like: failed pays never listed
>         };
>         const connection = seedPayInvoiceConnection(store, {
>             maxAmountSats: 10000,
>             totalSpendSats: 0,
>             budgetRenewal: 'daily',
>             activity: [
>                 {
>                     id: invoice,
>                     type: 'pay_invoice',
>                     payment_source: 'lightning',
>                     status: 'pending',
>                     satAmount: 10000,
>                     paymentHash,
>                     createdAt: new Date(Date.now() - 30 * 24 * 3600 * 1000)
>                 }
>             ]
>         });
> 
>         const changed = await (
>             store as any
>         ).reconcilePendingPayInvoiceActivities(connection);
>         expect(changed).toBe(false);
>         expect(connection.activity[0].status).toBe('pending');
>         expect(connection.pendingSpendSats).toBe(10000);
>         expect(connection.canSpend(1)).toBe(false);
> 
>         // budget renewal does not release the hold
>         connection.resetBudget();
>         expect(connection.canSpend(1)).toBe(false);
> 
>         // prune never evicts it
>         for (let i = 0; i < 120; i++) {
>             connection.activity.push({
>                 id: `fill-${i}`,
>                 type: 'make_invoice',
>                 payment_source: 'lightning',
>                 status: 'success',
>                 createdAt: new Date()
>             } as any);
>         }
>         (store as any).pruneConnectionActivity(connection);
>         const row = connection.activity.find((a: any) => a.id === invoice);
>         expect(row).toBeDefined();
> 
>         // replay guard treats it as unresolved forever -> empty-preimage success
>         expect(connection.isUnresolvedPayInvoiceActivity(row!)).toBe(true);
>     });
> });
> ```
> 
> Suggested shape: release the hold once the invoice's own expiry plus a grace period has passed, or after N days without the node listing the payment. Flip the row to `failed` with a non-timeout error so it exits the reconcile filter and becomes prunable. Either bound makes the hold safe to merge; without one, a single unlisted payment permanently consumes budget and permanently fake-succeeds that invoice, with connection deletion as the only way out.
> 
> ### 3. Replay guard response: fixed
> Returns the same `{preimage: '', fees_paid}` in-transit shape as the first request, and the locale key is gone. Note the interaction with item 2: a permanently unresolved invoice now reads as in-flight forever rather than failing loudly, which raises the value of a TTL.
> 
> ### 4. Raw failure_reason enum: fixed
> `paymentFailureReasonLocaleKey` handles both string names and numeric enums (the protobufjs reverse lookup works via the `Object.create(valuesById)` chain), all the `error.failureReason*` keys exist in `en.json`, and the stored value is a locale key rather than localized text. Nit: this duplicates the `FAILURE_REASON_*` map in `utils/ErrorUtils.ts`; fine to leave, but a shared map would keep the two from drifting.
> 
> ### 5. settledEvidence debiting external settles: reverted, not fixed
> [e5cc326](https://github.com/ZeusLN/zeus/commit/e5cc326ecdc98657b3eaaa4b24dca17865d85450) added a `!!paymentHash` guard here, then [fa2f764](https://github.com/ZeusLN/zeus/commit/fa2f7641f91877efd5d6aba60cc21144fb6ab757) removed it again, I assume because `payment_hash` is not populated on all backends and requiring it would reopen the missed-debit escape. That is the right call between the two, but the net behavior is identical to what I flagged: an invoice settled outside NWC still debits the connection budget once. I am fine accepting that as deliberate; please add a short code comment saying so (or spin a follow-up issue) so it does not read as an oversight later.
> 
> ### 6. Reconcile throw in the pay path: fixed
> try/catch in place, and the replay guard below it does not depend on reconcile succeeding, so the fetch-failure plus replay case stays closed.
> 
> ### 7. Locale checklist: moot
> The key was deleted, so the PR now has a net-zero `locales/en.json` diff. The PR body still ticks "added new locale text"; worth unticking.
> 
> One new minor note from this pass: the rewritten prune predicate no longer protects pending cashu `pay_invoice` rows. Unreachable today (the melt path never records pending, per the comment at its call site), but `pendingSpendSats` has no cashu exclusion, so if pending recording is ever added for cashu, prune would silently release counted holds. A one-line comment on the predicate would prevent that trap.
> 
> From my side, item 2 is the one thing left between this and merge, plus the small comment ask on item 5.

@kaloudis, Thanks for the detailed review. All mentioned issues are addressed in aea72d70b30bef8f59930e94e7d8ead64dbb7573 and 3bbf6c4f9c784c122ca265edfcab924a78912ded

kaloudis:
Re-reviewed at 3bbf6c4. CI is green, and since this branch predates the nostr-tools 2.x migration I verified locally against the merged result with master: the three NWC suites pass there (174 tests).

Closed out from last round:

- **Item 5 comment ask: done.** The `settledEvidence` debit is now marked deliberate at the call site.
- **Cashu prune note: done.**
- **Locale checklist: done**, matches the net-zero `en.json` diff.

### Item 2: right mechanism, wrong ordering; still blocking

`isPayInvoiceHoldAbandoned` is the shape I suggested: 7-day cap, invoice expiry plus 24h grace, flip to `failed` with a non-timeout error so the row exits the reconcile filter and becomes prunable, and the replay guard recovers. The new tests cover the never-listed case well.

But `reconcilePendingPayInvoiceActivities` now abandons stale holds *before* consulting the node's payment list (`stores/NostrWalletConnectStore.ts:2745-2749`). Two consequences, both reproduced with probe tests at this head:

1. **A stale hold whose payment the node lists as settled is abandoned instead of debited.** The spend is never debited (an abandoned row exits the reconcile filter for good), the row no longer trips the replay guard, and a replayed `pay_invoice` for that invoice passes the budget check and can pay again on backends that re-pay the same invoice. That is the exact overspend this PR set out to close, reopened for any hold older than the abandon bound. And the window is not just the 7-day cap: for a standard 1-hour-expiry invoice, the expiry-plus-grace path abandons at roughly 25 hours. "Payment settles a few minutes after the 125s timeout, app stays closed for a day, client retries on next launch" walks straight into it, the same idle-then-replay sequence as the PR description.

2. **A stale hold the node still lists as IN_FLIGHT is abandoned too**, releasing the budget hold while the HTLC is still out, and reconcile will never debit that row if it later settles.

Repro (both fail at 3bbf6c4; drop into `NostrWalletConnectStore.test.ts` where the pay_invoice helpers are in scope):

<details>
<summary>probe tests</summary>

```ts
describe('stale hold vs node listing', () => {
    const STALE = () =>
        new Date(
            Date.now() -
                NostrConnectUtils.UNRESOLVED_PAY_INVOICE_MAX_AGE_MS -
                1000
        );

    it('debits a stale pending hold when the node lists the payment as settled', async () => {
        const store = buildPayInvoiceTestStore();
        const invoice = 'lnbc1stale-but-settled';
        const paymentHash = hex64('a');
        const settled = new Payment({
            payment_request: invoice,
            payment_hash: paymentHash,
            status: 'SUCCEEDED',
            value_sat: 10000,
            fee_sat: '50',
            payment_preimage: hex64('9')
        });
        (store as any).paymentsStore = {
            payments: [settled],
            getPayments: jest.fn(async () => [settled])
        };
        const connection = seedPayInvoiceConnection(store, {
            maxAmountSats: 10000,
            totalSpendSats: 0,
            activity: [
                {
                    id: invoice,
                    type: 'pay_invoice',
                    payment_source: 'lightning',
                    status: 'pending',
                    satAmount: 10000,
                    paymentHash,
                    createdAt: STALE()
                }
            ]
        });

        await (store as any).reconcilePendingPayInvoiceActivities(connection);

        expect(connection.activity[0].status).toBe('success');
        expect(connection.activity[0].isBudgetDebited).toBe(true);
        expect(connection.totalSpendSats).toBe(10050);
        expect(connection.canSpend(1)).toBe(false);
    });

    it('keeps a stale hold pending when the node still lists the payment in-flight', async () => {
        const store = buildPayInvoiceTestStore();
        const invoice = 'lnbc1stale-but-inflight';
        const paymentHash = hex64('b');
        const inFlight = new Payment({
            payment_request: invoice,
            payment_hash: paymentHash,
            status: 'IN_FLIGHT',
            payment_preimage:
                '0000000000000000000000000000000000000000000000000000000000000000'
        });
        (store as any).paymentsStore = {
            payments: [inFlight],
            getPayments: jest.fn(async () => [inFlight])
        };
        const connection = seedPayInvoiceConnection(store, {
            maxAmountSats: 10000,
            totalSpendSats: 0,
            activity: [
                {
                    id: invoice,
                    type: 'pay_invoice',
                    payment_source: 'lightning',
                    status: 'pending',
                    satAmount: 10000,
                    paymentHash,
                    createdAt: STALE()
                }
            ]
        });

        await (store as any).reconcilePendingPayInvoiceActivities(connection);

        expect(connection.activity[0].status).toBe('pending');
        expect(connection.pendingSpendSats).toBe(10000);
    });
});
```

</details>

The fix is small: delete the pre-fetch abandon loop and let the existing `!payment` branch do the abandoning, so a hold is only released after the node had its chance to answer and did not list the payment:

```diff
         );
         if (pending.length === 0) return false;
 
-        for (const activity of pending) {
-            if (this.abandonStaleUnresolvedPayInvoiceHold(activity)) {
-                changed = true;
-            }
-        }
-
-        const stillUnresolved = connection.activity.filter((a) =>
-            connection.isUnresolvedPayInvoiceActivity(a)
-        );
-        if (stillUnresolved.length === 0) {
-            if (changed) {
-                this.scheduleMaxBudgetRefresh();
-                this.findAndUpdateConnection(connection);
-            }
-            return changed;
-        }
-
         const payments = await this.getPaymentsForPendingPayInvoiceRefresh(
             connection.id,
-            stillUnresolved
+            pending
         );
 
-        for (const activity of stillUnresolved) {
+        for (const activity of pending) {
             const payment = payments.find(
                 (p) =>
                     p.getPaymentRequest === activity.id ||
```

Verified at this head with that patch applied: both probes above pass (settled gets debited, in-flight stays held), your three new abandonment tests still pass (their mocked `getPayments` returns `[]`, so they take the `!payment` branch), and the three suites are green, 176 tests including my probes. Two properties come along for free: a `getPayments` throw now aborts reconcile without abandoning anything, so a Tor blip cannot release a hold that a successful listing would have debited; and the 30s refresh throttle is harmless here, since "not in a list fetched within the last 30 seconds" is a fine proxy for "not listed" against 25-hour and 7-day bounds.

That ordering change is the one thing left from my side.

kaloudis:
Re-reviewed at c35ab72. The reconcile reordering matches the suggested shape exactly: abandonment now only happens in the `!payment` branch, after the node's payment lookup had its chance to answer. The two new regression tests pin the cases from my last round (stale hold with a listed settlement gets debited, stale hold with a listed in-flight payment stays held), and I re-ran my pending-status probe variants on top of this head merged with master: all green alongside the three NWC suites, and CI is green.

With that, everything from all rounds is addressed:

- late-settle debit, CLN failed-shape handling, replay guard response, failure_reason locale keys, reconcile try/catch (rounds 1-3)
- deliberate `settledEvidence` debit comment, cashu prune note, locale checklist (round 4)
- stuck-hold TTL with listing-first ordering (rounds 4-5)

utACK c35ab72

Only non-blocking leftover is the `FAILURE_REASON_*` map duplication with `utils/ErrorUtils.ts`, fine as is or as a follow-up.

kaloudis (approved):

kaloudis:
Follow-up for the map duplication filed as #4490.

PR #4443, opened 2026-08-19
fix(nwc): prevent deletion race during in-flight pay_invoice

ajaysehwal:
# Description

Fixes a low-severity NWC robustness issue: deleting (or regenerating) a connection while `pay_invoice` is in flight could leave settled spend on an orphaned connection object, so the budget debit and activity record were never persisted. A concurrent delete could also cause `markConnectionUsed` to throw instead of returning a NIP-47 error, leaving the client hanging until timeout.
- **`NostrWalletConnectStore`**: track in-flight NIP-47 handlers per connection; on `deleteConnection`, reject new requests, unsubscribe, await in-flight work, then remove the record and save
- **`markConnectionUsed`**: return `false` when the connection is gone; `withGlobalHandler` responds with `UNAUTHORIZED` instead of throwing
- **`NWCConnectionDetails`**: regenerate snapshots `buildConnectionParams` after `deleteConnection` so spend/activity settled during teardown is copied onto the recreated connection
## Test plan
- [x] `yarn test stores/NostrWalletConnectStore.test.ts ` (37 passed)
- [x] Manual: start NWC `pay_invoice` with slow backend, delete connection mid-flight — client receives preimage; budget/activity reflect the spend after delete or regenerate
- [x] Manual: delete connection just before a new NWC request — client receives `UNAUTHORIZED`, not a silent hang

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -1182,6 +1182,37 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         jest.clearAllTimers();
         jest.useRealTimers();
     });
+
+    it('defers relay release when delete follows a subscribed connection even without in-flight handlers', async () => {
+        jest.useFakeTimers();
+        const store = buildStore();
+        const connection = seedConnection(store, {
+            expiresAt: new Date(Date.now() + 60_000)
+        });
+        (store as any).nwcWalletServices.set(OLD_RELAY, { close: jest.fn() });
+        (store as any).activeSubscriptions.set(connection.id, jest.fn());
+
+        await (store as any).withGlobalHandler(connection.id, async () => ({
+            result: { ok: true },
+            error: undefined
+        }));
+
+        jest.spyOn(store as any, 'unsubscribeFromConnection').mockRestore();
+
+        const releaseSpy = jest.spyOn(
+            store as any,
+            'releaseUnusedRelayService'
+        );
+        await store.deleteConnection(connection.id);
+
+        expect(releaseSpy).not.toHaveBeenCalled();
+
+        jest.advanceTimersByTime(RELAY_RELEASE_GRACE_MS);
+        expect(releaseSpy).toHaveBeenCalledWith(OLD_RELAY);
+
+        jest.clearAllTimers();
+        jest.useRealTimers();
+    });
 });
 
 function buildMakeInvoiceTestStore(invoices: Invoice[] = []) {
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -63,7 +63,8 @@ import Invoice from '../models/Invoice';
 import Base64Utils from '../utils/Base64Utils';
 import NostrConnectUtils, { Nip47ErrorCode } from '../utils/NostrConnectUtils';
 import NostrWalletConnectStore, {
-    NWC_CLIENT_KEYS
+    NWC_CLIENT_KEYS,
+    RELAY_RELEASE_GRACE_MS
 } from './NostrWalletConnectStore';
 
 const hex64 = (c: string) => c.repeat(64);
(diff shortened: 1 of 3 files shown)


kaloudis:
Concept ACK 3c621ee38. The gate protocol is sound and this fixes the reported race. One adjacent race survives the fix and the key manual test in the test plan is still unchecked, so holding off on a full ACK.

**What I verified**
- Walked the interleavings on the gate protocol: the gate is tracked before the existence/pending checks, and `pendingDeleteConnectionIds` is checked both on entry and again after the `markConnectionUsed` await. Every handler that passed the checks before delete started is in the `Promise.allSettled` snapshot; every handler registered after the snapshot bails on the pending check before touching budget state. I could not find a window where settled spend lands after the splice.
- `awaitInFlightHandlers` is bounded on the pay path (`MAX_PAYMENT_QUEUE_WAIT_MS` + `PAYMENT_TIMEOUT_SECONDS`), so delete cannot hang indefinitely.
- The `liveIndex` re-lookup fixes a latent stale-index splice bug that becomes essential once delete awaits in-flight work. Nice catch.
- `markConnectionUsed` throw-to-false is safe: `withGlobalHandler` is its only caller.
- The regenerate reorder is sound: params carry `totalSpendSats` / `activity` / `lastBudgetReset` / `id` / `createdAt`, `createConnection` honors all of them, and the view's `connection` reference stays valid after the splice.
- Ran `jest stores/NostrWalletConnectStore.test.ts` at head: 37/37 pass. `tsc` clean.

**Blocking ask: relay close can still eat the in-flight response**

The gate resolves when the handler resolves, but `@getalby/sdk` publishes the NIP-47 response after that: `subscribe.onevent` encrypts, signs, then fires `Promise.any(pool.publish(...))` without awaiting. Meanwhile `deleteConnection` resumes and calls `releaseUnusedRelayService(relayUrl)`, which `pool.close()`es the relay when no other connection references it. That is exactly the single-connection case, and also the regenerate flow. The socket can close before the response event is flushed, so the test plan's "client receives preimage" is likely still broken for the last connection on a relay. Budget/activity persistence (the core fix here) is unaffected.

Suggested: when handlers were awaited, defer `releaseUnusedRelayService` by a short grace period (a few seconds), or skip it during delete and let the normal lifecycle sweep unused services. Either way, please run the unchecked manual test (delete mid-flight with a slow backend, confirm the client actually receives the preimage) and report back.

**Non-blocking**
- `updateConnection`'s relay-change path has the same in-flight race and is untouched here: unsubscribe + `releaseUnusedRelayService(oldRelayUrl)` + budget mutation (including `totalSpendSats = 0` when a budget is removed) with no pending guard or await. Fine as a follow-up issue so it does not get lost.
- Delete/regenerate can now block up to ~150s worst case behind an in-flight payment with only a spinner. A "waiting for an in-flight request to finish" state on the button would help; users will assume the app froze.
- Nit: two overlapping `deleteConnection` calls for the same id: the first one's `finally` clears the `pendingDeleteConnectionIds` flag while the second may still be mid-flight, briefly dropping the reject-new-requests guard. Benign in practice since the record is already spliced by then, so late requests hit the missing-connection UNAUTHORIZED path anyway. Not worth restructuring unless you touch this again.

The two new tests are good, particularly the second one, which genuinely exercises the barrier ordering (delete blocked while a handler is pending, late request rejected, delete completes after release).

kaloudis:
utACK

assigned you https://github.com/ZeusLN/zeus/issues/4486 and https://github.com/ZeusLN/zeus/issues/4487 as follow-up issues @ajaysehwal

kaloudis (approved):

PR #4447, opened 2026-08-21
Activity: fix crash toggling filters while a date filter is set

kaloudis:
# Description

Fixes a reproducible hard crash on the Activity filter screen: once a start or end date filter is set, toggling any filter switch (received, sent, etc.) crashes the app, and because date filters are persisted the crash repeats on every attempt until filters are reset.

Root cause: `handleToggle` deep-clones the filter blob with `JSON.parse(JSON.stringify(filters))`, which converts the `startDate`/`endDate` `Date` instances into ISO strings. The next `setFilters` call then throws twice over: `ActivityFilterUtils` calls `filter.startDate.getFullYear()` on a string, and the filter screen render calls `toLocaleDateString()` on it. With no error boundary in the app, the render throw is fatal on release builds.

Fix: revive the dates during the clone with the same reviver `ActivityStore.getFilters` already uses for the persisted blob.

Repro (before the fix): Activity > filter > set a start date > toggle any switch.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

The crash is backend-agnostic (the throw happens before any backend call); device tests pending.

### Locales

### Other:

### DIFF
--- a/views/Activity/ActivityFilter.tsx
+++ b/views/Activity/ActivityFilter.tsx
@@ -118,7 +118,17 @@ export default class ActivityFilter extends React.Component<
         const { filters, setFilters } = ActivityStore;
         const locale = SettingsStore.settings.locale;
 
-        const newFilters = JSON.parse(JSON.stringify(filters));
+        // The JSON round-trip deep-clones the filters but turns the
+        // startDate/endDate Date instances into ISO strings, which throws
+        // downstream (getFullYear in ActivityFilterUtils, toLocaleDateString
+        // in render). Revive them like ActivityStore.getFilters does.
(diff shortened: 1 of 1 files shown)


shubhamkmr04 (approved):
tACK

PR #4448, opened 2026-08-21
ios: LncMobile: harden StreamingCallback against nil results

kaloudis:
# Description

Hardens the iOS LNC `StreamingCallback` against two native crash vectors, in response to a report of the app crashing on iOS over LNC (routing node, litd 0.17.0-alpha) while interacting with the Activity screens.

1. Nil results. The Go bridge declares the `sendResult:` parameter nullable (`LndmobileNativeCallback` in `Lndmobile.objc.h`), but `StreamingCallback` wrapped it straight into a dictionary literal; a nil result raises `NSInvalidArgumentException` and takes the whole app down. `StreamingCallback` carries every streaming RPC, including `SubscribeInvoices`/`SubscribeTransactions`, which only the lightning-node-connect backend opens and which are live exactly while the Activity screens are up. Since #4331 it also carries the three pairing register callbacks. #4331 added the equivalent guard to `Callback` (`data ?: @""`) but left `StreamingCallback` unguarded, and the Go side changed wholesale this release with the v0.21.2 artifact bump. Events are fire-and-forget, so a nil result is dropped rather than forwarded as an empty string; for the pairing callbacks an empty string could be persisted as a key by the credential store.

2. Dangling delegate. `delegate` was `assign` (unretained); a Go-side stream can outlive the bridge-owned `LncModule` (e.g. across a bridge reload), leaving the callback messaging a dangling pointer. It is now `weak`, so a late event becomes a no-op instead of undefined behavior.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

iOS-only native change; needs an iOS device test over LNC (fresh pairing plus Activity screen with live invoice/transaction events), ideally against litd 0.17.0-alpha.

### Locales

### Other:

### DIFF
--- a/ios/LncMobile/StreamingCallback.h
+++ b/ios/LncMobile/StreamingCallback.h
@@ -6,7 +6,9 @@
 #import "Lndmobile.xcframework/ios-arm64/Lndmobile.framework/Headers/Lndmobile.objc.h"
 
 @interface StreamingCallback : NSObject <LndmobileNativeCallback>
-@property (assign) id delegate;
+// weak: the delegate is the bridge-owned LncModule; the Go-side stream can
+// outlive it (e.g. bridge reload), and an assign reference would dangle.
+@property (weak) id delegate;
 @property NSString *eventId;
 -(void)setEventName:(NSString *)name;
 -(void)sendResult:(NSString *)data;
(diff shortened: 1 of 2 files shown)


shubhamkmr04 (approved):
tACK

PR #4450, opened 2026-08-21
stores: no node calls until a wallet is picked

### LINKED ISSUE #4016: bug: when Display > Select wallet on startup is enabled, last used wallet still spins up in the background
Let's make sure no connection happens until a wallet is selected from the list.

TheSeydiCharyyev:
# Description

Relates to issue: **#4016**

With `Display > Select wallet on startup` enabled, the wallet list opens without activating a wallet, but the previously used wallet was still contacted in the background.

`Wallet.tsx` handles the setting correctly — it returns before `fetchData()`. The requests come from somewhere else: `BalanceStore` sets up a reaction on the whole `settings` object in its constructor and fetches balances whenever `hasCredentials()` is true. During startup `getSettings()` runs several times, and each call assigns a freshly parsed settings object, so the reaction fires again and again. `updateNodeProperties()` has already loaded the credentials of `nodes[selectedNode]` by then, so those fetches go to the last used wallet.

Measured on an emulator with a local HTTP server standing in for the node: after a restart, with the wallet list open and nothing selected, the server received **10 requests** — five `/v1/balance/blockchain` + `/v1/balance/channels` pairs, then quiet. It is a burst during startup, not polling. With this change it receives **none**, and selecting a wallet still connects normally.

### The change

`SettingsStore` now carries `walletSelectionPending` (the latch). It is raised in `updateNodeProperties()` while startup is on and the setting is enabled — the same place the node credentials are loaded — so the latch is up before anything can react to them. My first attempt set it from `Wallet.tsx` instead, next to the navigation, and that was too late: the reaction had already fired five times with credentials by the time the view got there, which cut the requests from 10 to 6 rather than to 0.

The `BalanceStore` reaction returns early while the latch is up.

The latch is dropped when a wallet is activated:
- `setConnectingStatus(true)`, which the wallet list calls when it activates the picked wallet;
- the `nodeActive` branch in `Wallets.tsx`, which skips connecting;
- `fetchData` itself, when it starts and when the first connection finishes. Every activation ends up there, and it never runs while the startup list is open.

The POS branch in `Wallet.tsx` ends startup before it awaits `fetchData`, because the settings writes inside `fetchData` would otherwise raise the latch again while startup is on.

The latch has to stay up until the switch is done: during one `updateSettings`, `this.settings` is assigned three times (twice in `getSettings`, once in `setSettings`) before `updateNodeProperties` loads the picked wallet's credentials, so every reaction in that window sees the previous wallet. `onWalletPress` ignores repeat taps until the switch is done.

### Scope

Only wallets whose credentials are a macaroon or an access key were affected, since `hasCredentials()` checks those two. An LNDHub wallet has neither, so it never triggered this — I reproduced against LNDHub first and saw no requests at all.

The node process itself never started: `startLnd()` runs inside `fetchData()`, which the existing guard already skips. What leaked were the balance requests.

### Not covered

- iOS is not tested. I have no Mac and no iPhone, so I cannot run it at all.
- `Wallets.tsx` and `Wallet.tsx` have no component tests. The store tests pin the contract the views rely on. The view order was checked on the emulator for the normal switch, for the PIN flow and for the mid-connect toggle below.
- With a PIN set and a wallet deleted in the last session, the user has to pick a wallet twice on the next start: after the first pick the list opens again. This happens on master too and is not about the latch, so it is filed separately as #4684.
- On that same path, `handleJustDeletedWallet` drops the latch with `setConnectingStatus(true)` before its write, so the balance reaction can still run against the wallet that was persisted after the delete. On master it runs there too. Also in #4684.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

14 new tests: 3 in a new `stores/BalanceStore.test.ts` for the reaction gate, 6 in `stores/SettingsStore.test.ts` for when the latch is raised and dropped (including the POS contract), and 5 in a new `stores/startupWalletSelection.test.ts` that drives a real `SettingsStore` and `BalanceStore` through a switch. In that suite, removing the latch turns tests red, and so does dropping it while the previous wallet's credentials are still loaded.

- [ ] iOS

After the rebase (0d9c3583), with two LND (REST) wallets on separate stub servers and no lock: 45 s with the startup list open, 0 requests to either wallet; after picking the other wallet, 0 requests to the previous one while the picked one connected. With a PIN set and the active wallet deleted in the last session: the list opens twice (see above), and after the second pick the wallet connects and balances are fetched.

Mid-connect toggle, the case the `fetchData` drops exist for: with the stub holding `/v1/getinfo` for 20 s, I turned Select wallet on startup on while the first connection was still running, so the write raised the latch with startup still on. After the connection finished, every settings write sent a balance pair to the stub, so the reaction was alive, and the next restart opened the wallet list with 0 requests in 50 s. With the two `fetchData` drops taken out of my working tree and the same steps repeated, two settings writes sent nothing at all: the reaction stayed off for the session.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST) — pointed at local stub servers, which is what counts the requests
- [ ] Nostr Wallet Connect
- [x] LndHub — local stub, used to confirm this path is not affected

### Locales

### Other:

### DIFF
--- a/stores/BalanceStore.test.ts
+++ b/stores/BalanceStore.test.ts
@@ -1,4 +1,4 @@
-jest.mock('../stores/Stores', () => ({}));
+jest.mock('./Stores', () => ({}));
 jest.mock('react-native-blob-util', () => ({}));
 jest.mock('../ldknode/LdkNodeInjection', () => ({}));
 jest.mock('./SettingsStore', () => ({}));
--- a/stores/BalanceStore.test.ts
+++ b/stores/BalanceStore.test.ts
@@ -1,0 +1,86 @@
+jest.mock('../stores/Stores', () => ({}));
+jest.mock('react-native-blob-util', () => ({}));
+jest.mock('../ldknode/LdkNodeInjection', () => ({}));
+jest.mock('./SettingsStore', () => ({}));
+jest.mock('../utils/BackendUtils', () => ({
+    __esModule: true,
+    default: {
+        getBlockchainBalance: jest.fn(),
+        getLightningBalance: jest.fn(),
+        supportsOnchainBalance: jest.fn(() => true)
+    }
+}));
+
+import { observable, runInAction } from 'mobx';
+
+import BalanceStore from './BalanceStore';
+import BackendUtils from '../utils/BackendUtils';
+
+const blockchainBalance = {
+    total_balance: '0',
+    confirmed_balance: '0',
+    unconfirmed_balance: '0'
+};
+const lightningBalance = { balance: '0', pending_open_balance: '0' };
+
+const newSettingsStore = (walletSelectionPending: boolean) =>
+    observable({
+        settings: { nodes: [{}], selectedNode: 0 } as any,
+        walletSelectionPending,
+        hasCredentials: () => true
+    });
+
+// Writing settings again is what startup does: getSettings() runs several
+// times and assigns a freshly parsed object each time.
+const rewriteSettings = (settingsStore: any) =>
+    runInAction(() => {
+        settingsStore.settings = { nodes: [{}], selectedNode: 0 };
+    });
+
+describe('BalanceStore settings reaction', () => {
+    beforeEach(() => {
+        jest.clearAllMocks();
+        (BackendUtils.getBlockchainBalance as jest.Mock).mockResolvedValue(
+            blockchainBalance
+        );
+        (BackendUtils.getLightningBalance as jest.Mock).mockResolvedValue(
+            lightningBalance
+        );
+    });
+
+    it('does not reach a node while the user is still choosing a wallet on startup', () => {
+        const settingsStore = newSettingsStore(true);
+        new BalanceStore(settingsStore as any);
+
+        rewriteSettings(settingsStore);
+
+        expect(BackendUtils.getBlockchainBalance).not.toHaveBeenCalled();
+        expect(BackendUtils.getLightningBalance).not.toHaveBeenCalled();
+    });
+
+    it('fetches balances when no wallet selection is pending', () => {
+        const settingsStore = newSettingsStore(false);
+        new BalanceStore(settingsStore as any);
+
+        rewriteSettings(settingsStore);
+
+        expect(BackendUtils.getBlockchainBalance).toHaveBeenCalled();
+        expect(BackendUtils.getLightningBalance).toHaveBeenCalled();
+    });
+
+    it('fetches again once the wallet has been picked', () => {
+        const settingsStore = newSettingsStore(true);
+        new BalanceStore(settingsStore as any);
+
+        rewriteSettings(settingsStore);
+        expect(BackendUtils.getLightningBalance).not.toHaveBeenCalled();
+
+        runInAction(() => {
+            settingsStore.walletSelectionPending = false;
+        });
+        rewriteSettings(settingsStore);
+
+        expect(BackendUtils.getBlockchainBalance).toHaveBeenCalled();
+        expect(BackendUtils.getLightningBalance).toHaveBeenCalled();
+    });
+});
--- a/stores/BalanceStore.ts
+++ b/stores/BalanceStore.ts
@@ -29,6 +29,10 @@ export default class BalanceStore {
         reaction(
             () => this.settingsStore.settings,
             () => {
+                // While the user is still choosing a wallet on startup no
+                // wallet is active yet, but the credentials loaded from
+                // settings still point at the previously used one.
+                if (this.settingsStore.walletSelectionPending) return;
                 if (this.settingsStore.hasCredentials()) {
                     this.getBlockchainBalance(false, false);
                     this.getLightningBalance(false);
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -22,7 +22,10 @@ jest.mock('react-native-encrypted-storage', () => ({
     removeItem: jest.fn().mockResolvedValue(undefined)
 }));
 
-jest.mock('../utils/BackendUtils', () => ({}));
+jest.mock('../utils/BackendUtils', () => ({
+    __esModule: true,
+    default: { clearCachedCalls: jest.fn() }
+}));
 jest.mock('../utils/BiometricUtils', () => ({
     getSupportedBiometryType: jest.fn().mockResolvedValue(undefined)
 }));
@@ -33,7 +36,9 @@ jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
     migrateRgsDefaultToZeus: jest.fn().mockResolvedValue(undefined),
+    migrateSwapHostsToBoltz: jest.fn().mockResolvedValue(undefined),
     migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
+    migrateOlympusHostsToZeusLsp: jest.fn().mockResolvedValue(undefined),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
     storageMigrationV2: jest.fn().mockResolvedValue(undefined)
 }));
@@ -91,6 +96,65 @@ beforeEach(() => {
 
 afterEach(() => {
     logSpy.mockRestore();
+});
+
+describe('SettingsStore startup wallet selection', () => {
+    const nodeSettings = (selectNodeOnStartup: boolean) => ({
+        selectNodeOnStartup,
+        selectedNode: 0,
+        nodes: [{ implementation: 'lnd', macaroonHex: 'abcd' }]
+    });
+
+    it('holds off node access while a wallet is still to be picked', async () => {
+        seedSettings(nodeSettings(true));
+        const store = new SettingsStore();
+
+        await store.getSettings();
+
+        // credentials of the previously used wallet are loaded, so the
+        // latch is what keeps anything from reaching it
+        expect(store.hasCredentials()).toBe(true);
+        expect(store.walletSelectionPending).toBe(true);
+    });
+
+    it('holds nothing off when the setting is disabled', async () => {
+        seedSettings(nodeSettings(false));
+        const store = new SettingsStore();
+
+        await store.getSettings();
+
+        expect(store.walletSelectionPending).toBe(false);
+    });
+
+    it('does not raise the latch once startup has handed over', async () => {
+        seedSettings(nodeSettings(true));
+        const store = new SettingsStore();
+        store.setInitialStart(false);
+
+        await store.getSettings();
+
+        expect(store.walletSelectionPending).toBe(false);
+    });
+});
+
+describe('SettingsStore.setConnectingStatus', () => {
+    it('ends a pending startup wallet selection when a wallet is activated', () => {
+        const store = new SettingsStore();
+        store.walletSelectionPending = true;
+
+        store.setConnectingStatus(true);
+
+        expect(store.walletSelectionPending).toBe(false);
+    });
+
+    it('leaves the pending flag alone when connecting is turned off', () => {
+        const store = new SettingsStore();
+        store.walletSelectionPending = true;
+
+        store.setConnectingStatus(false);
+
+        expect(store.walletSelectionPending).toBe(true);
+    });
 });
 
 describe('SettingsStore.updateSettings', () => {
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -135,6 +135,28 @@ describe('SettingsStore startup wallet selection', () => {
 
         expect(store.walletSelectionPending).toBe(false);
     });
+
+    it('raises the latch again on the next settings write unless startup has ended', async () => {
+        seedSettings(nodeSettings(true));
+        const store = new SettingsStore();
+        await store.getSettings();
+        expect(store.walletSelectionPending).toBe(true);
+
+        // Dropping the latch on its own does not survive the next write
+        // while startup is still on: every write reloads the node
+        // properties, and that raises it again
+        store.setWalletSelectionPending(false);
+        await store.updateSettings({ lurkerMode: true });
+        expect(store.walletSelectionPending).toBe(true);
+
+        // Ending startup first is what keeps it down. The POS branch in
+        // Wallet.tsx relies on this, because that path never reaches
+        // setConnectingStatus(true)
+        store.setInitialStart(false);
+        store.setWalletSelectionPending(false);
+        await store.updateSettings({ lurkerMode: false });
+        expect(store.walletSelectionPending).toBe(false);
+    });
 });
 
 describe('SettingsStore.setConnectingStatus', () => {
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -37,6 +37,7 @@ jest.mock('../utils/MigrationUtils', () => ({
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
     migrateRgsDefaultsToV2: jest.fn().mockResolvedValue(undefined),
     migrateSwapHostsToBoltz: jest.fn().mockResolvedValue(undefined),
+    migrateRetiredSwapHosts: jest.fn().mockResolvedValue(undefined),
     migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
     migrateOlympusHostsToZeusLsp: jest.fn().mockResolvedValue(undefined),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -35,7 +35,7 @@ jest.mock('../utils/LocaleUtils', () => ({
 jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
-    migrateRgsDefaultToZeus: jest.fn().mockResolvedValue(undefined),
+    migrateRgsDefaultsToV2: jest.fn().mockResolvedValue(undefined),
     migrateSwapHostsToBoltz: jest.fn().mockResolvedValue(undefined),
     migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
     migrateOlympusHostsToZeusLsp: jest.fn().mockResolvedValue(undefined),
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1677,6 +1677,11 @@ export default class SettingsStore {
     @observable public lndDir?: string;
     @observable public isSqlite?: boolean;
     @observable public initialStart: boolean = true;
+    // Set while startup hands the user the wallet list because of
+    // Display > Select wallet on startup, and cleared once a wallet is
+    // picked. Node credentials still point at the previously used
+    // wallet during that window, so nothing may reach a node.
+    @observable public walletSelectionPending: boolean = false;
     @observable public embeddedLndStarted: boolean = false;
     @observable public walletJustCreated: boolean = false;
     @observable public lndFolderMissing: boolean = false;
@@ -1727,6 +1732,10 @@ export default class SettingsStore {
 
     public setInitialStart = (status: boolean) => {
         this.initialStart = status;
+    };
+
+    public setWalletSelectionPending = (status: boolean) => {
+        this.walletSelectionPending = status;
     };
 
     public setLndFolderMissing = (status: boolean) => {
@@ -1862,6 +1871,12 @@ export default class SettingsStore {
 
     @action
     private updateNodeProperties = (settings: Settings) => {
+        // The credentials loaded below belong to the previously used wallet.
+        // If the user asked to pick a wallet on startup, nothing may reach a
+        // node with them until a wallet is actually activated.
+        if (this.initialStart && settings?.selectNodeOnStartup) {
+            this.walletSelectionPending = true;
+        }
         const node: any =
             settings?.nodes?.length &&
             settings?.nodes[settings.selectedNode || 0];
@@ -2305,6 +2320,8 @@ export default class SettingsStore {
             this.error = false;
             this.errorMsg = '';
             this.lndFolderMissing = false;
+            // A wallet is being activated, so startup selection is over
+            this.walletSelectionPending = false;
             BackendUtils.clearCachedCalls();
             // remove fetchLock on reconnect
             this.fetchLock = false;
--- a/stores/startupWalletSelection.test.ts
+++ b/stores/startupWalletSelection.test.ts
@@ -1,0 +1,197 @@
+// Regression cover for #4016. With Display > Select wallet on startup
+// enabled, nothing may reach a node until the user picks one, and the
+// window that is easiest to lose is the switch itself: updateSettings
+// assigns this.settings inside setSettings, which is not an action, before
+// updateNodeProperties loads the newly selected wallet's credentials. Any
+// reaction on settings therefore runs once while the credentials still
+// belong to the previously used wallet.
+
+jest.mock('react-native-biometrics', () => ({ BiometryType: {} }));
+jest.mock('react-native-blob-util', () => ({
+    fs: {
+        dirs: { LibraryDir: '/lib', DocumentDir: '/docs' }
+    }
+}));
+jest.mock('react-native-encrypted-storage', () => ({
+    getItem: jest.fn().mockResolvedValue(null),
+    setItem: jest.fn().mockResolvedValue(undefined),
+    removeItem: jest.fn().mockResolvedValue(undefined)
+}));
+
+jest.mock('../utils/BackendUtils', () => ({
+    __esModule: true,
+    default: {
+        clearCachedCalls: jest.fn(),
+        getBlockchainBalance: jest.fn(),
+        getLightningBalance: jest.fn(),
+        supportsOnchainBalance: jest.fn(() => true)
+    }
+}));
+jest.mock('../utils/BiometricUtils', () => ({
+    getSupportedBiometryType: jest.fn().mockResolvedValue(undefined)
+}));
+jest.mock('../utils/LocaleUtils', () => ({
+    localeString: (s: string) => s
+}));
+jest.mock('../utils/MigrationUtils', () => ({
+    keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
+    migrateRgsDefaultToZeus: jest.fn().mockResolvedValue(undefined),
+    migrateSwapHostsToBoltz: jest.fn().mockResolvedValue(undefined),
+    migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
+    migrateOlympusHostsToZeusLsp: jest.fn().mockResolvedValue(undefined),
+    legacySettingsMigrations: jest.fn().mockResolvedValue({}),
+    storageMigrationV2: jest.fn().mockResolvedValue(undefined)
+}));
+jest.mock('../utils/TorUtils', () => ({
+    doTorRequest: jest.fn(),
+    RequestMethod: {}
+}));
+jest.mock('../utils/LdkNodeUtils', () => ({
+    DEFAULT_SCORER_URL: '',
+    DEFAULT_VSS_SERVER: '',
+    getDefaultEsploraServer: jest.fn().mockReturnValue(''),
+    getDefaultRgsServer: jest.fn().mockReturnValue('')
+}));
+jest.mock('../storage', () => {
+    const backing: Record<string, string> = {};
+    return {
+        _backing: backing,
+        getItem: jest.fn(async (key: string) => backing[key] ?? false),
+        setItem: jest.fn(async (key: string, value: any) => {
+            backing[key] =
+                typeof value === 'string' ? value : JSON.stringify(value);
+            return true;
+        }),
+        removeItem: jest.fn(async (key: string) => {
+            delete backing[key];
+            return true;
+        })
+    };
+});
+
+import { reaction } from 'mobx';
+
+import SettingsStore, { STORAGE_KEY } from './SettingsStore';
+import BalanceStore from './BalanceStore';
+import BackendUtils from '../utils/BackendUtils';
+
+const StorageMock: any = jest.requireMock('../storage');
+
+const previousWallet = { implementation: 'lnd', macaroonHex: 'aaaa' };
+const pickedWallet = { implementation: 'lnd', macaroonHex: 'bbbb' };
+const nodes = [previousWallet, pickedWallet];
+
+const seedStartupSettings = () => {
+    StorageMock._backing[STORAGE_KEY] = JSON.stringify({
+        selectNodeOnStartup: true,
+        selectedNode: 0,
+        nodes
+    });
+};
+
+// What Wallets.tsx commits when the user picks the second wallet
+const commitSwitch = (store: SettingsStore) =>
+    store.updateSettings({ nodes, selectedNode: 1 } as any);
+
+let logSpy: jest.SpyInstance;
+
+beforeEach(() => {
+    for (const key of Object.keys(StorageMock._backing)) {
+        delete StorageMock._backing[key];
+    }
+    jest.clearAllMocks();
+    (BackendUtils.getBlockchainBalance as jest.Mock).mockResolvedValue({
+        total_balance: '0',
+        confirmed_balance: '0',
+        unconfirmed_balance: '0'
+    });
+    (BackendUtils.getLightningBalance as jest.Mock).mockResolvedValue({
+        balance: '0',
+        pending_open_balance: '0'
+    });
+    logSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
+});
+
+afterEach(() => {
+    logSpy.mockRestore();
+});
+
+describe('startup wallet selection', () => {
+    it('raises the latch while loading the previously used credentials', async () => {
+        seedStartupSettings();
+        const settingsStore = new SettingsStore();
+
+        await settingsStore.getSettings();
+
+        expect(settingsStore.macaroonHex).toEqual('aaaa');
+        expect(settingsStore.walletSelectionPending).toBe(true);
+    });
+
+    it('exposes the previous wallet to settings reactions while a switch commits', async () => {
+        seedStartupSettings();
+        const settingsStore = new SettingsStore();
+        await settingsStore.getSettings();
+
+        const credentialsSeen: (string | undefined)[] = [];
+        const dispose = reaction(
+            () => settingsStore.settings,
+            () => credentialsSeen.push(settingsStore.macaroonHex)
+        );
+
+        await commitSwitch(settingsStore);
+        dispose();
+
+        // Every write during the switch lands before the new wallet's
+        // credentials do, which is why the latch has to stay up until the
+        // switch is confirmed. On an emulator this showed up as three
+        // balance pairs going to the previously used wallet.
+        expect(credentialsSeen.length).toBeGreaterThan(0);
+        expect(credentialsSeen.every((macaroon) => macaroon === 'aaaa')).toBe(
+            true
+        );
+        expect(settingsStore.macaroonHex).toEqual('bbbb');
+    });
+
+    it('does not fetch balances from the previous wallet while a switch commits', async () => {
+        seedStartupSettings();
+        const settingsStore = new SettingsStore();
+        new BalanceStore(settingsStore);
+        await settingsStore.getSettings();
+
+        await commitSwitch(settingsStore);
+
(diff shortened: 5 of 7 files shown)


kaloudis (commented):
Concept ACK. The root-cause analysis here is exactly right, and the PR body is a model writeup: measured evidence, scope analysis, and tests verified to fail without the fix.

I traced the full startup flow against the diff and confirmed the design holds: `BalanceStore` is the only store whose reaction tracks `settingsStore.settings` and reaches the node (the reactions in ChannelsStore, InvoicesStore, LSPStore, SwapStore, AlertStore, and NostrWalletConnectStore all track other observables), so gating that one reaction covers the leak, consistent with your 10-to-0 measurement. Raising the latch inside `updateNodeProperties()` is also the right call: it goes up inside the same `@action` that loads the credentials, before any `settings` assignment that could see them, so there is no window for the reaction to slip through. On the very first `getSettings()` the reaction fires at the initial `this.settings = parsedSettings` assignment before credentials exist, so `hasCredentials()` is false and nothing fetches; on every later call the latch is already up. No race.

Two findings before this is ready:

### 1. The unconditional clear in `Wallets.tsx` reintroduces a leak to the old wallet at selection time (should fix)

`onWalletPress` clears the latch before the `if (nodeActive)` branch, i.e. on every press. In the startup flow `nodeActive` is always false, since it is computed as `!this.state.fromStartup && ...`, so the else branch always runs: `await updateSettings({ nodes, selectedNode }).then(() => setConnectingStatus(true) ...)`.

Inside `updateSettings`, `setSettings` assigns `this.settings` before `updateNodeProperties(newSettings)` runs. Neither is wrapped in an action at that point, so the `BalanceStore` reaction fires synchronously at the assignment, with the latch already down (cleared at the top of `onWalletPress`) and `macaroonHex` still holding the previously used wallet's credentials. Result: when the user picks a different wallet from the startup list, one balance pair still goes out to the old wallet, which is exactly the traffic this PR sets out to eliminate. A stub server backing the selected wallet would not distinguish this from legitimate connection traffic, so the measurement would not have caught it.

Fix: move `setWalletSelectionPending(false)` inside the `if (nodeActive)` branch, matching what its comment says it is for. The else branch needs no explicit clear: the latch correctly gates the reaction during `updateSettings` (old credentials), then `setConnectingStatus(true)` clears it after `updateNodeProperties` has loaded the new wallet's credentials, and the real fetch happens via `fetchData` on the Wallet view. (The `nodeActive` clear is unreachable while `fromStartup` is true, so it is purely defensive, which is fine.)

### 2. POS mode can leave the latch stuck for the whole session (minor)

In `Wallet.tsx`, the branch chain checks POS before the `selectNodeOnStartup` guard: with POS enabled and `posStatus === 'unselected'`, the app calls `fetchData()` directly and never shows the wallet list. `getSettings()` has already raised the latch by then, `connecting` is initialized to `true`, and nothing on this path ever calls `setConnectingStatus(true)`. The latch therefore stays up for the entire session for users with both POS and Select wallet on startup enabled, silently disabling the balance-refresh-on-settings-change reaction. Low impact, but current behavior is that POS simply ignores the startup-selection setting, so the latch should not outlive startup there. Suggested fix: call `setWalletSelectionPending(false)` in the POS `unselected` branch before `fetchData()`, or more generally clear it in any branch that proceeds to `fetchData` without handing over to the wallet list.

### What holds up well

- Clearing coverage is complete for the non-POS paths: wallet selection, wallet creation/import (`WalletCreationUtils.ts`), seed recovery, the deleted-wallet restart path, and the retry button all go through `setConnectingStatus(true)`. The Lockscreen path correctly keeps the latch up until after login.
- The scope claims check out: `hasCredentials()` is `macaroonHex || accessKey`, which explains the LNDHub non-repro.
- The `MigrationUtils` mock fix is a genuine catch: `getSettings()` really does call `migrateSwapHostsToBoltz` and `migrateOlympusHostsToZeusLsp`, so the old mock made every existing `getSettings` test silently short-circuit into the catch block. That improves the pre-existing tests too.
- First-ever `BalanceStore` test file, and the fake-observable approach is sensible.

Nit: `stores/BalanceStore.test.ts` mocks `'../stores/Stores'`; `'./Stores'` would match the sibling-import style used elsewhere in the `stores/` tests.

Happy to re-review once finding 1 is addressed.

TheSeydiCharyyev:
Both findings were real. Fixed in `aec5794e`, and thank you for tracing the flow that carefully — finding 1 in particular was invisible to my measurement setup.

**1. The clear in `Wallets.tsx`.** Verified every step of your reading before changing anything: `nodeActive` is `!this.state.fromStartup && (...)`, so it is always false in the startup flow; `setSettings` is a plain `async` method, not an `@action`, and assigns `this.settings` before `updateNodeProperties(newSettings)` runs in `updateSettings`. So the reaction fires with the latch down and the old credentials still loaded.

You were right that my stub setup could not see this: both wallets pointed at the same server. I rebuilt the bench with the two wallets on separate stubs — the previously used wallet on one port, the wallet picked from the startup list on another — so any traffic to the old one is unambiguous. Picking a different wallet from the startup list:

```
old wallet, before the fix:  6 requests  (3 x /v1/balance/blockchain + /v1/balance/channels)
old wallet, after the fix:   0 requests
selected wallet:             connects normally in both cases
```

Three pairs rather than the one you predicted, since the reaction fires on more than one settings write during the switch. The clear now sits inside the `nodeActive` branch, as you suggested, and the comment says why that branch is the one that needs it.

**2. The POS path.** Confirmed: the POS `unselected` branch is checked before the `selectNodeOnStartup` guard, `connecting` starts as `true`, and the only `setConnectingStatus(true)` in `Wallet.tsx` is the retry button in `render`, not anything `fetchData` calls. So the latch would have outlived startup for the whole session. That branch now clears it before `fetchData`. I kept the fix to that one branch rather than generalising: every other branch either hands over to the wallet list, goes to the Lockscreen, or runs with the latch never raised in the first place.

**Nit** taken, the mock is now `./Stores`.

No new tests for either fix: both are placement changes in views, and the repo has no component tests, so the coverage here is the before/after measurement above plus the existing store tests. If you would rather have the wallet-switch case pinned down in code, the shape that would work is an integration test over `SettingsStore` and `BalanceStore` together, asserting no fetch while `updateSettings` commits a switch. Happy to add it if you want it in this PR.

kaloudis:
@TheSeydiCharyyev yes, let's add some tests for this. Would be nice to cover us from regressions.

TheSeydiCharyyev:
Added in `7e63b985`: `stores/startupWalletSelection.test.ts`, wiring a real `BalanceStore` to a real `SettingsStore` and driving the switch the way `Wallets.tsx` does.

Five cases: the latch goes up while the previously used credentials load, no balances are fetched while the switch commits, fetching resumes once `setConnectingStatus(true)` confirms it, and none of it engages when the setting is off. I checked that removing the latch turns the first and third red.

The fifth one documents the hazard rather than the fix. It watches `settings` reactions during the switch and asserts every one of them still sees the previous wallet's macaroon, since `setSettings` assigns before `updateNodeProperties` loads the new one. Writing it turned up something I had not measured precisely: that reaction fires **three** times per switch, not once — which lines up exactly with the three balance pairs the stub server recorded before the fix. So your one-pair estimate was conservative, and the unit test and the bench agree on the number.

I deliberately did not assert the count of three. A later change that legitimately reduces the number of settings writes during a switch should not turn this red; what matters is that every write during that window still carries the old credentials.

The view-level placement itself is still not covered by code — the repo has no component tests — so `Wallets.tsx` and the POS branch remain covered by the measurement rather than a test. If you want those pinned too, the honest way is component tests, which is a bigger conversation about test infrastructure than this PR should start.

kaloudis:
Round 2 review. Both round-1 findings are fixed correctly, and I verified them against the surrounding flows rather than just the diff. CI is green.

**Verified fixes from round 1:**

1. Latch clear ordering in `Wallets.tsx`: `nodeActive` is computed as `!fromStartup && ...`, so the startup flow always takes the else branch, where the clear happens via `setConnectingStatus(true)` inside `.then()` after `updateSettings` resolves. The settings assignment (which fires the reaction while the old wallet's macaroon is still loaded) happens with the latch still up. Correct.
2. The POS `unselected` branch now clears the latch before `fetchData`, so the initial raise no longer outlives startup on that path.

I also re-confirmed that `BalanceStore` is still the only store whose reaction tracks `settingsStore.settings` and reaches the node, and that `fetchData` calls the balance methods directly, so a wedged latch only disables the reaction-driven refresh. That keeps the blast radius of any remaining hole small.

**Should fix: the latch can be re-raised on the POS path and never cleared**

The raise condition is `initialStart && selectNodeOnStartup`, and it runs on every `getSettings()` / `updateSettings()` via `updateNodeProperties`. On the POS path, `initialStart` is only cleared at the end of `fetchData` (`Wallet.tsx:1331`), well after your clear. Several `updateSettings` calls sit inside `fetchData` before that point: the embedded-lnd `initialLoad` write (~line 869), `resetExpressGraphSyncOnStartup` (~line 888), `rescan` (~line 1093), and `recovery` (~lines 1105 and 1236). `applySettingsUpdate` calls both `getSettings()` and `updateNodeProperties`, so any of these re-raises the latch after the POS branch cleared it. POS never calls `setConnectingStatus(true)`, and `setConnectingStatus(false)` at the end of `fetchData` does not touch the latch, so the balance reaction is silently dead for the whole session, recurring on every launch.

Trigger combo: POS enabled + Select wallet on startup + an embedded LND or LDK wallet (or rescan/recovery flags set).

Suggested one-line fix: in the POS `unselected` branch, call `SettingsStore.setInitialStart(false)` alongside the latch clear, mirroring what the `selectNodeOnStartup` branch does at line 490. The only readers of `initialStart` are the two startup branches in `Wallet.tsx` and `Wallets.tsx`, and POS taking over is the startup handover, so ending it there is semantically right and kills the re-raise at its root.

**Nits (optional):**

- Double-tap race on the startup list: after the first tap sets `fromStartup: false` and `selectedNode`, a second tap on the same row hits the `nodeActive` branch and drops the latch before the first tap's settings assignment lands, which is precisely the "clearing it any earlier" hazard your code comment describes. Guarding `onWalletPress` on `this.state.isSelecting` would close it and also prevent duplicate `updateSettings` commits.
- Pre-existing, not introduced here: `handleJustDeletedWallet` calls `setConnectingStatus(true)` (clearing the latch) before its `updateSettings` assigns settings, so if `justDeletedWallet` persisted across a restart with the setting enabled, one balance pair could leak to the stale `nodes[selectedNode]`. Fine to leave for a follow-up, just noting it against this PR's own invariant.

**Quality notes:**

- The root-cause writeup and stub-server measurements (10 to 6 to 0 across attempts) are exemplary, and the commit messages document why the naive placements fail.
- The `MigrationUtils` mock fix is quietly important: without it, every prior test's `getSettings()` was silently throwing into its own catch, so the new tests would have been vacuous.
- `startupWalletSelection.test.ts` wiring real stores through the switch, including the test that documents the settings-assign-before-credentials hazard itself, is exactly the store contract the view code depends on.

Approach ACK. One should-fix above before merge. Given the finding, the POS + select-on-startup + embedded combo is the one device test I'd want to see; iOS also remains unchecked in the matrix.

TheSeydiCharyyev:
Round 2 in three commits on top, after a rebase on master.

06d25a1d tests: keep MigrationUtils mocks in step with master. Not from the review: getSettings on master now calls purgeRescueKeyFiles and migrateRgsDefaultsToV2, and on the rebased branch both suites failed (4 tests, "is not a function" thrown into the catch in getSettings). The green CI was from before those merges landed.

<sha2> views: end startup when POS takes over. The should-fix, as suggested: SettingsStore.setInitialStart(false) in the POS unselected branch, next to the latch clear. I checked every reader of initialStart: the raise in updateNodeProperties, the two startup branches in Wallet.tsx and Wallets.tsx, and WalletConfiguration, which only writes false. None of them is reached from the POS branch, so ending startup there changes nothing else. One side effect: a repeat getSettingsAndNavigate during fetchData can no longer take the selectNodeOnStartup branch once POS is active. The new test in SettingsStore.test.ts documents the contract the branch relies on: clearing the latch alone does not survive the next settings write while startup is on, ending startup first keeps it down. Its first assertion reproduces the re-raise with the real updateSettings.

<sha3> views: ignore repeat taps while a wallet switch commits. Nit 1. I did not guard on isSelecting: componentDidMount sets it to true when the list opens from startup (line 94), so that guard would block the first tap. The guard is an instance flag, set before updateSettings in the switch branch, cleared in a finally, and checked at the top of onWalletPress.

Nit 2 (handleJustDeletedWallet): left as is, agreed it is a follow-up.

Device test: not run for the POS + select-on-startup + embedded combo. My stand has no embedded wallet, and the settings writes that re-raise the latch live only in the embedded-lnd and ldk-node branches of fetchData, so the LndHub stub cannot reproduce it. The store test above is the proof I have. iOS is still unchecked.

kaloudis:
@TheSeydiCharyyev please rebase

kaloudis:
Round 3. The POS should-fix from round 2 is correct, and the repeat-tap guard closes the specific hazard I described. Two things before merge: a rebase that is not purely mechanical, and one wedge that is the same shape as the POS one.

### Verified fixed

`SettingsStore.setInitialStart(false)` in the POS `unselected` branch (`Wallet.tsx:476`) kills the re-raise at the root. I re-checked every `initialStart` reader on current master: the raise in `updateNodeProperties`, `Wallets.tsx:278`, `Wallet.tsx:495`, the end-of-`fetchData` clear at `Wallet.tsx:1363`, and `WalletConfiguration.tsx:1140` which only writes `false`. None is reachable from the POS branch. `posStatus` only returns to `unselected` from the POS settings dropdown, which re-enters the same branch and clears again, so there is no intra-session regression either.

I also re-confirmed against current master that `BalanceStore` is still the only store whose reaction tracks `settingsStore.settings`, and that `hasCredentials()` has no other caller. The gate still covers everything it needs to.

### Must fix: the rebase is load-bearing

Branch is CONFLICTING. The conflict itself is one hunk in `stores/SettingsStore.test.ts`: master consolidated the individual migrations into `MigrationsUtils.runSettingsMigrations(settings, fromQueue)` and added `keychainDesyncMigration` ahead of it. Take master's side there, and `60d54c44` becomes obsolete.

The part that is not mechanical: `stores/startupWalletSelection.test.ts` is a new file, so git merges it clean, but it carries the same stale mock. On the merged tree **3 of its 5 tests fail**, because `getSettings()` throws `keychainDesyncMigration is not a function` into its own catch and never reaches `updateNodeProperties`. I confirmed that replacing the five `migrate*` entries with

```ts
keychainDesyncMigration: jest.fn().mockResolvedValue(undefined),
runSettingsMigrations: jest
    .fn()
    .mockImplementation(async (settings: any) => settings),
```

turns all five green, with the rest of the suite green on the merged tree and `tsc` clean.

Please also copy the `afterEach` guard you added to `SettingsStore.test.ts` (the one that fails on a `'Could not load settings'` console.error) into `startupWalletSelection.test.ts`. This mock has gone stale three times now, and without the guard 2 of those 5 tests pass vacuously instead of failing.

### Should fix: `handleJustDeletedWallet` wedges the latch for the session

I flagged this in round 2 as pre-existing and agreed it was a follow-up. It is not pre-existing any more: the re-raise turns it into a wedge.

`Wallets.tsx:141` calls `setConnectingStatus(true)` first, which drops the latch, and the `updateSettings` that installs the new wallet only lands at line 151 after a 2 second sleep. The `justDeletedWallet` check at line 274 sits *above* `if (initialStart) setInitialStart(false)` at 278, so `initialStart` is still true when that write runs and `updateNodeProperties` raises the latch again. Nothing clears it afterwards: this path never reaches `setConnectingStatus(true)` a second time, and the only other call in `Wallet.tsx` is the retry button in `render`. The `BalanceStore` reaction is then dead for the rest of the session.

Confirmed against the real store rather than by reading, seeding `justDeletedWallet: true` + `selectNodeOnStartup: true`:

```ts
await store.getSettings();
expect(store.walletSelectionPending).toBe(true);

store.setConnectingStatus(true);                 // Wallets.tsx:141
expect(store.walletSelectionPending).toBe(false);
await store.updateSettings({ justDeletedWallet: false, selectedNode: 1 });

expect(store.walletSelectionPending).toBe(true); // re-raised, and nothing clears it
```

Reachable combo: delete the active wallet, kill the app from the wallet list (`justDeletedWallet` is persisted and is only cleared in this handler), relaunch with `Select wallet on startup` on, tap a wallet.

Suggested fix, which removes the class rather than patching the third instance of it. Make the raise an assignment rather than a one-way set in `updateNodeProperties` (`SettingsStore.ts:1877`):

```ts
this.walletSelectionPending = !!(
    this.initialStart && settings?.selectNodeOnStartup
);
```

Every settings write then recomputes the latch at the one moment where dropping it is safe, immediately after the new wallet's credentials have been loaded. A wedge becomes structurally impossible: once startup has ended the next write puts the latch down no matter which path got there, so a path that forgets to clear degrades to "cleared slightly late" instead of "never". The explicit clears in the `nodeActive` branch and the POS branch stay, since neither is guaranteed to be followed by a settings write, and keeping `setConnectingStatus(true)`'s clear as a safety net is fine.

Then hoist `if (initialStart) setInitialStart(false)` above the `justDeletedWallet` branch in `onWalletPress` so that write cannot re-raise. The single balance pair that still leaks on that path (latch already down at line 141 when the write's reaction fires) is the genuinely pre-existing part, and I am happy to leave that for a follow-up issue.

If you would rather keep this minimal, hoisting `setInitialStart(false)` alone fixes the wedge. I prefer the assignment form because this is now the third path where "who clears the latch" had to be traced by hand.

### Nits

1. **The repeat-tap guard is armed after an await.** `this.walletSwitchInFlight = true` is at `Wallets.tsx:337`, but the else branch awaits `setPersistentMode(false)` at lines 312 and 314 on Android when the current implementation is `ldk-node` or `embedded-lnd`. Two taps inside that window both pass the check at line 273 and both commit an `updateSettings` plus a `navigateAfterWalletSelection`. The hazard I raised in round 2 is closed, since `nodeActive` only becomes true after the `setState` at 331 and there is no await between there and the arm, so this is the leftover. Moving the arm to the top of the else branch (before the `setState` at 296) and widening the `try` closes it at no cost. The guard also does not cover `handleJustDeletedWallet`, which has the 2 second sleep inside it.

2. **`commitSwitch` fidelity.** `commitSwitch` in `startupWalletSelection.test.ts` does not replicate `Wallets.tsx:278-280`, which clears `initialStart` before the switch. The assertions hold either way today, but in the test the latch is up at the end because `updateNodeProperties` re-raised it, while in production it is up because nothing cleared it. Adding `store.setInitialStart(false)` to `commitSwitch` would make the test pass for the same reason production does.

### Testing

The POS re-raise is now pinned by `SettingsStore.test.ts`, and I checked that every `updateSettings` inside `fetchData` really is confined to the `embedded-lnd` and `ldk-node` branches, so your stub bench genuinely cannot reach it. I will take the store test as the proof for that one. Still owed: an Android run of the ordinary flow after the rebase, and iOS is still unchecked in the matrix.

Approach ACK. The rebase and the `handleJustDeletedWallet` wedge before merge. The analysis quality on this PR continues to be excellent.

TheSeydiCharyyev:
@kaloudis Rebased on master and pushed as 0d9c3583. I rebuilt the branch, so all hashes are new. Point by point:

**Rebase and mocks**
- `SettingsStore.test.ts`: took master's side, so 60d54c44 is gone.
- `startupWalletSelection.test.ts`: the `MigrationUtils` mock now matches the one in `SettingsStore.test.ts`, and the storage mock has `KEY_PREFIX` and `getRawItem`. With the old mock I got the same result as you: 3 of 5 red, and 2 green without a finished load.
- Copied the `afterEach` guard for `'Could not load settings'`. With a stale mock all 5 tests now fail.

**handleJustDeletedWallet**
I did not change this path. On the emulator the latch is not stuck for the session:
- With a PIN set, Lockscreen opens the list while startup is still on, and the write in `handleJustDeletedWallet` raises the latch again, as you described.
- But `popTo('Wallet')` then goes back to the Wallet screen under the list. That screen has not finished startup (`initialLoad` is still true), so `handleFocus` calls `getSettingsAndNavigate`, which sees `initialStart` and opens the wallet list a second time.
- The second tap takes the normal branch, and `setConnectingStatus(true)` drops the latch.

Checked on the emulator: delete the active wallet, kill the app from the list, start it again with a PIN and Select wallet on startup on, and tap a wallet. The list opened again, and after the second tap the wallet connected and its balances loaded. Without a lock the latch is not raised again at all, because `Wallet.tsx` ends startup before it opens the list.

So what the user sees on this path is "pick the wallet twice", and that happens on master too. Hoisting `setInitialStart(false)` removes the second list, but by the store code it also keeps the latch down for the whole write. Then all three settings assignments in that write run the balance reaction against the wallet that was persisted after the delete (two without the hoist), and if the write fails, the startup list stays open with the latch down. I think the double pick should be its own issue.

**The assignment in updateNodeProperties**
I tried it and did not keep it. `applySettingsUpdate` first calls `getSettings()`, and that ends with `updateNodeProperties()` for the old persisted settings. With the assignment the latch drops there, while the previous wallet's credentials are still loaded, and the next assignment of `this.settings` in the same write runs the reaction against the previous wallet. With `commitSwitch` ending startup first (your nit 2), "does not fetch balances from the previous wallet while a switch commits" fails: one balance fetch with the previous wallet's macaroon. Without nit 2 the test stays green, which is why it did not show up before. Without a lock it would drop even earlier: `Wallet.tsx` ends startup before it opens the list, so the `getSettings()` from `Wallets.refreshSettings` on mount would already drop the latch.

**Who clears the latch**
I agree that tracing this by hand is the weak spot, and there were more paths than the three we had. The plain activation at startup never calls `setConnectingStatus(true)`; it relies on `connecting` starting out true. For example, turning Select wallet on startup on while the first connection is still running raises the latch, because startup is still on, and nothing dropped it. 0d9c3583 now drops the latch in `fetchData`: when it starts, and next to `setInitialStart(false)` when the first connection finishes. Every activation ends up in `fetchData`, and it never runs while the startup list is open: without a lock the list replaces `Wallet`, and behind Lockscreen `Wallet` is not focused. The explicit clears stay. This is view code, so there is no test for it, and I did not run that toggle case on a device.

**Nits**
1. The guard is now armed before the first await. `onWalletPress` is a small wrapper that checks and arms it, and the old body moved unchanged into `selectWallet`. So a tap during `setPersistentMode` or during the pause in `handleJustDeletedWallet` is ignored too. The header Back button also calls `handleJustDeletedWallet`, but it is hidden while the list is open from startup.
2. `commitSwitch` now calls `setInitialStart(false)` first.

I also fixed some commit messages and the test file header. During one write `this.settings` is assigned three times (twice in `getSettings`, once in `setSettings`), not once, and not every activation path calls `setConnectingStatus(true)`.

**Testing**
- On the pushed tree: tsc, lint, prettier, and jest with 84 suites and 1802 tests. CI is green.
- Android, Pixel 9 Pro emulator on Android 16, two LND (REST) wallets on local stub servers, Select wallet on startup on, no lock. With the startup list open for 45 s there were 0 requests to either wallet. After I picked the other wallet there were 0 requests to the previous one, and the picked one connected.
- The PIN flow above.
- iOS is still not tested.

Should I open a separate issue for the double pick after a deleted wallet?

kaloudis:
Round 4. Everything from round 3 is addressed, and two of my three asks were addressed by showing me I was wrong rather than by changing the code. Both rebuttals check out against master. Code is good; what is left is test evidence.

### Conceded: the assignment form in `updateNodeProperties` is worse

You are right and I retract the suggestion. `applySettingsUpdate` calls `getSettings(false, true)` first, and that call ends at `SettingsStore.ts:2058` with `updateNodeProperties(this.settings)` against the *old* persisted blob. Under the assignment form the latch drops there, with the previous wallet's credentials loaded, and then `setSettings` at 2132 assigns `this.settings` and the reaction fires before `updateNodeProperties(newSettings)` at 2142 loads the new credentials. That is one balance pair straight to the previous wallet, which is the bug this PR exists to close. My proposal traded a wedge for a leak. Keeping the raise one-way is correct.

Good catch that nit 2 was what exposed it. That the test only went red after `commitSwitch` started clearing `initialStart` is a fair argument for the nit having been worth making.

### Conceded: `handleJustDeletedWallet` is not a session wedge

Traced it and your account is right. With a lock, `popTo('Wallet')` lands on a `Wallet` instance whose `_replacedForWalletSelection` is still false, so `handleFocus` reaches `getSettingsAndNavigate`, `initialStart` is still true at `Wallet.tsx:489`, and the list opens a second time. The second tap has `justDeletedWallet` already cleared, so it takes the normal branch and `setConnectingStatus(true)` drops the latch. Without a lock the latch is never re-raised in the first place, because `Wallet.tsx:490` ends startup before the list opens. Either way the reaction recovers, so "dead for the rest of the session" was wrong.

I also agree with not hoisting `setInitialStart(false)`. It removes the second list but widens the leak from two assignments to three, and on a failed write it leaves the startup list open with the latch down. That is a bad trade.

Yes, please open the issue. Put both halves in it: the double pick, and the two balance pairs that still reach the wallet persisted after the delete because `Wallets.tsx:140` drops the latch two seconds before the write lands. Link it here.

### Verified: the `fetchData` drops

The reasoning for making `fetchData` the universal drop point holds up. I checked the three ways an unfocused or replaced `Wallet` could reach it while the startup list is open, and all three are closed already: `getSettingsAndNavigate` returns at `Wallet.tsx:405` on `_replacedForWalletSelection`, `handleAppStateChange` gates on `navigation.isFocused()` at `Wallet.tsx:372`, and `handleFocus` cannot fire on a screen that is not focused. So the unconditional drop at the top of `fetchDataCore` cannot fire early.

I also re-confirmed the gate is still complete on current master: `BalanceStore` is the only store whose reaction tracks `settingsStore.settings`. `SwapStore` tracks `getHost`, which is settings-derived, but its own `fetchedRatesHost` guard already keeps it off the wire during startup.

The `nodeActive` explicit drop is still needed, since that branch writes no settings and so cannot rely on `triggerSettingsRefresh` to pull `handleFocus` into `fetchData`. That is correctly reasoned in the comment.

### Nits, neither blocking

1. `SettingsStore.setWalletSelectionPending(false)` in the POS branch at `Wallet.tsx:568` is now redundant: the `await this.fetchData()` three lines down drops it anyway. `setInitialStart(false)` there is still load-bearing. Removing the redundant line would cut one drop site from five to four, but keeping it as belt-and-braces is defensible given how much hand-tracing this latch has needed. Your call.

2. The end-of-connection drop at `Wallet.tsx:1368` sits inside `if (connecting && start != null)`. A settings write that raises the latch during a `fetchData` where `connecting` was false would leave it up until the next `fetchData`. I could not construct a reachable path, since `setInitialStart(false)` lives in that same block and every activation route sets `connecting` first, so this is theoretical.

### Testing: the two things still owed

The store side is well pinned. `raises the latch again on the next settings write unless startup has ended` is the contract the whole view-side design rests on, and it is now covered. The `afterEach` guard against a silently swallowed load is the right backstop for a mock that has gone stale three times.

What is not covered is the change that is new in this round:

- **The toggle case.** Turning `Select wallet on startup` on while the first connection is still running is the scenario the `fetchData` drops exist for, and it is the one you flagged as not run on a device. Please run it: fresh start with the setting off, toggle it on mid-connect, confirm balances still land and the next restart opens the list with no requests.
- **iOS.** Still unchecked in the matrix. This is startup and navigation code and the lock path differs, so an iOS pass is required before merge, not optional.

Once those two are in the description this is a tACK from me. The store tests plus your stub-server counts cover the rest.

Approach ACK, and thanks for pushing back on both of the round-3 asks with evidence rather than just applying them. That is what kept a real regression out.

TheSeydiCharyyev:
On iOS: I cannot run it. I have no Mac and no iPhone, so I can only build and test on Android. That is why the iOS box has been empty on all my PRs.

Everything else from round 4 I will do: the issue for the double pick and for the two balance pairs that reach the wallet persisted after the delete, and the toggle run on Android (start with the setting off, turn it on mid-connect, check that balances still land, then restart and check the list opens with no requests).

Can someone else take the iOS pass? I can write out the steps and what each screen should show.

TheSeydiCharyyev:
Pushed c1ae113f, CI is green.

**The issue is up: #4684.** It has both halves: the double pick with a lock, and the balance requests that reach the wallet the deletion selected, because `Wallets.tsx:140` drops the latch two seconds before the write lands. It also says that on master all three assignments in that write reach the reaction, while with this PR two of them do.

**Nit 1: removed.** `SettingsStore.setWalletSelectionPending(false)` is gone from the POS branch, `setInitialStart(false)` stays. Nothing between it and the `await this.fetchData()` three lines below writes settings, so the latch cannot come back up in that gap, and `fetchData` drops it anyway. Drop sites go from five to four.

**Nit 2:** I could not build that path either. Ending startup sits in the same block, and every activation route sets `connecting` first, so a raise during a `fetchData` with `connecting` false would need a settings write from something that is not an activation. If that ever shows up, the next `fetchData` drops it.

**The toggle case, on the emulator.** The stub held `/v1/getinfo` for 20 s, so the app stayed in its first connection while I turned Select wallet on startup on, and that write raised the latch because startup was still on. After the connection finished, every settings write sent a balance pair to the stub, and the restart after that opened the wallet list with 0 requests in 50 s.

Then I took the two `fetchData` drops out of my working tree and repeated the same steps: two settings writes in a row sent nothing at all, so the reaction was off for the session. That is the before and after for this round's change.

The description now carries both runs and the link to #4684.

**iOS** is the one thing I cannot do, as in my comment above: there is no Mac or iPhone here. Everything else from round 4 is done.

kaloudis:
Round 5. Everything from round 4 that was yours to do is done. I re-verified the removal in `c1ae113f` rather than taking my own nit on faith, and re-ran the suite against current master. Code is merge-ready; the only thing still open is the iOS pass, which is on us, not on you.

### Verified: dropping the POS clear is safe

The clear is now implicit, so the question is whether the POS branch always reaches the drop. `fetchData` (`Wallet.tsx:552`) takes the single-flight lock first and returns at `:557` when `acquireFetchLock` hands back `null`, which is *before* `fetchDataCore` and its drop at `:597`. So the branch is only safe if that early return is unreachable there. Checked:

- `fetchData` has exactly two call sites in the tree, `Wallet.tsx:485` and `:521`, both inside `getSettingsAndNavigate`. Nothing else in `views/`, `stores/` or `utils/` calls it, so the lock can only be held by another `getSettingsAndNavigate`.
- Both entry points to `getSettingsAndNavigate` set `_navigating` synchronously before the call and clear it in `.finally` (`:295`, `:384`), so one instance cannot re-enter itself while `fetchData` is awaited.
- A second live `Wallet` instance is not reachable on this path: the startup-selection instance returns at `:507` without ever calling `fetchData`, and `replace`/`popTo` keep one `Wallet` on the stack.
- The one unguarded `getSettingsAndNavigate()` call, the iOS Restart button at `:1621`, runs `setConnectingStatus(true)` first, which clears `fetchLock` at `SettingsStore.ts:2413` and drops the latch itself.

Also confirmed the ordering claim in the commit message: `setPosStatus` (`SettingsStore.ts:2435`) is a plain observable assignment with no settings write, and the three statements between `setInitialStart(false)` at `:477` and `await this.fetchData()` at `:485` are `startListeners` and two `setState` calls. Nothing can re-raise the latch in that gap.

One optional follow-up, not for this PR: the comment at `:594` says "reaching fetchData means a wallet is being activated", but the drop sits in `fetchDataCore`, one lock-guarded call deeper. If anyone later calls `fetchData` from a second place (pull-to-refresh, a retry button) while one is already running, the `seq === null` return skips the drop. Moving the line above `acquireFetchLock` in `fetchData` is behaviourally identical today and puts the code where the comment says it is.

### Verified: the rest of round 4

- **Mid-connect toggle.** The run in the body is the case I asked for, and the negative control is what makes it convincing: with both `fetchData` drops removed from your tree, two settings writes sent nothing and the reaction stayed off for the session. That pins the drops as load-bearing rather than defensive.
- **#4684.** Both halves are there and the reading matches mine: the double pick from `initialStart` still being true when `popTo('Wallet')` lands, and the balance requests reaching the wallet the deletion selected because `Wallets.tsx:141` drops the latch two seconds before the write. Correctly filed as pre-existing.
- **Nit 2 stays open by design.** The end-of-connection drop is inside `if (connecting && start != null)`, so a first connection that returns early (the LDK VSS hard-fail at `:761`, the embedded-lnd folder-missing alert) leaves `initialStart` true and a later raise undropped. That is still theoretical and self-healing: the same `initialStart` that allows the raise also sends the next focus down the `:495` branch, which opens the list and drops the latch on the pick. Not worth more code.

### Re-verified against current master (`bbb779b8f`, 39 commits ahead of your base)

- `BalanceStore` is still the only reaction tracking `settingsStore.settings`. Re-walked all eleven store reactions: `AlertStore` tracks `networkInfo`, `ChannelsStore` tracks its own channel lists, `InvoicesStore` tracks `pay_req`, `LSPStore` tracks `channelsStore.channels`, `NostrWalletConnectStore` tracks `connecting` and `transactionsStore.loading`, and `SwapStore` tracks `getHost` but stays off the wire at startup via `fetchedRatesHost`.
- No rebase needed. Master has not touched `Wallet.tsx`, `Wallets.tsx` or `BalanceStore.ts` since your base, and the only `SettingsStore.ts` change is three added `default*` fields in `getLspConfigForNetwork`. No new calls inside `getSettings`, so the mock drift that bit the last rebase cannot recur here.
- Merged master into your head locally and ran `yarn jest stores/`: 11 suites, 156 tests, all green.
- Mutation check on the suite: commenting out the `walletSelectionPending` guard in the `BalanceStore` reaction turns 3 of the 8 new tests red. The `afterEach` guard on `'Could not load settings'` is doing its job too, so these cannot silently pass on an unfinished load.

### tACK `c1ae113f`

Conditional only on an iOS pass, which is not yours to do. I read the diff for platform exposure: the only branch is the Android-only `setPersistentMode` dismissal in `selectWallet`, which iOS skips, and the repeat-tap guard is armed synchronously so the shorter iOS path is strictly easier on it. Everything else is JS and store state. I will run the iOS pass myself unless someone beats me to it: startup list with two remote wallets and no requests to either, the switch, and the PIN flow.

kaloudis:
iOS pass done. **All six scenarios pass, and the negative control reproduces the Android result.** This closes the last condition on my round-5 tACK.

### Setup

iPhone 17 Pro simulator, iOS 26.5, Xcode 27.0. Debug build of your head `c1ae113f` merged with current master (`bbb779b8f`), so master's 39 commits are in the picture too. Two LND (REST) wallets (`walletA`, `walletB`) pointed at local HTTP stubs on `127.0.0.1:8090` and `:8091` that log every request with a timestamp, same idea as your Android setup. Wallet picks were delivered by invoking the row's real `onPress` through the live React tree, and every count below is from the stub logs.

### Results

**1. Baseline, setting off.** Cold start connects `walletA`: 13 requests to A, 0 to B. Plumbing confirmed.

**2. Setting on, list open.** Cold start opens the wallet list. **0 requests to either stub in 45 s.** Store state at that moment is exactly the hazard the PR describes: `walletSelectionPending: true`, `macaroonHex` still the previously used wallet's, `selectedNode: 0`.

**3. Pick the other wallet.** 12 requests to B, **0 to A**, latch drops, credentials switch to B. The previous wallet is never contacted during the switch.

**4. PIN path.** With a PIN set the start goes Lockscreen first. Latch is up there and `initialStart` is still **true**, because `Lockscreen.tsx:342-354` does its own `replace('Wallets', {fromStartup: true})` and `Wallet.tsx:496` never runs. 0 requests through the lock screen and through the unlock. Picking `walletA` from that list: 11 requests to A, **0 to B**. This is also the path that makes #4684's double pick possible, and the reading there matches what I see.

**5. Mid-connect toggle, the case the `fetchData` drops exist for.** Stub holding `/v1/getinfo` for 20 s. Turned the setting on 7 s into the first connection, with `initialStart` still true, so the write raised the latch. After the connection finished the latch was **false**, and an inert settings write produced a fresh balance pair on the stub, so the reaction was alive. Next restart opened the list with **0 requests in 45 s**.

One detail from that run worth recording: the toggle itself emits one balance pair, because `getSettings` assigns `this.settings` before `updateNodeProperties` raises the latch. Those requests go to the wallet that is currently active and connected, so they are correct traffic, not a leak. It is the same ordering that makes the assignment form in `updateNodeProperties` wrong, from round 4.

**6. Negative control.** Commented out both `fetchData` drops (`Wallet.tsx:597` and `:1371`) and repeated scenario 5. The latch stayed **true** after the connection finished, and **two settings writes produced 0 requests**: the reaction was off for the session. Same result you measured on Android. Restored the file and re-confirmed scenario 5 passes.

**7. Repeat-tap guard.** Two `onPress` calls 10 ms apart both entered `onWalletPress`; the previous wallet received **0** requests and exactly one switch committed. At 60 ms and 150 ms the second tap found no row at all, because without the Android `setPersistentMode` await iOS commits the switch and navigates away inside one frame. So on iOS the guard's window is sub-frame, which matches what the comment says about the Android path being the wider one.

### tACK `c1ae113f`

Android by you, iOS by me, negative control on both. Nothing further owed from your side. The `fetchData` vs `fetchDataCore` placement note from round 5 stays a follow-up, not a blocker.

kaloudis (approved):
tACK

really nice approach on this one. well done

PR #4455, opened 2026-08-21
Currency modal: dismiss dialog window before navigating to Select Currency

### LINKED ISSUE #4451: Other currencies not selectable in Currency Converter view from keypad
### Describe the bug

When using currency converter from the keypad view, curriencies are not added to the list after pressing the plus sign (nothing happens). Currency converter works as expected when opened from the settings page. This is true for both favorites and non favorites currencies. Tested on 13.2.0-rc1.

[21_08-11-01-41_539.zip](https://github.com/user-attachments/files/31299613/21_08-11-01-41_539.zip)

### Reproduce

1. Open the app
2. Go to keypad view from main screen
3. Select fiat currency
4. Go to Currency Converter tab
5. Press "+" sign
6. Select a fiat currency
7. Nothing happens

### ZEUS version

v13.2.0

### Node interface

LDK Node

### Network

Clearnet

### Device

Samsung S25 Ultra

### Device operating system

Android 16

### Log output

```shell
Logs attached (file too large)
```

kaloudis:
# Description

Fixes #4451: on Android, tapping any currency (favorites included) on the Select Currency screen reached from the keypad currency modal's Converter tab did nothing, stranding the user on the list. The Settings > Currency Converter entry point was unaffected.

Root cause, established from the reporter's logcat and screen recording: the converter's add button starts the ModalBox slide-down animation and calls `navigation.navigate('SelectCurrency')` in the same tick. The ModalBox is an RN `Modal` (`coverScreen`), i.e. a separate Android dialog window, and that window is only dismissed once the close animation's completion callback flips `isAnimateClose` so `visible` goes false. Android stops driving the animation once the pushed screen occludes the dialog window, so the invisible dialog stayed alive (3.5 seconds in the reporter's trace) and received every tap meant for Select Currency. The logcat shows the taps landing on the dialog's window (`ViewPostIme` on the modal's `VRI[MainActivity]@b8d78ba`) while Select Currency was on screen, with no JS `onPress` ever firing; gesture handling was left wedged afterwards ("Can't cancel already finished gesture").

Fix:
- New `ModalBox.closeImmediate(onDone?)`: commits the closed state without animating (so the dialog window is torn down deterministically) and invokes `onDone` after the commit.
- The converter's navigate-away path now dismisses the modal via `closeImmediate` and navigates from its callback, so Select Currency is only pushed once the dialog is gone. The animated close is still used everywhere the modal closes without a navigation (Currencies tab select, backdrop press, swipe).
- The reopen-on-focus flow (`pendingReopen`) is unchanged and already worked in the reporter's recording.

Repro (before the fix, Android): keypad > tap the fiat chip > Currency Converter tab > "+" > tap any currency. The tap does nothing. After the fix the currency is added and the modal reopens on the Converter tab.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

## Locales

- [ ] I've added new translatable text, wrapped with the `localeString` function
- [x] I haven't added any new translatable text
- [ ] I acknowledge that all new translations are handled on Transfix and not in-repo

## Third Party Dependencies and Packages

- [ ] Contributors will need to run `yarn` after this change
- [x] No new dependencies introduced
- [ ] New dependencies introduced and vetted
- [ ] Verified that the package.json and yarn.lock files are in sync

## Other

- [ ] I've documented new flows or processes in the repo's README or docs folder
- [x] No documentation changes needed

### DIFF
--- a/components/ModalBox.tsx
+++ b/components/ModalBox.tsx
@@ -760,4 +760,49 @@ export default class ModalBox extends React.PureComponent<
             }
         }
     }
+
+    /*
+     * Close without animating. The animated close only dismisses the
+     * underlying RN Modal once the slide-down finishes, and on Android that
+     * animation stops advancing when the dialog window is occluded (e.g. by
+     * a navigation push), leaving an invisible dialog that swallows every
+     * tap on the screen below. Callers that navigate right after closing
+     * must use this instead of close(); onDone fires once the closed state
+     * has been committed, so navigation can safely start.
+     */
+    closeImmediate(onDone?: () => void) {
+        if (this.props.isDisabled) return;
+        if (
+            !this.state.isOpen &&
+            !this.state.isAnimateOpen &&
+            !this.state.isAnimateClose
+        ) {
+            if (onDone) onDone();
+            return;
+        }
+        this.stopAnimateOpen();
+        this.stopAnimateClose();
+        if (this.backHandlerSubscription) {
+            this.backHandlerSubscription.remove();
+            this.backHandlerSubscription = null;
+        }
+        this.setState(
+            {
+                isOpen: false,
+                isAnimateOpen: false,
+                isAnimateClose: false,
+                isAnimateBackdrop: false
+            },
+            () => {
+                this.state.position.setValue(
+                    this.props.entry === 'top'
+                        ? -this.state.containerHeight
+                        : this.state.containerHeight
+                );
+                this.state.backdropOpacity.setValue(0);
+                if (this.props.onClosed) this.props.onClosed();
+                if (onDone) onDone();
+            }
+        );
+    }
 }
--- a/components/ModalBox.tsx
+++ b/components/ModalBox.tsx
@@ -768,10 +768,11 @@ export default class ModalBox extends React.PureComponent<
      * a navigation push), leaving an invisible dialog that swallows every
      * tap on the screen below. Callers that navigate right after closing
      * must use this instead of close(); onDone fires once the closed state
-     * has been committed, so navigation can safely start.
+     * has been committed, so navigation can safely start. Ignores
+     * isDisabled: that prop guards user-initiated open/close, and a caller
+     * navigating away needs the dialog torn down unconditionally.
      */
     closeImmediate(onDone?: () => void) {
-        if (this.props.isDisabled) return;
         if (
             !this.state.isOpen &&
             !this.state.isAnimateOpen &&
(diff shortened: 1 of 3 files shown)


components/ModalBox.tsx
```diff
@@ -760,4 +760,49 @@ export default class ModalBox extends React.PureComponent<
...
+     * Close without animating. The animated close only dismisses the
+     * underlying RN Modal once the slide-down finishes, and on Android that
+     * animation stops advancing when the dialog window is occluded (e.g. by
+     * a navigation push), leaving an invisible dialog that swallows every
+     * tap on the screen below. Callers that navigate right after closing
+     * must use this instead of close(); onDone fires once the closed state
+     * has been committed, so navigation can safely start.
+     */
+    closeImmediate(onDone?: () => void) {
+        if (this.props.isDisabled) return;
```
shubhamkmr04: nit: this return skips `onDone`, so the caller's `navigate()` is dropped and `pendingReopen` stays true — the same dead-tap symptom this PR fixes. Not reachable today (`CurrencySelectorModal` never sets `isDisabled`), but worth firing `onDone` on every exit path.

components/ModalBox.tsx
```diff
@@ -760,4 +760,49 @@ export default class ModalBox extends React.PureComponent<
...
+        if (
+            !this.state.isOpen &&
+            !this.state.isAnimateOpen &&
+            !this.state.isAnimateClose
+        ) {
+            if (onDone) onDone();
+            return;
+        }
+        this.stopAnimateOpen();
+        this.stopAnimateClose();
```
shubhamkmr04, line 785: nit: `animBackdrop` isn't stopped here, so an in-flight fade-in overwrites the `backdropOpacity.setValue(0)` below and leaves it at 1 — the next open then has no backdrop fade. Reachable on any device by hitting "+" during the open animation. `close()` avoids it because `animateBackdropClose()` stops `animBackdrop` first.

shubhamkmr04 (commented):
The fix is working as I have done the device testing.
Two nits on the new method below.

components/ModalBox.tsx
kaloudis: Good catch, thanks for testing. Addressed in ba98f6549, though slightly differently than suggested: firing onDone on the disabled path would navigate with the dialog still up, which recreates the zombie-window bug this PR fixes. Instead I dropped the isDisabled guard from closeImmediate entirely. That prop guards user-initiated open/close; a programmatic force-close ahead of navigation should always run, which both tears down the dialog and fires onDone.

components/ModalBox.tsx
kaloudis: Confirmed, your mechanism is exactly right: stopAnimateOpen/Close only stop the position animation, so an in-flight fade-in keeps driving backdropOpacity past the setValue(0) and parks it at 1, and the pendingReopen reopen then shows the backdrop with no fade. Fixed in ba98f6549: closeImmediate now stops animBackdrop before committing the closed state, mirroring animateBackdropClose.

shubhamkmr04 (approved):
tACK

PR #4459, opened 2026-08-22
Show status for unconfirmed on-chain transactions

### LINKED ISSUE #4457: Transaction detail view shows no status for unconfirmed transactions
### Feedback

From the same emailed user feedback as #4456 (2026-08-21): after making an on-chain send, the user opened the Transaction view before the transaction was mined and found no "Number of confirmations" field or any other status. The field only appeared after the first confirmation, so while the transaction was pending the page gave them no indication of its state.

### Current behavior

The Activity list does label pending transactions: `views/Activity/Activity.tsx:243-245` renders "On-chain: Unconfirmed" when `num_confirmations == 0`. But tapping through to the detail view loses that information:

- `views/Transaction.tsx:328` gates the confirmations row on `!!num_confirmations`, so `0` (falsy) hides the row exactly when the user most needs it. The `isNull` guard on the same line is redundant, and the `color={num_confirmations > 0 ? 'green' : 'red'}` red branch is dead code since the guard means it can only render for positive counts
- The Status row (`views/Transaction.tsx:337`) is gated on `!!status`, and only LDK Node populates a `status` field; lnd-family and CLN transactions have none, so on those backends a pending transaction renders neither row
- `models/Transaction.ts` `getStatusDisplay` falls through to `''` when there is no `status` field, so simply un-gating the Status row isn't enough on its own
- `block_hash` and block height are also empty pre-confirmation, so the page collapses to hash + timestamp + addresses, visually indistinguishable from a confirmed transaction apart from the small pending clock glyph next to the amount (which opens an info modal when tapped, but carries no label)

### Proposed fix

Small, view + model only, no new locale strings (`general.unconfirmed`, `general.confirmed`, and `views.Transaction.numConf` all exist):

1. `views/Transaction.tsx`: change the confirmations guard to `num_confirmations != null` so "Number of Confirmations: 0" renders (in red, making the existing color branch live), and render the Status row unconditionally
2. `models/Transaction.ts` `getStatusDisplay`: return `general.unconfirmed` when not confirmed and no backend `status` string is present, so lnd-family and CLN show "Unconfirmed" instead of a blank value

### Adjacent bugs worth fixing in the same pass

- `views/UTXOs/UTXO.tsx:156` has the identical `!!getConfs &&` hide bug, with the same dead red branch (`isUnconfirmed ? 'red' : 'green'`)
- CLN confirmation math (`backends/CoreLightningRequestHandler.ts:288,299`) computes `getinfo.blockheight - block_height`, which is off by one (a transaction mined in the tip block shows 0 confirmations and would render as unconfirmed), and if the recorded block height is 0/absent the subtraction yields roughly the full chain height as the confirmation count. Both become user-visible once the field stops being hidden

### Prior art

#1699 looks similar but is a different problem: incoming unconfirmed transactions not appearing at all on embedded LND, closed as a Neutrino constraint (Neutrino can't see arbitrary mempool transactions). That constraint doesn't apply here: this report concerns the user's own send, which the node knows about immediately; the app has the data and hides the row.

kaloudis:
# Description

Relates to issue: **#4457**

Fixes the Transaction detail view giving no status for unconfirmed on-chain transactions. The confirmations row was gated on `!!num_confirmations`, hiding it exactly when the count is 0, and the Status row only rendered on LDK Node (the one backend that populates a `status` field), so on the lnd family and CLN a pending transaction showed neither. The Activity list labels the same transaction "Unconfirmed"; tapping through lost that information.

Changes:

- `views/Transaction.tsx`: render the confirmations row whenever the field is present ("Number of Confirmations: 0" now shows in red, making the previously dead color branch reachable) and render the Status row unconditionally, colored green/red by confirmation state
- `models/Transaction.ts`: `getStatusDisplay` falls back to "Unconfirmed" instead of `''` when no backend status string exists; unknown backend statuses still pass through
- `views/UTXOs/UTXO.tsx` + `models/Utxo.ts`: same hide bug fixed on the UTXO detail view; `getConfs` now returns 0 instead of NaN when the confirmations field is absent, which also repairs `isUnconfirmed` for that case
- `backends/CoreLightningRequestHandler.ts`: CLN confirmation math was off by one (a transaction mined in the tip block showed 0 confirmations and would have rendered as unconfirmed) and produced roughly the full chain height when the recorded block height was 0/absent; the tip block now counts as one confirmation and a missing height maps to 0

No new locale strings; `general.confirmed`, `general.unconfirmed`, and `views.Transaction.numConf` already exist.

Screenshots to follow with device testing.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Added `models/Transaction.test.ts` covering `isConfirmed` and `getStatusDisplay` across lnd-style (no status field), LDK Node (pending/confirmed), and unknown-status shapes, including the empty-string regression this PR fixes.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -10,6 +10,7 @@ import {
     getBalance,
     getChainTransactions,
     getOffchainBalance,
+    getUTXOs as getChainUTXOs,
     listPeers,
     listClosedChannels,
     listPeerChannels
@@ -617,7 +618,7 @@ export default class CLNRest {
             feebase: data.base_fee_msat,
             feeppm: data.fee_rate
         });
-    getUTXOs = () => this.postRequest('/v1/listfunds');
+    getUTXOs = async () => await getChainUTXOs();
     signMessage = (message: string) =>
         this.postRequest('/v1/signmessage', {
             message
--- a/backends/CoreLightningRequestHandler.test.ts
+++ b/backends/CoreLightningRequestHandler.test.ts
@@ -1,0 +1,107 @@
+const mockPostRequest = jest.fn();
+
+// the handler constructs its CLNRest instance at import time, so postRequest
+// has to resolve the mock lazily
+jest.mock('./CLNRest', () => ({
+    __esModule: true,
+    default: class {
+        postRequest(...args: Array<any>) {
+            return mockPostRequest(...args);
+        }
+    }
+}));
+// only used by getChainTransactions, but importing it pulls in the whole
+// store graph
+jest.mock('../utils/AddressUtils', () => ({ __esModule: true, default: {} }));
+
+import { getUTXOs } from './CoreLightningRequestHandler';
+
+const listfunds = (outputs: Array<any>) => ({ outputs, channels: [] });
+
+const respond = (funds: any, info?: any) =>
+    mockPostRequest.mockImplementation((route: string) => {
+        if (route === '/v1/listfunds') return Promise.resolve(funds);
+        if (route === '/v1/getinfo')
+            return info
+                ? Promise.resolve(info)
+                : Promise.reject(new Error('getinfo failed'));
+        return Promise.reject(new Error(`unexpected route ${route}`));
+    });
+
+describe('CoreLightningRequestHandler.getUTXOs', () => {
+    beforeEach(() => mockPostRequest.mockReset());
+
+    it('derives a confirmation count from the blockheight and the tip', async () => {
+        // listfunds outputs carry no confirmations field; before this was
+        // normalized every CLN UTXO reported 0 confirmations
+        respond(listfunds([{ txid: 'a', blockheight: 108 }]), {
+            blockheight: 200
+        });
+
+        const { outputs } = await getUTXOs();
+
+        expect(outputs[0].confirmations).toBe(93);
+    });
+
+    it('counts the tip block as one confirmation', async () => {
+        respond(listfunds([{ txid: 'a', blockheight: 200 }]), {
+            blockheight: 200
+        });
+
+        const { outputs } = await getUTXOs();
+
+        expect(outputs[0].confirmations).toBe(1);
+    });
+
+    it('reports 0 for an output with no blockheight', async () => {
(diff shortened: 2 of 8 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (commented):
Image

UTXO view on CLNRest node always shows 0 confirmations

kaloudis:
Good catch, and thanks for testing this on a real node — confirmed and fixed in 3f5c1cb.

You surfaced a pre-existing gap that this PR made visible. CLN's `listfunds` outputs carry a `blockheight` but no `confirmations` field, and `models/Utxo.ts` only ever read `confirmations` — so on CLNRest it was `Number(undefined)` → `NaN`. On master the `!!getConfs` gate hid the row entirely, so the gap never showed; once this PR dropped that gate, the `|| 0` fallback turned it into a confident, wrong "0".

`backends/CLNRest.ts` now routes `getUTXOs` through `CoreLightningRequestHandler`, which derives the count from the tip the same way `LdkNode` and `getChainTransactions` already do, with the tip block counting as one confirmation:

- confirmed → real count, green
- still in the mempool (no `blockheight`) → `0`, red — the case this PR is actually trying to surface
- `getinfo` failed → count omitted and the row hidden, rather than painting every UTXO red on a transient failure

The UTXO in your screenshot at block height 108 should now read `tip - 107` in green.

Added `backends/CoreLightningRequestHandler.test.ts` covering all three paths. I mutation-checked it rather than trusting green: restoring the old `listfunds` passthrough fails 3 of the 6 cases, dropping the `+1` fails 2, dropping the tip guard fails 1.

Mind giving it another run on your regtest node?

shubhamkmr04 (approved):
tACK

PR #4460, opened 2026-08-23
fix(swaps): re-derive the preimage when rescuing reverse swaps

kaloudis:
# Description

Relates to issue: **ZEUS-0000** (internal security audit finding M7)

`SwapStore.getRescuableSwaps` rebuilds a rescued swap's claim keys from the restored rescue key, but a rescued **reverse** swap still could not be claimed. It reached `ReverseClaimTransaction.build` missing two things it needs, and neither of them can come back from the host.

**The preimage.** A reverse swap's preimage is chosen by ZEUS and never leaves the device: the host is given only `sha256(preimage)`, which it uses as the hold invoice's payment hash. So for a swap restored on another device the preimage existed nowhere — not in the restore response, not in storage. `ReverseClaimTransaction.build` then fell back to `preimageHex: ''` (`preimageHexFrom` returns `''` for a missing preimage, and preimage was not among the required fields), and `createReverseClaimTransaction` broadcast that claim on all ten retries before giving up, logging each failure to the console only.

**The destination address.** Same shape of problem, caught in review by @shubhamkmr04. `createReverseSwap` posts only `invoiceAmount`/`to`/`from`/`claimPublicKey`/`preimageHash` and attaches the user's address client-side afterwards, so the host is never told where we want to be paid and `/swap/restore` has nothing to return. `build` already required `destinationAddress`, so a rescued claim returned `null` regardless of the preimage.

Net effect: **a reverse swap could not be completed from a restored rescue key.** No claim was ever accepted, so the host reclaimed the on-chain lockup once its timeout elapsed. The failure was entirely silent from the UI — the screen just never progressed.

### Preimage

The derivation is deterministic — `sha256` of the rescue key's child private key at the swap's `keyIndex`, which the restore endpoint returns in `claimDetails` — and the rescue path *already* derives that exact child key in order to rebuild `keys`. So reproducing the preimage is a matter of calling the derivation that the creation path was doing on its own:

- **`utils/SwapUtils.ts`**: new `deriveSwapPreimage(privateKey)`.
- **`stores/SwapStore.ts`**: both `derivePreimageFromRescueKey` (creation) and the reverse branch of `getRescuableSwaps` (rescue) now call it.

Sharing one function here is the substance of the fix rather than a tidy-up: the two paths must derive byte-identically or the claim is unspendable, and the bug was precisely that the expression lived in only one of them. Note the rescue path derives from the **user-supplied restore mnemonic** already in hand, not from `Storage.getItem(SWAPS_RESCUE_KEY)` — during a restore on a fresh install that key is not yet stored, so calling `derivePreimageFromRescueKey` here would have been wrong.

Only the reverse branch is touched. Submarine swaps have no ZEUS-chosen preimage — the host picks it and reveals it on settlement — so there is nothing to re-derive there.

### Destination address

Resolved lazily at claim time, in `SwapDetails.resolveDestinationAddress`: a fresh address from the wallet in use — the same source the swap creation screen defaults to — persisted against the stored swap by a new `SwapStore.updateSwapDestinationAddress`, so a later attempt pays out to the address already chosen instead of generating another.

Claim time rather than during the rescue, because `/swap/restore` returns every swap for the xpub including long-settled ones: resolving during the rescue would spend an address per restored swap, where this spends one on the swap that actually claims. It also means swaps rescued *before* this PR get an address on their next claim attempt rather than needing a re-rescue.

Deriving an address rather than prompting for one matches what the app already does when a reverse swap is created normally. If a picker is wanted, it can layer on top of this without changing the fallback.

### Fail closed on a missing preimage, and say so

`ReverseClaimTransaction.build` now treats an empty preimage as a missing required field. A reverse claim spends the lockup *by revealing the preimage*, so an empty one can never be accepted by definition; this converts the class of bug from ten guaranteed-rejected broadcasts into a single bail.

That guard on its own was still invisible to the user, which review flagged: `if (!claim) return false;` set no error and left the screen spinning — the same user-visible behaviour as the bug it was added for. Three paths in `createReverseClaimTransaction` now set `error`: nothing to claim to, `build` returning `null`, and all ten broadcasts failing. The websocket is deliberately left open so a later `transaction.confirmed` still retries.

The submarine `createClaimTransaction`'s equivalent `if (!claim) return false;` is left alone on purpose: that claim is cooperative and the swap still settles without it, so surfacing an error there would mislead.

This also updates an existing test that had pinned the old preimage behaviour (`falls back to empty string for null preimage`) — that assertion was encoding the bug.

### Scope note for review

Swaps rescued **before** this change are stored without a preimage and are not repaired by it; they will fail closed with a visible error rather than silently retrying. Re-running the rescue flow re-derives them correctly. I deliberately did not add a migration to backfill stored swaps — that would be storage-touching and wants maintainer sign-off; happy to add one if you'd prefer it.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

`yarn verify` is green (1304 tests). New coverage:

- `utils/SwapUtils.test.ts` — `deriveSwapPreimage` pinned against BIP39-canonical-mnemonic vectors at `m/44/0/0/0/{0,7}`, including the resulting payment hash (i.e. the value the host committed to at creation, which the rescued claim has to reproduce), determinism, index-scoping, and `Uint8Array` input as `@scure/bip32` returns it.
- `models/ClaimTransaction.test.ts` — `build` returns `null` for missing / empty-buffer / empty-string / null / undefined preimages.
- `stores/SwapStore.test.ts` (new) — `getRescuableSwaps` derives the pinned preimage for a rescued reverse swap end to end and leaves submarine swaps without one; `updateSwapDestinationAddress` records the address on a reverse swap, finds a rescued swap still filed under `SWAPS_KEY` (rescued swaps land there whatever their type until the next `fetchAndUpdateSwaps` re-files them), overwrites an existing address, and writes nothing when no swap matches.

**The 5 new `ClaimTransaction` assertions fail against the pre-fix code and pass after it**, as does the new `getRescuableSwaps` preimage test (both verified by reverting the change and re-running), per the CONTRIBUTING "Test Coverage" requirement.

`package.json` gains `uint8array-tools` in the jest `transformIgnorePatterns` allowlist — `ecpair`, which `SwapStore` imports, pulls it in as ESM. No dependency change.

The `SwapDetails` changes carry no unit coverage; the repo has no view tests, so that part rests on the device run below.

- [ ] Android
- [ ] iOS

**Not yet device-tested — please don't merge on CI alone.** The end-to-end check that matters is: create a reverse swap, restore the rescue key on a second device, and confirm the rescued swap claims on-chain to a freshly derived wallet address rather than stalling. I'd also want to confirm the pre-existing rescued-swap case now surfaces an error instead of spinning.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

The claim path is backend-independent (it builds and broadcasts on-chain via the swap host), so this wants one backend that can pay a hold invoice rather than the full matrix. The address fallback does touch the backend: it is gated on `supportsOnchainReceiving()` and errors cleanly on a wallet that cannot receive on-chain.

### Locales

Three new English strings, all failure messages on the swap details screen: `views.SwapDetails.claimAddressFailed`, `views.SwapDetails.claimFailed`, `views.SwapDetails.claimBroadcastFailed`.

### Other:


stores/SwapStore.ts
```diff
@@ -958,9 +959,18 @@ export default class SwapStore {
...
+                                    // rescued claim is built with an empty
+                                    // preimage, fails every retry, and the
+                                    // host reclaims the lockup at timeout.
                                     return {
                                         ...rescuedSwapBase,
-                                        type: SwapType.Reverse
+                                        type: SwapType.Reverse,
+                                        preimage: deriveSwapPreimage(
+                                            childKey.privateKey
+                                        )
```
shubhamkmr04, line 1022: I don't think this alone gets a rescued swap to a broadcastable claim, destinationAddress is still missing. createReverseSwap only posts invoiceAmount/to/from/claimPublicKey/preimageHash and attaches the address client-side afterwards, so the host never has it and it can't come back from /swap/restore. ReverseClaimTransaction.build already requires it, so the rescued claim still returns null and the screen still stalls. The rescue flow needs to prompt for a claim address or fall back to a derived wallet address. 
Also worth noting: if (!claim) return false; in SwapDetails.createReverseClaimTransaction sets no error and leaves submitted false, so the fail-closed path spins silently, same user-visible behaviour as the bug being fixed.

kaloudis: Both correct — thanks, that's the PR's headline claim unmet. Fixed in 5872231.

**destinationAddress.** Right on every count: it's attached client-side after `createReverseSwap` posts, so the host never has it and `/swap/restore` can't return it, and `build` requiring it meant the rescued claim returned `null` whatever the preimage did.

I took the derived-wallet-address option rather than a prompt, since that's what the creation screen already defaults to, and resolve it at claim time rather than in the rescue flow: `/swap/restore` returns every swap for the xpub, including long-settled ones, so resolving during the rescue spends an address per restored swap where claim time spends one on the swap that actually claims. It also means swaps rescued before this PR pick up an address on their next attempt instead of needing a re-rescue. New `SwapStore.updateSwapDestinationAddress` persists it against the stored swap so a retry pays to the same address rather than generating another (it searches both storage keys — rescued swaps land under `SWAPS_KEY` whatever their type until `fetchAndUpdateSwaps` re-files them). Gated on `supportsOnchainReceiving()`. If you'd rather the user picked the address explicitly, a picker layers on top of this cleanly — happy to add it.

**The silent fail-closed.** Also right, and it made the guard indistinguishable from the bug it was guarding against. Three paths in `createReverseClaimTransaction` now set `error`: nothing to claim to, `build` returning `null`, and all ten broadcasts failing. I left the websocket open so a later `transaction.confirmed` still retries.

I deliberately left the submarine `createClaimTransaction`'s `if (!claim) return false;` alone — that claim is cooperative and the swap still settles without it, so an error there would mislead. Shout if you read that differently.

New `stores/SwapStore.test.ts` covers the rescued-reverse preimage derivation end to end through `getRescuableSwaps` (fails with the derivation removed) plus the `updateSwapDestinationAddress` cases. The `SwapDetails` changes have no unit coverage — no view tests in the repo — so that part still rests on the device run, which I haven't done yet.

shubhamkmr04 (approved):
utACK - not device tested as swap endpoints are currently disabled

PR #4461, opened 2026-08-23
fix(cashu): describe the cashu seed by where it actually came from

kaloudis:
# Description

Relates to issue: **ZEUS-0000** (internal security audit finding M12)

`views/Cashu/CashuSeed.tsx` described every wallet's cashu seed the same way:

> The following 12 words are your Cashu wallet backup. You do not have to save these if you have your 24 word embedded LND wallet seed phrase backed up, but it is useful if you want to import funds into another Cashu wallet.

Three different kinds of seed reach that screen, and that sentence is true of only one of them. `CashuStore.getCDKMnemonic` produces the cashu seed three ways, and all three are persisted under the same `v2-bip39` seed version, so nothing downstream could tell them apart:

| Wallet | Cashu seed | What the old copy told them |
|---|---|---|
| 12-word LDK | the wallet mnemonic, **verbatim** (`CashuStore.ts:509-512`) | nothing — that these words are ecash-scoped |
| 24-word LDK / LND | derived from bytes `[48:64]` of the BIP-39 seed | correct |
| Remote nodes (no wallet seed) | freshly generated, **only copy in existence** | that they don't need to save it |

The last row is the one the old copy actively misled, and it is reachable today: a remote-node user is told their wallet seed backs up ecash that it demonstrably cannot, and is invited not to save the only backup they have.

## Fix

New `classifyCashuSeedOrigin` in `utils/CashuUtils.ts` mirrors the three branches in `getCDKMnemonic`. The screen now describes the words actually on it:

- **derived** — a plain note that the wallet seed backup covers these words
- **independent** — a plain note that these are the ONLY backup and must be saved separately
- **wallet seed** — a full-width `ErrorMessage`: these words are also the wallet seed, and control everything in the wallet rather than just the ecash

The wallet-seed warning is repeated inside the copy-to-clipboard modal, which is the most dangerous action on the screen.

That last string deliberately says *only* the blast radius, and no longer tells the user not to share the words. The screen already renders `views.Settings.Seed.text2` ("anyone who sees these words can steal your funds") and `text3` ("Beware of scams…") for all three origins, so sharing advice here would be redundant — and, worse, would contradict the screen's own purpose, since `text1` tells the user these words are what they use to import their ecash into another Cashu wallet. An earlier draft also warned against sharing the seed with a mint, which is not a reachable action: NUT-13 derives secrets and blinding factors client-side, and the mint only ever sees blinded points. Attaching sharing advice to this branch alone would also imply the derived and independent seeds are safe to hand out, which they are not.

Also fixes the modal's word count: it borrows `views.Settings.Seed.dangerousText1` ("your 24 word seed") but the seed here is always 12 words. `views/Settings/Seed.tsx` already does this same `.replace('24', '12')` substitution for its own 12-word case.

## Important scope note — please read before merging

**The wallet-seed branch is currently unreachable through the UI, and I found that while writing this.** `views/Tools/CashuTools.tsx:67` gates the only navigation entry point to `CashuSeed` on `seedVersion === 'v2-bip39' && !isLdk`, and `ldk-node` is the only implementation that has a 12-word `ldkMnemonic`. So the exact cohort whose cashu seed *is* their wallet seed cannot open this screen at all. That gate arrived incidentally with `2b3c99a79` ("backend: LDK Node"), not as a deliberate control for this.

That means the audit finding's headline framing — "user exporting/sharing their cashu seed phrase hands over the entire LDK node wallet" — overstates what is reachable today. I have corrected the audit entry accordingly.

I still think this is worth merging, for three reasons, but the wallet-seed leg is explicitly defense-in-depth and I'd understand a NACK on that part:

1. The **independent-seed** case is a live, reachable data-loss bug for remote-node users, and is fixed here.
2. The protection for the LDK case currently lives in one caller's boolean. Putting the disclosure on the screen that reveals the seed means a future navigation path (or a change to that list gate) cannot silently reopen it.
3. The classifier is the piece worth having regardless — it is the thing that knows the difference, and it is unit-tested.

**What this PR deliberately does NOT do:** the two substantive halves of M12 are untouched, because both are derivation changes rather than copy and both are storage-touching. (a) A 12-word LDK wallet's mnemonic is still written verbatim into a second keychain entry (`<nodeDir>-cashu-seed-phrase`), doubling the number of at-rest locations holding the full wallet seed — which compounds with #4385. (b) The LDK passphrase is still excluded from cashu derivation, so the same mnemonic under two different passphrases derives the *identical* cashu wallet and can collide on proof secrets at a shared mint. Both want maintainer sign-off and a migration plan; happy to take them separately.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

`yarn verify` green. 6 new tests in `utils/CashuUtils.test.ts` covering all three origins, whitespace tolerance in the wallet mnemonic, the LND path, and that `WalletSeed` is never reported without a cashu seed to compare against (1289 → 1295, no existing tests touched).

- [ ] Android
- [ ] iOS

**Not device-tested — copy and layout changes want eyes on both platforms.** Worth checking specifically: the `ErrorMessage` block sits between two `Text` blocks on a screen that has no `ScrollView` in the `!understood` state and an absolutely-positioned button at `bottom: 35`, so the longest of the three strings is the one to check for overlap at large font sizes. The reachable cases to exercise are a remote node (independent) and embedded LND (derived); reaching the wallet-seed case needs the `!isLdk` gate temporarily removed.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

Three new keys, `locales/en.json` only. `views.Cashu.CashuSeed.text1` is reworded, so its existing translations will need refreshing.

### Other:


shubhamkmr04 (approved):
tACK

PR #4463, opened 2026-08-23
fix(nwc): stop returning the invoice as a cashu payment preimage

kaloudis:
# Description

Relates to issue: **ZEUS-0000** (internal security audit finding M11, part a)

The NIP-47 `pay_invoice` response for a **cashu** payment fell back to the bolt11 invoice when the melt produced no preimage (`stores/NostrWalletConnectStore.ts:2573-2577`):

```ts
preimage:
    cashuInvoice.getPreimage ||
    cashuInvoice.getPaymentRequest ||
    '',
```

NIP-47 clients treat `preimage` as proof of payment. A bolt11 string is not one — it's the *request*, it's public, and the client already has it, since it's what they sent us. Returning it means:

- a payment that settled without a recoverable preimage is indistinguishable from one that produced a real receipt, and
- anything downstream verifying `sha256(preimage) === payment_hash` receives a value that cannot possibly match.

## Fix

The lightning leg had this exact fallback and lost it in **#4334**, which tightened it to `preimage || payment?.getPreimage || ''` (`:2406`). This brings the cashu leg in line: prefer the melt's preimage, fall back to the payments-list entry — already looked up two lines above to compute `feeSats`, so no new work — and otherwise return empty rather than something false.

Worth noting the *activity record* written by `finalizePayment` here was already honest: #4334 gave `buildSettledPaymentFallback` the same treatment, and `:2560` passes `preimage: cashuInvoice.getPreimage` with no fallback. So internal accounting was correct and only the wire response lied — which is why this survived the earlier NWC work.

This is part (a) of M11. Parts (b) partial-MPP reported as success and (c) the unvalidated single-mint `meltResult.state` are untouched and remain open; both are in `CashuStore` and are substantially larger.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

`yarn verify` green (1289 → 1292). Three new tests in `stores/NostrWalletConnectStore.test.ts` drive `handleCashuPayInvoice` past its guards to the response shape: the melt preimage is returned when present, the invoice is **never** returned as the preimage, and the payments-list fallback works as it does on the lightning leg.

**Two of the three fail against the pre-fix code** (verified by stashing the store change and re-running), per the CONTRIBUTING "Test Coverage" requirement.

Note `isCashuConfigured` is a MobX computed and can't be `jest.spyOn`'d with `'get'` — the test satisfies its inputs (`cashuEnabled` plus a stubbed `isProperlyConfigured()`) instead. Flagging it because it's a trap for anyone extending these tests.

- [ ] Android
- [ ] iOS

Not device-tested. A real check needs an NWC client paying a lightning invoice from an ecash balance and inspecting the `pay_invoice` response: the `preimage` field should be a real 32-byte hex preimage, or empty — never the bolt11 string that was sent.

Behaviour change worth a maintainer opinion: a settled melt with no recoverable preimage now returns `preimage: ""` instead of the invoice. Clients that (incorrectly) treated any non-empty value as success will now see an empty one. That matches what the lightning leg has done since #4334, so this makes the two consistent rather than introducing something new.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Cashu-specific and backend-independent — it needs an ecash balance and an NWC connection with `pay_invoice`, on any backend.

### Locales

No new strings.

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -1410,3 +1410,93 @@ describe('NostrWalletConnectStore sign_message', () => {
         expect(response.result.methods).toEqual(['get_info']);
     });
 });
+
+describe('NostrWalletConnectStore cashu pay_invoice preimage', () => {
+    const invoice = 'lnbc1cashupreimagetest';
+    const preimage = 'a'.repeat(64);
+
+    // Drives handleCashuPayInvoice past its guards to the response shape.
+    // The melt has already succeeded by that point, so only the fields the
+    // response is built from need to be real.
+    const buildCashuStore = (
+        cashuInvoice: any,
+        payments: any[] = []
+    ): NostrWalletConnectStore => {
+        const store = buildPayInvoiceTestStore();
+        (store as any).cashuStore = {
+            selectedMintUrl: 'https://mint.example',
+            totalBalanceSats: 100_000,
+            paymentError: false,
+            paymentErrorMsg: '',
+            payReq: new Invoice({
+                payment_request: invoice,
+                num_satoshis: '100'
+            } as any),
+            getPayReqError: null,
+            getPayReq: jest.fn().mockResolvedValue(undefined),
+            payLnInvoiceFromEcash: jest.fn().mockResolvedValue(cashuInvoice),
+            isProperlyConfigured: () => true,
+            payments
+        };
+        // isCashuConfigured is a computed, so satisfy its inputs rather
+        // than trying to redefine it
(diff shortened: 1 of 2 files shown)


ajaysehwal (approved):
tACK

PR #4464, opened 2026-08-23
fix(nwc): mask the connection string like every other credential

kaloudis:
# Description

Relates to issue: **ZEUS-0000** (internal security audit finding L5)

The Nostr Wallet Connect URL was the only credential field on the wallet configuration screen rendered in the clear.

Every sibling — LndHub password (`:2214`), macaroon (`:2437`), rune (`:2538`), LNC pairing phrase (`:2756`) — passes `secureTextEntry={this.state.hidden}` and sits beside a `ShowHideToggle`. The `nostrWalletConnectUrl` input passed neither.

That's the wrong field to leave visible. `nostr+walletconnect://…` carries the connection **secret** in its query string, so it isn't an identifier — it's the spending credential for that connection. Anyone who reads it off the screen can pay from the wallet up to the connection's budget. Shoulder surfing, screen sharing, and screenshots on a settings screen users are told is safe to open.

## Fix

Applies the existing pattern verbatim: wrap the input in the same `flexDirection: 'row'` container, mask it with the shared `this.state.hidden`, and add the same `ShowHideToggle`. `ShowHideToggle` was already imported (`:55`).

Nothing else changes — the `nostr+walletconnect://` prefix validation, the whitespace trimming, the error colouring, and `locked={loading}` are all untouched.

### Note on the shared toggle

`hidden` is a single piece of state shared by all masked fields on this screen, so the toggle reveals whichever credential field is currently rendered. That's pre-existing behaviour and only one implementation's fields are visible at a time, so joining the pattern is the consistent choice rather than introducing per-field state. Flagging it in case you'd rather these were independent — that would be a separate, broader change.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

No utility files touched — this is a view-only change. `yarn verify` green (1289 tests, unchanged).

- [ ] Android
- [ ] iOS

**Not device-tested, and this one genuinely needs eyes** — it's a layout change and there are no screenshots yet, which I know is thin for a UI PR. What to check: with `implementation = nostr-wallet-connect`, the URL renders masked by default, the toggle reveals and re-hides it, the input still resizes correctly beside the toggle (NWC URLs are long, so the `flex: 1` / `marginRight: 15` split is the thing to eyeball), pasting a URL still validates and the error colour still appears for a malformed one, and the field is still usable while `locked={loading}`.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Only the NWC client configuration screen is affected.

### Locales

No new strings.

### Other:

### DIFF
--- a/views/Settings/WalletConfiguration.tsx
+++ b/views/Settings/WalletConfiguration.tsx
@@ -1968,30 +1968,51 @@ export default class WalletConfiguration extends React.Component<
                                             'views.Settings.WalletConfiguration.nostrWalletConnectUrl'
                                         )}
                                     </Text>
-                                    <TextInput
-                                        placeholder={'nostr+walletconnect://'}
-                                        textColor={
-                                            nostrWalletConnectUrlError
-                                                ? themeColor('error')
-                                                : themeColor('text')
-                                        }
-                                        value={nostrWalletConnectUrl}
-                                        autoCapitalize="none"
-                                        onChangeText={(text: string) =>
-                                            this.setState({
-                                                nostrWalletConnectUrl: text
-                                                    .trim()
-                                                    .replace(/\s+/g, ' '),
-                                                nostrWalletConnectUrlError:
-                                                    !text.startsWith(
-                                                        'nostr+walletconnect://'
-                                                    ),
-                                                saved: false
-                                            })
-                                        }
-                                        locked={loading}
-                                        autoCorrect={false}
-                                    />
+                                    <View
+                                        style={{
+                                            flexDirection: 'row',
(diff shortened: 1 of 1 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4466, opened 2026-08-23
fix(backends): anchor the WebSocket scheme rewrite to the scheme

kaloudis:
# Description

Relates to issue: **ZEUS-0000** (internal security audit finding L1)

`getURL` built its `wss://` URL with bare string replaces (`backends/LND.ts:221`, `backends/CLNRest.ts:201`):

```ts
baseUrl = baseUrl.replace('https', 'wss').replace('http', 'ws');
```

`String.replace` with a **string** pattern rewrites the first match *anywhere in the string*, not just the scheme. So any host whose own name contains `http` gets mangled, and the macaroon then rides a WebSocket to a different — attacker-registrable — hostname.

Both replaces can bite, and they compound in a way the audit didn't capture:

| Input | Before | After |
|---|---|---|
| `https://node.example.com:8080` | `wss://node.example.com:8080` | unchanged |
| `http://192.168.1.5:8080` | `ws://192.168.1.5:8080` | unchanged |
| `https://xyz.onion:10009` | `wss://xyz.onion:10009` | unchanged |
| `https://httpbin.org:8080` | **`wss://wsbin.org:8080`** | `wss://httpbin.org:8080` |
| `https://api.http-relay.com` | **`wss://api.ws-relay.com`** | `wss://api.http-relay.com` |
| `http://myhttpserver.local:3000` | **`ws://mywsserver.local:3000`** | `ws://myhttpserver.local:3000` |

The third failing row is the one worth noting: a plain `http://` host hits the **first** rule, because `myhttpserver` contains the substring `https`. The scheme ends up right and the hostname ends up wrong.

## Fix

Anchor both replaces to `^`.

The ordering still holds: an `https` URL becomes `wss` and the second rule then can't match (the string now starts `wss`), while an `http` URL falls through to it. `hostPath` guarantees a scheme is present (`host.includes('://') ? host : 'https://' + host`), so anchoring is always correct.

Hosts without an embedded `http` are completely unaffected — which is every host this has ever been exercised against, and is why it went unnoticed.

## Scope: which call sites are live

I re-checked rather than trusting the audit's note, and it holds:

- **LND — three live callers**: `subscribeCustomMessages` (`:316`), `openChannelStream` (`:481`), `initChanAcceptor` (`:851`).
- **LND `wsReq` (`:160`) is dead** — no callers anywhere in the tree. Only its locale string `backends.LND.wsReq.warning` is reused by the three live methods.
- **CLNRest is dead** — its single `getURL` call (`:178`) passes no `ws` argument, so the branch is unreachable.

I fixed both files anyway rather than leaving a bug waiting for a caller. Deleting the dead `wsReq` and the unreachable CLNRest branch is a reasonable follow-up, but it's a cleanup decision rather than part of this fix, and I'd rather not bundle a deletion into a security patch. Happy to do it separately if you want it.

Note this is the *scheme* half of the WebSocket problems. The Tor-bypass half of the same code (L2 in the audit) is separate and is what #4372 addresses with an enable-time warning.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

`yarn verify` green (1289 → 1306).

**`backends/LND.test.ts` and `backends/CLNRest.test.ts` are the first test files under `backends/`.** 17 tests; **6 fail against the pre-fix code**, verified by checking the parent commit's files back out and re-running:

```
● LND.getURL     › leaves a host containing "http" in its name intact
● LND.getURL     › leaves an http host containing "https" in its name intact
● LND.getURL     › rewrites only the scheme, never a later occurrence
● LND.getURL     › defaults a schemeless host to https, then to wss
● CLNRest.getURL › leaves a host containing "http" in its name intact
● CLNRest.getURL › leaves an http host containing "https" in its name intact
```

The rest pin the behaviour the fix must not disturb: normal hosts, `.onion`, schemeless hosts, port assembly, trailing-slash stripping, and the non-`ws` path.

### Why there were no backend tests before, and what unblocks them

Nothing under `backends/` was testable because these modules pull native code in at import time — `react-native-blob-util` directly, and `react-native-nitro-tor` through `utils/TorUtils`. Three `jest.mock` calls cover it.

`CLNRest` needs a fourth, and it's worth flagging on its own: `CoreLightningRequestHandler` imports `CLNRest` and calls `new CLNRest()` **at module scope**, while `CLNRest` imports back from it. With `CLNRest` as the entry point the cycle resolves with the class still `undefined` and the import throws. That circularity is pre-existing and doesn't bite the app, because nothing there imports `CLNRest` first — but it will bite the next person who tries to test this file, so the mock carries a comment explaining it. Worth untangling separately.

All four mocks are import-time scaffolding only; `getURL` is a pure function of its arguments and touches none of them. The same three-mock pattern should open up the rest of `backends/` for testing.

CLNRest's `ws` branch is unreachable today, so its tests exist to keep the two implementations in step and stop a future caller reviving the bug.

- [ ] Android
- [ ] iOS

Not device-tested. What to exercise: a remote LND node over a normal hostname, confirming the three live WebSocket paths still connect — batch channel opens (`openChannelStream`), LSPS custom messages (`subscribeCustomMessages`), and the channel acceptor (`initChanAcceptor`). Those are pass-through cases the fix must not disturb; the failing rows in the table need a host with `http` in its name, which is easiest to fake with a hosts-file entry.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Remote LND only in practice — `EmbeddedLND` and `LightningNodeConnect` override these methods natively, CLNRest's branch is unreachable, and LndHub inherits `getURL` but never opens a WebSocket.

### Locales

No new strings.

### Other:

### DIFF
--- a/backends/CLNRest.test.ts
+++ b/backends/CLNRest.test.ts
@@ -1,0 +1,83 @@
+// Import-time scaffolding, as in LND.test.ts. CoreLightningRequestHandler
+// additionally has to be stubbed because it and CLNRest import each other
+// and it does `new CLNRest()` at module scope: with CLNRest as the entry
+// point the cycle resolves with CLNRest still undefined. getURL touches
+// none of these.
+jest.mock('react-native-blob-util', () => ({
+    __esModule: true,
+    default: { fetch: jest.fn(), config: jest.fn() }
+}));
+jest.mock('../stores/Stores', () => ({
+    settingsStore: { settings: {} },
+    nodeInfoStore: { nodeInfo: {} }
+}));
+jest.mock('../utils/TorUtils', () => ({
+    doTorRequest: jest.fn(),
+    isOnionHttpsUrl: jest.fn(),
+    RequestMethod: {}
+}));
+jest.mock('./CoreLightningRequestHandler', () => ({
+    getBalance: jest.fn(),
+    getChainTransactions: jest.fn(),
+    getOffchainBalance: jest.fn(),
+    listPeers: jest.fn(),
+    listClosedChannels: jest.fn(),
+    listPeerChannels: jest.fn()
+}));
+
+import CLNRest from './CLNRest';
+
+// CLNRest's own `getURL` is not reachable with ws=true today — its single
+// caller passes no ws argument — so these lock in the anchored behaviour
+// against the day it acquires one, and keep it in step with LND.getURL.
+describe('CLNRest.getURL', () => {
+    const cln = new CLNRest();
+    const url = (host: string, port: string | number = '', ws = false) =>
+        cln.getURL(host, port, '/v1/route', ws);
+
+    describe('scheme rewriting for WebSocket URLs', () => {
+        it('rewrites https to wss and http to ws', () => {
+            expect(url('https://node.example.com', 3010, true)).toBe(
+                'wss://node.example.com:3010/v1/route'
+            );
+            expect(url('http://192.168.1.5', 3010, true)).toBe(
+                'ws://192.168.1.5:3010/v1/route'
+            );
+        });
+
+        it('leaves a host containing "http" in its name intact', () => {
+            expect(url('https://httpbin.org', 3010, true)).toBe(
+                'wss://httpbin.org:3010/v1/route'
+            );
+        });
+
+        it('leaves an http host containing "https" in its name intact', () => {
+            expect(url('http://myhttpserver.local', 3010, true)).toBe(
+                'ws://myhttpserver.local:3010/v1/route'
+            );
+        });
+    });
+
+    describe('without the ws flag', () => {
+        it('leaves the scheme alone', () => {
+            expect(url('https://node.example.com', 3010)).toBe(
+                'https://node.example.com:3010/v1/route'
+            );
+            expect(url('https://httpbin.org', 3010)).toBe(
+                'https://httpbin.org:3010/v1/route'
+            );
+        });
+
+        it('defaults a schemeless host to https', () => {
+            expect(url('node.example.com', 3010)).toBe(
+                'https://node.example.com:3010/v1/route'
+            );
+        });
+
+        it('strips a trailing slash before appending the route', () => {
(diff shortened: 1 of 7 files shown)


ajaysehwal (commented):
@kaloudis, can we also update views/Swaps/SwapDetails.tsx to avoid the unanchored replace here?

https://github.com/ZeusLN/zeus/blob/484254aad0d8c947147fcb4e6743e03ec5839d9b/views/Swaps/SwapDetails.tsx#L261

There is the same pattern at line 453 as well. It would be safer and more consistent to anchor it, for example:

`endpoint.replace(/^https/, 'wss') + '/ws'
`

This is less severe than the previous getURL issue since there is only one replacement, but anchoring it would still be a nice defense-in-depth improvement.

kaloudis:
> @kaloudis, can we also update views/Swaps/SwapDetails.tsx to avoid the unanchored replace here?
> 
> https://github.com/ZeusLN/zeus/blob/484254aad0d8c947147fcb4e6743e03ec5839d9b/views/Swaps/SwapDetails.tsx#L261
> 
> There is the same pattern at line 453 as well. It would be safer and more consistent to anchor it, for example:
> 
> `endpoint.replace(/^https/, 'wss') + '/ws' `
> 
> This is less severe than the previous getURL issue since there is only one replacement, but anchoring it would still be a nice defense-in-depth improvement.

good catch. covered in https://github.com/ZeusLN/zeus/pull/4466/commits/3673f89ae450fbbb6ce54d0bd827819f0ac47f18

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4474, opened 2026-08-24
[Feature] Add remote phoenixd wallet interface

newtonick:
# Description

Adds [phoenixd](https://phoenix.acinq.co/server) (ACINQ's server-side Phoenix node) as a new remote wallet interface.

My expectation of this pull request getting merged is low. I didn't check ahead of time what the Zeus projects feeling or thoughts would be on adding Phoenixd (ACINQ) as a remote node. Phoenix/ACINQ is some ways is a direct competitor to Zeus.

I'm also aware that phoenixd can already be reached indirectly through LNbits via LNDHub/NWC on Zeus, but only its Lightning interface is possible. Connecting to it natively unlocks the three things that indirection hides:

1. On-chain receiving via phoenixd's swap-in address
2. On-chain sending by splicing out of the channel
3. Payment and invoice history straight from the node, including the fees ACINQ charges

Verified against phoenixd v0.9.0 on mainnet. The published API docs using Kotlin source (`Api.kt`, `json/JsonSerializers.kt`) and confirmed against a live mainnet test node.

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

Disclosure: most of the code and this pull request description were written by Claude (Fable) and then reviewed and edited by me.

## What's supported

| Area | Support |
|---|---|
| Receive | BOLT 11 invoices, BOLT 12 offer, on-chain via `getswapinaddress` |
| Send | BOLT 11, BOLT 12 offers, Lightning addresses, on-chain splice-out with an explicit sat/vB fee |
| Balances | Lightning and on-chain, plus pending swap-in funds |
| Activity | Last 100 incoming (including unpaid invoices) and last 100 outgoing, with ACINQ's fees shown |
| Config | Base URL + HTTP password, with automatic detection of phoenixd's read-only password |

Deliberately not supported in this pull request: channel listing and closing, coin control and UTXO selection, fee bumping, ecash, and on-chain send-max. Mainnet only.

## Files

New:

- `backends/Phoenixd.ts`: the backend class
- `utils/PhoenixdRequestHandler.ts`: pure phoenixd-JSON to Zeus-shape mappers
- `backends/Phoenixd.test.ts` and `utils/PhoenixdRequestHandler.test.ts`: 44 tests

Modified: `utils/BackendUtils.ts`, `stores/SettingsStore.ts`,
`views/Settings/WalletConfiguration.tsx`, `views/Wallet/BalancePane.tsx`,
`views/Receive.tsx`, `views/Invoice.tsx`, `models/Invoice.ts`,
`components/LayerBalances/index.tsx`, `locales/en.json`.

### Lightning and on-chain balances are the same sats, on purpose

This is the choice most likely to look like a bug, so it is worth stating plainly. phoenixd has no separate on-chain wallet. Every on-chain payment is a splice out of the Lightning channel, so the funds spent on-chain are the same funds that back the Lightning balance. Both balance rows report the channel balance, and they legitimately show the same number.

Summing them, as the wallet total normally does, would double-count. Rather than branch on `implementation` (which the codebase reserves for connection lifecycle, not features), this adds a capability flag:

```ts
hasOverlappingBalances = () => true;
```

`views/Wallet/BalancePane.tsx` reads it and reports the total without adding the two together. Every other backend omits the flag and keeps today's behavior exactly.

The on-chain row is not purely a duplicate. It also carries swap-in funds that have arrived on-chain but have not been swept into the channel yet, reported as unconfirmed.

### Read-only passwords are detected, not configured

phoenixd exposes two types of passwords: `http-password` (full access) and `http-password-limited-access`, which can create invoices and read history but is rejected on every payment endpoint. Instead of asking users to declare which one they pasted, the backend probes once during `getMyNodeInfo`. An empty-bodied `POST /payinvoice` returns 401 for the limited password, because auth rejects it before the handler runs, and 400 for the full one, because the `invoice` parameter is missing. No payment can occur either way. When limited access is detected, `supportsLightningSends()` and `supportsOnchainSends()` return false and Zeus hides every send path rather than letting it fail at call time.

### Splice-outs and liquidity purchases appear as on-chain transactions

phoenixd returns one combined outgoing list, tagged by `subType`. Lightning payments become `Payment` models. `splice_out`, `splice_cpfp`, `auto_liquidity`, `manual_liquidity` and `channel_close` become `Transaction` models carrying the real txid. That is what puts the mining and liquidity fees ACINQ charges into their own activity rows, and it keeps splice-outs from appearing twice.

### `sendtoaddress` success is detected by shape, not status code

phoenixd answers `POST /sendtoaddress` with HTTP 200 in every case: a bare txid on success, and a plain-text reason on failure. Confirmed against a live node, where a bad address returns `Invalid address` with status 200. `sendCoins` therefore tests the body against a txid pattern and throws otherwise. Trusting the status code here would report failed on-chain sends as successful.

### Payments are raced against a client-side timeout

phoenixd blocks on `/payinvoice` until the payment resolves and has no server-side deadline, so the window is entirely client-imposed. Following the LndHub precedent, the call is raced against the payment-timed-out shape so a slow payment surfaces as "may be in transit, check Activity" rather than a retryable-looking transport error. A retry here would risk double-sending.

### Tor is disabled for this interface

The Tor toggle is hidden on the phoenixd config form, and switching the implementation dropdown to phoenixd clears any `enableTor` value carried over from another interface. Without that, it would persist with no way to turn it off. The backend's Tor branch is kept rather than deleted so that a config arriving with the flag set still routes over Tor instead of silently falling back to clearnet.

### Address-reuse warning on the Receive screen

`getswapinaddress` returns the first *unused* derived address, so it only rotates once the current one receives coins, unlike backends that mint a fresh address per request. A warning now appears under the on-chain QR and under the unified QR, which embeds the same address, gated on a new`reusesOnchainAddress()` flag.

## Shared-code changes

Three changes touch code outside phoenixd. All are guarded so other backends behave identically:

- `models/Invoice.ts` adds a `getFee` getter reading a `fees_sat` field that only phoenixd sets, so the fee ACINQ takes on an incoming payment can be displayed. It returns `0` everywhere else. Invoice CSV export is unaffected, since its `getFee` column applies only to transactions.
- `views/Wallet/BalancePane.tsx` gets the `hasOverlappingBalances()` branch described above.
- `components/LayerBalances/index.tsx` is a bug fix. The "Move on-chain funds to lightning" button rendered whenever the Lightning balance was zero and the on-chain balance was not, navigating to `OpenChannel`. It was unreachable before, because the only backends without channel management also report no on-chain balance. phoenixd reaches it on a fresh wallet with funds sent to the swap-in address and nothing swept yet, so it is now gated on `supportsChannelManagement()`. Happy to split this into its own PR if preferred.

## Known limitations

- No on-chain send-max. `sendtoaddress` requires an explicit `amountSat` and has no `sendAll` parameter, checked on v0.9.0 and master, so `supportsOnchainSendMax()` is false and the fund-max switch is hidden.
- Incoming on-chain payments to the swap-in address are reflected in the balance but never appear in Activity, at any confirmation stage. phoenixd's /payments/incoming endpoint returns only Lightning payments.
- History caps at 100 each. The API supports `from`, `to` and `offset` pagination; wiring it up is left for a follow-up.
- Offers are partly synthetic. `listOffers` returns phoenixd's single seed-derived node offer. Offers created through `createoffer` are not stored by phoenixd, so they cannot be listed afterwards, and their `single_use` and `label` values are echoed from the request rather than read back.
- `disableOffer` is a no-op, matching `LdkNode`. phoenixd's node offer is derived from the seed and cannot be revoked. The "Disable Offer" button still reports success. This is a pre-existing issue shared with LDK Node that would be fixed properly by a `supportsDisablingOffers()` flag, deliberately left out of this PR to keep its scope contained.
- Settlement is polled, not streamed. Invoice and on-chain receive detection poll on a timer. phoenixd's `/websocket` `payment_received` event is the better long-term answer.
- Fee bumping is skipped, though phoenixd exposes `/bumpfee`. This is because fee bumping bumps all current on-chain transactions, not just one.

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

44 new tests. `utils/PhoenixdRequestHandler.test.ts` covers the JSON mapping: paid, unpaid and expired invoices, the Payment versus Transaction split across every `subType`, millisecond to second and msat to sat conversions, balance projection, mainnet rejection, and list ordering. `backends/Phoenixd.test.ts` covers the transport and every send and receive type, including the HTTP-200-with-failure-body case, the payment timeout race, and both outcomes of the limited-access probe. Fixtures are real responses captured from a live node.

Note: `backends/Phoenixd.test.ts` is the first test for any backend class in the repo. Network is mocked at `react-native-blob-util`, matching the existing utility tests.

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

Three new keys in `locales/en.json` only:
`views.Settings.AddEditNode.phoenixdPassword`,
`views.Receive.warningAddressReuse`, `views.PayCodes.nodeOffer`.

No new dependencies.

### Other:

Both worth a maintainer's opinion: the README lists supported node types, and
onboarding lists wallet interfaces, so each may want a phoenixd mention.


kaloudis:
Sorry, will not support.

phoenixd vendor locks liquidity provisioning to ACINQ. Furthermore, Phoenix and phoenixd degrade user privacy by using Trampoline routing with no ability to opt out.. This method outsources route calculation for payments to ACINQ meaning they know your payment amount and destination when paying to an invoice without blinded routes. This coupled with the vendor-locked liquidity essentially means no privacy.

We don't believe in adding tech debt for something that both goes against our principles and supports a competitor. This being said, ZEUS is FOSS. Feel free to create your own fork supporting this.

newtonick:
Appreciate the response and understand the decision

PR #4475, opened 2026-08-24
config: default LDK RGS servers to v2 endpoints

kaloudis:
# Description

Relates to issue: #4470 (keystore-op reduction) — and fixes BOLT 12 offer payments on LDK Node (field report: Olympus channel, paying ACINQ Phoenix offers).

## Problem

BOLT 12 offer payments from LDK Node wallets consistently time out. Root cause: **our default RGS server serves v1 snapshots, which contain no node announcement data.** LDK's `DefaultMessageRouter` never relays onion messages multi-hop — to deliver a BOLT 12 `invoice_request` it looks up the offer's introduction node in the network graph and requires `announcement_info` (onion-message feature bit + addresses) to direct-connect ([messenger.rs](https://github.com/lightningdevkit/rust-lightning/blob/main/lightning/src/onion_message/messenger.rs)). With a v1-sourced graph, `announcement_info` is `None` for every node, so the `invoice_request` is never sent and the payment fails deterministically.

Only RGS **format v2** snapshots carry node announcements (features + addresses):

```
$ curl -s https://rgs.zeusln.com/snapshot/0            | xxd -l4  →  4c 44 4b 01  ("LDK" v1)
$ curl -s https://rgs.zeusln.com/snapshot/v2/0                    →  404
$ curl -s https://rapidsync.lightningdevkit.org/snapshot/v2/0     →  4c 44 4b 02  ("LDK" v2) ✓
$ curl -s https://rapidsync.lightningdevkit.org/testnet/v2/snapshot/0  →  4c 44 4b 02  ("LDK" v2) ✓
```

Note the LDK server's path asymmetry: testnet v2 lives at `/testnet/v2/snapshot`, not `/testnet/snapshot/v2`. An earlier revision of this PR probed the latter, got a 404, and wrongly concluded no testnet v2 snapshot exists; the testnet default and migration are now included.

Since the retired `migrateRgsDefaultToZeus` and wallet creation (`WalletConfiguration.tsx`) both pinned the default URL into each node's `ldkRgsServer`, changing the constants alone would not reach any existing wallet — hence the migration.

## Changes

- `utils/LdkNodeUtils.ts`: mainnet `RGS_SERVERS_MAINNET` entries now point at the `/snapshot/v2` endpoints (ZEUS + LDK); `RGS_SERVERS_TESTNET` now points at `/testnet/v2/snapshot`.
[...]

Issue #4470, opened 2026-08-23
Startup performs dozens of serialized keystore ops; slow KeyMint devices see multi-minute boots and Android 16 freezer kills


kaloudis:
## Summary

A field report (Galaxy S25, SM-S931B/DS, Android 16, ZEUS v13.2.0, remote LND REST) of 9-minute app startups was root-caused via a full Android bug report: the device's hardware keystore (KeyMint/TEE) services operations in 1 to 16.7 seconds instead of milliseconds. The system keystore2 watchdog fired 759 times in a ~20 minute logcat window, and continued firing at 16-22 reports/minute in a control window where ZEUS was not running, so the degradation is device-wide and not caused by ZEUS. But ZEUS is disproportionately affected because its startup path performs dozens of serialized keystore round trips before the wallet renders, and each `Keychain.getInternetCredentials` read expands to multiple KeyMint operations (key fetch, cipher begin/update/finish).

Measured timeline from the same report: process start to React root render took 1.7 seconds; everything after that, roughly 2 minutes of black screen plus 5 minutes on the loading screen, is keystore wait.

## Why ZEUS multiplies the damage

Startup currently performs, before the wallet is interactive:

- `zeus-settings-v2` blob read at least 3 times: `components/StealthModeWrapper.tsx` (gates the entire app tree on Android, the black-screen phase), FiatStore's constructor-time `getSettings()`, and `getSettingsAndNavigate`
- `stores/NotesStore.ts` `loadNoteKeys()` at module eval: 1+N keychain reads, serialized with everything else on react-native-keychain's global native mutex
- 4 `EncryptedStorage` migration-flag reads on every `getSettings()` call (`migrateRgsDefaultToZeus`, `migrateSwapHostsToBoltz`, `migrateInvoiceExpiryDisplay`, `migrateOlympusHostsToZeusLsp`), each an EncryptedSharedPreferences/keystore touch
- An unconditional `updateSettings({ supportedBiometryType })` in `views/Wallet/Wallet.tsx` `componentDidMount`: full blob read plus full blob write every launch, even when the value is unchanged (`SettingsStore.checkBiometricsStatus` already shows the diff-first pattern)
- `stores/UnitsStore.ts` read at module eval

At normal keystore speeds this is invisible. At seconds per operation it is minutes, and the app gives no feedback while it happens.

## Secondary finding: Android 16 freezer kills ZEUS mid-boot

[...]

PR #4477, opened 2026-08-24
refactor(migrations): stamp settings blob, single write

kaloudis:
# Description

Relates to issue: #4470 (implements the "fold the one-shot migration flags" checkbox).

## Problem

Startup pays for the migration system on every boot (#4470 field report: 9-minute boots on a Galaxy S25 with a degraded KeyMint):

- The modern load path reads **5 EncryptedStorage migration flags on every `getSettings()` call**, and `getSettings()` runs 3+ times per boot, so 15+ serialized keystore touches in steady state, forever.
- A first run after upgrade can add up to 5 full-blob writes + 5 flag writes (two migrations wrote the blob **unconditionally**).
- The legacy upgrade path runs 8 MOD_KEY blocks, each doing flag read → **full-blob `setSettings`** → flag write, before `storageMigrationV2` persists the blob again: ~12 redundant blob writes.

## Design

**Version-stamp the blob content.** `settingsVersion` (new top-level scalar, `SETTINGS_VERSION = 1`, defined in the dependency-free `utils/SettingsVersion.ts` and re-exported from `SettingsStore`) lives in the blob itself:

- **Stamped blob → migrations skipped with zero keystore reads.** The check is on the already-parsed object.
- **Unstamped blob → one-time consolidation:** read the 5 retired flags once (so migrations a previous version already applied are not re-applied over values the user has since changed), apply the needed mutations in memory, and let **a single write** land values + stamp together. The retired flags are never written again; leftovers are inert and cleared by Clear All Data.
- **The consolidation write is serialized through `updateSettingsQueue`.** Queue membership is passed explicitly (`applySettingsUpdate` → `getSettings(fromQueue)` → `runSettingsMigrations(settings, inQueue)`). Inside the queue's critical section the consolidation only mutates in memory; the migrated values and stamp ride the enclosing update's own persist, so there is exactly one blob write. Outside the queue the snapshot is never touched: a no-op `updateSettings({})` is enqueued, and its critical section redoes the consolidation on fresh state. A concurrent settings update (e.g. the biometry-type write on first boot) can never be clobbered by a stale pre-migration snapshot, regardless of what else is in flight.
- **Legacy path:** the 8 MOD_KEY flags are read in one `Promise.all` batch, blocks are pure mutations, and the single persist stays in `storageMigrationV2`. MOD_KEY flags are also never written anymore (after migration the blob is stamped, so they're never consulted again).
- The five shared migrations become pure, individually-tested mutations (`applyRgsDefaultsToV2`, `applyOlympusHostsToZeusLsp`, `applySwapHostsToBoltz`, `applyRetiredSwapHosts`, `applyInvoiceExpiryDisplay`); the orchestrator (`runSettingsMigrations`) owns flag reads and write routing (the write itself belongs to `applySettingsUpdate` on the modern path and `storageMigrationV2` on the legacy path). **Migration logic is byte-for-byte unchanged.**

## Before / after

| | Before | After |
|---|---|---|
| Steady-state load, migrated install | 5 flag reads per `getSettings()`, 15+ keystore touches per boot, forever | **0** (stamp check on the parsed blob) |
| First boot after upgrade | Up to 5 blob writes + 5 flag writes, interleavable with a crash | One pass, **one write** carrying values + stamp |
| Crash mid-migration | Migrated values and done-flags can diverge | Atomic: nothing persists until the single write lands |
| Legacy-blob upgrade | ~12 redundant blob writes | **1** persist (in `storageMigrationV2`) |
| Concurrent update during migration | A mid-migration blob write could silently erase it | Serialized through `updateSettingsQueue` |
| Migration functions | 5 methods, each doing its own storage I/O | 5 pure mutations + 1 orchestrator |
| `SETTINGS_VERSION` in tests | Hardcoded copy inside a jest mock (would pin gating tests to v1 after a bump) | Real constant imported from a dependency-free module |

## Future migrations

Bump `SETTINGS_VERSION`, add one pure mutation to `applySettingsMigrations` gated on the blob's prior version (`priorVersion < 2`), and leave the existing blocks untouched. A blob below the current version runs every admitted block in file order and is re-stamped at the current version in the same single write, so users who skip releases catch up in one pass. Blocks are exactly-once: a gate admits a blob only if it was stamped below that block's version, so earlier blocks never re-run on later upgrades (the v1 blocks are gated too, protecting values a post-stamp user restored). Contracts for new migrations: pure, synchronous, in-memory; value-conditional and idempotent as defense in depth (pre-stamp app downgrades and crash retries still re-run them); gated on version, never on new keystore flags. A gate-consistency test asserts the highest gate equals `SETTINGS_VERSION`, failing CI on a forgotten bump or a bump without a gated block. A version number is frozen once any public build (alphas and TestFlight included) ships stamping it; migrations added after that take a fresh bump. Within that constraint, unreleased migrations may share a bump, as #4475 and #4542 folded into version 1 here.

## Migration plan

| Cohort | Effect |
|---|---|
| Fresh installs | Born stamped (inline defaults include `settingsVersion`); consolidation never runs |
| Current v2 users (flags set) | One consolidation pass: flags honored, nothing re-applied, stamp written in a single write; zero migration I/O afterward |
| v2 users missing some flags (skipped releases) | Un-flagged migrations apply (value-conditional, identical logic), stamp written, one write |
| Legacy-blob upgraders | Same end state as today in **one** persist; MOD_KEY flags from ancient installs still honored |
| iCloud-restored old blob (fresh device) | No stamp, no local flags → all migrations correctly re-run on the restored data. This is **more correct** than flag-based state, which lives apart from the data it describes |
| App downgrade after stamping | Old code ignores the extra field and re-runs its value-conditional migrations (idempotent); re-upgrade re-honors the flags it wrote. Best-effort, as today |

- **Crash mid-migration**: nothing persisted until the single write lands → clean retry next boot. Strictly better than today, where blob writes and flag writes could interleave with a crash.
- **Idempotent**: all mutations are value-conditional; consolidation re-running is harmless.
- **Shallow-merge safety**: `settingsVersion` is a top-level scalar, carried by `updateSettings` spreads; no nested-group hazard.
- **No new keychain keys**; no registry changes apply (`DataClearUtils` wipes EncryptedStorage + blob wholesale already).
- **Revert story**: fully revert-safe. Reverting leaves a harmless extra field in blobs and the retired flags were only read, never removed; old code resumes flag-based behavior exactly where it left off (flags for migrations consolidated in the interim are unset → old code re-applies value-conditional logic, idempotent).

## Relationship to #4475 and #4542

Both landed on master first, so this PR absorbed them per the contingency documented in earlier revisions: the RGS mutation is `applyRgsDefaultsToV2`, carrying master's per-network v1 → v2 URL map verbatim and honoring `rgs-defaults-v2` (the superseded `rgs-default-zeus` flag is deliberately not consulted, matching master); #4542's retired-swap-host migration folded in as `applyRetiredSwapHosts`, honoring `swap-hosts-retired-eldamar`. `SETTINGS_VERSION` stays 1 because neither change shipped in a release before the stamp.

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [x] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

_Migration suite restructured: mutations are tested pure (all prior domain cases preserved: swap hosts, retired swap hosts, Olympus hosts, expiry-triplet repair, RGS defaults), plus a `runSettingsMigrations` orchestrator suite covering stamp short-circuit, single-write consolidation, retired-flag honoring, queue routing (direct in-queue persist, no stale-snapshot write, concurrent-update clobber repro), object-not-string persistence, and steady-state silence. `stores/SettingsStore.test.ts` now fails loudly if `getSettings` swallows a load error, closing the silent-mock-drift trap._

- [ ] Android
- [ ] iOS

_Device testing round pending; versions will be recorded here._

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

_Settings-load path affects all backends equally; upgrade testing on one embedded + one remote wallet should cover it, versions to be recorded here._

### Locales

### Other:

### DIFF
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -32,8 +32,8 @@ jest.mock('../utils/LocaleUtils', () => ({
 jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
-    migrateRgsDefaultToZeus: jest.fn().mockResolvedValue(undefined),
-    migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
+    runSettingsMigrations: jest.fn().mockResolvedValue(undefined),
+    migrateRetiredSwapHosts: jest.fn().mockResolvedValue(undefined),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
     storageMigrationV2: jest.fn().mockResolvedValue(undefined)
 }));
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -33,7 +33,6 @@ jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
     runSettingsMigrations: jest.fn().mockResolvedValue(undefined),
-    migrateRetiredSwapHosts: jest.fn().mockResolvedValue(undefined),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
     storageMigrationV2: jest.fn().mockResolvedValue(undefined)
 }));
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -35,9 +35,9 @@ jest.mock('../utils/MigrationUtils', () => ({
     // getSettings adopts the return value (the queue's authoritative
     // object when consolidation routes through updateSettings), so the
     // stub must hand the settings back rather than resolve undefined
-    runSettingsMigrations: jest.fn().mockImplementation(
-        async (settings: any) => settings
-    ),
+    runSettingsMigrations: jest
+        .fn()
+        .mockImplementation(async (settings: any) => settings),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
     storageMigrationV2: jest.fn().mockResolvedValue(undefined)
 }));
@@ -85,15 +85,27 @@ const seedSettings = (settings: any) => {
 const persistedSettings = () => JSON.parse(StorageMock._backing[STORAGE_KEY]);
 
 let logSpy: jest.SpyInstance;
+let errorSpy: jest.SpyInstance;
 
 beforeEach(() => {
     for (const key of Object.keys(StorageMock._backing)) {
         delete StorageMock._backing[key];
     }
     logSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
+    errorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
 });
 
 afterEach(() => {
+    // getSettings swallows load failures (console.error, then falls
+    // through with defaults), which has repeatedly let a stale
+    // MigrationUtils mock pass this suite without ever completing the
+    // load path; fail loudly instead
+    expect(
+        errorSpy.mock.calls.filter(
+            ([message]) => message === 'Could not load settings'
+        )
+    ).toEqual([]);
+    errorSpy.mockRestore();
     logSpy.mockRestore();
 });
 
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -32,7 +32,12 @@ jest.mock('../utils/LocaleUtils', () => ({
 jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
-    runSettingsMigrations: jest.fn().mockResolvedValue(undefined),
+    // getSettings adopts the return value (the queue's authoritative
+    // object when consolidation routes through updateSettings), so the
+    // stub must hand the settings back rather than resolve undefined
+    runSettingsMigrations: jest.fn().mockImplementation(
+        async (settings: any) => settings
+    ),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
     storageMigrationV2: jest.fn().mockResolvedValue(undefined)
 }));
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -25,6 +25,13 @@ import LoginRequest from '../models/LoginRequest';
 const LEGACY_STORAGE_KEY = 'zeus-settings';
 export const STORAGE_KEY = 'zeus-settings-v2';
 
+// Version stamp for the CONTENT of the settings blob (the storage key's
+// -v2 suffix versions the key namespace, not the content). Blobs stamped
+// with the current version skip all one-shot settings migrations at load
+// (#4470). Bump when adding a migration — see
+// MigrationUtils.runSettingsMigrations.
+export const SETTINGS_VERSION = 1;
+
 export const LEGACY_CURRENCY_CODES_KEY = 'currency-codes';
 export const CURRENCY_CODES_KEY = 'zeus-currency-codes';
 export const FAVORITE_CURRENCIES_KEY = 'zeus-favorite-currencies';
@@ -188,6 +195,8 @@ interface SwapsSettings {
 }
 
 export interface Settings {
+    // see SETTINGS_VERSION; absent on blobs predating the stamp
+    settingsVersion?: number;
     nodes?: Array<Node>;
     selectedNode?: number;
     justDeletedWallet?: boolean;
@@ -1486,6 +1495,7 @@ export const DEFAULT_SLIDE_TO_PAY_THRESHOLD = 10000;
 
 export default class SettingsStore {
     @observable settings: Settings = {
+        settingsVersion: SETTINGS_VERSION,
         privacy: {
             defaultBlockExplorer: 'mempool.space',
             customBlockExplorer: '',
@@ -1971,15 +1981,8 @@ export default class SettingsStore {
                 console.log('attempting to load modern settings');
                 const parsedSettings = JSON.parse(modernSettings);
                 this.settings = parsedSettings;
-                await MigrationsUtils.migrateRgsDefaultsToV2(parsedSettings);
-                await MigrationsUtils.migrateSwapHostsToBoltz(parsedSettings);
+                await MigrationsUtils.runSettingsMigrations(parsedSettings);
                 await MigrationsUtils.migrateRetiredSwapHosts(parsedSettings);
-                await MigrationsUtils.migrateInvoiceExpiryDisplay(
-                    parsedSettings
-                );
-                await MigrationsUtils.migrateOlympusHostsToZeusLsp(
-                    parsedSettings
-                );
                 this.settings = parsedSettings;
             } else {
                 console.log('attempting to load legacy settings');
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1399,7 +1399,7 @@ export const LEGACY_ZEUS_SWAP_HOST_MAINNET = 'https://swaps.zeuslsp.com/api/v2';
 export const LEGACY_ZEUS_SWAP_HOST_TESTNET =
     'https://testnet-swaps.zeuslsp.com/api/v2';
 // Swap providers that have shut down; users pinned to one are moved
-// back to the default host by MigrationUtils.migrateRetiredSwapHosts
+// back to the default host by MigrationUtils.applyRetiredSwapHosts
 export const RETIRED_SWAP_HOSTS_MAINNET = ['https://boltz-api.eldamar.icu/v2'];
 
 export const DEFAULT_NOSTR_RELAYS_2023 = [
@@ -1982,7 +1982,6 @@ export default class SettingsStore {
                 const parsedSettings = JSON.parse(modernSettings);
                 this.settings = parsedSettings;
                 await MigrationsUtils.runSettingsMigrations(parsedSettings);
-                await MigrationsUtils.migrateRetiredSwapHosts(parsedSettings);
                 this.settings = parsedSettings;
             } else {
                 console.log('attempting to load legacy settings');
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -8,6 +8,7 @@ import BackendUtils from '../utils/BackendUtils';
 import { getSupportedBiometryType } from '../utils/BiometricUtils';
 import { localeString } from '../utils/LocaleUtils';
 import MigrationsUtils from '../utils/MigrationUtils';
+import { SETTINGS_VERSION } from '../utils/SettingsVersion';
 import { doTorRequest, RequestMethod } from '../utils/TorUtils';
 import {
     DEFAULT_SCORER_URL,
@@ -26,11 +27,10 @@ const LEGACY_STORAGE_KEY = 'zeus-settings';
 export const STORAGE_KEY = 'zeus-settings-v2';
 
 // Version stamp for the CONTENT of the settings blob (the storage key's
-// -v2 suffix versions the key namespace, not the content). Blobs stamped
-// with the current version skip all one-shot settings migrations at load
-// (#4470). Bump when adding a migration — see
-// MigrationUtils.runSettingsMigrations.
-export const SETTINGS_VERSION = 1;
+// -v2 suffix versions the key namespace, not the content). Defined in a
+// dependency-free module so version-gating tests exercise the real
+// value; re-exported here for consumers of the settings schema.
+export { SETTINGS_VERSION };
 
 export const LEGACY_CURRENCY_CODES_KEY = 'currency-codes';
 export const CURRENCY_CODES_KEY = 'zeus-currency-codes';
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1981,8 +1981,12 @@ export default class SettingsStore {
                 console.log('attempting to load modern settings');
                 const parsedSettings = JSON.parse(modernSettings);
                 this.settings = parsedSettings;
-                await MigrationsUtils.runSettingsMigrations(parsedSettings);
-                this.settings = parsedSettings;
+                // when consolidation routes through the updateSettings
+                // queue the authoritative object is the queue's, not this
+                // call's snapshot, so adopt the return value
+                this.settings = await MigrationsUtils.runSettingsMigrations(
+                    parsedSettings
+                );
             } else {
                 console.log('attempting to load legacy settings');
 
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -1969,7 +1969,14 @@ export default class SettingsStore {
         }
     };
 
-    public getSettings = async (silentUpdate: boolean = false) => {
+    // fromQueue is true only for the call inside applySettingsUpdate's
+    // critical section; runSettingsMigrations branches on it to decide
+    // whether a consolidation may piggyback on that update's write or
+    // must enqueue one.
+    public getSettings = async (
+        silentUpdate: boolean = false,
+        fromQueue: boolean = false
+    ) => {
         if (!silentUpdate) this.loading = true;
         try {
             await MigrationsUtils.keychainCloudSyncMigration();
@@ -1985,7 +1992,8 @@ export default class SettingsStore {
                 // queue the authoritative object is the queue's, not this
                 // call's snapshot, so adopt the return value
                 this.settings = await MigrationsUtils.runSettingsMigrations(
-                    parsedSettings
+                    parsedSettings,
+                    fromQueue
                 );
             } else {
                 console.log('attempting to load legacy settings');
@@ -2061,7 +2069,9 @@ export default class SettingsStore {
     ) => {
         this.settingsUpdateInProgress = true;
         try {
-            const existingSettings = await this.getSettings();
+            // fromQueue: this is the queue's critical section, so a
+            // pending consolidation folds into this update's write
+            const existingSettings = await this.getSettings(false, true);
             // Functional updates read the settings inside the critical
             // section, so callers whose new value depends on the current
             // one (delete node X from the array) cannot act on a snapshot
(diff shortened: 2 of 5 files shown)


kaloudis:
Rebased onto master (`1073c8e92`). `#4475` landed first, so this took the branch that PR's body called out:

- `applyRgsDefaultToZeus` is now **`applyRgsDefaultsToV2`** and carries master's per-network v1 -> v2 URL mapping verbatim (including the testnet path asymmetry, `/testnet/v2/snapshot`). Cross-network URLs stay untouched.
- The honored retired flag is **`rgs-defaults-v2`**, not `rgs-default-zeus`. An install that already ran `#4475`'s migration has that flag set, so consolidation correctly skips the rewrite; an install predating it gets the rewrite in the same single write as the stamp. The superseded `rgs-default-zeus` flag is still deliberately not consulted, matching master.
- `SETTINGS_VERSION` stays at 1: the RGS v2 rewrite is folded into version 1 rather than becoming version 2, since it has not shipped in a release yet.

Also picked up from master and left intact: `purgeRescueKeyFiles` (`#4373`) is unchanged and still runs ahead of the blob read; `setSettings` now only commits to memory when the write lands (`58ee40597`), which the single-write path inherits for free.

Tests: master's `migrateRgsDefaultsToV2` suite is folded in. Its domain cases (both mainnet defaults, testnet default, no cross-network application, custom URLs untouched, unset untouched, idempotency) became pure-mutation tests on `applyRgsDefaultsToV2`; its flag-gating and single-persist cases moved to the `runSettingsMigrations` orchestrator suite, which gained a case asserting `rgs-defaults-v2` gates the rewrite while `rgs-default-zeus` does not. `stores/SettingsStore.test.ts`'s MigrationUtils mock was updated to `runSettingsMigrations` (it still named `migrateRgsDefaultToZeus`, stale since `#4475`, which meant `getSettings()` was silently throwing and swallowing under test).

`yarn verify` green: tsc, lint, prettier clean; 1457/1458 tests pass, the one failure being the known PhotoUtils relative-path artifact of running in a nested worktree.

Merge blockers from the description are otherwise unchanged: maintainer sign-off on the migration plan, and the two-platform upgrade testing. Item 3, sequencing with `#4475`, is now resolved.

utils/MigrationUtils.ts
```diff
@@ -463,21 +445,77 @@ class MigrationsUtils {
...
+    // ------------------------------------------------------------------
+    public async runSettingsMigrations(settings: any) {
+        if ((settings?.settingsVersion ?? 0) >= SETTINGS_VERSION) {
+            return settings;
+        }
+
+        await this.applySettingsMigrations(settings);
+
+        // single write: migrated values and stamp land atomically
+        await settingsStore.setSettings(settings);
```
shubhamkmr04: This write runs outside `updateSettingsQueue`, and `settings` is a snapshot taken several awaits earlier (the blob read, plus the four flag reads in `applySettingsMigrations`). That is the read-merge-write hazard the queue comment in `SettingsStore` describes.

Concretely, on the first boot after upgrade: `IntroSplash` calls `getSettings()` fire-and-forget while `Wallet.componentDidMount` does `await updateSettings({ supportedBiometryType })`. If that update lands during the flag reads, this write puts the pre-update blob back — both on disk and in `this.settings`.

On master a fully-migrated user's `getSettings()` writes nothing (each migration is flag-guarded and `if (changed)`-guarded), so the window is new for that cohort. The stamp does need persisting, so the fix isn't to skip the write — it should go through the same queue, or re-read and merge right before writing. The mutations are value-conditional and idempotent, so re-applying them onto fresh state is safe.

stores/SettingsStore.test.ts
```diff
@@ -32,8 +32,7 @@ jest.mock('../utils/LocaleUtils', () => ({
 jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
-    migrateRgsDefaultToZeus: jest.fn().mockResolvedValue(undefined),
-    migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
+    runSettingsMigrations: jest.fn().mockResolvedValue(undefined),
```
shubhamkmr04: This mock is missing `migrateRetiredSwapHosts`, so every `getSettings()` throws `TypeError: _MigrationUtils.default.migrateRetiredSwapHosts is not a function` and the `catch` swallows it. All 4 tests pass while never reaching the modern load path or `updateNodeProperties()`.

Pre-existing (master's mock is missing it too), but worth fixing here since the PR edits these exact lines — and because with `runSettingsMigrations` stubbed to a no-op, this suite cannot exercise the new consolidation write at all.

stores/SettingsStore.ts
```diff
@@ -1948,15 +1958,8 @@ export default class SettingsStore {
                 console.log('attempting to load modern settings');
                 const parsedSettings = JSON.parse(modernSettings);
                 this.settings = parsedSettings;
-                await MigrationsUtils.migrateRgsDefaultsToV2(parsedSettings);
-                await MigrationsUtils.migrateSwapHostsToBoltz(parsedSettings);
+                await MigrationsUtils.runSettingsMigrations(parsedSettings);
                 await MigrationsUtils.migrateRetiredSwapHosts(parsedSettings);
```
shubhamkmr04: `migrateRetiredSwapHosts` is the one shared migration left out of the stamp, so the modern path still reads its flag on every call — plus 2 reads in `keychainCloudSyncMigration` and 1 in `purgeRescueKeyFiles`. That is 4 serialized keystore reads per `getSettings()`, or 12+ per boot in steady state: the same figure the description says goes to 0.

Its mutation is value-conditional like the other four, so it folds into `applySettingsMigrations` cleanly. Worth doing, since that number is what the storage gate is being asked to sign off on.

utils/MigrationUtils.test.ts
```diff
@@ -30,6 +30,7 @@ jest.mock('../stores/LSPStore', () => ({}));
 jest.mock('../utils/BackendUtils', () => ({}));
 
 jest.mock('../stores/SettingsStore', () => ({
+    SETTINGS_VERSION: 1,
```
shubhamkmr04: Hardcoding the version decouples these tests from the real constant. After the bump to 2 that #4475 plans, the orchestrator suite keeps asserting v1 gating and stays green even if a v2 migration is wrongly gated and never runs for existing v1 users.

Mocking the module is unavoidable given the RN deps, but `SETTINGS_VERSION` could live in a dependency-free constants file that both the store and this test import for real. Same coupling on `settingsVersion: 1` at line 159.

shubhamkmr04 (commented):
Read through the refactor and ran both suites (48/48 and 4/4 green).

The extraction holds up: the four mutations are semantically identical to the originals (guard-clause inversion and `return changed` only), no migration is dropped from either path, the `applySwapHostsToBoltz` → `migrateRetiredSwapHosts` ordering is preserved, and moving the expiry fix to the end of the legacy path is safe — the blocks it now runs after (POS, neutrino peers, payments) don't touch `settings.invoices`. No stale callers of the renamed methods.

Four notes inline. The first is a blocker; the rest are worth folding in while the branch is open.

utils/MigrationUtils.ts
kaloudis: Fixed in 73ee02ade. runSettingsMigrations now branches on settingsStore.settingsUpdateInProgress: inside the queue's critical section (applySettingsUpdate -> getSettings -> here) it persists directly, since it is already serialized with every other update there, and enqueueing would deadlock behind the very updateSettings task awaiting it. Outside the queue it no longer touches the stale snapshot at all: it enqueues a no-op updateSettings({}), whose critical section re-runs getSettings on fresh state and re-enters the direct branch. getSettings adopts the return value so memory tracks the queue's authoritative object rather than the pre-queue snapshot. I went with the queue over re-read-and-merge because the latter still leaves an in-flight window between the re-read and the native write. Two new orchestrator cases pin the routing contract, including one reproducing the concurrent-update clobber.

One nit on the trigger: IntroSplash's getSettings only fires while that screen has focus, so for an upgrading user with a wallet the likelier overlap is Wallet.getSettingsAndNavigate's direct getSettings racing componentDidMount's biometry update. Same mechanism and same fix either way.

stores/SettingsStore.ts
kaloudis: Folded in ebab5ce17. applyRetiredSwapHosts is now the fifth pure mutation, ordered after applySwapHostsToBoltz as on master, and swap-hosts-retired-eldamar joins the consolidation flag reads. The flag is honored so an install that already ran #4542's migration is not re-migrated over a since-changed value, and SETTINGS_VERSION stays 1 since #4542 has not shipped in a release. The legacy path picks it up automatically through applySettingsMigrations. Steady state on the modern path is back to zero migration keystore reads.

On the arithmetic: I count 3 residual reads per getSettings before this fix rather than 4; keychainCloudSyncMigration is a single flag read once migrated. The remaining two (cloud-sync flag and rescue-file purge flag) cannot be gated by the blob stamp because they must run before the blob is read, but they could be latched in memory per session if that residual matters. I left them alone here to keep this PR scoped to the settings-blob migrations.

stores/SettingsStore.test.ts
kaloudis: The literal fix fell out of the fold (thread above): after ebab5ce17 the store no longer calls migrateRetiredSwapHosts, so the mock's method list is complete again. The durable problem you identified, the catch swallowing mock drift while every test stays green, now has a backstop in 2c7e2473c: the suite spies console.error and fails any test where 'Could not load settings' was logged. Verified it trips by renaming a mocked method, which now fails all 4 tests instead of none. The consolidation write itself is exercised in the runSettingsMigrations orchestrator suite, including the new queue-routing cases added with the blocker fix.

utils/MigrationUtils.test.ts
kaloudis: Done in 2c7e2473c. SETTINGS_VERSION now lives in dependency-free utils/SettingsVersion.ts and is re-exported from SettingsStore for schema consumers. MigrationUtils and its test import it directly, so the suite's stores/SettingsStore mock no longer shadows it, and the settingsVersion literals (the line 159 seed included) reference the real constant. A future bump therefore flows into the gating tests automatically. Minor note: #4475 is already merged and its rewrite is folded into version 1, so the first real bump will come with whichever migration lands next.

kaloudis:
Thanks for the thorough round. All four notes were on the mark; pushed as a rebase onto master (`05d721877`) plus three commits so each change reviews on its own:

- `76995a28b` is the original refactor rebased. Master had grown `migrateRetiredSwapHosts` (#4542) inside the block this PR rewrites, so the branch genuinely conflicted despite what the mergeable badge said; the resolution keeps master's call intact at this commit.
- `ebab5ce17` folds it: `applyRetiredSwapHosts` becomes the fifth pure mutation and `swap-hosts-retired-eldamar` joins the consolidation flag reads (honored, since #4542 may ship before this merges; `SETTINGS_VERSION` stays 1). Modern path is back to zero migration keystore reads in steady state.
- `73ee02ade` fixes the blocker. The consolidation write now runs inside `updateSettingsQueue`: direct persist when already in the critical section (enqueueing there would deadlock behind the task awaiting it), otherwise a no-op `updateSettings({})` whose critical section redoes the consolidation on fresh state. `getSettings` adopts the return value so memory follows the queue's authoritative object. New orchestrator cases cover both branches plus the concurrent-update clobber.
- `2c7e2473c` un-mocks `SETTINGS_VERSION` (now in dependency-free `utils/SettingsVersion.ts`, re-exported from `SettingsStore`) and adds the swallowed-load guard to `stores/SettingsStore.test.ts`: any test that logs `Could not load settings` now fails, which turns the silent-mock-drift trap you hit into 4 loud failures.

`yarn verify`: tsc, lint, prettier clean; 1632/1633 tests pass, the one failure being the known PhotoUtils relative-path artifact of running in a nested worktree.

shubhamkmr04 (commented):
Re-reviewed at 2c7e247. Three of the four are cleanly fixed:

- **Retired swap hosts folded in** — now gated inside the same `Promise.all` batch, ordering after `applySwapHostsToBoltz` preserved, and the description's figure corrected to the honest "5 flag reads → 0".
- **Test mock** — the stub returns settings, and the suite now fails when `Could not load settings` is logged. That guard is the right call; it's the trap that bit twice.
- **Hardcoded version** — `utils/SettingsVersion.ts` is exactly the decoupling needed, and the suite imports the real constant.

Both suites green here: 54/54.

The queue-routing fix doesn't hold, though — the guard reads a global flag that doesn't mean what the branch needs it to mean, so the lost-update window survives in precisely the case it was meant to close. Detail inline, with a test against this head that demonstrates it.

One aside, not for this PR: `keychainCloudSyncMigration` (2 reads) and `purgeRescueKeyFiles` (1) still run per `getSettings()`, so the boot is at ~9 keystore touches rather than 0. Out of scope here, but worth knowing the stamp doesn't get the load path all the way down.

utils/MigrationUtils.ts
```diff
@@ -463,19 +446,92 @@ class MigrationsUtils {
...
+    // Adding a future migration: bump SETTINGS_VERSION, gate the new
+    // mutation on the blob's prior version (settings.settingsVersion ??
+    // 0), and leave the existing blocks untouched.
+    // ------------------------------------------------------------------
+    public async runSettingsMigrations(settings: any) {
+        if ((settings?.settingsVersion ?? 0) >= SETTINGS_VERSION) {
+            return settings;
+        }
+
+        if (settingsStore.settingsUpdateInProgress) {
```
shubhamkmr04: `settingsUpdateInProgress` is a global flag set across the whole of `applySettingsUpdate`. It means "some update is in flight", not "I am on the queue's call stack".

So an outside `getSettings()` running while an **unrelated** update is in flight takes this branch and does the unserialized `setSettings(staleSnapshot)` the commit set out to remove. The guard picks the unsafe path exactly when a concurrent update exists, and only enqueues when there is nothing to clobber.

This passes against this head:

```ts
it('unrelated in-flight update -> outside caller writes stale snapshot', async () => {
    EncryptedStorage.getItem.mockResolvedValue(null);
    settingsStore.settingsUpdateInProgress = true; // not us; we are outside
    const stale: any = {
        swaps: { hostMainnet: 'https://boltz-api.eldamar.icu/v2' }
    };

    await MigrationUtils.runSettingsMigrations(stale);

    expect(settingsStore.updateSettings).not.toHaveBeenCalled();
    expect(settingsStore.setSettings).toHaveBeenCalledWith(stale);
});
```

Passing the fact explicitly avoids inferring it from shared state — there's only one call site inside the critical section:

```ts
// applySettingsUpdate
await this.getSettings(false, true); // fromQueue
// getSettings -> runSettingsMigrations(parsedSettings, fromQueue)
// runSettingsMigrations(settings, inQueue = false)
if (inQueue) { /* persist directly */ }
return await settingsStore.updateSettings({});
```

Minor: the queued re-entry makes the consolidation boot do two blob writes and re-run `keychainCloudSyncMigration` / `purgeRescueKeyFiles`.

utils/MigrationUtils.test.ts
```diff
@@ -757,363 +549,455 @@ describe('MigrationUtils', () => {
...
-                'rgs-defaults-v2',
-                'true'
-            );
+            expect(settingsStore.updateSettings).not.toHaveBeenCalled();
         });
 
-        it('leaves unset values unset so runtime falls back to new defaults', async () => {
-            EncryptedStorage.getItem.mockResolvedValue(null);
+        it('routes the write through the updateSettings queue when outside it', async () => {
+            settingsStore.settingsUpdateInProgress = false;
```
shubhamkmr04: All three routing cases set `settingsUpdateInProgress = false` first, so they only cover the quiet case — the one where the branch above is safe either way. Hence green.

The inverse is the one worth pinning: flag `true`, called from outside the queue.

utils/MigrationUtils.ts
kaloudis: Good catch, confirmed: the flag says an update is in flight somewhere, not that we are on the queue's call stack, so the guard picked the direct write exactly when there was a concurrent update to clobber. Fixed in 1baa3de99 with your suggested shape: applySettingsUpdate passes fromQueue through getSettings into runSettingsMigrations(settings, inQueue), and settingsUpdateInProgress is no longer consulted. Outside the queue the snapshot is never touched regardless of flag state.

Your minor fell out of the same commit: the in-queue branch no longer writes at all. applySettingsUpdate persists right after its getSettings returns and the top-level stamp rides that spread, so the migrated values and stamp land in that update's single serialized write. Consolidation is back to one blob write on both routes. The residual cost of the enqueued re-entry is the keychainCloudSyncMigration and purgeRescueKeyFiles re-run, 2 flag reads once on the consolidation boot only.

utils/MigrationUtils.test.ts
kaloudis: Added in 1baa3de99 as 'enqueues even while an unrelated update is in flight': flag set true, called from outside, asserting no direct write, no flag reads, and the enqueued no-op update. Verified it fails against the previous guard (setSettings receives the stale snapshot) before passing with the fix. The consolidation cases now pass inQueue explicitly and pin the inverse contract too: inside the queue, mutate and stamp in memory with no write and no enqueue.

shubhamkmr04 (approved):
tACK 1baa3de99 - Device tested on android and iOS, works correctly

kaloudis:
Round 3 (5fe9f95) hardens the future-migration contract. The five v1 blocks are now gated on `priorVersion < 1` in addition to their retired flags: without the gate, every future `SETTINGS_VERSION` bump would send v1-stamped blobs back through them, and since the flags are never written under the stamp system, they would re-apply over old-default-looking values a user deliberately restored after stamping (at-least-once where the flag system gave exactly-once). The gate also skips the five flag reads for stamped blobs.

A new source-scan test pins the highest gate to `SETTINGS_VERSION`, so a gate added without a bump (silently unreachable for existing users) or a bump without a gated block fails CI. A companion behavior test proves a v1-stamped blob passes through `applySettingsMigrations` untouched with zero keystore reads.

The orchestrator comment now states the version-freeze rule: a version is frozen once any public build (alphas included) ships stamping it; migrations added after that take a fresh bump. PR body's Future migrations section updated to match.

ajaysehwal (approved):
tACK

PR #4488, opened 2026-08-27
Cashu: don't consume tokens before the self-custody sweep is ready

kaloudis:
# Description

Fixes the v13.2.0 reports of "Receive to self-custody" failing with a red "Token has already been spent" error while the token's value shows up in the ecash mint balance, leading users to believe their funds were lost.

## Root cause

`CashuStore.claimToken(toSelfCustody)` ran a non-atomic four-step flow that consumed the token as its very first step:

1. `receiveTokenCDK` (swaps the token into the wallet's own CDK balance at the mint; irreversible)
2. create the sweep invoice on the user's node
3. create a melt quote (fee probe)
4. melt to pay the invoice

Any failure in steps 2-4 (node not ready, mint can't route to a private mobile node, network blips) surfaced as a claim failure, with the claim buttons still enabled. A retry then re-ran step 1 and could only fail with "token already spent" against the wallet's own successful receive from the first attempt. The funds were safe in the mint balance the whole time, but nothing told the user that.

Additionally, the melt amount was computed from the token's face value (`decoded.getAmount`). On mints that charge input fees, the receive credits less than face value, so the melt could fail with insufficient funds every time.

Before the CDK migration (v12.x, cashu-ts) the invoice and melt quote were created before the token proofs were touched, so early failures left the token retryable. This restores that ordering under CDK.

## Changes

- `stores/CashuStore.ts`: prepare the sweep invoice and melt fee probe before receiving the token, so failures up to the point of no return leave the token unclaimed and retryable. The fee-exceeds-amount check also now happens before the receive.
- Size the sweep from the amount actually credited by the receive instead of token face value, fixing the melt on mints with input fees.
- If the sweep fails after the token has been received, return a partial-success warning ("redeemed to your ecash balance at {mint}, sweep did not complete, funds are safe") instead of a retryable error. The `CashuToken` view surfaces it via the existing `WarningMessage` banner and hides the claim buttons, so the "already spent" retry loop can no longer happen.
- `stores/CashuStore.test.ts` (new): regression tests for the ordering, the received-amount sizing, and the partial-success path. 5 of the 7 tests fail against the previous implementation.
- `locales/en.json`: new `stores.CashuStore.selfCustodySweepFailed` string.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

## Device test plan (pending)

1. Happy path: scan a token, "Receive to self-custody", verify LN balance credit and activity entry.
2. Fee-charging mint: same, verify melt succeeds with fee-adjusted amount.
3. Sweep failure: claim with channels that cannot receive (or airplane-mode the node after invoice creation); verify the warning banner names the mint, claim buttons are hidden, and the mint balance shows the funds.
4. Genuinely spent token: verify the "already spent" error still appears.

Follow-ups (out of scope): a manual "sweep mint to self-custody" entry point so stranded balances have a one-tap exit, and an own-wallet check on "already spent" errors after app restart.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -2359,6 +2359,7 @@
     "stores.CashuStore.noPaymentRequestAvailable": "No payment request available",
     "stores.CashuStore.invalidMultimintPaymentAmount": "Invalid multimint payment amount. Please use a whole sat amount.",
     "stores.CashuStore.feeExceedsAmt": "Melt fees match or exceed token amount",
+    "stores.CashuStore.selfCustodySweepFailed": "The token was redeemed to your ecash balance at {{mintName}}, but the sweep to self-custody did not complete. Your funds are safe in your ecash balance.",
     "stores.CashuStore.autoSweep": "Cashu auto-sweep",
     "stores.CashuStore.notProperlyConfigured": "Cashu wallet is not properly configured. Please add at least one mint.",
     "stores.CashuStore.errorDiscoveringMints": "Failed to discover mints from Nostr relays",
--- a/stores/CashuStore.test.ts
+++ b/stores/CashuStore.test.ts
@@ -1,0 +1,270 @@
+jest.mock('../stores/Stores', () => ({
+    activityStore: {},
+    connectivityStore: {
+        isOffline: false,
+        start: jest.fn(),
+        onReconnect: jest.fn()
+    }
+}));
+
+jest.mock('../utils/LocaleUtils', () => ({
+    localeString: (key: string) => key
+}));
+
+jest.mock('../utils/BackendUtils', () => ({
+    __esModule: true,
+    default: {}
+}));
+
+jest.mock('../utils/MigrationUtils', () => ({
+    __esModule: true,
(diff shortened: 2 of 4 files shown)


kaloudis:
Follow-up filed as #4491: switch this flow to cdk-ffi's prepareMeltToken (already in the shipped 0.17.4 bindings, needs RN bridge exposure) so the token is only consumed when the sweep payment actually executes, plus a NUT-07 checkProofsSpent pre-check on scanned tokens. This PR stays the minimal JS-only fix for the point release.

PR #4404, opened 2026-08-13
ui: disable pay on zero balance, grey out disabled button icons

kaloudis:
# Description

Cleans up the disabled states of the pay buttons on the Lightning and Cashu payment request views.

**`components/Button.tsx`**
- When a `Button` is disabled, its icon is now rendered in the same grey `@rneui/themed` uses for the disabled title text (`hsl(208, 8%, 63%)`, i.e. `color(theme.colors.disabled).darken(0.3)`). Previously the icon kept its full-color styling while the label and background greyed out. Applies to every disabled `Button` with an icon app-wide.

**`views/PaymentRequest.tsx`**
- The "Pay this invoice" button and the slide-to-pay variant are now disabled when the Lightning balance is 0. Previously the zero-balance case only kept the button disabled on embedded-lnd (via `checkIfLndReady` never marking Lightning ready); every other backend enabled the button immediately, even while the "no Lightning balance" warning was showing above it.
- The send icon is now always passed to the button, so it shows greyed out while disabled instead of disappearing while Lightning is not ready to send.

**`views/Cashu/CashuPaymentRequest.tsx`**
- Same zero-balance gating for consistency: with a total mint balance of 0, the SwipeButton is replaced by the greyed static button (above the slide-to-pay threshold) and the "Pay this invoice" button is disabled (below it). Previously the view showed the no-balance warning but still allowed triggering the payment.

No new strings, no dependency changes, no storage changes.

This pull request is categorized as a:

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

I have tested this PR on the following platforms (please specify OS version and phone model/VM):

- [ ] Android
- [x] iOS

I have tested this PR with the following types of nodes (please specify node version and API version where appropriate):

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] LND (Lightning Node Connect)
[...]

Issue #4286, opened 2026-07-28
NWC: add a first-class lookupPayment backend method instead of scanning the payment list


warioishere:
### Summary

ZEUS has no backend method for "get the payment with this hash". Every feature
that needs one currently fetches a page of the payment list and scans it in JS.
Proposing a first-class `lookupPayment` across the seven backends.

### Why this exists

NIP-47 `lookup_invoice` is defined over both directions — `"incoming"` for
invoices, `"outgoing"` for payments — and must resolve by `payment_hash` or
`invoice`. `BackendUtils` has `lookupInvoice`, but nothing for the payment side,
so #4273 answers the payment half by fetching up to 100 payments and matching in
memory. That works, but it is the wrong shape for a by-hash question, and it
scales with history rather than with the request.

The same gap shows up elsewhere: `getActivities` reconciles pending NWC payments
by fetching the payment list and searching it (`getPaymentsForPendingPayInvoiceRefresh`),
and `TransactionsStore` resolves a just-sent payment the same way.

### What each backend could actually do

Verified against the vendored protos in this repo and the Core Lightning schemas,
not assumed:

| Backend | Targeted lookup available |
|---|---|
| LND (REST), Embedded LND, LNC | **No request/response option.** `ListPaymentsRequest` in `proto/lightning.proto` has `include_incomplete`, `index_offset`, `max_payments`, `reversed` and creation-date bounds — no payment-hash filter. The hash-targeted call is `TrackPaymentV2`, which returns `stream lnrpc.Payment` (`proto/routerrpc/router.proto`). Usable, but it means opening a stream for a one-shot query — WebSocket for LND REST, `sendStreamCommand` plus the event emitter for embedded. |
| CLNRest | **Yes, directly.** `listsendpays` accepts `payment_hash` and `bolt11` as filters (`doc/schemas/listsendpays.json` in ElementsProject/lightning). Cheaper and more correct than the current `/v1/sql` scan, which is also hardcoded to `limit 150`. |
| LDK Node | **Yes, free.** `listPayments()` is in memory; filtering by hash and `direction` is a local operation. |
| LndHub | **No.** `/gettxs` returns the transaction list with no filter; `/checkpayment/:payment_hash` answers only `{paid: boolean}` and carries no amount, invoice or timestamps. A scan stays necessary here. |
| NostrWalletConnect | **Yes.** `lookup_invoice` over NIP-47 resolves either direction, which is what the remote wallet is for. |

[...]

Issue #2577, opened 2024-11-26
SendPayment: Improve handling over in-transit payments


kaloudis:
Per Leo in Telegram:
```
One problem I've run into with Zeus is that a payment I'm making is being "held" for whatever reason by the recipient node, most likely by accident. In the current UI, a held payment is indistinguishable from a failed payment, so in my previous cases (and again today), i just try making another payment, which is again held, and another payment, which is again held. All three payments eventually go through and I'm stuck with three double-scoop ice cream cones.
Some error/notification that's different from success/failure might be useful here
```

We have some code to explicitly catch Zaplocker payments and display an appropriate message, but not in-transit messages in general (despite highlighting them on the main wallet view).

Should be able to make some changes here: https://github.com/ZeusLN/zeus/blob/master/stores/TransactionsStore.ts#L566

kaloudis:
This is now being addressed by #4492 (with #4495 stacked on top).

The failure mode described here (a held payment surfacing as a failure, prompting retries that all eventually settle) comes from the send path treating client-side timeouts and non-terminal (`IN_FLIGHT`) results as terminal. #4317 documents the mechanics, including why this also creates a keysend double-pay window.

What the PRs change:
- #4492: client-side timeouts become outcome-unknown instead of failures. On lookup-capable backends (LND REST, embedded LND, LNC) the app shows the in-transit screen and polls a new `lookupPayment` method until the payment reaches `SUCCEEDED`/`FAILED`, then flips the screen to the real outcome. The send guard is held through the ambiguity, so the retry-and-double-pay loop described here is closed rather than just relabeled.
- #4495: extends `lookupPayment` to CLNRest, LDK Node, and NWC.

Related: #1865 (hold invoice payments stuck on "sending" on embedded LND) is the same underlying gap and is also linked from #4492.

PR #4492, opened 2026-08-27
Track in-flight payments to a terminal state before releasing the send guard

### LINKED ISSUE #4317: Track in-flight payments to a terminal state instead of releasing the send guard on timers
## Background

PR #4301 added a `paymentInFlight` guard in `TransactionsStore.sendPaymentInternal` to prevent accidental double-submission of Lightning payments (double-tap or re-fired swipe). Keysend is the dangerous case: each invocation generates a fresh preimage, so duplicate submissions have different payment hashes and defeat every backend's same-hash duplicate protection.

Review discussion: https://github.com/ZeusLN/zeus/pull/4301#discussion_r3701653514

The guard is currently cleared by `handlePayment`, `handlePaymentError`, `reset()`, and a timed backstop (`timeout_seconds + 60s`). The reviewer correctly pointed out that `handlePayment` clears the flag even when the payment result is non-terminal (`IN_FLIGHT`).

## The gap

There is a real window where the guard releases while an HTLC may still settle:

- **LND REST**: `restReq` parses the last ndjson line of the `/v2/router/send` stream. If the stream ends while HTLCs are pending, the last update is `IN_FLIGHT`. This is the state `SendingLightning.inTransit()` and `Payment.isInTransit` already model. `handlePayment` clears `paymentInFlight`, send buttons re-enable, and a keysend retry (fresh preimage, new hash) can double-pay if the original HTLC later succeeds.
- **Client-side timeouts**: the LND REST `forcedTimeout` race resolves with `{ payment_error: 'payment timed out' }` (no status, no htlcs), and `restReq`'s timeout rejection lands in `handlePaymentError`. Both clear the flag while the payment may still be in flight on the node. Any status-based check in `handlePayment` misses these paths entirely.
- **LNC**: the view-level subscriptions filter out `IN_FLIGHT` events before calling `handlePayment`, so the flag is held until the backstop, but nothing ever observes the actual terminal state either.

Simply holding the flag on an `IN_FLIGHT` result (as suggested in the review) is not sufficient: nothing event-driven remains to clear it once the request/stream has ended, so the timed backstop becomes the only release. That converts "retry possible immediately" into "retry possible in ~60 seconds," while a stuck HTLC can pend for hours (until CLTV expiry). It also locks the user out of all unrelated sends in the meantime, since `paymentInFlight` gates the whole send path.

## Proposed fix

When a payment dispatch ends in a non-terminal or ambiguous state (an `IN_FLIGHT` result, or a client-side timeout where the outcome is unknown), keep `paymentInFlight` set and actively track the payment to a terminal state instead of relying on timers:

- LND-based backends: poll `trackPaymentV2` / lookup by payment hash until `SUCCEEDED` or `FAILED`, then clear the guard (and update the in-transit UI state accordingly).
- Other backends: use the closest equivalent payment lookup where available; fall back to the existing backstop where no lookup exists.
- Keep the timed backstop only as a last resort for backends or states where the terminal outcome genuinely cannot be observed.

This closes the double-pay window instead of shrinking it, and gives `SendingLightning`'s in-transit screen a real resolution path (it could flip to success/failure when the tracked payment settles).

## Scope note

PR #4301 intentionally targets accidental double-submission. Retrying from the in-transit screen requires deliberately navigating back and re-initiating, and invoice retries remain protected by node-side same-hash dedup. The remaining exposure is deliberate keysend retry while a prior attempt is still in flight, which is a pre-existing condition that this issue tracks.

kaloudis:
# Description

Relates to issue: **#4317** (fixes)
Fixes issue: **#2577** (a held payment was indistinguishable from a failed one, so users retried and double-paid; client-side timeouts now surface the in-transit screen and are tracked to their real outcome)
Fixes issue: **#2467** (LNC payment showed a failure with a retry button while it had actually settled, and the retry double-paid; the LNC stream-death timeout marker plus lookup tracking here holds the send guard and resolves the payment to its real outcome)
Relates to issue: **#1865** (hold invoice payments on embedded LND stuck on "sending": the timeout marker plus tracking here surfaces the in-transit screen and follows the payment to its real outcome)

PR #4301 added a `paymentInFlight` guard against accidental double-submission, but the guard was released by `handlePayment` on non-terminal (`IN_FLIGHT`) results, by `handlePaymentError` on client-side timeouts and transport errors, and by a blind timed backstop. In all three cases an HTLC may still settle after release, and a keysend retry generates a fresh preimage, so node-side same-hash dedup cannot prevent a double-pay.

This PR holds the guard through ambiguity and actively tracks the payment to a terminal state, as proposed in the issue:

- **New `lookupPayment({ payment_hash })` backend method** on `lnd` (REST), `embedded-lnd`, and `lightning-node-connect`, gated by a new `supportsPaymentLookup` flag. All three scan the newest 50 payments (`include_incomplete`), since LND REST's only per-payment query (`TrackPaymentV2`) is a stream. `LndHub` explicitly overrides the flag to `false` (inheritance hazard); `cln-rest`, `ldk-node`, and `nostr-wallet-connect` are explicit `false` with follow-up notes (CLN's `/v1/listpays` accepts `payment_hash` and could support this later).
- **`TransactionsStore.trackPaymentToTerminal`**: polls `lookupPayment` every 5s until `SUCCEEDED`/`FAILED`, then surfaces the real outcome through `handlePayment` (which releases the guard and flips the SendingLightning screen to the true success/failure state). Exits with the guard released: on a terminal state, when the node answers that it has no record of the payment (dispatch never reached it), after 3 consecutive unobservable lookups, or at a 10-minute ceiling (a stuck HTLC can pend until CLTV expiry; holding the guard that long would lock the user out of all sends).
- **Client-side timeouts are now outcome-unknown, not failures**: the forced-timeout shapes in `LND.payLightningInvoice`, lndmobile's `sendPaymentV2Sync`, and LNC's stream-death resolution gain a `payment_timed_out: true` marker; `restReq`'s `Request timeout` rejection is normalized to the same shape as a safety net (with master's +5s transport headroom the forced timeout normally wins the race). On lookup-capable backends these now show the in-transit screen and get tracked instead of showing "Payment timed out" while the node may still be paying. The `payment_error` string is unchanged, so existing consumers (donation flow, NWC service string-match) behave as before.
- **Rejected dispatches are verified before release**: `handlePaymentError` still surfaces the error immediately, but on lookup-capable backends it verifies against the node before releasing the guard. If the payment turns out to be pending, the UI flips to in-transit; if it settled anyway (e.g. Tor connection dropped after the request went through), it flips to the real success outcome. Genuine pre-dispatch errors release after a single lookup round-trip.
- **The timed backstop remains** for lost completion callbacks and non-lookup backends, but where lookup is available it now attempts observation instead of blindly clearing. Together with the marker on LNC's stream-death shape, this gives LNC dropped-stream payments a real terminal observation path for the guard.
- AMP payments settle under per-attempt hashes, so they remain untrackable and keep the timed backstop. The tracked hash is derived with `deriveExpectedPaymentHash` from `utils/LncPayUtils.ts`, reusing the correlation logic that landed with the LNC payment fix rather than duplicating it.

Review questions:
1. Is the 10-minute tracking ceiling the right balance between double-pay protection and not locking the user out of unrelated sends? (The old backstop was `timeout_seconds + 60s`.)
2. While a tracked payment is unresolved, other sends are silently ignored by the guard, same as before but potentially for longer; the in-transit screen is showing during this time. The disabled-state work in #4404 complements this.

Lays groundwork for #4286 (first-class `lookupPayment` for the NWC service's payment scans).

New unit test suite `stores/TransactionsStore.test.ts` (8 tests) covers: guard held on `IN_FLIGHT` and tracked to `SUCCEEDED`, timeout marker tracked to `FAILED`, transport error replaced by tracked success, release when the node has no record, give-up after consecutive failed lookups, unchanged behavior on non-lookup backends, second send ignored while tracking, and the unscoped LNC view path stopping the tracker.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

Device tests pending: LND REST (clearnet + Tor, keysend + invoice, forced timeout), embedded LND, and LNC dropped-stream scenarios.

### DIFF
--- a/backends/LdkNode.ts
+++ b/backends/LdkNode.ts
@@ -2178,6 +2178,9 @@ export default class LdkNode {
     supportsOnchainReceiving = () => true;
     supportsLightningSends = () => true;
     supportsKeysend = () => true;
+    // follow-up: awaitPaymentCompletion already polls listPayments; a
+    // lookup by hash could resolve payments its timeout leaves pending
+    supportsPaymentLookup = () => false;
     supportsChannelManagement = () => true;
     supportsCircularRebalancing = () => false;
     supportsForceClose = () => true;
--- a/stores/TransactionsStore.test.ts
+++ b/stores/TransactionsStore.test.ts
@@ -1,0 +1,272 @@
+jest.mock('../stores/Stores', () => ({}));
+jest.mock('react-native-blob-util', () => ({}));
+jest.mock('bitcoinjs-lib', () => ({}));
+jest.mock('react-native-randombytes', () => ({
+    randomBytes: (n: number) => require('crypto').randomBytes(n)
+}));
+jest.mock('./SettingsStore', () => ({}));
+jest.mock('./NodeInfoStore', () => ({}));
+jest.mock('./ChannelsStore', () => ({}));
+jest.mock('./BalanceStore', () => ({}));
+jest.mock('./ModalStore', () => ({}));
+jest.mock('../utils/BackendUtils', () => ({
+    __esModule: true,
+    default: {
+        payLightningInvoice: jest.fn(),
+        sendKeysend: jest.fn(),
+        lookupPayment: jest.fn(),
+        supportsPaymentLookup: jest.fn(() => true),
+        isLNDBased: jest.fn(() => true)
+    }
+}));
+jest.mock('../utils/Bolt11Utils', () => ({
+    __esModule: true,
+    default: {
+        decode: jest.fn(() => ({ payment_hash: 'ab'.repeat(32) }))
+    }
+}));
+jest.mock('../utils/GraphSyncUtils', () => ({
+    checkGraphSyncBeforePayment: jest.fn(() => true)
+}));
+jest.mock('../utils/LocaleUtils', () => ({
+    localeString: (s: string) => s
+}));
+jest.mock('../utils/ErrorUtils', () => ({
+    errorToUserFriendly: (error: any) =>
+        typeof error === 'string' ? error : error?.message
+}));
+jest.mock('../utils/UrlUtils', () => ({
+    __esModule: true,
+    default: {}
+}));
+jest.mock('../utils/RatingUtils', () => ({
+    RATING_MODAL_TRIGGER_DELAY: 1000
+}));
+
+import TransactionsStore, {
+    PAYMENT_TRACK_POLL_MS,
+    PAYMENT_TRACK_MAX_FAILURES
+} from './TransactionsStore';
+import BackendUtils from '../utils/BackendUtils';
+
+// 64-char hex: normalizePaymentHash rejects anything that isn't a real hash
+const HASH = 'ab'.repeat(32);
+
+const newStore = () =>
+    new TransactionsStore(
+        { implementation: 'lnd', enableTor: false, settings: {} } as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        { checkAndTriggerRatingModal: jest.fn() } as any
+    );
+
+const flush = () => jest.advanceTimersByTimeAsync(0);
+const advancePoll = () => jest.advanceTimersByTimeAsync(PAYMENT_TRACK_POLL_MS);
+
+describe('TransactionsStore payment tracking (issue #4317)', () => {
+    beforeEach(() => {
+        jest.useFakeTimers();
+        jest.clearAllMocks();
+        (BackendUtils.supportsPaymentLookup as jest.Mock).mockReturnValue(true);
+    });
+
+    afterEach(() => {
+        jest.useRealTimers();
+    });
+
+    it('holds the guard on an IN_FLIGHT result and tracks to SUCCEEDED', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+        // the first lookup fires immediately when tracking starts; the
+        // terminal result lands on the second poll
+        (BackendUtils.lookupPayment as jest.Mock)
+            .mockResolvedValueOnce({ status: 'IN_FLIGHT', payment_hash: HASH })
+            .mockResolvedValueOnce({ status: 'IN_FLIGHT', payment_hash: HASH })
+            .mockResolvedValueOnce({
+                status: 'SUCCEEDED',
+                payment_hash: HASH,
+                payment_preimage: 'deadbeef'
+            });
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+
+        // stream ended non-terminal: guard must stay armed
+        expect(store.paymentInFlight).toBe(true);
+        expect(store.status).toBe('IN_FLIGHT');
+
+        await advancePoll();
+        expect(store.paymentInFlight).toBe(true);
+
+        await advancePoll();
+        expect(store.paymentInFlight).toBe(false);
+        expect(store.status).toBe('SUCCEEDED');
+        expect(store.error).toBe(false);
+    });
+
+    it('treats a client-side timeout as unknown and surfaces the tracked FAILED outcome', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockResolvedValue({
+            payment_error: 'views.SendingLightning.paymentTimedOut',
+            payment_timed_out: true
+        });
+        // hold the first lookup open so the ambiguous state is observable
+        let resolveLookup: (value: any) => void = () => {};
+        (BackendUtils.lookupPayment as jest.Mock).mockReturnValue(
+            new Promise((resolve) => {
+                resolveLookup = resolve;
+            })
+        );
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+
+        // outcome unknown: no error yet, guard held, in-transit UI
+        expect(store.paymentInFlight).toBe(true);
+        expect(store.error).toBe(false);
+        expect(store.status).toBe('IN_FLIGHT');
+
+        resolveLookup({
+            status: 'FAILED',
+            failure_reason: 'FAILURE_REASON_TIMEOUT',
+            payment_hash: HASH
+        });
+        await flush();
+        expect(store.paymentInFlight).toBe(false);
+        expect(store.error).toBe(true);
+        expect(store.status).toBe('FAILED');
+    });
+
+    it('replaces a transport error with success when the payment settled anyway', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockRejectedValue(
+            new Error('connection closed')
+        );
+        (BackendUtils.lookupPayment as jest.Mock).mockResolvedValue({
+            status: 'SUCCEEDED',
+            payment_hash: HASH,
+            payment_preimage: 'deadbeef'
+        });
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+
+        expect(store.paymentInFlight).toBe(false);
+        expect(store.status).toBe('SUCCEEDED');
+        expect(store.error).toBe(false);
+    });
+
+    it('releases the guard when the node has no record of the payment', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockRejectedValue(
+            new Error('invoice is invalid')
+        );
+        (BackendUtils.lookupPayment as jest.Mock).mockResolvedValue(null);
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+
+        expect(store.paymentInFlight).toBe(false);
+        expect(store.error).toBe(true);
+        expect(store.error_msg).toBe('invoice is invalid');
+    });
+
+    it('gives up after consecutive failed lookups and releases the guard', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+        (BackendUtils.lookupPayment as jest.Mock).mockRejectedValue(
+            new Error('node unreachable')
+        );
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+        expect(store.paymentInFlight).toBe(true);
+
+        for (let i = 0; i < PAYMENT_TRACK_MAX_FAILURES; i++) {
+            await advancePoll();
+        }
+        expect(store.paymentInFlight).toBe(false);
+        expect(BackendUtils.lookupPayment).toHaveBeenCalledTimes(
+            PAYMENT_TRACK_MAX_FAILURES
+        );
+    });
+
+    it('keeps the pre-tracking behavior on backends without payment lookup', async () => {
+        (BackendUtils.supportsPaymentLookup as jest.Mock).mockReturnValue(
+            false
+        );
+        (BackendUtils.payLightningInvoice as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+
+        expect(store.paymentInFlight).toBe(false);
+        expect(store.status).toBe('IN_FLIGHT');
+        expect(BackendUtils.lookupPayment).not.toHaveBeenCalled();
+    });
+
+    it('ignores a second send while a tracked payment is unresolved', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+        (BackendUtils.lookupPayment as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+        expect(store.paymentInFlight).toBe(true);
+
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+        expect(BackendUtils.payLightningInvoice).toHaveBeenCalledTimes(1);
+    });
+
+    it('lets an unscoped handlePayment (Rebalance view path) stop tracking and clear the guard', async () => {
+        (BackendUtils.payLightningInvoice as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+        (BackendUtils.lookupPayment as jest.Mock).mockResolvedValue({
+            status: 'IN_FLIGHT',
+            payment_hash: HASH
+        });
+
+        const store = newStore();
+        store.sendPayment({ payment_request: 'lnbc1fake' });
+        await flush();
+        expect(store.paymentInFlight).toBe(true);
+
+        store.handlePayment({
+            status: 'SUCCEEDED',
+            payment_hash: HASH,
+            payment_preimage: 'deadbeef'
+        });
+        expect(store.paymentInFlight).toBe(false);
+        expect(store.status).toBe('SUCCEEDED');
+
+        // the orphaned tracking loop must exit without resurrecting state
+        const calls = (BackendUtils.lookupPayment as jest.Mock).mock.calls
+            .length;
+        await advancePoll();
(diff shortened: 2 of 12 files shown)


kaloudis:
Rebased on master to pick up the #4306 timeout headroom, the LNC payment correlation work, and the #4435/#4443 NWC merges. Integration changes beyond conflict resolution:

- `LND.payLightningInvoice` keeps master's `timeoutSeconds` parse and +5s transport headroom; the `Request timeout` normalization is retained as a safety net and the forced-timeout shape now carries the `payment_timed_out` marker.
- LNC's stream-death resolution (which already resolves the timed-out shape instead of rejecting) also gets the marker, so those payments are now tracked to terminal instead of surfacing as failures.
- The store now derives the tracked hash with `deriveExpectedPaymentHash` from `utils/LncPayUtils.ts` instead of duplicating the keysend/bolt11/AMP hash logic.

`yarn verify` green after the rebase (65 suites, 1425 tests).

stores/TransactionsStore.ts
```diff
@@ -727,6 +760,107 @@ export default class TransactionsStore {
...
+    // Polls lookupPayment until the guard-owning payment reaches SUCCEEDED
+    // or FAILED, then surfaces the real outcome through handlePayment
+    // (which also releases the guard). This closes the double-pay window
+    // left by releasing the guard on timers while HTLCs are still pending:
+    // a keysend retry generates a fresh preimage, so node-side same-hash
+    // dedup can't protect against it (issue #4317). Exits released: on a
+    // terminal state, when the payment provably never reached the node,
+    // after PAYMENT_TRACK_MAX_FAILURES unobservable polls, or at the
+    // PAYMENT_TRACK_MAX_MS ceiling.
+    private trackPaymentToTerminal = async (seq: number) => {
```
ajaysehwal, line 810: @kaloudis, I think there’s still a race in `trackPaymentToTerminal`.

The “no record found → safe to release” path relies on a single `lookupPayment()` result, but all three implementations only scan the latest 50 payments.

A pending payment could fall outside that window during the 10-minute tracking period and be incorrectly treated as never dispatched. That would release the guard while the HTLC could still settle, allowing another payment to start.

I think we should be more conservative here — e.g. require multiple consecutive `not found` results before releasing the guard, or search further back before concluding the payment was never dispatched.

kaloudis: Good catch, agreed. The guard only serializes this app's sends, so on a shared node (or via the NWC service's guard-exempt background sends) newer payments could evict the tracked one from the newest-50 page, and a single no-record answer can also race the dispatch itself: over a slow Tor circuit the send request may still be in transit when a fresh-connection lookup returns.

Addressed in d05f01acd with both of your suggestions:

- Lookups are now bounded instead of paging from the newest: the store records the dispatch time and passes `creation_date_start` (dispatch minus a 10 minute clock-skew slack), and all three implementations scan ascending from that bound, so the target payment stays on the first page regardless of node traffic. LND REST takes it as a query param, lndmobile forwards it on `ListPayments`, and LNC passes it through `listPayments`.
- The never-dispatched conclusion now requires 3 consecutive no-record answers with no sighting in between (`PAYMENT_TRACK_MAX_NOT_FOUND`), instead of releasing on the first one. Genuine pre-dispatch failures still surface their error immediately; only the guard release waits the extra polls (about 10s at the 5s cadence).

Two tests cover the new behavior: release only after repeated no-record answers (with the `creation_date_start` bound asserted), and a no-record answer followed by the payment appearing and being tracked to settlement.

kaloudis: https://github.com/ZeusLN/zeus/pull/4492/commits/d05f01acd87e3b7dbda54f15209ddfbdcae98270

ajaysehwal (commented):
@kaloudis, there’s still the same ambiguous timeout race in `CLNRest.payWithTimeout`: `Promise.race([forcedTimeout, call()])` can return the client-side timeout before the actual `/v1/xpay` or `/v1/pay` response arrives, but CLN never marks this as `payment_timed_out`.

Since `supportsPaymentLookup()` is false for CLN, `canTrackPayment` is always false and `handlePaymentError` releases `paymentInFlight` immediately. If the payment settles on the node moments later, the user can retry — especially for keysend, where the retry uses a fresh preimage — and potentially pay twice.

This looks like the same fund-loss scenario this PR is fixing for LND, but it remains open for CLN. Could we apply the same `payment_timed_out` handling to the CLN path?

backends/LdkNode.ts
ajaysehwal, line 1: @kaloudis, there’s a similar gap for LDK Node: `awaitPaymentCompletion` already documents the risk of a double-pay on keysend/BOLT12 when its client-side polling times out, but `supportsPaymentLookup()` is still `false` for LDK Node.

When `awaitPaymentCompletion` throws `error.paymentTimedOut`, `handlePaymentError` sees `canTrackPayment(seq) === false` and immediately calls `clearPaymentInFlight(seq)`. If the original payment settles after that, a retry can potentially double-pay.

This is the exact scenario called out in the existing `awaitPaymentCompletion` comment. Could we mark the timeout as trackable / keep the send guard in place for LDK Node

stores/TransactionsStore.ts
```diff
@@ -727,6 +773,130 @@ export default class TransactionsStore {
...
+    // Fires when the backstop timer elapses without a completion callback
+    // having cleared the guard (dropped LNC stream, hung request). Rather
+    // than blindly releasing the guard while an HTLC may still settle, try
+    // to track the payment to its terminal state first.
+    private backstopPaymentInFlight = (seq: number) => {
+        if (this.inFlightOwnerSeq !== seq) return;
+        if (this.trackingSeq === seq) return; // tracking owns the release
+        if (this.canTrackPayment(seq)) {
+            this.trackPaymentToTerminal(seq);
+        } else {
```
ajaysehwal, line 796: @kaloudis, one remaining gap in the tracked-backend path: `lookupPayment` only scans a single 50-item page. On a busy/shared node, or with NWC background sends that bypass the guard, 50+ payments can be created within the ~10-minute lookup window and push the tracked payment beyond that first page.

The tracker then gets consecutive “not found” results, assumes the payment was never dispatched, and calls `clearPaymentInFlight` without knowing the actual outcome. If the original keysend is still pending or later settles, a retry with a fresh preimage can double-pay.

Could we paginate/deepen the lookup, or otherwise avoid treating a single 50-item page as proof that the payment doesn’t exist?

kaloudis: this is addressed in the follow up I believe: https://github.com/ZeusLN/zeus/pull/4495

kaloudis:
On the CLN timeout: agreed, and the resolve path is why it needs explicit handling. `payWithTimeout`'s forced shape resolves rather than rejects, so it lands in `handlePayment`, which only holds the guard for `IN_FLIGHT` or `payment_timed_out`. That is implemented in the stacked follow-up #4495 (53499bbd1): the forced shape carries `payment_timed_out`, `lookupPayment` goes through `/v1/listpays` with the `payment_hash` filter, and `supportsPaymentLookup` flips to true. The same commit covers LDK Node from your other comment.

This PR stays scoped to the LND family on purpose, with #4495 finishing the rest, so I would rather not widen it here.

One gap in that follow-up worth calling out, because it lands exactly on the keysend case you raise. `sendPaymentInternal` derives the tracked hash from a preimage Zeus generates locally, but `CLNRest.sendKeysend` transmits neither the preimage nor a TLV carrying it: `/v1/keysend` and `/v1/xkeysend` mint their own. So a CLN keysend timeout would poll a hash the node never had, collect three consecutive no-record answers, and release the guard with a false never-dispatched verdict about 15 seconds later, which is worse than not tracking at all. Bolt11 is fine there, since the hash comes from the invoice. Unless keysend or xkeysend can be made to accept our preimage, which I do not believe they can, the fix is to keep CLN keysend out of the never-dispatched release path, and I will do that in #4495 before it is ready for review.

backends/LdkNode.ts
kaloudis: Agreed on the risk. Implemented in the stacked follow-up #4495 (53499bbd1), which adds `LdkNode.lookupPayment` over the in-memory `listPayments` and flips `supportsPaymentLookup` to true. No marker is needed on this path: `awaitPaymentCompletion` throws, so it lands in `handlePaymentError`, which already tracks whenever `canTrackPayment(seq)` is true, timeout or not.

Two caveats I hit while checking your scenario:

- The keysend half does not work from the flag alone. The store derives the tracked hash from a preimage it generates, but `sendKeysend` calls `spontaneousPayment().send`, which mints its own preimage and hash, so the lookup would poll a hash the node never had and release the guard on the third no-record answer. ldk-node exposes `spontaneousPayment().sendWithPreimage` in the uniffi bindings on both platforms, but our React Native bridge only wires `send` (`LdkNodeModule.kt`, `LdkNodeModule.swift`). Wiring the preimage through makes the hash authoritative and matches the LND semantics, and it also restores node-side same-hash dedup if a retry reuses the preimage. Until that lands, LDK and CLN keysend should be kept out of the never-dispatched release path rather than tracked against a hash we invented.
- The BOLT12 half sits outside this guard entirely. `payBolt12` in `views/Send.tsx` goes through `fetchInvoiceFromOffer`, which completes the payment inside the backend and never enters `sendPaymentInternal`, so `paymentInFlight` is never set for it. That needs separate work, adjacent to #4307.

stores/TransactionsStore.ts
kaloudis: Adding detail to that, since part of this is already in this PR and part of it is genuinely still open.

The newest-50 framing predates the first review round. d05f01acd anchored the scan instead of paging from the newest: the store records the dispatch time and passes `creation_date_start` (dispatch minus 10 minutes of clock slack), and every implementation scans ascending from that bound with `reversed: false`, so the tracked payment sits at the head of the first page no matter how busy the node is or how many NWC background sends land in between. #4495 then makes the lookup targeted (TrackPaymentV2 for LND REST, embedded and LNC, `/v1/listpays` filtered by hash for CLN, an in-memory filter for LDK), leaving the scan as a fallback for the in-flight and transport-failure cases.

Two residual ways to get a false no-record out of that fallback remain, and I would rather harden them here than lean on the follow-up:

- More than 50 payments created on the node inside the 10 minute pre-dispatch window still pushes the target off the first page.
- Clock skew in the wrong direction. The bound comes from the device clock, `creation_date` comes from the node. A device more than 10 minutes ahead of the node puts every candidate before the bound, so the lookup misses permanently.

The fix is to stop reading a full page as an answer: if the scan returns `maxPayments` results with no match, the range was truncated, so it is inconclusive rather than a no-record answer and must not count toward the never-dispatched conclusion. I will pair that with a reversed newest-50 pass before returning null, so a null requires both views to miss.

stores/TransactionsStore.ts
```diff
@@ -727,6 +773,130 @@ export default class TransactionsStore {
...
+                        ? lnrpc.Payment.PaymentStatus[payment.status]
+                        : payment?.status;
+
+                if (status === 'SUCCEEDED' || status === 'FAILED') {
+                    // clear any transport error shown while the outcome
+                    // was unknown; handlePayment re-derives error state
+                    // from the payment's actual terminal result
+                    this.clearPaymentError();
+                    this.handlePayment(payment, seq);
+                    return;
```
ajaysehwal, line 858: @kaloudis, I think there’s still a UI state issue when payment tracking gives up.

After `payment_timed_out`, `handlePayment` sets `status = 'IN_FLIGHT'` without an error. If tracking then ends without a terminal result (e.g. 3 not-found results or the 10-minute ceiling), `finally` only clears the guard, leaving the payment in `IN_FLIGHT`.

`SendingLightning` renders the in-transit state whenever `status === 'IN_FLIGHT' && !error`, so the user can remain stuck seeing “in transit” even though tracking has stopped and the final payment outcome is unknown.

On the give-up path, could we set an appropriate error/unknown state before clearing the guard, so the UI doesn’t present the payment as still in transit?

stores/TransactionsStore.test.ts
ajaysehwal, line 1: @kaloudis, one test coverage gap worth adding:

* A full lookup page where the target payment isn’t present.
* A `creation_date_start` later than the payment’s `creation_date`.
* The final `status` / `error` state after tracking gives up.
* Unit tests for the backend `lookupPayment` implementations, which currently aren’t covered.

Not blocking, but these would make the tracker’s edge cases and give-up behavior more explicit.

stores/TransactionsStore.ts
kaloudis: Good catch. The tracker's `finally` only released the guard, so a payment diverted as `payment_timed_out` or `IN_FLIGHT` stayed on the in-transit screen after tracking stopped. Fixed in b9f6833bc.

I did not set an error on every give-up path, because an error shows Try Again and that retry is the double-pay from #2577. The screen now depends on why tracking stopped:

- Three consecutive no-record answers: the node never received the payment. It is shown as not sent (`views.SendingLightning.paymentNotSent`), and `status` is cleared. If `handlePaymentError` already showed a dispatch error, that message is kept because it is more specific. Retrying is safe here.
- Three failed lookups, or the 10 minute ceiling: the outcome is unknown, and at the ceiling the HTLC is usually still pending (a hold invoice, for example). A failed state would offer the retry that can double-pay, so the in-transit view stays. A new `paymentOutcomeUnverified` flag adds a line telling the user that ZEUS stopped checking and to look at Activity before sending again. The flag resets on the next send and in `reset()`.

stores/TransactionsStore.test.ts
kaloudis: Added in b9f6833bc, along with the full-page change I promised in the earlier thread:

- Full page without the target: `lookupPayment` for LND REST, embedded LND and LNC now goes through a shared `findPaymentByHash` (`utils/PaymentLookupUtils.ts`). If a full bounded page has no match, it rejects with a tagged inconclusive error instead of resolving null. The tracker counts that as neither a no-record answer nor a failed lookup, so it keeps polling until the ceiling. Covered in `utils/PaymentLookupUtils.test.ts`, plus a store test that checks the guard is still held after more inconclusive answers than either limit allows.
- `creation_date_start` later than the payment's `creation_date` (device clock more than the 10 minute slack ahead of the node): a second pass over the newest 50 payments now finds it. It has its own test case.
- Final `status`/`error` after giving up: tested for each exit. The failed-lookups test now asserts the unverified in-transit state, a new test covers timeout followed by no-record (shown as not sent), and the no-record test after a dispatch error asserts that the original message is kept.
- Backend `lookupPayment`: `backends/LND.test.ts` now checks the two REST request URLs and case-insensitive matching. I left embedded LND and LNC for #4495, because TrackPaymentV2 replaces their scan there and tests written against the scan would be rewritten.

locales/en.json
```diff
@@ -972,6 +972,8 @@
     "views.SendingLightning.tryAgain": "Try Again",
     "views.SendingLightning.copyPaymentHash": "Copy Hash to Clipboard",
     "views.SendingLightning.inTransit": "Payment is in transit. It may complete momentarily or it may go on to fail. Check the Activity view to track its progress.",
+    "views.SendingLightning.paymentNotSent": "Payment was not sent. Your node has no record of it, so it is safe to try again.",
```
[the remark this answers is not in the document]
kaloudis: Fixed in cbc5dfdbf. The copy was the bigger problem, but the transport mechanism is narrower than described:

- LND REST (clearnet and Tor): `call()` isn't cancelled in JS, but its native request is (`RNBlobUtil` `timeout` / nitro-tor `timeout_ms`, both T+5s). The third no-record lookup ran at T+11s or later, after that deadline. That held only because two poll intervals happened to exceed the 4s headroom, and nothing tied the two together.
- Embedded LND: in-process call, so the request can't be delivered late.
- LNC: the forced resolve is at T+60s and the `sendPaymentV2` stream is never cancelled, so late delivery through the mailbox can't be ruled out. This is the case where the scenario applies.

Changes:
- Timed-out results now carry `dispatch_deadline_ms`, and no-record answers only count toward the never-dispatched verdict from that time on. LND REST reports its native request deadline and embedded reports dispatch time.
- LNC omits the deadline, so repeated no-record answers end tracking as unverified (no Try Again, "check Activity" copy). The backstop path, where the send never settled, gets the same treatment.
- The `handlePaymentError` path counts from the rejection, because the request has settled by then. That path also never shows `paymentNotSent`: the dispatch error stays on screen.
- `paymentNotSent` no longer says a retry is safe: "Payment was not sent. Your node has no record of it. Check the Activity view before trying again."
- Two new store tests: no-record answers before the deadline don't count, and a timed-out send without a deadline ends unverified.

PR #4495, opened 2026-08-27
Complete lookupPayment across capable backends and use it for the NWC pending refresh

### LINKED ISSUE #4286: NWC: add a first-class lookupPayment backend method instead of scanning the payment list

kaloudis:
# Description

Fixes issue: **#4286**

**Stacked on #4492** (its commits are included here). Review only `53499bbd1`, `3e3f5350b`, and `9d8b674e9` until #4492 merges, after which this rebases onto master.

#4492 introduced `lookupPayment` / `supportsPaymentLookup` but only implemented the LND-family backends, which have no targeted per-payment query and scan their newest 50 payments. This PR completes #4286's proposal:

- **`CLNRest.lookupPayment`**: `/v1/listpays` with its `payment_hash` filter, a true targeted query. `listpays` returns one entry per retry group, so a settled attempt is preferred, then a pending one, over stale failed groups. Status is mapped to LND's enum strings (`complete` to `SUCCEEDED`, `failed` to `FAILED`, else `IN_FLIGHT`) so `trackPaymentToTerminal` and the NWC reconcile helpers read it uniformly; the remaining fields keep CLN's native names, which `models/Payment` already handles. CLN's `payWithTimeout` forced-timeout shape also gains the `payment_timed_out: true` marker from #4492, so a client-side timeout on `xpay`/`pay`/keysend is now tracked to its terminal state instead of surfacing as a failure while CLN may still be retrying inside `retry_for`.
- **`LdkNode.lookupPayment`**: filters the in-memory `listPayments()` result by hash (outbound, non-onchain) and reuses `formatPayment`, which already emits LND-shaped statuses. This also gives the send-guard tracker a real observation path for payments `awaitPaymentCompletion`'s timeout leaves pending, the follow-up noted in #4492.
- **`NostrWalletConnect.lookupPayment`**: NIP-47 `lookup_invoice` via the SDK's raw client (the webln wrapper discards `state` and amounts), which resolves either direction by `payment_hash`. `NOT_FOUND` maps to the `null` no-record contract; msat amounts are converted to sats for parity with the webln-mapped `getPayments` list; wallets that predate `state` fall back to `settled_at`.
- **`LND.lookupPayment` (REST)**: now targeted as well (review suggestion). lnd has no unary payment lookup (lightningnetwork/lnd#1018), but `TrackPaymentV2`'s REST mapping (`GET /v2/router/track/{payment_hash}?no_inflight_updates=true`) emits one update and closes the stream when the payment is already terminal, so the buffered transport resolves it like a unary call; an unknown hash errors with NOT_FOUND ("payment isn't initiated"), an authoritative no-record `null` with no scan-page eviction caveat. An in-flight payment holds the stream open, so a 10s timeout falls back to the dispatch-anchored page scan from #4492, which also backs up unexpected transport failures. No version gating needed: paying already depends on `/v2/router/send`.
- **`EmbeddedLND.lookupPayment` / `LightningNodeConnect.lookupPayment`**: also targeted now. Both transports support real streaming, so the lookup takes the payment's current state from the first TrackPaymentV2 update carrying its hash (`no_inflight_updates` false, so an in-flight payment answers immediately) instead of scanning a page. Both event channels are multiplexed by method name with no per-request correlation, so updates are matched by hash; NOT_FOUND errors carry no hash and could belong to a concurrent lookup's stream, so absence is confirmed against the dispatch-anchored scan before answering `null`, and any other failure degrades to that scan. The underlying streams can't be cancelled client-side and stay open node-side until the payment is terminal; only the listener is removed after the first matching update.
- **NWC service**: `getPaymentsForPendingPayInvoiceRefresh` now answers the by-hash question through `lookupPayment` (one call per unique pending hash) instead of fetching and scanning the full payment list, so `lookup_invoice`, `list_transactions`, and budget reconciliation scale with the request rather than with payment history. The existing per-connection min-age and refresh-interval gates are unchanged. The list scan remains the fallback for backends without lookup (LndHub), pending activities without a recorded payment hash (matching needs the list's payment_request strings), and failed lookups. A `null` lookup result is authoritative absence, handled by the existing stale-hold abandonment.

All seven backends now resolve #4286's capability table: six targeted lookups (CLN via `listpays`, LDK Node in memory, NWC via `lookup_invoice`, and LND REST, embedded LND, and LNC via TrackPaymentV2), one genuine impossibility (LndHub, explicit `false`), and the #4492 dispatch-anchored page scan surviving only as the shared fallback path.

Six new unit tests cover the targeted refresh: settle-without-list-fetch, the three fallback paths (no capability, missing hash, lookup failure), authoritative-absence handling, and per-hash dedup. Seven more in `backends/LND.test.ts` cover the TrackPaymentV2 lookup: terminal-without-scan, streamed and thrown NOT_FOUND to `null`, timeout to anchored scan, unanchored scan, unexpected-shape backup, and double-failure propagation. Seven more in `lndmobile/index.test.ts` cover the embedded stream helper (request encoding, hash-filtered first update through real protobuf round-trips, case-insensitivity, NOT_FOUND surfacing, EOF/transient tolerance, timeout, listener removal), and eleven in `utils/LncPayUtils.test.ts` cover the LNC track-event classifier.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

Device tests pending: CLN (xpay timeout tracked to terminal, NWC service lookup_invoice on a pending payment), LDK Node (timeout then settle), NWC backend (pay via remote wallet with dropped transport), and an LndHub regression pass on the scan fallback.

### DIFF
--- a/backends/LND.ts
+++ b/backends/LND.ts
@@ -535,6 +535,16 @@ export default class LND {
             return response;
         };
 
+        // payment_timed_out marks the outcome as unknown on the node (the
+        // client gave up, not the payment), so the caller can track the
+        // payment to its terminal state instead of reporting failure
+        const timedOutResponse = () => ({
+            payment_error: localeString(
+                'views.SendingLightning.paymentTimedOut'
+            ),
+            payment_timed_out: true
+        });
+
         // The request timeout must exceed the forcedTimeout below, or the
         // generic transport rejection wins the race and the user sees a
         // retryable-looking error instead of the payment-timed-out shape
@@ -547,19 +557,32 @@ export default class LND {
                     allow_self_payment: true
                 },
                 (timeoutSeconds + 5) * 1000
-            );
+            ).catch((err: any) => {
+                // safety net if the transport deadline ever fires first:
+                // a client-side timeout is outcome-unknown, not a failure
+                if (err?.message === 'Request timeout')
+                    return timedOutResponse();
+                throw err;
+            });
 
         const result: any = await Promise.race([
-            forcedTimeout((timeoutSeconds + 1) * 1000, {
-                payment_error: localeString(
-                    'views.SendingLightning.paymentTimedOut'
-                )
-            }),
+            forcedTimeout((timeoutSeconds + 1) * 1000, timedOutResponse()),
             call()
         ]);
 
         return result;
     };
+    // scans the newest payments because LND REST has no non-streaming
+    // per-payment lookup (TrackPaymentV2 streams until terminal)
+    lookupPayment = (data: { payment_hash: string }) =>
+        this.getPayments({ maxPayments: 50, reversed: true }).then(
+            (response: any) =>
+                response?.payments?.find(
+                    (payment: any) =>
+                        payment.payment_hash?.toLowerCase() ===
+                        data.payment_hash.toLowerCase()
+                ) ?? null
+        );
     closeChannel = (urlParams?: Array<string>) => {
         let requestString = `/v1/channels/${urlParams && urlParams[0]}/${
             urlParams && urlParams[1]
@@ -1014,6 +1037,7 @@ export default class LND {
     supportsOnchainReceiving = () => true;
     supportsLightningSends = () => true;
     supportsKeysend = () => true;
+    supportsPaymentLookup = () => true;
     supportsChannelManagement = () => true;
     supportsCircularRebalancing = () => true;
     supportsForceClose = () => true;
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -815,6 +815,8 @@ export default class CLNRest {
     supportsOnchainReceiving = () => true;
     supportsLightningSends = () => true;
     supportsKeysend = () => true;
+    // follow-up: /v1/listpays accepts payment_hash and could support this
+    supportsPaymentLookup = () => false;
     supportsChannelManagement = () => true;
     supportsCircularRebalancing = () => true;
     supportsForceClose = () => false;
--- a/backends/CLNRest.ts
+++ b/backends/CLNRest.ts
@@ -403,6 +403,39 @@ export default class CLNRest {
             };
         });
 
+    // targeted lookup via listpays' payment_hash filter. listpays returns
+    // one entry per retry group, so prefer a settled attempt, then a
+    // pending one, over stale failed groups from earlier retries. The
+    // status is mapped to LND's enum strings so trackPaymentToTerminal
+    // and the NWC reconcile helpers read it uniformly.
+    lookupPayment = (data: { payment_hash: string }) =>
+        this.postRequest('/v1/listpays', {
+            payment_hash: data.payment_hash
+        }).then((response: any) => {
+            const pays = response?.pays || [];
+            if (pays.length === 0) return null;
+            const pay =
+                pays.find((p: any) => p.status === 'complete') ||
+                pays.find((p: any) => p.status === 'pending') ||
+                pays[pays.length - 1];
+            return {
+                payment_hash: pay.payment_hash,
+                status:
+                    pay.status === 'complete'
+                        ? 'SUCCEEDED'
+                        : pay.status === 'failed'
+                        ? 'FAILED'
+                        : 'IN_FLIGHT',
+                destination: pay.destination,
+                created_at: pay.created_at?.toString(),
+                bolt11: pay.bolt11,
+                bolt12: pay.bolt12,
+                amount_msat: pay.amount_msat,
+                amount_sent_msat: pay.amount_sent_msat,
+                preimage: pay.preimage
+            };
+        });
+
     getNewAddress = (data: any) => {
         let addresstype: string | undefined;
 
@@ -515,7 +548,10 @@ export default class CLNRest {
             forcedTimeout((timeoutSeconds + 1) * 1000, {
                 payment_error: localeString(
                     'views.SendingLightning.paymentTimedOut'
-                )
+                ),
+                // outcome unknown on the node: lets the caller track the
+                // payment to a terminal state instead of reporting failure
+                payment_timed_out: true
             }),
             call
         ]);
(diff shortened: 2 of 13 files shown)


backends/LND.ts
```diff
@@ -547,19 +557,32 @@ export default class LND {
...
-                )
-            }),
+            forcedTimeout((timeoutSeconds + 1) * 1000, timedOutResponse()),
             call()
         ]);
 
         return result;
     };
+    // scans the newest payments because LND REST has no non-streaming
+    // per-payment lookup (TrackPaymentV2 streams until terminal)
```
kaloudis, line 576: https://lightning.engineering/api-docs/api/lnd/lightning/lookup-invoice/

kaloudis: Implemented in 3e3f5350b. LookupInvoice itself only reads the invoice DB (it can't see outgoing payments), and lnd still has no unary payment foil for it (lightningnetwork/lnd#1018), but TrackPaymentV2's REST mapping gets us the same targeted semantics: `GET /v2/router/track/{payment_hash}?no_inflight_updates=true` emits a single update and closes the stream once the payment is in a terminal state, so the buffered transport resolves it like a unary call, and an unknown hash errors with NOT_FOUND ("payment isn't initiated"), which is an authoritative no-record answer with none of the scan-page eviction caveats.

The one state it can't answer non-blockingly is IN_FLIGHT (the stream just stays open), so after a 10s timeout it falls back to the dispatch-anchored page scan, which now only serves that case. No version gating needed since paying already depends on `/v2/router/send`. Covered by new unit tests in `backends/LND.test.ts`.

Embedded LND and LNC still use their scans; they have real streaming transports, so they could subscribe to TrackPaymentV2 directly and take the first update. Happy to do that here or as a follow-up.

kaloudis: Follow-up in 9d8b674e9: embedded LND and LNC now use their real streaming transports for the lookup too. Both subscribe to TrackPaymentV2 with `no_inflight_updates` false and take the first update carrying their hash, so an in-flight payment answers immediately and terminal states come straight from the router. The wrinkle on both transports is that events are multiplexed by method name with no per-request correlation (the embedded emitter and the LNC event channel behave identically here), so payment updates are matched by hash, and since error events carry no hash, a NOT_FOUND is confirmed against the dispatch-anchored scan before answering null in case it belonged to a concurrent lookup's stream. The scan survives only as that fallback. Neither transport can cancel a stream client-side, so a lookup on a still-pending payment leaves its stream open node-side until the payment is terminal; that's bounded by the payment's own lifetime.

All three LND-family backends are now targeted, so #4286's table is fully precise lookups plus LndHub's explicit false. Unit tests cover the embedded helper through real protobuf round-trips and the LNC event classifier; device tests still owed on all three.

shubhamkmr04:
@kaloudis please resolve conflicts

PR #4500, opened 2026-08-28
fix(nwc): guard updateConnection against in-flight handler races

### LINKED ISSUE #4486: NWC: updateConnection races with in-flight handlers (relay change can drop responses, budget edits race settlement)
### Summary

PR #4443 fixes the race between `deleteConnection` and in-flight NIP-47 handlers, but the same race exists in `updateConnection`'s relay-change and budget-edit paths and is untouched there. Splitting it out per review discussion so it does not get lost: https://github.com/ZeusLN/zeus/pull/4443#issuecomment-5350308753

### Problem

`updateConnection` (`stores/NostrWalletConnectStore.ts:677`) mutates connection state and tears down relay plumbing with no guard against handlers that are currently executing for that connection:

1. **Relay change can eat an in-flight response.** When `relayUrlChanged`, it unsubscribes and calls `releaseUnusedRelayService(oldRelayUrl)` immediately (`stores/NostrWalletConnectStore.ts:808`). If the connection was the last one on the old relay, the pool socket is closed while `@getalby/sdk` may still be flushing a NIP-47 response (the SDK publishes fire-and-forget after the handler resolves). Same mechanism as the delete-path bug fixed in #4443, where it is now handled with a grace period.
2. **Budget edits race settlement.** Removing a budget sets `connection.totalSpendSats = 0` and clears `lastBudgetReset` (`stores/NostrWalletConnectStore.ts:800`) while a `pay_invoice` may be mid-settle. The settle then records spend on top of the rewritten budget state, or the payment was approved against the old budget and lands under the new one. `Object.assign(connection, updates)` mid-flight has the same problem for permission and expiry edits.
3. **No pending guard.** New requests arriving during the update window are not rejected, so they can interleave with the mutation.

### Suggested fix

Once #4443 merges, its primitives make this small:

- mark the id pending (analogous to `pendingDeleteConnectionIds`) so `withGlobalHandler` rejects new requests during the update window
- `await awaitInFlightHandlers(connectionId)` before applying budget/permission mutations and before re-subscribing
- when handlers were in flight or the connection was subscribed, release the old relay via `scheduleReleaseUnusedRelayService(oldRelayUrl)` instead of immediately

Severity is low: the window requires the user to be editing a connection at the exact moment a client request is in flight. Same class as the #4443 issue.

ajaysehwal:
# Description

Relates to issue: **#4486**

PR #4443 fixed the race between `deleteConnection` and in-flight NIP-47 handlers, but the same window existed in `updateConnection` — relay changes could tear down the old relay while the SDK was still flushing a response, and budget/permission/expiry edits could interleave with a mid-settle `pay_invoice`.
This applies the same mutation guard pattern to `updateConnection`:
- Rename `pendingDeleteConnectionIds` → `pendingConnectionMutationIds` and use it for both delete and update, so `withGlobalHandler` rejects new requests during either operation
- `awaitInFlightHandlers()` before applying connection mutations (budget, permissions, relay rotation, etc.)
- On relay change, unsubscribe first, then drain handlers, then mutate
- Defer old-relay release via `scheduleReleaseUnusedRelayService` when handlers were in flight or the connection had been subscribed (same rule as delete)

No behavior change for the common case where no request is in flight.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -1260,6 +1260,127 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         jest.clearAllTimers();
         jest.useRealTimers();
     });
+
+    it('lets an in-flight handler finish before updateConnection mutates budget', async () => {
+        const store = buildStore();
+        jest.spyOn(store as any, 'subscribeToConnection').mockResolvedValue(
+            undefined
+        );
+        const connection = seedConnection(store, {
+            maxAmountSats: 5000,
+            totalSpendSats: 42
+        });
+
+        let releaseHandler: (value: { result: { ok: boolean } }) => void = () =>
+            undefined;
+        const handlerDone = (store as any).withGlobalHandler(
+            connection.id,
+            () =>
+                new Promise((resolve) => {
+                    releaseHandler = resolve;
+                })
+        );
+
+        await Promise.resolve();
+
+        let updated = false;
+        const updateDone = store
+            .updateConnection(connection.id, { maxAmountSats: 0 })
+            .then((result) => {
+                updated = result.success;
+            });
+
+        await Promise.resolve();
+        expect(updated).toBe(false);
+        expect(connection.maxAmountSats).toBe(5000);
+
+        const lateResponse = await (store as any).withGlobalHandler(
+            connection.id,
+            async () => ({
+                result: { should: 'not-run' },
+                error: undefined
+            })
+        );
+        expect(lateResponse.error?.code).toBe('UNAUTHORIZED');
+
+        releaseHandler({ result: { ok: true } });
+        await handlerDone;
+        await updateDone;
+
+        expect(updated).toBe(true);
+        expect(connection.maxAmountSats).toBe(0);
+    });
+
+    it('defers relay release when updateConnection awaited in-flight handlers', async () => {
+        jest.useFakeTimers();
+        const store = buildStore();
+        jest.spyOn(store as any, 'subscribeToConnection').mockResolvedValue(
+            undefined
+        );
+        jest.spyOn(
+            store as any,
+            'unsubscribeFromConnection'
+        ).mockImplementation(() => undefined);
+        jest.spyOn(store as any, 'generateConnectionSecret').mockReturnValue({
+            connectionUrl: `nostr+walletconnect://${SERVICE_PUB}?relay=${encodeURIComponent(
+                NEW_RELAY
+            )}`,
+            connectionPrivateKey: OLD_PRIVKEY,
+            connectionPublicKey: OTHER_PUBKEY
+        });
+        jest.spyOn(store as any, 'storeClientKeys').mockResolvedValue(
+            undefined
+        );
+        jest.spyOn(store as any, 'deleteClientKeys').mockResolvedValue(
+            undefined
+        );
+        (store as any).nwcWalletServices.set(OLD_RELAY, { close: jest.fn() });
+        (store as any).nwcWalletServices.set(NEW_RELAY, {
+            connected: true,
+            close: jest.fn()
+        });
+        (store as any).publishedRelays.add(OLD_RELAY);
+        (store as any).publishedRelays.add(NEW_RELAY);
+        const connection = seedConnection(store, {
+            expiresAt: new Date(Date.now() + 60_000)
+        });
+
+        let releaseHandler: (value: { result: { ok: boolean } }) => void = () =>
+            undefined;
+        const handlerDone = (store as any).withGlobalHandler(
+            connection.id,
+            () =>
+                new Promise((resolve) => {
+                    releaseHandler = resolve;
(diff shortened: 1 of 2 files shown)


kaloudis (commented):
Reviewed the diff against `withGlobalHandler` / `awaitInFlightHandlers` / the #4443 delete path. This implements what #4486 asked for, and the core mechanics check out:

- The drain cannot miss a handler: `withGlobalHandler` registers its gate promise synchronously before any checks, so handlers that entered before the guard was set are seen by `awaitInFlightHandlers`, and later arrivals are rejected at the pending-set check, including the re-check after the `markConnectionUsed` await.
- The deferral rule matches delete exactly: `hadActiveSubscription` captured before the unsubscribe, defer when `hadInFlight || hadActiveSubscription`, immediate release otherwise.
- No self-deadlock: `updateConnection` is only called from `AddOrEditNWCConnection`, never from inside a NIP-47 handler.
- The old relay's deferred release survives the final re-subscribe, since `subscribeToConnection` only cancels the scheduled release for the connection's (new) relay.
- Rename covers all five references; guard cleanup in `finally` is correct on every throw path; test 2 correctly pre-seeds `publishedRelays` so the fake-timer run skips the 500ms publish sleep.

One round-2 item and two follow-up suggestions:

**1. Relay-change failure path now strands the connection unsubscribed (please fix in this PR).** The unsubscribe moved from after the fallible rotation work to before it. If `loadWalletServiceKeys`, `generateConnectionSecret`, or `storeClientKeys` throws, the catch returns `{ success: false }` with the connection already unsubscribed from the old relay: dead until restart or a successful retry, even though `relayUrl` is unchanged. Pre-PR, a `storeClientKeys` failure left the old subscription running. The early unsubscribe is not load-bearing for the drain (the pending set already rejects new requests), so either restore the old ordering (unsubscribe after the keychain work, before the drain) or re-subscribe on the failure path. The comment "a failed attempt leaves relayUrl unchanged and retries still detect the change" is now only half true; please update it with whichever fix lands. A test for "rotation failure leaves the connection subscribed" would lock this in.

**2. Update windows reuse the delete rejection (fine as a follow-up).** A budget edit that coincides with a client request returns UNAUTHORIZED with "connection not found". For delete that is semantically right; for an update it is a transient-busy condition, and a client that treats UNAUTHORIZED as revocation may drop the pairing over a routine edit. The relay-change window can span publish retries with backoff, so it is not always tiny. A `Map<connectionId, 'delete' | 'update'>` returning INTERNAL_ERROR (retryable) for update windows would be safer. #4486 did not specify this, so no objection to deferring.

**3. Concurrent delete + update on the same id can drop the guard early (nit).** Both operations add the same id to one Set, and whichever finishes first removes it in its `finally`, unguarding the other mid-mutation. The shape predates this PR but now has two entry points. A refcount map would close it; at minimum worth a comment.

Also: the PR template checkboxes are all unchecked (CI confirms the four attestations pass, and the category should be Bug fix). Before merge this also wants an NWC device smoke test: budget edit while a payment settles, and a relay change on an active connection.

Approach ACK. The guard pattern is applied correctly and consistently and the tests mirror the #4443 patterns well; item 1 is the only blocker.

kaloudis (approved):

kaloudis (commented):
Round 2 verified at 426f580.

**Item 1 (blocker) is resolved.** `unsubscribeFromConnection` now runs only after the fallible rotation work (`loadWalletServiceKeys`, `generateConnectionSecret`, `storeClientKeys`) and the publish attempt, so a failed rotation leaves the old subscription live and `relayUrl` unchanged, and a retry still detects the relay change and returns a fresh pairing URL. The rewritten comment matches the new ordering, and the new assertions in the store-key-failure test (both `unsubscribeFromConnection` and `subscribeToConnection` uncalled) lock in exactly the stranded-subscription regression. The deferred-release test exercises the in-flight path correctly under fake timers.

One behavior note from the reorder, not blocking: requests arriving during the publish-with-retry window used to be dropped because the connection was already unsubscribed; now they are answered UNAUTHORIZED over the still-live old subscription. That makes round-1 item 2 slightly more user-visible, so I filed it, together with the item 3 shared-guard nit, as #4503 to keep both from getting lost.

Re-checked at head: rename covers every reference with no stragglers, guard cleanup in `finally` holds on all throw paths, the deferral rule still mirrors delete, and CI is green across test/tsc/lint/prettier.

utACK 426f580

Before merge, please confirm the two device smoke tests from round 1: a budget edit while a payment settles, and a relay change on an active connection with a live client attached.

PR #4501, opened 2026-08-28
components: extract shared swipeable wallet row shell

TheSeydiCharyyev:
# Description

Relates to issue: **none** — follow-up to #4433, asked for in that review ("please do open it").

`LightningSwipeableRow`, `OnchainSwipeableRow` and `EcashSwipeableRow` each carried a copy of the same shell: the `Swipeable` setup, the open/closed state, the ref plumbing and, since #4433, the accessibility wiring. That PR copied the block a third time. This PR moves the shared parts into one place: +381/−553 across 5 files, no new dependencies.

Four commits:

**components: extract shared swipeable wallet row shell** — two new files under `components/LayerBalances/`:

- `SwipeableRowContainer` — the shell: gesture config, `expanded` state driven by `onSwipeableWillOpen`/`onSwipeableWillClose`, `accessibilityRole="button"`, `accessibilityState` and the `expand`/`collapse` accessibility actions. It hands `open` and `close` back to the row (`close` as a `renderLeftActions` argument, `open` as the `onPress` argument), so a row can still close itself from one of its own actions and open itself from a tap.
- `SwipeableRowAction` — the reveal animation and layout of one action behind a row. Each row now passes the icon and the press handler for every action directly. Before, the handler and the icon were picked by comparing the action label against `localeString(...)` values; that indirection is gone.

**components: announce disabled locked wallet rows** — nit from the #4433 review: locked rows that keep a press handler now carry `accessibilityState={{ disabled }}` next to their button role.

**components: keep action styles in SwipeableRowAction** — from the review: the style sheet had a single consumer, so it lives in `SwipeableRowAction.tsx` again and the separate `swipeableRowStyles.ts` is gone.

**components: tidy swipeable row imports and x comment** — two review nits: the Ecash SVG imports are one group again, as on master, and the comment on `x` now says what it is (a translateX offset), which the `x={200}` call sites did not match.

No behavior change is intended. The accessibility wiring from #4433 moved into the container as-is.

### How this was checked

Components have no unit tests in this repo (coverage is `utils/`, `models/`, `lndmobile/`), so the evidence is manual. On the emulator, on this branch rebased on current master, for both the Lightning and the On-chain row: the row renders as before and is one `android.widget.Button` node with its label (`Lightning, 250,000, sats`); tap-to-open reveals the actions; swipe-to-close hides them; swipe-to-open reveals them again; pressing Receive opens the Receive screen, and after going back the row is closed. Same as master before the change.

Not exercised: the `disabled` state on locked rows (second commit) — those rows only appear in `PaymentMethodList`, which my stub setup does not reach; the change there is one prop next to the existing button role.

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [x] LndHub — local stub server, only to bring the wallet screen up; the change does not touch any backend

### Locales

### Other:


components/LayerBalances/swipeableRowStyles.ts
kaloudis, line 1: We only pull this into one file. Leave the styles in `SwipeableRowAction.tsx`

kaloudis (changes requested):

components/LayerBalances/swipeableRowStyles.ts
TheSeydiCharyyev: Done in 741fead3. The style sheet was only used by SwipeableRowAction, so it is back at the bottom of that file, matching the other components, and swipeableRowStyles.ts is removed. PR description updated to match.

components/LayerBalances/EcashSwipeableRow.tsx
```diff
@@ -21,6 +18,9 @@ import { cashuStore } from '../../stores/Stores';
 import SyncStore from '../../stores/SyncStore';
 
 import MintToken from '../../assets/images/SVG/MintToken.svg';
+import SwipeableRowAction from './SwipeableRowAction';
+import SwipeableRowContainer from './SwipeableRowContainer';
```
kaloudis, line 21: nit: these two imports split the SVG group here (MintToken above, Mint/Receive/Send below). Lightning and Onchain keep them in a group of their own; same here would be cleaner.

components/LayerBalances/SwipeableRowAction.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react';
+import { Animated, StyleSheet, Text, View } from 'react-native';
+import { RectButton } from 'react-native-gesture-handler';
+
+import { themeColor } from './../../utils/ThemeUtils';
+
+interface SwipeableRowActionProps {
+    text: string;
+    // Width of the whole action strip: the action slides in from there
```
kaloudis: nit: true for most call sites, but not the legacy x={200} ones (the Ecash strip is 280 wide). Something like "translateX offset the action slides in from" would be strictly accurate.

kaloudis (commented):
Two non-blocking nits from a read-through.

components/LayerBalances/EcashSwipeableRow.tsx
TheSeydiCharyyev: Done in <sha>. The container imports moved above the SVG group, so MintToken, Mint, Receive and Send are together again in the master order.

components/LayerBalances/SwipeableRowAction.tsx
TheSeydiCharyyev: Done in <sha>, reworded to "translateX offset the action slides in from". You are right about the x={200} call sites — the old wording was only true where x equals the strip width.

kaloudis (approved):

PR #4502, opened 2026-08-28
Receive: force LSP on when creating wrapped invoice

kaloudis:
# Description

|LSP JIT Settings|Receive view from 'create a wrapped invoice'|
|-|-|
|Simulator Screenshot - iPhone 17 Pro - 2026-08-28 at 09 28 53|Simulator Screenshot - iPhone 17 Pro - 2026-08-28 at 09 28 58|

The "Create a wrapped invoice" item in LSP (Flow 2.0) settings navigated to the Receive view with no params, so the screen simply used the persisted `enableLSP` setting. If the user had the LSP toggled off, tapping "Create a wrapped invoice" landed them on a screen that would generate a plain, unwrapped invoice, contradicting the button they just tapped.

This PR:

- passes a new `forceLsp` route param (following the existing `forceLn`/`forceOnChain` convention) from the LSP settings entry, along with `selectedIndex: 1` so the user lands on the Lightning tab, since wrapped invoices are Lightning-specific
- makes Receive activate the LSP for the session when `forceLsp` is set, regardless of the persisted setting; the persisted `enableLSP` setting itself is not modified
- renders the LSP switch disabled (locked in the on position) when entering via this path

Because the override is state-only and flows through `lspIsActive`, downstream behavior follows automatically: the memo field hides, expiry locks to 1 hour, and the amount input's inbound-liquidity check uses the LSP path. The entry point is only reachable when the Flow LSP is configured (the list is gated on `flowLspNotConfigured` in `views/Settings/LSP.tsx`), so the forced-on state cannot conflict with an unconfigured LSP.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Test plan: with the LSP toggle off in LSP settings, tap "Create a wrapped invoice" and verify the Receive screen opens on the Lightning tab with the LSP switch on and locked, then generate an invoice and confirm it is wrapped (payee is the LSP, not the node). Also verify normal Receive entry still respects the persisted setting and the toggle remains interactive there.

### Locales

### Other:

### DIFF
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -117,6 +117,7 @@ interface ReceiveProps {
             autoGenerateChange?: boolean;
             forceLn?: boolean; // when coming from the LN balance slider action
             forceOnChain?: boolean; // when coming from the on-chain balance slider action
+            forceLsp?: boolean; // when coming from Create a wrapped invoice in LSP settings
             account: string;
             addressType?: string;
             selectedIndex: number;
@@ -307,8 +308,10 @@ export default class Receive extends React.Component<
 
         const expirationIndex = expirationIndexFromSeconds(newExpirySeconds);
 
+        const forceLsp = !!route.params?.forceLsp;
+
         const lspIsActive =
-            settings?.enableLSP &&
+            (settings?.enableLSP || forceLsp) &&
             BackendUtils.supportsFlowLSP() &&
             !flowLspNotConfigured;
 
@@ -339,7 +342,7 @@ export default class Receive extends React.Component<
             routeHints,
             ampInvoice,
             blindedPaths,
-            enableLSP: settings?.enableLSP,
+            enableLSP: settings?.enableLSP || forceLsp,
             lspIsActive,
             flowLspNotConfigured
         });
@@ -2094,6 +2097,11 @@ export default class Receive extends React.Component<
                                                             <Switch
                                                                 value={
                                                                     enableLSP
+                                                                }
+                                                                disabled={
+                                                                    !!route
+                                                                        .params
+                                                                        ?.forceLsp
                                                                 }
                                                                 onValueChange={async () => {
                                                                     this.setState(
--- a/views/Receive.tsx
+++ b/views/Receive.tsx
@@ -282,6 +282,7 @@ export default class Receive extends React.Component<
             InvoicesStore,
             SettingsStore,
             LightningAddressStore,
+            LSPStore,
             NodeInfoStore,
             route
         } = this.props;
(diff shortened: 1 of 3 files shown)


views/Receive.tsx
```diff
@@ -2095,6 +2098,11 @@ export default class Receive extends React.Component<
                                                                 value={
                                                                     enableLSP
                                                                 }
+                                                                disabled={
+                                                                    !!route
+                                                                        .params
+                                                                        ?.forceLsp
+                                                                }
                                                                 onValueChange={async () => {
```
ajaysehwal, line 2115: @kaloudis, nice clean fix — I traced it end-to-end and `forceLsp` correctly threads through to the actual `noLsp` param on invoice creation, not just the UI.

One defense-in-depth thought: the switch’s `onValueChange` doesn’t check `forceLsp` itself and relies on `disabled` to prevent the callback. RN’s native Switch should reliably avoid firing `onValueChange` when disabled, so this likely won’t happen in practice.

But if it ever did, it would call `updateSettings({ enableLSP: false })` and persist a change to the real setting, which conflicts with the PR’s guarantee that `forceLsp` doesn’t modify the persisted setting.

Might be worth adding a one-line early return in the handler so that guarantee doesn’t depend solely on the Switch’s disabled behavior.

shubhamkmr04 (commented):
https://github.com/user-attachments/assets/71dcca4a-7d1c-4722-ac9d-50f6ae0b8d18

wrapped invoice is NOT getting generated at all when we force LSP from settings to receive view.
as seen in the video, there is no LSP card showing on invoice QR view when we try to create an invoice while tapping 'Create a wrapped invoice' in settings. also i inspected the invoice separately as well, its a unwrapped one

kaloudis:
@shubhamkmr04 good catch, confirmed. The forced flag only overrode Receive state, while InvoicesStore gates both wrap steps on the persisted enableLSP, so the wrap was silently skipped. Fixed in d4809441c5d704270e4fa7282eef56630aa397df: forceLsp now threads through to both gates, and Receive fetches LSP info on mount when forced (Wallet only fetches it when the setting is on).

views/Receive.tsx
kaloudis: Good call, guard added in d4809441c5d704270e4fa7282eef56630aa397df. Turned out the persisted setting also gated the wrap inside InvoicesStore, so the same commit threads forceLsp through to the store.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4510, opened 2026-08-29
Wallet: halt fetchData when LNC or NWC connection fails

kaloudis:
# Description

Fixes the "Lightning address error: Pubkey not defined" error thrown on failed connections.

Root cause: in `Wallet.tsx` `fetchData`, the `lightning-node-connect` and `nostr-wallet-connect` branches do not bail out when `connect()` reports an error. Every other backend branch returns early on connection failure, but these two only skip their own data fetching (`if (!error)`) and then fall through to the post-connection work. With `NodeInfoStore.nodeInfo` empty, the ZEUS Pay status call posts `pubkey: undefined` to the server, which responds with `Pubkey not defined`; `LightningAddressStore.status()` stores that as its error state and re-throws. The dead connection also triggered pointless follow-up work (LSP info fetch, swaps fetch) before the connecting status was finally cleared.

Changes:
- LNC and NWC branches now return early when `connect()` reports an error, clearing the connecting status first so the error screen (driven by `SettingsStore.error`, which `connect()`/`connectNWC()` already set) is shown, matching the LND (REST) branch's failure handling.
- Their post-connect fetch `catch` blocks now also clear the connecting status before returning, instead of stranding the connecting spinner.
- The Lightning address block is additionally guarded on `NodeInfoStore.nodeInfo?.identity_pubkey`: every ZEUS Pay call authenticates with the node pubkey, so without it the calls can only fail server-side.

Only the LNC path can currently reach the Lightning address block on a failed connection (`supportsLightningAddress()` is false for NWC), but the NWC branch had the same fall-through defect, so both are fixed.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Wallet/Wallet.tsx
+++ b/views/Wallet/Wallet.tsx
@@ -1185,32 +1185,39 @@ export default class Wallet extends React.Component<WalletProps, WalletState> {
             if (connecting) {
                 error = await connect();
             }
-            if (!error) {
-                try {
-                    await BackendUtils.checkPerms();
-                    await NodeInfoStore.getNodeInfo();
-                    if (BackendUtils.supportsAccounts())
-                        await UTXOsStore.listAccounts();
-                    await BalanceStore.getCombinedBalance();
-                    if (BackendUtils.supportsChannelManagement())
-                        await ChannelsStore.getChannels();
-                } catch (connectionError) {
-                    console.log('LNC connection failed:', connectionError);
-                    return;
-                }
+            if (error) {
+                setConnectingStatus(false);
+                return;
+            }
+            try {
+                await BackendUtils.checkPerms();
+                await NodeInfoStore.getNodeInfo();
+                if (BackendUtils.supportsAccounts())
+                    await UTXOsStore.listAccounts();
+                await BalanceStore.getCombinedBalance();
+                if (BackendUtils.supportsChannelManagement())
+                    await ChannelsStore.getChannels();
+            } catch (connectionError) {
+                console.log('LNC connection failed:', connectionError);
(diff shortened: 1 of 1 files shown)


shubhamkmr04 (commented):
One thing on the LNC catch: `setConnectingStatus(false)` now runs unconditionally, including on invocations where `connecting` is false (pull-to-refresh, app resume). Since `connecting` is snapshotted at the top of `getSettingsAndNavigate`, a stale refresh whose `getNodeInfo()` rejects on a dead session can clear the connecting flag for a *newer* wallet connection that's still in flight — switch wallets right after a refresh and the spinner drops early. Every other branch in this function already gates on the local `connecting` snapshot (1177, 1190, 1272, 1307, 1325), so `if (connecting) setConnectingStatus(false);` would keep it consistent.

Also, `NodeInfoStore.handleGetNodeInfoError()` there looks redundant — `getNodeInfo()` is the only call in that `try` that can reject, and it already calls `handleGetNodeInfoError()` itself before rejecting (`stores/NodeInfoStore.ts:107`), with a staleness check this call site doesn't have.

Non-blocking either way; the fall-through fix itself and the `identity_pubkey` guard both look right to me.

PR #4518, opened 2026-08-29
Receive: generate imported account addresses under the account's key scope

### LINKED ISSUE #2223: Address Generated in Zeus for Airgapped xPub account does not match SeedSigner/Sparrow
### Describe the bug

After adding an account to the LND Embedded node for an air gapped hardware wallet (SeedSigner in this case) the receive address generated does not match SeedSigner/Sparrow/Import. The Import screen in Zeus shows the correct receive and change addresses, but when I tap the "receive" button under/next to the account I get a bitcoin address I don't recognize at all.

Example Import:
IMG_3396

First Receive Address:
IMG_3398

The receive address looks like a Segwit Address but I imported a taproot wallet.

### Reproduce

IMG_3395

Then click receive:

image

Click create invoice (it's showing an error related to account 'SeedSigner' not found (see screenshot):
image

Address show is not valid
IMG_3398

### ZEUS version

v0.9.0

### Node interface

Embedded LND

### Network

Clearnet

### Device

iPhone 13 mini

### Device operating system

17.4.1

### Log output

_No response_

### LINKED ISSUE #2932: External account no addresses
### Describe the bug

I have three external watch only accounts and I cannot seem to generate receive addresses for them or even view their addresses. I confirmed the addresses were correct when importing the watch only accounts. When I click "Receive" on them, I get the following message with nothing below it:

Image

### Reproduce

1. Import external watch only account
2. attempt to generate a receive address for it

### ZEUS version

v0.10.0

### Node interface

Embedded LND

### Network

Clearnet

### Device

Pixel 7

### Device operating system

GrapheneOS 2025032100

### Log output

```shell

```

kaloudis:
# Description

Relates to issue: **ZEUS-2223**, **ZEUS-2932**

Fixes #2223
Fixes #2932

Full root cause analysis: https://github.com/ZeusLN/zeus/issues/2223#issuecomment-5465433173

Receiving to imported (watch-only) accounts was broken in two ways:

1. **"account not found" for taproot and nested segwit imports.** lnd never infers an account's key scope from its name: `NewAddress`/`NextAddr` default to `WITNESS_PUBKEY_HASH` (BIP84) when no type is given, and btcwallet rejects the request when the account lives under another scope. The ZEUS-2396 workaround of stripping the address type from account requests therefore only ever worked for BIP84 (zpub) imports; taproot (BIP86) and nested segwit (BIP49) accounts could not receive at all. It also discarded the correct address type that the OnChainAddresses view already passes.
2. **Unified invoices silently used the default account.** `createUnifiedInvoice` and the Create Invoice button never passed the account, so the on-chain leg of the invoice came from the node's own hot wallet while the Receive screen labeled the address with the imported account's name. This is the unrecognized segwit address in the ZEUS-2223 screenshots, and funds sent to it landed in the embedded node's wallet instead of the user's airgapped wallet.

Changes:

- `utils/LndUtils`: new `toWalletrpcAddressTypeName` / `toWalletrpcAddressTypeNum` helpers. walletrpc.AddressType names and numbering differ from lnrpc.AddressType (walletrpc `WITNESS_PUBKEY_HASH=1` vs lnrpc `0`), REST/LNC decode enum names while the embedded proto decode yields raw numbers, so account address types need explicit normalization for each RPC's enum space.
- `stores/InvoicesStore`: `getNewAddress` honors account-derived walletrpc address types and only strips types that don't come from the account (the settings preference that motivated ZEUS-2396). `createUnifiedInvoice` accepts an `account` so the on-chain leg comes from the requested account, never silently from `default`.
- `stores/UTXOsStore`: new `getAccountAddressType` resolves an account's type from ListAccounts (cache first). The post-import address pre-generation now normalizes types through the walletrpc helpers instead of `walletrpc.AddressType[...]` lookups that sent walletrpc numbers into lnrpc `NewAddress` on REST.
- `views/Receive`: resolves the account's address type on mount and threads it through auto-generated receive/change addresses, tab-switch generation, and all `createUnifiedInvoice` call sites. Non-default account requests never fall back to the picker/settings address type.
- `backends/EmbeddedLND` + `backends/LightningNodeConnect`: `NextAddr` change-address types are now converted into walletrpc enum space. The embedded path previously fed enum-name strings into the protobuf encoder (coerced to NaN, serialized as 0/UNKNOWN); LNC used a map with lnrpc numbering, so a nested segwit request became walletrpc `WITNESS_PUBKEY_HASH`.

Behavior notes:

- BIP84 (zpub) imported accounts behave exactly as before.
- If an account's type cannot be resolved (e.g. ListAccounts unavailable), the request is sent without a type, which matches today's behavior and surfaces lnd's error rather than showing an address from the wrong wallet.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

New tests: walletrpc conversion helpers in `utils/LndUtils.test.ts`; account address-type handling in `stores/InvoicesStore.test.ts` (honors account-derived types, normalizes embedded numeric types, drops settings-preference types, default account untouched, errors surfaced).

- [ ] Android
- [x] iOS

On-device
- [ ] LDK Node
- [x] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Testing performed (iOS Simulator, embedded LND mainnet): imported the taproot xpub from the ZEUS-2223 report (verified against an independent local BIP86 derivation, which also matches the reporter's screenshots) plus the BIP84 and BIP49 spec test vector zpub/ypub with the existing-account toggle and birthday height. Verified: swipe-receive generates the exact expected address for all three scopes with no "account not found" errors; unified invoice creation from the account context puts the account's next address in the BIP21 URI; the 50+50 import pre-generation completes for all scopes, with receive and change lists matching the derivations (including nested segwit change, which previously hit the protobuf enum trap); the OnChainAddresses regenerate path works; default-account receive is unchanged. No platform-specific code is touched. 

Additional testing (Polar regtest, iOS Simulator): over LND (REST) and Lightning Node Connect (litd via Polar, admin session, default mailbox), imported regtest accounts derived from the BIP39 spec mnemonic (verified against independent local derivations): taproot tpub over LNC and BIP84 vpub over REST, both with birthday height and address pre-generation. Swipe-receive matched the exact expected next index on both backends, the LNC change list matched the derivation (exercising the rewritten NextAddr mapping), and an end-to-end receive worked over REST: coins sent from another Polar node to a Zeus-generated watch-only account address, mined, and reflected in the account balance. The LNC testing also surfaced a pre-existing bug fixed in this PR: existing-account imports over LNC always failed because Zeus-side fields (addresses_to_generate, birthday_height) were sent to litd's proto JSON unmarshaler, which rejects unknown fields.

### Locales

### Other:

### DIFF
--- a/stores/InvoicesStore.ts
+++ b/stores/InvoicesStore.ts
@@ -12,6 +12,7 @@ import ChannelInfo from '../models/ChannelInfo';
 import SettingsStore from './SettingsStore';
 import LSPStore from './LSPStore';
 import BackendUtils from '../utils/BackendUtils';
+import { toWalletrpcAddressTypeName } from '../utils/LndUtils';
 import { localeString } from '../utils/LocaleUtils';
 import { errorToUserFriendly } from '../utils/ErrorUtils';
 import LdkNodeInjection from '../ldknode/LdkNodeInjection';
@@ -184,6 +185,7 @@ export default class InvoicesStore {
         routeHints,
         routeHintChannels,
         addressType,
+        account,
         customPreimage,
         noLsp,
         forceLsp,
@@ -198,6 +200,7 @@ export default class InvoicesStore {
         routeHints?: boolean;
         routeHintChannels?: Channel[];
         addressType?: string;
+        account?: string;
         customPreimage?: string;
         noLsp?: boolean;
         forceLsp?: boolean;
@@ -234,11 +237,15 @@ export default class InvoicesStore {
                 }
                 const { rHash, paymentRequest } = result;
                 if (BackendUtils.supportsOnchainReceiving() && !skipOnchain) {
-                    return this.getNewAddress(
-                        addressType
-                            ? { type: addressType, unified: true }
-                            : { unified: true }
-                    )
+                    // ZEUS-2223: the on-chain leg must come from the
+                    // requested account; falling back to the default
+                    // account would silently receive into the node's own
+                    // wallet instead of e.g. an imported watch-only account
+                    const addressRequest: any = { unified: true };
+                    if (addressType) addressRequest.type = addressType;
+                    if (account && account !== 'default')
+                        addressRequest.account = account;
+                    return this.getNewAddress(addressRequest)
                         .then((onChainAddress: string) => {
                             runInAction(() => {
                                 this.onChainAddress = onChainAddress;
@@ -619,10 +626,16 @@ export default class InvoicesStore {
             this.creatingInvoice = true;
             this.error_msg = null;
         }
-        // ZEUS-2396
-        // https://github.com/ZeusLN/zeus/issues/2396
+        // ZEUS-2223 / ZEUS-2932: imported accounts exist under a single key
+        // scope, and lnd requires the request's address type to match it;
+        // it never infers the scope from the account name. Honor an
+        // account-derived (walletrpc) address type; drop anything else (like
+        // the user's preferred type from settings, see ZEUS-2396) so lnd's
+        // default applies instead of a mismatched scope.
         if (params.account && params.account !== 'default') {
-            delete params.type;
+            const accountType = toWalletrpcAddressTypeName(params.type);
+            if (accountType) params.type = accountType;
+            else delete params.type;
         }
         this.onChainAddress = null;
         return BackendUtils.getNewAddress(params)
--- a/stores/InvoicesStore.ts
+++ b/stores/InvoicesStore.ts
@@ -629,16 +629,18 @@ export default class InvoicesStore {
         // ZEUS-2223 / ZEUS-2932: imported accounts exist under a single key
         // scope, and lnd requires the request's address type to match it;
         // it never infers the scope from the account name. Honor an
-        // account-derived (walletrpc) address type; drop anything else (like
+        // account-derived (walletrpc) address type; omit anything else (like
         // the user's preferred type from settings, see ZEUS-2396) so lnd's
         // default applies instead of a mismatched scope.
-        if (params.account && params.account !== 'default') {
-            const accountType = toWalletrpcAddressTypeName(params.type);
-            if (accountType) params.type = accountType;
-            else delete params.type;
-        }
+        const { type, ...rest } = params;
+        const resolvedType =
+            rest.account && rest.account !== 'default'
+                ? toWalletrpcAddressTypeName(type)
+                : type;
+        const request =
+            resolvedType === undefined ? rest : { ...rest, type: resolvedType };
         this.onChainAddress = null;
-        return BackendUtils.getNewAddress(params)
+        return BackendUtils.getNewAddress(request)
             .then((data: any) => {
                 const address =
                     data.address ||
@@ -646,8 +648,8 @@ export default class InvoicesStore {
                     data.p2tr ||
                     (data[0] && data[0].address);
                 runInAction(() => {
-                    if (!params.unified) this.onChainAddress = address;
-                    if (!params.unified) this.creatingInvoice = false;
+                    if (!request.unified) this.onChainAddress = address;
+                    if (!request.unified) this.creatingInvoice = false;
                 });
                 return address;
             })
@@ -671,15 +673,15 @@ export default class InvoicesStore {
             this.error_msg = null;
         }
 
-        params.change = true;
+        const request = { ...params, change: true };
 
         this.onChainAddress = null;
-        return BackendUtils.getNewChangeAddress(params)
+        return BackendUtils.getNewChangeAddress(request)
             .then((data: any) => {
                 const address = data.addr;
                 runInAction(() => {
-                    if (!params.unified) this.onChainAddress = address;
-                    if (!params.unified) this.creatingInvoice = false;
+                    if (!request.unified) this.onChainAddress = address;
+                    if (!request.unified) this.creatingInvoice = false;
                 });
                 return address;
             })
--- a/stores/UTXOsStore.ts
+++ b/stores/UTXOsStore.ts
@@ -246,9 +246,20 @@ export default class UTXOsStore {
             data.birthday_height = this.start_height;
         }
 
-        console.log('importAccount req', data);
+        // birthday_height and addresses_to_generate drive the Zeus-side
+        // rescan and address pre-generation above; they aren't part of
+        // lnd's ImportAccountRequest. Embedded (protobufjs) and REST
+        // silently drop unknown fields but LNC's proto JSON unmarshaler
+        // rejects the whole request over them.
+        const {
+            birthday_height: _birthdayHeight,
+            addresses_to_generate: _addressesToGenerate,
+            ...importRequest
+        } = data;
 
-        return BackendUtils.importAccount(data)
+        console.log('importAccount req', importRequest);
+
+        return BackendUtils.importAccount(importRequest)
             .then(async (response: any) => {
                 if (!data.dry_run) {
                     if (this.start_height) {
--- a/stores/UTXOsStore.ts
+++ b/stores/UTXOsStore.ts
@@ -5,11 +5,10 @@ import SettingsStore from './SettingsStore';
 import SyncStore from './SyncStore';
 
 import BackendUtils from '../utils/BackendUtils';
+import { toWalletrpcAddressTypeName } from '../utils/LndUtils';
 
 import Account from '../models/Account';
 import Utxo from '../models/Utxo';
-
-import { walletrpc } from '../proto/lightning';
 
 export const LEGACY_HIDDEN_ACCOUNTS_KEY = 'hidden-accounts';
 export const HIDDEN_ACCOUNTS_KEY = 'zeus-hidden-accounts';
@@ -201,6 +200,29 @@ export default class UTXOsStore {
             });
     };
 
+    // Resolves an imported account's address type (as the walletrpc enum
+    // name) so address requests can match the account's key scope; lnd
+    // rejects requests whose type doesn't match it (ZEUS-2223 / ZEUS-2932).
+    public getAccountAddressType = async (
+        name: string
+    ): Promise<string | undefined> => {
+        const cached = this.accounts.find(
+            (account: any) => account.name === name
+        );
+        const cachedType = toWalletrpcAddressTypeName(cached?.address_type);
+        if (cachedType) return cachedType;
+        try {
+            const data = await BackendUtils.listAccounts();
+            const match = (data?.accounts || []).find(
+                (account: any) => account.name === name
+            );
+            return toWalletrpcAddressTypeName(match?.address_type);
+        } catch (e) {
+            console.error('error resolving account address type', e);
+            return undefined;
+        }
+    };
+
     @action
     public importAccount = (data: any) => {
         this.errorMsg = '';
@@ -235,9 +257,9 @@ export default class UTXOsStore {
                             this.addresses_to_generate_progress = i + 1;
                             await BackendUtils.getNewAddress({
                                 account: this.accountToImport.account.name,
-                                type: walletrpc.AddressType[
+                                type: toWalletrpcAddressTypeName(
                                     this.accountToImport.account.address_type
-                                ]
+                                )
                             }).then((response: any) => {
                                 console.log(
                                     `generated address ${i}`,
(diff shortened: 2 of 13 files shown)


ajaysehwal (commented):
Image

@kaloudis, can we disable the **Import Account** button until all required fields are valid?

kaloudis:
Good catch, done in 84ff30c.

The Import account button is now held until the form can actually produce an import request:

- **Account name** and **Extended public key** must be non-empty (trimmed).
- **Master key fingerprint** stays optional, but a supplied one has to be 8 hex characters. `Base64Utils.reverseMfpBytes` reverses `mfp.match(/.{2}/g)`, so anything that is not a whole number of bytes silently loses its trailing nibble and the account gets imported with a wrong fingerprint. The field is marked in error (red border) while it does not parse, so the disabled button is not a mystery. Every scan path that lands here already produces 8 hex characters (`mfpIntToBytes` pads to 8, the descriptor and string-export parsers take the `[...]` prefix), so this only catches hand-entered values.
- **Block height**, when the existing-account toggle is on. Clearing that field yields `Number("") === 0`, and a 0 birthday height starts a rescan from the genesis block.
- Also held while an import is already in flight, so a double tap cannot fire two dry runs.

Submitted values are trimmed rather than passed through raw.

Deliberately not gating on the xpub prefix: `AddressUtils.xpubFormat` is missing `upub`, so gating on it would block BIP49 testnet imports. Length and base58 validation belong to lnd, which reports them accurately.

Ride-along in bc2b411: `autoCapitalize="none"` and `autoCorrect={false}` on the extended public key and fingerprint fields. They defaulted to sentence capitalization on iOS, so a hand-typed key became `Xpub...` and was rejected. Same treatment as 372ae9b.

No new locale strings needed. The disabled-button pattern matches Add watchtower. `tsc`, `lint`, `prettier` and `test` are clean locally; I have not re-run the device import matrix for these two commits yet, so consider that outstanding.

stores/InvoicesStore.ts
```diff
@@ -614,10 +621,16 @@ export default class InvoicesStore {
...
-        // https://github.com/ZeusLN/zeus/issues/2396
+        // ZEUS-2223 / ZEUS-2932: imported accounts exist under a single key
+        // scope, and lnd requires the request's address type to match it;
+        // it never infers the scope from the account name. Honor an
+        // account-derived (walletrpc) address type; drop anything else (like
+        // the user's preferred type from settings, see ZEUS-2396) so lnd's
+        // default applies instead of a mismatched scope.
         if (params.account && params.account !== 'default') {
-            delete params.type;
+            const accountType = toWalletrpcAddressTypeName(params.type);
```
ajaysehwal: @kaloudis, `getNewAddress` currently mutates the caller's `params` object with `params.type = ...` / `delete params.type`. Whereas the rest of this PR (createUnifiedInvoice's addressRequest, Receive.tsx's buildAddressRequest) builds fresh objects instead of mutating inputs. It would be safer to do the same here, e.g. `const request = { ...params, type: ... }`, so the function doesn't unexpectedly change the object passed by the caller.

ajaysehwal (commented):
@kaloudis, I confirmed that the default-account change-address path (`autoGenerateChange` when `account === 'default'`) still sends the raw picker value (`'0'` / `'1'` / `'4'`) directly to `/v2/wallet/address/next` for remote LND REST, without normalization. This differs from the `NewAddress` (lnrpc) path, which is now fully normalized. This is pre-existing behavior and is unchanged by this PR.

stores/InvoicesStore.ts
kaloudis: Good catch, taken in ab546c58f, and I took it a bit further than the one line.

You are right that no caller is harmed today: `createUnifiedInvoice`'s `addressRequest` and Receive's `buildAddressRequest` both hand over freshly built literals, and `views/Receive.tsx`'s own `getNewAddress` wrapper turned out to have no callers at all. So this was a latent hazard rather than a live bug. But since the store was the one layer still writing through to its input, it was worth fixing properly rather than patching the single line.

Rather than swap `delete` for a spread, I dropped `delete` from the address path entirely:

- `InvoicesStore.getNewAddress` now destructures `type` off and builds the outgoing request, omitting `type` when no account derived walletrpc name resolves. The `.then`/`.catch` blocks read `unified` off the built request too, so they are not straddling two objects.
- `InvoicesStore.getNewChangeAddress` had the same shape one function down with `params.change = true`, so it now builds `{ ...params, change: true }`. `change` has to survive the copy, since `LND.getNewChangeAddress` posts the body straight through to `/v2/wallet/address/next`.
- `LND.getNewAddress` was already safe, since it copied before deleting, but it is the same address type normalization one layer down, and leaving it would have made the sweep half done. It builds its request the same way now.
- Removed the unused `getNewAddress` wrapper in `views/Receive.tsx`.

Omitting the key rather than sending `type: undefined` is equivalent across every backend: LND, LNC and EmbeddedLND all funnel through `toLnrpcAddressType*(undefined) -> undefined`, and CLNRest's switch falls through to `undefined`. No behavior change anywhere, so the device matrix already recorded on this PR still holds.

Added two regression tests asserting the caller's object comes back untouched, one per method.

One related thing I deliberately left out of scope: `BalanceStore.getBlockchainBalance` does `delete accounts.default` on the backend response. It is benign today, since the delete is gated on `set` and the only caller that consumes the return value passes `set=false`, so no reader ever sees the stripped object. Different subsystem from receive, and folding it in would drag balance display back into a device matrix that is already closed here. Happy to do it as a separate PR if you would like it cleaned up.

stores/UTXOsStore.ts
```diff
@@ -224,9 +246,18 @@ export default class UTXOsStore {
...
-        console.log('importAccount req', data);
-
-        return BackendUtils.importAccount(data)
+        // birthday_height and addresses_to_generate drive the Zeus-side
+        // rescan and address pre-generation above; they aren't part of
+        // lnd's ImportAccountRequest. Embedded (protobufjs) and REST
+        // silently drop unknown fields but LNC's proto JSON unmarshaler
+        // rejects the whole request over them.
+        const {
+            birthday_height: _birthdayHeight,
```
shubhamkmr04, line 257: `birthday_height` isn't Zeus-side on the embedded path — it's a real request field there: `proto/walletrpc/walletkit.proto:678` declares it, and `lndmobile/wallet.ts` forwards it to `WalletKitImportAccount`. Upstream lnd stops at `dry_run = 5`, which is why litd rejects it.

Stripping it unconditionally also makes lines 245-246 dead code. `ImportingAccount` sends the real import without a height, so those two lines exist only to re-attach the one captured during the dry run — embedded accounts now get imported with no birthday at all.

Suggest dropping `addresses_to_generate` always, and `birthday_height` only for the backends that can't parse it.

stores/UTXOsStore.ts
```diff
@@ -279,6 +310,9 @@ export default class UTXOsStore {
                         this.error = false;
                         this.success = true;
                     });
+                    // refresh so the new account shows up on the balance
+                    // pane without a manual pull-to-refresh
+                    this.listAccounts();
```
shubhamkmr04, line 326: This refresh never runs on REST or LNC. `rescan` is a Zeus lnd-fork RPC, so it's only defined on `EmbeddedLND`; `BackendUtils.call` returns `false` for the others and line 297 does `false.then(...)`.

Any import with a birthday height throws there and lands in the outer catch: `errorMsg` shows `TypeError: false.then is not a function`, the import is reported as failed even though it succeeded, and `accountToImport = null` makes `ImportingAccount`'s render destructure throw. The toggle that sets the height isn't backend-gated, so this is reachable on the LNC path this PR fixes.

Pre-existing, but worth guarding since it swallows the new `listAccounts()`:

```ts
const rescanRequest = BackendUtils.rescan({ start_height: this.start_height });
if (rescanRequest) {
    rescanRequest
        .then((response: any) => console.log('rescan resp', response))
        .catch((err: Error) => console.log('rescan err', err));
}
```

shubhamkmr04 (commented):
The enum split and threading the account's key scope through `getNewAddress` / `NextAddr` / `createUnifiedInvoice` look right, and the LNC `NEXT_ADDR_MAP` removal is a genuine fix on its own. Two notes on the import path inline.

stores/UTXOsStore.ts
kaloudis: Confirmed on both counts, and it goes further than dead code. I traced the field through the fork: `walletkit_server.go` resolves `req.BirthdayHeight` to a block stamp unconditionally, and btcwallet's `maybeUpdateBirthdayBlock` only ever moves the wallet birthday backward. With the field stripped, height 0 resolves to the genesis block, so every embedded real import was dragging the whole wallet's birthday back to genesis, not just importing without one. Dry runs were unaffected (the fork derives preview addresses without persisting).

Taken in 82d0977e4, split exactly as you suggested: `addresses_to_generate` is stripped everywhere (no backend knows it), and `birthday_height` now rides the import request only where the fork extensions exist, behind a new `supportsAccountImportRescan` capability (embedded only, declared false on LND and LNC per the usual pattern). The dry-run-captured re-attach is live again, reworked so it no longer mutates the caller's `data`. Regression tests in the new `stores/UTXOsStore.test.ts` cover the keep, the strip, the re-attach, and the no-mutation contract.

One adjacent pre-existing bug this surfaced, which Zeus can't fix client-side: proto3 can't distinguish an unset `uint32` from 0, so importing a *new* account on embedded (existing-account toggle off) has always sent 0 and reset the wallet birthday to genesis. That needs the fork to treat 0 as unset; filed as #4640.

stores/UTXOsStore.ts
kaloudis: Confirmed. Fixed in 82d0977e4, gated on the same `supportsAccountImportRescan` capability as the birthday fix rather than the falsy return, so the fork's two import extensions share one flag and the skip is self-documenting. Regression test asserts a non-fork import with a birthday completes with `success` set, no rescan call, and the `listAccounts` refresh intact.

One correction this forced on my earlier testing claims: the crash is mostly masked because the outer catch resolves the promise, so ImportingAccount's `popTo('Wallet')` still navigates away immediately. That's why my Polar round over REST and LNC looked clean: the node-side import, pre-generation, and address verification all genuinely passed, but the store must have ended in this error state unnoticed, so the success message and the post-import `listAccounts` refresh were never actually exercised there. I'll re-run the birthday-height import matrix (embedded + LNC) on-device before this merges and report back.

Worth noting the residual gap once this is guarded: remote backends get no post-import rescan at all, since upstream lnd has no rescan RPC. Pre-generated addresses are watched going forward, but historical transactions on a remote node still need an out-of-band rescan. Fork-only limitation, out of scope here.

views/Tools/Accounts/ImportAccount.tsx
ajaysehwal, line 1: @kaloudis, one UX gap here: the `existing_account` toggle in `ImportAccount.tsx` isn’t gated by `BackendUtils.supportsAccountImportRescan()`, but the copy says:

> “If importing an existing account, ZEUS will initiate a rescan from the block height specified below”

This screen is also available for LND-REST and LNC, where `supportsAccountImportRescan()` is `false`. In those cases, `UTXOsStore.importAccount` strips `birthday_height` and skips the rescan, so a user can enable “Existing account”, enter a block height, and believe a rescan will happen when it actually won’t.

Since this is already marked “EXPERIMENTAL”, I think the cleanest approach would be to hide the toggle and block-height field when `supportsAccountImportRescan()` is false, so the option only appears where it actually does something.

kaloudis: Good catch, fixed in e4a298db5. One clarification so this doesn't read as a regression from this PR: the toggle and the rescan-promising copy already exist on master, where the behavior is worse (LNC rejects the whole import request over the unknown `birthday_height` field, and the unconditional `rescan()` call hits the `call()`-returns-false trap, reporting successful imports as failed). The capability gating in 82d0977e4 fixed that breakage; this commit addresses the leftover dishonest copy.

I went with a variant of your suggestion: rather than hiding the toggle entirely, only the block height field and rescan copy are gated on `supportsAccountImportRescan()`. The address pre-generation still runs on remote backends and is worth keeping there, since it makes the node watch the first N external and change addresses of an existing account beyond lnd's lookahead. A new note on those backends now states that past transactions won't be detected.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4520, opened 2026-08-30
views: NWC: show in-flight wait on delete/regenerate

### LINKED ISSUE #4487: NWC: connection delete/regenerate can block ~150s behind an in-flight payment with only a spinner
### Summary

Follow-up to PR #4443 (split out from review discussion: https://github.com/ZeusLN/zeus/pull/4443#issuecomment-5350308753).

With #4443, `deleteConnection` correctly awaits in-flight NIP-47 handlers before removing the connection, so settled spend and activity are persisted. The await is bounded by the payment queue wait plus the payment timeout, which is roughly 150s worst case behind a slow `pay_invoice`.

### Problem

During that window the delete and regenerate flows in `views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx` show only a spinner. A user deleting a connection that happens to have a payment in flight sees the app apparently hang for up to a couple of minutes. The likely reaction is force-killing the app, which is exactly the case the fix is trying to protect: killing the process mid-settle is what loses the budget debit and activity record.

### Suggested fix

When delete or regenerate is blocked on in-flight work, surface a waiting state instead of a bare spinner:

- disable the button and show text along the lines of "waiting for an in-flight request to finish" (new locale string)
- apply to both the delete confirmation flow and the regenerate flow, since regenerate calls `deleteConnection` first

A simple way to plumb it: an observable on `NostrWalletConnectStore` set while `awaitInFlightHandlers` is pending for a delete, read by the view.

ajaysehwal:
# Description

Relates to issue: **#4487**

Follow-up to #4443: `deleteConnection` correctly waits for in-flight NIP-47 handlers before removing a connection, but delete and regenerate in `NWCConnectionDetails` showed only a spinner during that window (up to ~125s behind a slow `pay_invoice`). Users may force-kill the app, which is exactly what the handler-await fix is meant to prevent.

This PR:
- Adds `waitingForInFlightHandlers` on `NostrWalletConnectStore`, set while `deleteConnection` awaits in-flight handlers
- Shows spinner + “Waiting to finish a pending request...” above the action buttons during delete or regenerate
- Disables both buttons while either operation is in progress

Regenerate is covered because it calls `deleteConnection` before `createConnection`.

Simulator Screenshot - iPhone 17 Pro - 2026-08-31 at 17 20 30

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -2047,6 +2047,7 @@
     "views.Settings.NostrWalletConnect.connectionSecret": "Connection Secret",
     "views.Settings.NostrWalletConnect.connectionSecretDescription": "Scan this QR code or copy the connection secret in your Nostr app to link this connection.",
     "views.Settings.NostrWalletConnect.waitingForAppToConnect": "Waiting for app to connect...",
+    "views.Settings.NostrWalletConnect.waitingForInFlightRequest": "Waiting for an in-flight request to finish...",
     "views.Settings.NostrWalletConnect.noEventReceivedTimeoutDescription": "We didn't receive any event from your Nostr client, but you can still close this screen and use your connection. The connection will work once your Nostr client sends a request.",
     "views.Settings.NostrWalletConnect.connectionTimeout": "Connection Timeout",
     "views.Settings.NostrWalletConnect.keepWaiting": "Keep Waiting",
--- a/locales/en.json
+++ b/locales/en.json
@@ -2047,7 +2047,7 @@
     "views.Settings.NostrWalletConnect.connectionSecret": "Connection Secret",
     "views.Settings.NostrWalletConnect.connectionSecretDescription": "Scan this QR code or copy the connection secret in your Nostr app to link this connection.",
     "views.Settings.NostrWalletConnect.waitingForAppToConnect": "Waiting for app to connect...",
-    "views.Settings.NostrWalletConnect.waitingForInFlightRequest": "Waiting for an in-flight request to finish...",
+    "views.Settings.NostrWalletConnect.waitingToFinishPendingRequest": "Waiting to finish a pending request...",
     "views.Settings.NostrWalletConnect.noEventReceivedTimeoutDescription": "We didn't receive any event from your Nostr client, but you can still close this screen and use your connection. The connection will work once your Nostr client sends a request.",
     "views.Settings.NostrWalletConnect.connectionTimeout": "Connection Timeout",
     "views.Settings.NostrWalletConnect.keepWaiting": "Keep Waiting",
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -1170,6 +1170,7 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         await new Promise((resolve) => setImmediate(resolve));
         expect(deleted).toBe(false);
         expect(store.connections).toHaveLength(1);
+        expect(store.waitingForInFlightHandlers).toBe(true);
 
         const lateResponse = await (store as any).withGlobalHandler(
             connection.id,
@@ -1186,6 +1187,7 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
 
         expect(deleted).toBe(true);
         expect(store.connections).toHaveLength(0);
+        expect(store.waitingForInFlightHandlers).toBe(false);
     });
 
     it('defers relay release when delete awaited in-flight handlers', async () => {
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -437,6 +437,7 @@ export default class NostrWalletConnectStore {
                 this.walletServiceKeys = null;
                 this.errorMessage = '';
                 this.waitingForConnection = false;
+                this.waitingForInFlightHandlers = false;
                 this.currentConnectionId = undefined;
                 this.connectionJustSucceeded = false;
                 this.lastConnectionAttempt = 0;
@@ -667,20 +668,16 @@ export default class NostrWalletConnectStore {
                     connectionId
                 );
                 this.makeInvoiceTimestampsByConnection.delete(connectionId);
-                const hasInFlightHandlers =
-                    !!this.inFlightHandlersByConnection.get(connectionId)?.size;
-                if (hasInFlightHandlers) {
+                let hadInFlight = false;
+                if (this.inFlightHandlersByConnection.get(connectionId)?.size) {
                     runInAction(() => {
                         this.waitingForInFlightHandlers = true;
                     });
-                }
-                let hadInFlight = false;
-                try {
-                    hadInFlight = await this.awaitInFlightHandlers(
-                        connectionId
-                    );
-                } finally {
-                    if (hasInFlightHandlers) {
+                    try {
+                        hadInFlight = await this.awaitInFlightHandlers(
+                            connectionId
+                        );
+                    } finally {
                         runInAction(() => {
                             this.waitingForInFlightHandlers = false;
                         });
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -147,6 +147,7 @@ export default class NostrWalletConnectStore {
         new Map();
     @observable private publishedRelays: Set<string> = new Set();
     @observable public waitingForConnection = false;
+    @observable public waitingForInFlightHandlers = false;
     @observable public currentConnectionId?: string;
     @observable public connectionJustSucceeded = false;
     @observable private walletServiceKeys: WalletServiceKeys | null = null;
@@ -666,9 +667,25 @@ export default class NostrWalletConnectStore {
                     connectionId
                 );
                 this.makeInvoiceTimestampsByConnection.delete(connectionId);
-                const hadInFlight = await this.awaitInFlightHandlers(
-                    connectionId
-                );
+                const hasInFlightHandlers =
+                    !!this.inFlightHandlersByConnection.get(connectionId)?.size;
+                if (hasInFlightHandlers) {
+                    runInAction(() => {
+                        this.waitingForInFlightHandlers = true;
+                    });
+                }
+                let hadInFlight = false;
+                try {
+                    hadInFlight = await this.awaitInFlightHandlers(
+                        connectionId
+                    );
+                } finally {
+                    if (hasInFlightHandlers) {
+                        runInAction(() => {
+                            this.waitingForInFlightHandlers = false;
+                        });
+                    }
+                }
                 await this.deleteClientKeys(connection.pubkey);
                 const { index: liveIndex } = this.getConnection({
                     connectionId
--- a/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
+++ b/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
@@ -714,7 +714,7 @@ export default class NWCConnectionDetails extends React.Component<
                                         ]}
                                     >
                                         {localeString(
-                                            'views.Settings.NostrWalletConnect.waitingForInFlightRequest'
+                                            'views.Settings.NostrWalletConnect.waitingToFinishPendingRequest'
                                         )}
                                     </Text>
                                 </View>
--- a/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
+++ b/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
@@ -322,7 +322,10 @@ export default class NWCConnectionDetails extends React.Component<
         const { navigation, NostrWalletConnectStore } = this.props;
         const { loading, regenerating, deleting, error, connection } =
             this.state;
-        const { loading: storeLoading } = NostrWalletConnectStore;
+        const { loading: storeLoading, waitingForInFlightHandlers } =
+            NostrWalletConnectStore;
+        const showInFlightWait =
+            waitingForInFlightHandlers && (deleting || regenerating);
 
         return (
             <Screen>
@@ -691,12 +694,31 @@ export default class NWCConnectionDetails extends React.Component<
                         </ScrollView>
 
                         <View style={styles.bottomContainer}>
+                            {showInFlightWait && (
+                                <View style={styles.inFlightWaitContainer}>
+                                    <LoadingIndicator size={24} />
+                                    <Text
+                                        style={[
+                                            styles.inFlightWaitText,
+                                            {
+                                                color: themeColor(
+                                                    'secondaryText'
+                                                )
+                                            }
+                                        ]}
+                                    >
+                                        {localeString(
+                                            'views.Settings.NostrWalletConnect.waitingForInFlightRequest'
+                                        )}
+                                    </Text>
+                                </View>
+                            )}
                             <Button
                                 title={localeString(
                                     'views.Settings.NostrWalletConnect.regenerateConnection'
                                 )}
                                 onPress={this.confirmRegenerateConnection}
-                                disabled={regenerating}
+                                disabled={regenerating || deleting}
                                 secondary={regenerating}
                                 noUppercase
                             />
@@ -708,7 +730,7 @@ export default class NWCConnectionDetails extends React.Component<
                                     this.deleteConnection(connection)
                                 }
                                 warning
-                                disabled={loading || deleting}
+                                disabled={loading || deleting || regenerating}
                                 secondary={deleting}
                                 noUppercase
                             />
@@ -766,5 +788,18 @@ const styles = StyleSheet.create({
         paddingHorizontal: 20,
         paddingVertical: 15,
         gap: 10
+    },
+    inFlightWaitContainer: {
+        flexDirection: 'row',
+        alignItems: 'center',
+        justifyContent: 'center',
+        gap: 10,
+        paddingVertical: 4
+    },
+    inFlightWaitText: {
+        flexShrink: 1,
+        fontSize: 14,
+        fontFamily: 'PPNeueMontreal-Book',
+        textAlign: 'center'
     }
 });
--- a/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
+++ b/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
@@ -56,6 +56,7 @@ export default class NWCConnectionDetails extends React.Component<
     NWCConnectionDetailsState
 > {
     private unsubscribeFocus?: () => void;
+    private isUnmounted = false;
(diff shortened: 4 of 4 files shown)


kaloudis:
utACK f4334a56f

Nice follow-up to #4443, and thanks for picking up #4487. Verified the flag is set iff the await will actually block (the sync pre-check matches the internal `!!set?.size` check in `awaitInFlightHandlers`, with no gap before the await), the `try/finally` guarantees cleanup, the view is an `@observer` so the destructured observable is tracked, and regenerate is covered since it calls `deleteConnection` before `createConnection`. The mutual button disabling also closes a real hole: previously Regenerate was tappable while a delete was awaiting handlers. CI is green at head.

A few non-blocking suggestions:

1. **Assert the new observable in the existing test.** `stores/NostrWalletConnectStore.test.ts` already has "delete waits for in-flight handler" with a deferred handler promise; adding `expect(store.waitingForInFlightHandlers).toBe(true)` after the delete starts and `.toBe(false)` after `deleteDone` would pin the new behavior for two lines.

2. **Clear the flag in `reset()`.** `reset()` clears `waitingForConnection` but not `waitingForInFlightHandlers`. If a wallet switch happens mid-await, the flag only clears when the orphaned handler promise eventually settles. Worth adding `this.waitingForInFlightHandlers = false;` alongside the existing line for hygiene.

3. **Shape nit.** Conditionally setting and clearing the flag around an unconditional await duplicates the size check. Equivalent and a bit tighter:

   ```ts
   let hadInFlight = false;
   if (this.inFlightHandlersByConnection.get(connectionId)?.size) {
       runInAction(() => (this.waitingForInFlightHandlers = true));
       try {
           hadInFlight = await this.awaitInFlightHandlers(connectionId);
       } finally {
           runInAction(() => (this.waitingForInFlightHandlers = false));
       }
   }
   ```

   `awaitInFlightHandlers` returns `false` exactly when the set is empty, so skipping the call in the empty case is behavior-preserving.

4. **Scope note.** The `updateConnection` relay-change path has the same silent wait (tracked in #4486). The global boolean is fine while `NWCConnectionDetails` is the only caller of `deleteConnection`, but if #4486's UI lands later it may want a per-connection shape.

None of these need to hold up merge; 1 and 2 would be nice to fold in while the PR is open.

ajaysehwal:
> utACK [f4334a5](https://github.com/ZeusLN/zeus/commit/f4334a56f8d0c2d0ea3a49e6f4a472506e6beb65)
> 
> Nice follow-up to #4443, and thanks for picking up #4487. Verified the flag is set iff the await will actually block (the sync pre-check matches the internal `!!set?.size` check in `awaitInFlightHandlers`, with no gap before the await), the `try/finally` guarantees cleanup, the view is an `@observer` so the destructured observable is tracked, and regenerate is covered since it calls `deleteConnection` before `createConnection`. The mutual button disabling also closes a real hole: previously Regenerate was tappable while a delete was awaiting handlers. CI is green at head.
> 
> A few non-blocking suggestions:
> 
> 1. **Assert the new observable in the existing test.** `stores/NostrWalletConnectStore.test.ts` already has "delete waits for in-flight handler" with a deferred handler promise; adding `expect(store.waitingForInFlightHandlers).toBe(true)` after the delete starts and `.toBe(false)` after `deleteDone` would pin the new behavior for two lines.
> 2. **Clear the flag in `reset()`.** `reset()` clears `waitingForConnection` but not `waitingForInFlightHandlers`. If a wallet switch happens mid-await, the flag only clears when the orphaned handler promise eventually settles. Worth adding `this.waitingForInFlightHandlers = false;` alongside the existing line for hygiene.
> 3. **Shape nit.** Conditionally setting and clearing the flag around an unconditional await duplicates the size check. Equivalent and a bit tighter:
>    ```ts
>    let hadInFlight = false;
>    if (this.inFlightHandlersByConnection.get(connectionId)?.size) {
>        runInAction(() => (this.waitingForInFlightHandlers = true));
>        try {
>            hadInFlight = await this.awaitInFlightHandlers(connectionId);
>        } finally {
>            runInAction(() => (this.waitingForInFlightHandlers = false));
>        }
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    `awaitInFlightHandlers` returns `false` exactly when the set is empty, so skipping the call in the empty case is behavior-preserving.
> 4. **Scope note.** The `updateConnection` relay-change path has the same silent wait (tracked in [NWC: updateConnection races with in-flight handlers (relay change can drop responses, budget edits race settlement) #4486](https://github.com/ZeusLN/zeus/issues/4486)). The global boolean is fine while `NWCConnectionDetails` is the only caller of `deleteConnection`, but if [NWC: updateConnection races with in-flight handlers (relay change can drop responses, budget edits race settlement) #4486](https://github.com/ZeusLN/zeus/issues/4486)'s UI lands later it may want a per-connection shape.
> 
> None of these need to hold up merge; 1 and 2 would be nice to fold in while the PR is open.

fixed in 4702fa9096f9c54245d1caadf1dad3f7d220e2cc

kaloudis:
Re-reviewed 4702fa909. All three suggestions landed, and I checked them rather than eyeballing:

- **Test assertions.** Non-vacuous: I mutated the store to set `waitingForInFlightHandlers = false` instead of `true` and the new `expect(...).toBe(true)` fails, so it genuinely pins the behavior. 60/60 in the suite locally.
- **`reset()`.** The clear sits with the existing `waitingForConnection` / `errorMessage` writes in the same teardown block, so it inherits their semantics and introduces nothing new. It is also correctly inside the `!skipTeardown` branch, so the iOS-active-connection / Android-foreground-service path that deliberately keeps NWC attached will not clobber a live wait.
- **Shape nit.** Adopted, and exactly behavior-preserving: `awaitInFlightHandlers` returns `false` iff `!set?.size` and has no side effects beyond the `Promise.allSettled`, so skipping the call in the empty case cannot change `hadInFlight` or which relay-release branch runs.

Also re-confirmed two things the feature depends on. The flag is set synchronously (no `await` between `deleteConnection` entry and the `runInAction`, since `unsubscribeFromConnection` is not awaited), so the row appears in the same tick as the tap. And the row lives in `bottomContainer`, which is outside the `loading || storeLoading || regenerating || deleting` ternary (that one only swaps the header `rightComponent`), so it is genuinely visible during the wait rather than being replaced by a full-screen spinner.

CI green. utACK 4702fa909.

One thing I turned up while re-reviewing that I think is worth folding in here, since it is in a handler this PR already touches.

### Leaving the screen mid-wait fires a stale `goBack()`

The delete confirm callback does:

```ts
NostrWalletConnectStore.deleteConnection(connection.id)
    .then(() => { navigation.goBack(); })
```

Nothing stops the user leaving during the wait: the header Back button is unconditional, `NWCConnectionDetails` is registered without `gestureEnabled: false` so iOS swipe-back works, and Android hardware back works. The store call keeps running, which is #4443 behaving as designed, but when it resolves (up to ~150s later) `goBack()` fires from a screen that is no longer in the stack.

I expected React Navigation to no-op that, since the action carries `source`. It does not. `useNavigationCache` sets `source: route.key` and never sets `target`, and `StackRouter`'s `POP` only honors `source` when `action.target === state.key`. With `target` undefined it falls back to `state.index`, so it pops whatever is on top. Against the shipped router (`@react-navigation/routers` 7.21.10):

```
stack [Settings, List], stale GO_BACK with source 'Details-1'
  -> [Settings]     // popped List; not a no-op
```

So: open a connection, tap Delete, see the new waiting row, back out to the connections list, and some time later the list pops out from under you into Settings. If the user navigated elsewhere in the meantime, an arbitrary screen gets popped instead.

This predates the PR (the `.then(goBack)` is unchanged, and the long wait came from #4443), so I am not calling it a merge blocker. But this PR is the one that tells the user the wait is real and expected, which makes "back out and let it finish" the natural response instead of the force-kill you are trying to prevent. `views/LSPS/LSPPaymentAwait.tsx` already has the idiom:

```ts
private isUnmounted = false;
componentWillUnmount() { this.isUnmounted = true; }
...
.then(() => { if (!this.isUnmounted) navigation.goBack(); })
```

### Scope note on regenerate

`regenerateConnection` has the same staleness on `navigation.navigate('NWCConnectionQR', ...)`, but the same guard is not obviously correct there: `NWCConnectionQR` takes `nostrUrl` as a route param and is only reachable immediately after create or regenerate, so silently skipping the navigate would leave a regenerated connection whose URI the user can never see again. Yanking them to the QR screen late is bad; stranding the URI is worse. That probably wants its own issue (make a connection's URI re-viewable from the details screen, then guard the navigate) rather than a fix in this PR. Happy to file it.

### Optional copy nit

"in-flight request" is the only use of that phrasing in `en.json`. Something like "Finishing a pending request..." may read better for non-technical users, but no strong feeling, and it is a Transifex-visible string either way.

kaloudis:
Re-reviewed 8cc32037c. Both round-2 items landed, and the guard is complete rather than partial, which I checked rather than assumed.

### The guard covers every exit from this screen

`isUnmounted` only helps if unmount is the only way the screen can lose its place in the stack. While a delete is running, it is. The header `rightComponent` swaps to a `LoadingIndicator` when `loading || storeLoading || regenerating || deleting`, which removes both of the screen's forward-navigation entry points (the activity `ClockIcon` and the edit pencil) for the entire wait. The rest of the screen is `KeyValue` rows and `Amount`, none of which navigate, and `showWarning` goes through `ModalStore` rather than the navigator. So the only ways out are header Back, the iOS swipe gesture, and Android hardware back, and all three unmount. There is no residual "user pushed a screen, then the delete resolved and popped it" case hiding behind the fix.

Good call giving the `.catch` the same guard and not just the `.then`. A failure after the user leaves now logs and stops instead of calling `setState` on a dead component.

### Copy nit

Key renamed cleanly. `git grep waitingForInFlightRequest` on the branch is empty and the only reference to the new key is the single call site, so nothing dangles. Locale keys are not type-checked, so that is worth confirming rather than trusting the tsc job for.

CI green on 8cc32037c (lint, prettier, test, tsc).

utACK 8cc32037c.

### Follow-up filed

I opened #4525 for the regenerate half I flagged last round, since that one needs a connection's URI to be re-viewable before the same guard becomes safe. Confirmed while writing it that the URI is fully reconstructible from state we already persist: `NWC_CLIENT_KEYS[connection.pubkey]` holds the client private key, and `NostrConnectUtils.buildWalletConnectConnectionUrl` is a pure function of that plus the relay and service pubkey. Only the route param is ephemeral, not the URI. Out of scope here.

### Two cosmetic things

Neither is worth a push on its own:

- the screenshot in the PR description still shows the old string
- the locale rename rode along in the "guard delete goBack after screen unmount" commit, whose message does not mention it. ZEUS keeps individual commits through merge, so if you end up pushing again for another reason, splitting or rewording it would be nice.

ajaysehwal:
> Re-reviewed [8cc3203](https://github.com/ZeusLN/zeus/commit/8cc32037cd3363e08faaeaf4ddedaf7e240d643c). Both round-2 items landed, and the guard is complete rather than partial, which I checked rather than assumed.
> 
> ### The guard covers every exit from this screen
> `isUnmounted` only helps if unmount is the only way the screen can lose its place in the stack. While a delete is running, it is. The header `rightComponent` swaps to a `LoadingIndicator` when `loading || storeLoading || regenerating || deleting`, which removes both of the screen's forward-navigation entry points (the activity `ClockIcon` and the edit pencil) for the entire wait. The rest of the screen is `KeyValue` rows and `Amount`, none of which navigate, and `showWarning` goes through `ModalStore` rather than the navigator. So the only ways out are header Back, the iOS swipe gesture, and Android hardware back, and all three unmount. There is no residual "user pushed a screen, then the delete resolved and popped it" case hiding behind the fix.
> 
> Good call giving the `.catch` the same guard and not just the `.then`. A failure after the user leaves now logs and stops instead of calling `setState` on a dead component.
> 
> ### Copy nit
> Key renamed cleanly. `git grep waitingForInFlightRequest` on the branch is empty and the only reference to the new key is the single call site, so nothing dangles. Locale keys are not type-checked, so that is worth confirming rather than trusting the tsc job for.
> 
> CI green on [8cc3203](https://github.com/ZeusLN/zeus/commit/8cc32037cd3363e08faaeaf4ddedaf7e240d643c) (lint, prettier, test, tsc).
> 
> utACK [8cc3203](https://github.com/ZeusLN/zeus/commit/8cc32037cd3363e08faaeaf4ddedaf7e240d643c).
> 
> ### Follow-up filed
> I opened #4525 for the regenerate half I flagged last round, since that one needs a connection's URI to be re-viewable before the same guard becomes safe. Confirmed while writing it that the URI is fully reconstructible from state we already persist: `NWC_CLIENT_KEYS[connection.pubkey]` holds the client private key, and `NostrConnectUtils.buildWalletConnectConnectionUrl` is a pure function of that plus the relay and service pubkey. Only the route param is ephemeral, not the URI. Out of scope here.
> 
> ### Two cosmetic things
> Neither is worth a push on its own:
> 
> * the screenshot in the PR description still shows the old string
> * the locale rename rode along in the "guard delete goBack after screen unmount" commit, whose message does not mention it. ZEUS keeps individual commits through merge, so if you end up pushing again for another reason, splitting or rewording it would be nice.

fixed

kaloudis:
Re-reviewed 7b1ef016e. The branch was force-pushed since my last ACK, so this is mostly a re-confirmation, plus one thing I turned up.

### The push is a pure commit split

`git diff 8cc32037c 7b1ef016e` is empty. The tree I utACK'd last round is byte-identical to the tree here, so nothing about the behavior needed re-reviewing, and the two rounds of verification above still stand as-is.

The split itself is clean and bisectable, which is worth stating since it was the point of the exercise. `0b88dc8bb` is only the `isUnmounted` guard on the delete `.then`/`.catch`, and `7b1ef016e` moves the locale key **and** its single call site together, so there is no intermediate commit that references a key that does not exist in `en.json`. Old key is gone; the new one has exactly one reference.

Screenshot in the description is updated too (the alt text is stamped 17:20 IST, three minutes after the copy commit, and the image shows the new string).

CI green at `7b1ef016e` specifically, not just "on the PR": I checked the run's `head_sha` rather than trusting the checks tab. Lint, prettier, test, tsc all pass. Master has picked up 8 commits since your base and none of them touch these four files (`en.json` is untouched upstream; only `zh_CN.json` moved), so the clean merge is real and not just a stale mergeability flag.

utACK 7b1ef016e. This is ready to merge from my side.

### One note, non-blocking, better suited to #4486

Now that the round-3 guard makes backing out mid-wait a safe and reasonable thing to do, "back out and come back to check on it" becomes a natural user response. That path currently hides the very state this PR exists to show.

During the wait the connection is still in `connections`, since the splice happens after the await. Your own test pins this at `expect(store.connections).toHaveLength(1)` while the flag is true. So the list still lists it, and tapping it re-opens the details screen.

The row is gated on `waitingForInFlightHandlers && (deleting || regenerating)`. The store flag survives the remount, but `deleting` does not: it is constructor state on a fresh instance. The re-entered screen therefore shows Status "Active", no wait row, and both buttons enabled, which reads as "nothing is happening" while the delete is still running.

I checked what a second Delete tap from that screen actually does, with a bystander connection present:

```
mid-wait   storeFlag=true   listed=[conn-expiry, conn-bystander]
settled    [fulfilled, fulfilled]
after      remaining=[conn-bystander]   flag=false
```

Benign, not corrupting. The `liveIndex !== -1` re-lookup stops the second delete from splicing the bystander, `releaseUnusedRelayService` early-returns while any connection still references the relay so the double release is idempotent, and the flag still clears correctly. It is a display inconsistency, not a data bug, which is why I am not asking for it here.

The tidy fix is the per-connection shape I raised in round 1 note 4: key the wait state by `connectionId` instead of a global boolean, so both the row and the disabled buttons survive a remount. That is the shape #4486's UI wants anyway. I would rather it landed there than as a fifth push on this PR. I can fold it into #4486, or file it separately if you would prefer to pick it up on its own.

kaloudis:
Correction to my last comment: I pointed the follow-up at #4486, but that issue was closed as completed on 2026-08-28 by #4500, which is already in this branch's base. Its scope was store-level only, so the UI half was never covered there and my suggestion had nowhere to land.

Filed as #4527 instead. While writing it up I found the same gap has a second half worth fixing in one go: `updateConnection` awaits in-flight handlers at `stores/NostrWalletConnectStore.ts:775` without ever setting the flag, so a relay change or budget edit saved on top of an in-flight payment blocks `AddOrEditNWCConnection` behind its generic spinner with no explanation. That is the problem this PR solved for delete, on the path #4500 made blocking. Keying the wait state by `connectionId` instead of using a global boolean fixes both that and the remount case in one change.

One correction to the numbers while I am here: the "~150s" bound in the description (and in #4487's title) is an overestimate. `PAYMENT_TIMEOUT_SECONDS` is 120 and the handler waits `PAYMENT_TIMEOUT_SECONDS + 5`, so the real worst case is ~125s. Does not change anything about this PR.

Nothing here affects the review. utACK 7b1ef016e stands and this is still ready to merge.

ajaysehwal:
> Correction to my last comment: I pointed the follow-up at #4486, but that issue was closed as completed on 2026-08-28 by #4500, which is already in this branch's base. Its scope was store-level only, so the UI half was never covered there and my suggestion had nowhere to land.
> 
> Filed as #4527 instead. While writing it up I found the same gap has a second half worth fixing in one go: `updateConnection` awaits in-flight handlers at `stores/NostrWalletConnectStore.ts:775` without ever setting the flag, so a relay change or budget edit saved on top of an in-flight payment blocks `AddOrEditNWCConnection` behind its generic spinner with no explanation. That is the problem this PR solved for delete, on the path #4500 made blocking. Keying the wait state by `connectionId` instead of using a global boolean fixes both that and the remount case in one change.
> 
> One correction to the numbers while I am here: the "~150s" bound in the description (and in #4487's title) is an overestimate. `PAYMENT_TIMEOUT_SECONDS` is 120 and the handler waits `PAYMENT_TIMEOUT_SECONDS + 5`, so the real worst case is ~125s. Does not change anything about this PR.
> 
> Nothing here affects the review. utACK [7b1ef01](https://github.com/ZeusLN/zeus/commit/7b1ef016e4dca3fbf2ab7cd30dbbf902f55f6d77) stands and this is still ready to merge.

Updated the maximum payment timeout in the PR description from **150 seconds to 120 seconds**.

PR #4521, opened 2026-08-30
Offer channel migration export when the embedded node won't start

kaloudis:
# Description

A user on embedded LND reported that the "Export Channels to other device" option was missing from Tools. Their node fails to start, and they wanted to migrate to another device before resorting to disaster recovery.

The option was gated on `hasChannels`, which is computed from `ChannelsStore` and only populates from a running node. A dead node means channels never load, so the migration feature was unreachable in exactly the scenario it was built for. The local zip export doesn't need a running node at all: it stops LND (tolerating already-stopped) and zips the on-disk graph directory.

Changes:

- **Tools**: when the node is connected (`nodeInfo.identity_pubkey` present), keep trusting the node's channel list, preserving the intent of the `hasChannels` gate from v13.0.0. When it isn't, fall back to checking for channel data on disk (`channelDataExists` helper in `ChannelMigrationUtils`), so export stays reachable with a dead node.
- **Network source**: `isTestnet` now comes from the persisted `embeddedLndNetwork` wallet config instead of `nodeInfo.isTestNet`, which is empty when the node is down and would silently fall back to the mainnet directory for testnet wallets.
- **Olympus upload**: only offered while the node is running, since it signs auth challenges with the node key via `BackendUtils.signMessage`. With a dead node the local zip export is offered alone.
- Applied the same `isTestnet`/Olympus changes to the identical caller in `views/Settings/Seed.tsx` for consistency, plus a defensive default on `seedPhrase`.

Draft until device tests: (1) embedded wallet with channels, node prevented from starting, verify export appears and local zip completes; (2) node running, verify Olympus + local options unchanged; (3) fresh embedded wallet with no channel history, verify option stays hidden while the node is up; (4) testnet wallet with node down, verify the testnet graph dir is exported.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/utils/ChannelMigrationUtils.ts
+++ b/utils/ChannelMigrationUtils.ts
@@ -29,6 +29,25 @@ const getGraphDir = (lndDir: string, isTestnet: boolean): string => {
     return `${rootPath}/${lndDir}/data/graph/${network}`;
 };
 
+/**
+ * Checks whether channel data exists on disk, so export can be offered
+ * even when the node isn't running.
+ */
+export const channelDataExists = async (
+    lndDir: string,
+    isTestnet: boolean
+): Promise<boolean> => {
+    try {
+        const graphDir = getGraphDir(lndDir, isTestnet);
+        if (!(await RNFS.exists(graphDir))) return false;
+        const entries = await RNFS.readDir(graphDir);
+        return entries.length > 0;
+    } catch (e) {
+        console.error('Failed to check for channel data on disk:', e);
+        return false;
+    }
+};
+
 const stopLndSafely = async (): Promise<void> => {
     try {
         await stopLnd();
@@ -770,8 +789,8 @@ export const exportChannelDb = async (
 };
 
 /**
- * Prompts the user to export channel backup — either to Olympus (SQLite only)
- * or as a local zip file.
+ * Prompts the user to export channel backup — either to Olympus (SQLite
+ * only, requires a running node) or as a local zip file.
  */
 export const handleExportChannels = ({
     isSqlite,
@@ -779,6 +798,7 @@ export const handleExportChannels = ({
     isTestnet,
     pubkey,
     seedPhrase,
+    canUploadToOlympus,
     setStatus
 }: {
     isSqlite: boolean;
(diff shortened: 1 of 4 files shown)


utils/ChannelMigrationUtils.ts
```diff
@@ -29,6 +29,25 @@ const getGraphDir = (lndDir: string, isTestnet: boolean): string => {
     return `${rootPath}/${lndDir}/data/graph/${network}`;
 };
 
+/**
+ * Checks whether channel data exists on disk, so export can be offered
+ * even when the node isn't running.
+ */
+export const channelDataExists = async (
```
shubhamkmr04: should we rename it to `graphDataExists` instead? as it can't distinguish whether the node had channels or not

shubhamkmr04 (changes requested):
@kaloudis Seed.tsx is not wired up to give us option to export graph data for when lnd is dead, like we did in Tools.tsx, is it expected?

utils/ChannelMigrationUtils.ts
shubhamkmr04: also lets add tests for the new util

kaloudis: Good call on both. Renamed to `graphDataExists` (plus the mirrored Tools state, now `graphDataOnDisk`) and noted in the JSDoc that lnd creates the graph DB files on first start, so the check can't distinguish whether channels ever existed. Unit tests added in 144155e11 covering the missing-dir, empty-dir, populated-dir, testnet path, custom lndDir, and throw cases.

kaloudis:
Not intended, good catch. The first commit fixed the params Seed.tsx passes to `handleExportChannels` but left both gates (the backup-complete reminder alert and the export button) on `hasChannels` from ChannelsStore, so they were unreachable with a dead node. 76c35142e wires Seed.tsx up the same way as Tools: trust the node's channel list when it's up, otherwise fall back to graph data on disk. The inactive-wallet and rescue-key paths keep their existing guards, since the disk check uses the active wallet's lndDir.

One behavior note that applies to both screens: since lnd creates the graph DB files on first start, a dead node that never had channels will now also see the export option. That fail-open seems right for a disaster-recovery path (exporting an empty db is harmless), and it's why the util is now named `graphDataExists`.

shubhamkmr04 (approved):
tACK

Issue #2794, opened 2025-02-04
Clean up negative positioning/margin


myxmaster:
### Describe your enhancement idea

A lot of those issues were already cleaned up, but in the following files we can still improve layouting by removing negative values for positioning/margin:
- ~~PaymentPath.tsx~~ (fixed with https://github.com/ZeusLN/zeus/pull/3771)
- ~~NostrContacts.tsx~~ (fixed with https://github.com/ZeusLN/zeus/pull/3774)
- ~~Payment.tsx~~ (fixed with https://github.com/ZeusLN/zeus/pull/3348)
- ~~AddNotes.tsx~~ (fixed with https://github.com/ZeusLN/zeus/pull/3775)
- ~~ContactDetails.tsx~~ (fixed with PR above)
- ~~ImportAccount.tsx~~ (fixed with PR above)
- ~~Channel.tsx~~ (fixed with https://github.com/ZeusLN/zeus/pull/3783)
- ~~ProductDetails.tsx~~ (fixed with PR above)
- ~~PointOfSale.tsx~~ (fixed with PR above)
- views\Settings\LSPS1\index.tsx

techaakritisha:
Hello @myxmaster I would love to work on this issue

techaakritisha:
ok thank you and where i can discuss my doubts

techaakritisha:
ok

avinxshKD:
I would like to work upon the issue, could you please assign

kaloudis:
Status update:

The last unchecked item here is done. `views/Settings/LSPS1/index.tsx` was moved to `views/LSPS1/index.tsx`, and its remaining negative margins were removed in 8066f4837 (ui: Hide Settings and History icons on LSPS1/LSPS7 order info view).

However, a fresh sweep of `views/` and `components/` (as of ebbb89ead) shows roughly 30 new instances have crept in since this issue was filed. Updated list:

- [ ] views/Swaps/index.tsx (786, 805, 816, 1151, 1517) - includes the same `marginTop: -10` header-icon pattern that 8066f4837 just fixed in LSPS1/LSPS7
- [ ] views/Tools/CurrencyConverter.tsx (99, 113, 122)
- [ ] components/CurrencyConverterContent.tsx (432, 446, 455) - same hack duplicated when the converter content was extracted into a shared component
- [ ] views/Send.tsx (955, 1583)
- [ ] views/LightningAddress/index.tsx (1151-1153)
[...]

PR #4526, opened 2026-08-31
views/Send: tie send max toggle to the UTXO selection

kaloudis:
# Description

Follow-up to a user report: when sending a single UTXO in full, the send fails because the fee has nowhere to come from, and there is no obvious way to have it subtracted.

The capability already exists. The **Use all possible funds** toggle on the on-chain Send screen is selection-aware: with UTXOs picked in coin control, it maxes out the *selection* rather than the whole balance, producing a single-output sweep with the fee subtracted and no change. On LND-family backends that goes out as `SendCoins` with `outpoints` + `send_all` (needs lnd v0.18.3+, gated by `supportsOnchainSendMax`); CLN uses `withdraw satoshi=all utxos=[...]`, LDK Node uses `sendAllToOnchainAddressWithUtxos`.

Nothing on screen communicated that. The toggle sat directly under the amount field while the UTXO picker sat well below it, past the multi-output section, and the label never changed when a selection was active. So the feature reads as absent.

This PR:

1. **Fixes a stale amount.** `selectUTXOs` updated `utxos`/`utxoBalance` but not `satAmount`. Enabling the toggle *before* picking UTXOs therefore left `satAmount` at the full confirmed balance. The amount field itself looked right because it derives from `utxoBalance`, but the stale value was forwarded to `VerifyOnChain`, whose total comes from `satAmount`, so the confirmation screen showed the whole wallet balance instead of the selection. The broadcast was never wrong, since `send_all` drops the amount, but the number shown before signing was. Selecting a non-default account (which force-disables send max) now also restores the previously entered amount instead of leaving `amount` empty with a non-zero `satAmount`.

2. **Improves toggle placement.** The UTXO picker and the toggle now sit together immediately under the amount field, so the switch is adjacent both to the amount it locks and the selection it applies to. `Add output` moved below the toggle, so enabling send max no longer makes a control disappear from above the switch and shift it under the user's finger.

3. **Changes the label when UTXOs are selected.** With a selection active the toggle reads **Use all selected UTXOs**; with no selection it keeps the existing string. `views.OpenChannel.fundMax` is shared with the channel-open screen, so it is left untouched and one new key is added rather than changing an existing translated string.

Also replaces the `marginTop: -20 / top: 20` hack that faked the label and switch onto one row with a flex row, per #2794. That hack also collided with the switch on longer translations; the label now wraps cleanly.

Not addressed here: nothing in the flow yet states that the fee is deducted, and #4508's fee estimate deliberately skips coin control and send max. Worth a follow-up.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

One new key: `views.Send.fundMaxUtxos` = "Use all selected UTXOs". No existing string was modified.

### Other:

Device tests pending. Suggested passes:

* Select one UTXO, enable the toggle, confirm the label reads "Use all selected UTXOs" and that `VerifyOnChain` shows the selection total rather than the wallet balance.
* Enable the toggle *first*, then select a UTXO. This is the ordering that was broken.
* Change the selection while the toggle is on, and clear the selection while it is on. The amount should track each time.
* Switch to a non-default account with the toggle on. It should turn off and restore the amount previously entered.
* Broadcast a single-UTXO send to confirm one output, no change, fee deducted.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -926,6 +926,7 @@
     "views.Send.zaplockerWarning": "This is a Zaplocker invoices that will hold payment up to 24 hours. Open ZEUS regularly after payment to help mitigate the risk of a force closed channel. Proceed at your own risk.",
     "views.Send.addOutput": "Add output",
     "views.Send.removeOutput": "Remove output",
+    "views.Send.fundMaxUtxos": "Use all selected UTXOs",
     "views.Send.payBolt12.offerFetchFailure": "Couldn't fetch offer",
     "views.Send.payBolt12.invoiceFetchFailure": "Couldn't fetch invoice",
     "views.Send.payBolt12.specifyAmount": "Amount must be specified",
--- a/views/Send.tsx
+++ b/views/Send.tsx
@@ -298,12 +298,29 @@ export default class Send extends React.Component<SendProps, SendState> {
         utxoBalance: number,
         account: string
     ) => {
-        this.setState((prevState) => ({
-            utxos,
-            utxoBalance,
-            account,
-            fundMax: account === 'default' ? prevState.fundMax : false
-        }));
+        const { confirmedBlockchainBalance } = this.props.BalanceStore;
+
+        this.setState((prevState) => {
+            // send max is only offered on the default account
+            const fundMax = account === 'default' ? prevState.fundMax : false;
+
+            let amount = prevState.amount;
+            let satAmount = prevState.satAmount;
+
+            if (fundMax) {
+                // send max is bound to the selection, so the amount it will
+                // send has to follow the selection as it changes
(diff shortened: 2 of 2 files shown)


ajaysehwal (approved):
tACK

Issue #4524, opened 2026-08-31
Embedded LND: fatal SQLite error "no such table: graph_source_nodes"


ThomasFarstrike:
### Describe the bug

A week or two ago I started the Embedded LND after having it off for a few weeks, and now it keeps telling me "Restart required" after a few minutes of starting up. Restarting it doesn't fix the issue.

The logs show:

`[ERR] LTND: Shutting down due to error in main method rev="" err="unable to fetch source node: unable to fetch source node: unknown sqlite error: SQL logic error: no such table: graph_source_nodes (1)"
` 

So now my EmbLND doesn't work anymore.

Question: what's the best way to recover from this? Just delete the wallet and restore from backup?

I don't expect a fix or technical support from the Zeus team, of course. I just thought I'd create this issue to log it (in case anyone else stumbles upon it) and to give a heads up, in case anyone wants any debug info before I delete the wallet.

### Reproduce

To reproduce the issue now, I just start Zeus with EmbLND and get the issue.

I'm not sure what triggered it. The most plausible to me seems that the Embedded LND instance got killed right in the middle of doing a migration or some other critical SQLite operations, which broke it.

I don't remember exactly but it's possible that I switched to a different Zeus wallet while the EmbLND was starting up, or I closed Zeus while it was starting up, or I move to a different app while it was starting up...

### ZEUS version

v13.2.0

### Node interface

Embedded LND

### Network

Clearnet

### Device

Motorola

### Device operating system

Android

### Log output

```shell
10:19:42.701 [INF] LTND: REST API is disabled!
10:19:42.702 [INF] LTND: Listening on the p2p interface is disabled!
10:19:42.704 [INF] LTND: Version Info rev="" version=0.21.2-beta commit=v0.21.2-beta-zeus debuglevel=production logging=info
10:19:42.705 [INF] LTND: Network Info rev="" active_chain=Bitcoin network=mainnet
10:19:42.706 [INF] RPCS: RPC server listening on bufconn
10:19:42.720 [INF] LTND: Opening the main database, this might take a few minutes...
10:19:42.756 [INF] LTND: Creating local graph and channel state DB instances
10:19:42.758 [INF] CHDB: Checking for schema update: latest_version=35, db_version=35
10:19:42.758 [INF] CHDB: Applying 2 optional migrations
10:19:42.758 [INF] CHDB: Checking for optional update: name=prune_revocation_log
[...]

PR #4528, opened 2026-08-31
lndmobile: make EGS graph reset non-destructive, repair bricked graph DBs

kaloudis:
# Description

Relates to issue: **#4524**

Implements the ZEUS-side fixes from @ziggie1984's root-cause analysis in #4524. The Express Graph Sync reset for native-SQL wallets dropped LND's graph tables, deleted `migration_tracker` rows 9/10, and rewound `schema_migrations` to 7. LND derives its database version from the single highest `migration_tracker` row, so with rows 11-18 still present the graph migrations were skipped forever and every subsequent start failed with `no such table: graph_source_nodes`. Restarting could never recover.

## What changed

**Reset is now non-destructive and atomic.** `DEBUG_resetGraphDb` is replaced by `resetGraphDb`, which empties the 13 graph tables with `DELETE`s (children first) inside a single transaction. `migration_tracker` and `schema_migrations` are treated as LND-owned internals and are never touched. Since a fully-migrated empty graph schema is exactly the state of a fresh wallet, LND re-creates its source node on start and EGS repopulates the graph.

**Bricked wallets are repaired automatically.** A new `repairGraphDb` runs on every embedded-LND sqlite startup, before EGS or `startLnd` open `lnd.sqlite`. If `migration_tracker` says the graph schema should exist (version >= 11) but the tables are missing, it transactionally rebuilds the schema exactly as the fork's `000008_graph` + `000009_graph_v2` migrations define it, restores tracker rows 9/10, and sets `schema_migrations` to the true schema version (15). Affected wallets, including the reporter's, recover on first launch of the patched build without a seed restore.

**Fail closed on unknown schemas.** Both operations refuse to act if `migration_tracker` reports a version above 18 or if unrecognized `graph_*` tables exist, so a future fork bump can never be silently corrupted; the embedded DDL must be re-reviewed against the fork's migrations when the pin moves (noted in code comments).

**Reset flag semantics.** The reset flag is now only cleared after a successful reset, so failures are retried on the next startup instead of being silently dropped. The pre-existing interrupted-reset hazard (index drops without table drops) is also healed: the reset path re-creates any missing indexes via `CREATE INDEX IF NOT EXISTS`.

## Verification

Simulation against the real fork migrations (sqlite fixture built by executing `000001`-`000015` from `v0.21.2-beta-zeus`, tracker rows 1-18, then applying the exact v13.2.0 reset statements to reproduce the bricked state):

- repair rebuilds a schema **identical** to the real-migration schema: columns, foreign keys, index definitions, normalized SQL text, tracker rows, and `schema_migrations` all match
- reset empties all graph tables and leaves schema + trackers byte-identical
- reset heals an interrupted legacy reset (indexes dropped, tables intact) back to the exact index set
- the Java and Swift SQL statement lists are verified equal

Tested on iOS and Android devices.

Not covered here (follow-ups from ziggie's writeup): moving reset/repair into lndmobile behind an LND-fork API, the speedloader's `ApplyAllMigrations`/`MigrateGraphToSQL` split, and upstream migration-tracker continuity validation.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A (native SQL paths; schema equivalence verified by simulation against the fork's real migration files, script in PR discussion)

- [x] Android
- [x] iOS

On-device
- [ ] LDK Node
- [x] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/android/app/src/main/java/com/zeus/LndMobileTools.java
+++ b/android/app/src/main/java/com/zeus/LndMobileTools.java
@@ -441,8 +441,163 @@ class LndMobileTools extends ReactContextBaseJavaModule {
     return fileOrDirectory.delete() && success;
   }
 
+  // Graph SQL database maintenance.
+  //
+  // The embedded LND fork stores the gossip graph in native-SQL tables inside
+  // lnd.sqlite. LND decides which migrations to run from the single highest
+  // version in migration_tracker, so deleting individual tracker rows never
+  // makes it re-run a migration. The pre-v13.2.1 reset dropped the graph
+  // tables and deleted tracker rows 9/10 while rows 11-18 remained, leaving
+  // databases LND could never open again (issue #4524). The reset now only
+  // DELETEs rows inside one transaction and treats migration_tracker and
+  // schema_migrations as LND-owned. Repair rebuilds the schema for wallets
+  // the old reset already damaged.
+  //
+  // The DDL below mirrors the lnd fork's migrations 000008_graph and
+  // 000009_graph_v2 and must stay in sync with the fork pinned in
+  // fetch-libraries-versions.json. If the fork's highest migration version
+  // moves past GRAPH_KNOWN_MAX_DB_VERSION, repair fails closed until these
+  // definitions are reviewed against the new migrations.
+  private static final int GRAPH_V2_MIGRATION_VERSION = 11;   // 000009_graph_v2
+  private static final int GRAPH_KNOWN_MAX_DB_VERSION = 18;   // 000015_chain_params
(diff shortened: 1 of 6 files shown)


ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4533, opened 2026-08-31
chore: replace deprecated InteractionManager with requestIdleCallback

### LINKED ISSUE #4196: Migrate InteractionManager → requestIdleCallback (RN 0.86 deprecation)
## Description

`InteractionManager` is deprecated as of React Native 0.86. The dev-mode warning fires at runtime:

> `InteractionManager has been deprecated and will be removed in a future release. Please refactor long tasks into smaller ones, and use 'requestIdleCallback' instead.`

The replacement is `requestIdleCallback`, but the semantics differ — `runAfterInteractions` waits for any in-flight animations/gestures to settle, while `requestIdleCallback` schedules during browser-idle-equivalent time. Each call site deserves a quick look to confirm the new behaviour is acceptable.

## Call sites

- `stores/CashuStore.ts:3030` — `checkPendingItems()` defers a pending-invoice/quote sweep
- `views/Cashu/CashuSendingLightning.tsx:111` — defers `payLnInvoiceFromEcash` after the screen mounts so the navigation transition isn't janked
- `views/Wallet/Wallet.tsx:925` — iOS-only: defers the battery-saver modal until after navigation settles

## Suggested approach

Drop-in replacement for most sites:

```ts
// Before
InteractionManager.runAfterInteractions(() => doWork());

// After
requestIdleCallback(() => doWork());
```

For sites where the work explicitly needs to wait for animations (e.g., the `CashuSendingLightning` payment kickoff, which must not start before the screen transition completes to avoid frame drops), consider keeping a short `setTimeout(..., 0)` or using `InteractionManager` while it remains available, and noting the rationale.

## Acceptance

- [ ] All three call sites migrated or explicitly justified
- [ ] No `InteractionManager has been deprecated` warning in Android/iOS dev logs
- [ ] Manual smoke: cashu invoice flow, ecash payment, wallet startup with battery-saver toggled on iOS

## Refs

- Surfaced during the RN 0.85 → 0.86 + reanimated 4.5 + gesture-handler 3.1.0-nightly bump (Android ReanimatedSwipeable rendering fix).

kaloudis:
# Description

Relates to issue: **#4196**

`InteractionManager` is deprecated in React Native 0.86 and logs a warning the moment it is imported. Worth knowing before reading the diff: in 0.86 `InteractionManager` is no longer an implementation at all, it is a stub whose `runAfterInteractions` just calls `setImmediate` (`Libraries/Interaction/InteractionManager.js`), which under bridgeless is `queueMicrotask`. So all three call sites had already silently lost the wait-for-animations behaviour they were written for when the RN bump landed. This restores real deferral where it was wanted and removes the wrapper where it was not.

### New: `utils/SchedulingUtils.runWhenIdle`

A thin wrapper over the global `requestIdleCallback` that:

- holds its own timer alongside the idle callback, so the task still runs on a thread that never goes idle. React Native declares the `timeout` option on `requestIdleCallback` but never forwards it to the scheduler, so the option on its own is not a backstop: see [this thread](https://github.com/ZeusLN/zeus/pull/4533#discussion_r3916159460)
- returns a cancel function, safe to call after the task has run
- logs task failures rather than leaving an unhandled rejection behind (nothing observes the return value of an idle callback)
- falls back to a `setTimeout` macrotask where no idle scheduler exists: Jest, and the legacy runtime scheduler, where `requestIdleCallback` throws `requestIdleCallback is not supported in legacy runtime scheduler`

React Native ships no TypeScript declarations for `requestIdleCallback` and Zeus does not pull in the DOM lib, so `typings/idle-callback.d.ts` mirrors the Flow types from `NativeIdleCallbacks`. The handle is typed opaque on purpose: bridgeless returns a `jsi::Object`, legacy JSTimers returns a number.

### Call sites

**`stores/CashuStore.ts` `checkPendingItems()`** migrated. The pending invoice and unspent token sweep now waits for real idle time instead of a microtask, so its mint round trips no longer compete with startup. Still fire and forget, which the added doc comment now says out loud, pointing at `checkSentTokensSpentStatus` for the awaitable variant.

**`views/Cashu/CashuSendingLightning.tsx`** migrated with a 500 ms backstop. Two things worth reviewing:

- the kickoff is cancelled in `componentWillUnmount`, so a screen the user has already backed out of cannot start a payment. Previously the payment fired in a microtask so there was no window; with idle deferral there is one.
- no blank frame results from the delay: `render()` gates the sending animation on `!paymentSuccess && !paymentError`, not on `loading`, so `SendingLoadingView` is up from first paint either way.

**`views/Wallet/Wallet.tsx`** wrapper dropped, `setTimeout(..., 500)` kept. The timer is what actually lets the navigation transition settle before the iOS battery saver modal appears; the `InteractionManager` wrapper around it contributed nothing beyond a microtask. Behaviour is unchanged here by design, per the "keep a short timeout and note the rationale" option in the issue.

No source file imports `InteractionManager` any more, and no dependency imports React Native's `InteractionManager` on native (`react-native-gesture-handler` has a class by the same name, but it is its own web only implementation), so the deprecation warning should be gone.

- [ ] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

`utils/SchedulingUtils.test.ts`, 10 cases: idle scheduling, timeout backstop and default, cancellation in both directions, both fallback paths, and error handling for sync throws and rejected promises.

- [x] Android
- [x] iOS

Both platforms are green across the full [test plan](https://github.com/ZeusLN/zeus/pull/4533#issuecomment-5481825789): deprecation warning gone on a full bundle reload and confirmed still present on `master` as a baseline, repeated ecash Lightning payments all kicking off with no blank frame, failure path and fast back-out both clean, Cashu pending invoice and spent token sweeps still settling, and the embedded LND battery saver modal appearing unchanged under battery saver / Low Power Mode.

On-device
- [ ] LDK Node
- [x] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

Settled in review: the `timeout` option is dropped by React Native, so as originally written neither number affected scheduling at all. `runWhenIdle` now keeps its own timer as the deadline. The payment kickoff stays at 500 ms; the background sweep default moved from 2000 ms to 10 s, since a timer that really does fire would otherwise drop the mint round trips onto a startup that is still busy, which is what deferring them was for. Both are still judgement calls rather than measurements, happy to tune either.

### DIFF
--- a/utils/SchedulingUtils.ts
+++ b/utils/SchedulingUtils.ts
@@ -26,6 +26,12 @@ const runTask = async (task: () => void | Promise<void>) => {
 /**
  * Run `task` once the JS thread goes idle, or after `timeoutMs` at the latest.
  *
+ * `timeoutMs` is a deadline, not a delay. It is handed to
+ * `requestIdleCallback` as its `timeout` option, so it only bounds how long
+ * the idle scheduler may hold the task back. It has no effect on the fallback
+ * path, where there is no idle scheduler to wait on and the task runs on the
+ * next macrotask, well inside the deadline.
+ *
  * Returns a cancel function. Call it (from `componentWillUnmount`, say) to
  * drop a task that has not run yet; calling it afterwards is a no-op.
  */
@@ -61,7 +67,9 @@ export const runWhenIdle = (
 
     // No idle scheduler (Jest, legacy runtime). A macrotask at least gets the
     // work off the current call stack, which is all the deprecated
-    // `runAfterInteractions` stub did anyway.
+    // `runAfterInteractions` stub did anyway. Not `timeoutMs`: that is the
+    // latest the task may run, and with nothing to wait for there is no
+    // reason to hold it back.
     const timeoutId = setTimeout(run, 0);
     return () => {
         if (settled) return;
--- a/utils/SchedulingUtils.ts
+++ b/utils/SchedulingUtils.ts
@@ -13,9 +13,15 @@
 // first, short enough that a deferred sweep is not put off indefinitely.
 const DEFAULT_IDLE_TIMEOUT_MS = 2000;
 
+// A task may resolve to anything: the value is ignored, but the promise is
+// awaited so a rejection lands in the catch below. Typing this as
+// `Promise<void>` would reject the common `() => store.doSomething()` shape
+// and push callers into dropping the promise instead.
+type IdleTask = () => void | PromiseLike<unknown>;
+
 // Never rejects: nothing observes the result of an idle callback, so a
 // rejected task would otherwise surface as an unhandled rejection.
-const runTask = async (task: () => void | Promise<void>) => {
+const runTask = async (task: IdleTask) => {
     try {
         await task();
     } catch (error) {
@@ -34,9 +40,12 @@ const runTask = async (task: () => void | Promise<void>) => {
  *
  * Returns a cancel function. Call it (from `componentWillUnmount`, say) to
  * drop a task that has not run yet; calling it afterwards is a no-op.
+ *
+ * Return the promise from an async `task` rather than dropping it, so a
+ * rejection is logged here instead of surfacing as an unhandled rejection.
  */
 export const runWhenIdle = (
-    task: () => void | Promise<void>,
+    task: IdleTask,
     timeoutMs: number = DEFAULT_IDLE_TIMEOUT_MS
 ): (() => void) => {
     // Guards both directions: a cancel after the task ran, and a callback
--- a/utils/SchedulingUtils.ts
+++ b/utils/SchedulingUtils.ts
@@ -5,13 +5,23 @@
  * is now a stub that resolves on the next `setImmediate` tick, so it no longer
  * waits for animations or gestures to settle. `requestIdleCallback` is the
  * supported replacement: it runs the task when the JS runtime has spare time
- * in a frame, and it takes a timeout so the task still runs on a thread that
- * never goes idle.
+ * in a frame.
+ *
+ * Its `timeout` option is not a working backstop, so we keep our own timer
+ * alongside it. In `NativeIdleCallbacks::requestIdleCallback` the local
+ * `timeout` is declared and never assigned: `options.timeout` only feeds
+ * `expirationTime`, which sets the `didTimeout` flag we ignore. Every idle
+ * task is therefore scheduled with the default idle expiration of 5 minutes.
+ * Nothing fires a task when its expiration passes either: expiration is only
+ * the task queue's sort key, so a busy queue can hold an idle task back for
+ * as long as higher priority work keeps sorting ahead of it. Verified in
+ * 0.86.0, still present in 0.87.1.
  */
 
-// Backstop for background work. Long enough to let a busy startup finish
-// first, short enough that a deferred sweep is not put off indefinitely.
-const DEFAULT_IDLE_TIMEOUT_MS = 2000;
+// Backstop for background work. Long enough that a busy startup gets to
+// finish first rather than having a sweep dropped on top of it, short enough
+// that the work is not put off indefinitely.
+const DEFAULT_IDLE_TIMEOUT_MS = 10000;
 
 // A task may resolve to anything: the value is ignored, but the promise is
 // awaited so a rejection lands in the catch below. Typing this as
@@ -32,11 +42,12 @@ const runTask = async (task: IdleTask) => {
 /**
  * Run `task` once the JS thread goes idle, or after `timeoutMs` at the latest.
  *
- * `timeoutMs` is a deadline, not a delay. It is handed to
- * `requestIdleCallback` as its `timeout` option, so it only bounds how long
- * the idle scheduler may hold the task back. It has no effect on the fallback
- * path, where there is no idle scheduler to wait on and the task runs on the
- * next macrotask, well inside the deadline.
+ * `timeoutMs` is a deadline, not a delay: a thread with spare time runs the
+ * task well before it. The deadline is held by a timer of ours, not by
+ * `requestIdleCallback`, whose `timeout` option React Native drops (see the
+ * note at the top of this file). It has no effect on the fallback path, where
+ * there is no idle scheduler to wait on and the task runs on the next
+ * macrotask, well inside the deadline.
  *
  * Returns a cancel function. Call it (from `componentWillUnmount`, say) to
  * drop a task that has not run yet; calling it afterwards is a no-op.
@@ -51,27 +62,66 @@ export const runWhenIdle = (
     // Guards both directions: a cancel after the task ran, and a callback
     // that fires anyway after a cancel.
     let settled = false;
+    let timeoutId: ReturnType<typeof setTimeout> | undefined;
+    // Opaque `IdleCallbackHandle` (see typings/idle-callback.d.ts), so its
+    // presence is tracked separately rather than by testing it for undefined.
+    let idleHandle: unknown;
+    let hasIdleHandle = false;
+
+    // Whichever of the two fires first drops the other.
+    const clearPending = () => {
+        if (timeoutId !== undefined) {
+            clearTimeout(timeoutId);
+            timeoutId = undefined;
+        }
+        if (hasIdleHandle) {
+            hasIdleHandle = false;
+            if (typeof cancelIdleCallback === 'function') {
+                cancelIdleCallback(idleHandle);
+            }
+        }
+    };
+
     const run = () => {
         if (settled) return;
         settled = true;
+        clearPending();
         runTask(task);
     };
 
-    if (typeof requestIdleCallback === 'function') {
-        try {
-            const handle = requestIdleCallback(run, { timeout: timeoutMs });
-            return () => {
-                if (settled) return;
-                settled = true;
-                if (typeof cancelIdleCallback === 'function') {
-                    cancelIdleCallback(handle);
-                }
-            };
-        } catch {
-            // The legacy runtime scheduler throws 'requestIdleCallback is not
-            // supported in legacy runtime scheduler'. Fall through to the
-            // timer below rather than dropping the task on the floor.
+    const runFromIdle = () => {
+        // The scheduler is already running this callback, so the handle is
+        // spent and there is nothing left to cancel.
+        hasIdleHandle = false;
+        run();
+    };
+
+    const cancel = () => {
+        if (settled) return;
+        settled = true;
+        clearPending();
+    };
+
+    try {
+        // The feature detect is inside the try on purpose. React Native
+        // installs these globals as lazy getters over
+        // `TurboModuleRegistry.getEnforcing`, so a missing native module
+        // throws on the property read itself, not on the call.
+        if (typeof requestIdleCallback === 'function') {
+            // Still passed for correctness, and so this becomes a priority
+            // hint if React Native ever wires the option up. The timer below
+            // is what actually bounds the wait.
+            idleHandle = requestIdleCallback(runFromIdle, {
+                timeout: timeoutMs
+            });
+            hasIdleHandle = true;
+            timeoutId = setTimeout(run, timeoutMs);
+            return cancel;
         }
+    } catch {
+        // The legacy runtime scheduler throws 'requestIdleCallback is not
+        // supported in legacy runtime scheduler'. Fall through to the
+        // timer below rather than dropping the task on the floor.
     }
 
     // No idle scheduler (Jest, legacy runtime). A macrotask at least gets the
@@ -79,10 +129,6 @@ export const runWhenIdle = (
     // `runAfterInteractions` stub did anyway. Not `timeoutMs`: that is the
     // latest the task may run, and with nothing to wait for there is no
     // reason to hold it back.
-    const timeoutId = setTimeout(run, 0);
-    return () => {
-        if (settled) return;
-        settled = true;
-        clearTimeout(timeoutId);
-    };
+    timeoutId = setTimeout(run, 0);
+    return cancel;
 };
--- a/utils/SchedulingUtils.ts
+++ b/utils/SchedulingUtils.ts
@@ -1,0 +1,71 @@
+/**
+ * Helpers for deferring work that should not compete with rendering.
+ *
+ * React Native 0.86 deprecated `InteractionManager`. Its `runAfterInteractions`
+ * is now a stub that resolves on the next `setImmediate` tick, so it no longer
+ * waits for animations or gestures to settle. `requestIdleCallback` is the
+ * supported replacement: it runs the task when the JS runtime has spare time
+ * in a frame, and it takes a timeout so the task still runs on a thread that
+ * never goes idle.
+ */
+
+// Backstop for background work. Long enough to let a busy startup finish
+// first, short enough that a deferred sweep is not put off indefinitely.
+const DEFAULT_IDLE_TIMEOUT_MS = 2000;
+
+// Never rejects: nothing observes the result of an idle callback, so a
+// rejected task would otherwise surface as an unhandled rejection.
+const runTask = async (task: () => void | Promise<void>) => {
+    try {
+        await task();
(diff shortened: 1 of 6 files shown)


kaloudis:
## Device test plan

Build note: run `yarn` first if your `node_modules` predates the RN 0.86 merge. This branch depends on globals that only exist on 0.86.

Grouped by risk, highest first.

### A. Deprecation warning is actually gone

1. Trigger a **full** JS bundle reload, not Fast Refresh. `warnOnce` dedupes per bundle load, so a hot reload will not re-emit the warning and you will get a false pass.
2. Watch Metro output, or `adb logcat -s ReactNativeJS` on Android, or the Xcode console on iOS.
3. Expected: no `InteractionManager has been deprecated` line.
4. Worth doing once as a baseline: the same reload on `master` should still print it. That is what makes step 3 meaningful rather than just quiet.

### B. Ecash Lightning payment

The riskiest change. `CashuSendingLightning` now waits for idle with a 500 ms backstop and cancels the kickoff on unmount. Needs a Cashu wallet with a spendable balance.

5. Pay a small LN invoice from ecash, watching the moment of transition into the sending screen.
   * Expected: sending animation is on screen from the first frame, no blank or white flash, payment completes as before.
   * Failure signal: a visible gap before the animation, or the animation sitting there with the payment never starting.
6. Repeat five more times back to back. This is specifically checking that the idle callback always fires. A payment that silently never kicks off is the one regression this change could introduce, and it would be intermittent by nature.
7. Pay an invoice that will fail. An expired invoice is the easiest to arrange. Expected: error view renders as before.
8. Tap pay, then hardware back (Android) or swipe back (iOS) as fast as you can.
   * Expected: either the payment ran normally, or nothing happened at all and the balance is unchanged.
   * Never acceptable: a payment landing on a screen you already left, or a double spend.
9. If donations are enabled, run one donation payment through the same screen. Not changed here, but it shares the component and the second `payLnInvoiceFromEcash` path.

### C. Cashu pending sweep

`checkPendingItems` now defers to real idle with a 2000 ms backstop instead of a microtask. It runs on wallet focus and from ActivityStore.

10. Create a mint quote (receive ecash over Lightning) and leave the invoice unpaid. Navigate away from the wallet and back. Expected: no crash, no hang, wallet renders normally.
11. Pay that invoice from another wallet, then return to the Zeus wallet screen. Expected: the pending invoice flips to paid and the balance updates. It may take a beat longer than before, which is the intended effect.
12. Send an ecash token out, have it redeemed elsewhere, then return to the wallet screen. Expected: the activity entry flips to spent.

### D. iOS battery saver modal

Behavior here is meant to be **identical**. The 500 ms timer is unchanged; only the no-op wrapper around it was removed. Note the code path: this block sits inside `implementation === 'embedded-lnd'` and `connecting === true`, so it only fires on a fresh connect, meaning app launch or wallet switch, not a return from background.

13. iOS Settings > Battery > Low Power Mode ON.
14. Force quit Zeus and relaunch into an embedded LND wallet. Expected: battery saver modal appears shortly after the wallet loads, exactly as before.
15. Same launch on Android with battery saver on. Expected: modal appears. That branch is synchronous and untouched, so this is pure regression cover.

### E. Sanity

16. One cold start per platform with Cashu enabled, confirming nothing wedges during startup.

### Coverage notes

B and C need an ecash wallet and any node. D needs embedded LND specifically. A and E are backend agnostic, so a single remote LND pass covers the remainder of the matrix.

kaloudis:
iOS pass: all of A through E green.

- **A** warning gone on a full bundle reload, confirmed against `master` as the baseline
- **B** repeated ecash Lightning payments all kicked off, animation from the first frame, failure path and fast back-out both clean
- **C** pending invoice and spent token sweeps still settle
- **D** embedded LND battery saver modal appears unchanged under Low Power Mode
- **E** cold starts clean with Cashu enabled

Android still to run, so leaving this in draft.

kaloudis:
Android pass: all of A through E green as well, so both platforms are now covered.

Marking ready for review. CI is green on lint, prettier, test and tsc.

The one thing a reviewer may want to push back on: the 500 ms backstop on the ecash payment kickoff and the 2000 ms default for background sweeps are judgement calls rather than measurements. Both were exercised repeatedly on device without the backstop ever being the thing that fired, but I am happy to tune either.

utils/SchedulingUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+            // The legacy runtime scheduler throws 'requestIdleCallback is not
+            // supported in legacy runtime scheduler'. Fall through to the
+            // timer below rather than dropping the task on the floor.
+        }
+    }
+
+    // No idle scheduler (Jest, legacy runtime). A macrotask at least gets the
+    // work off the current call stack, which is all the deprecated
+    // `runAfterInteractions` stub did anyway.
+    const timeoutId = setTimeout(run, 0);
```
ajaysehwal: @kaloudis, nice write-up and test coverage on this one. One tiny nit: in the fallback path, it uses `setTimeout(run, 0)` instead of `setTimeout(run, timeoutMs)`, so `timeoutMs` is ignored there.
I think this is fine functionally since it preserves the existing behavior, but a short docstring note that `timeoutMs` only applies to the idle-scheduler path would make this clearer for future changes.

views/Cashu/CashuSendingLightning.tsx
```diff
@@ -106,13 +108,17 @@ export default class CashuSendingLightning extends React.Component<
...
 
-        // Reset payment state and start payment after navigation completes
+        // Reset payment state and start the payment once the JS thread is
+        // free, so the melt does not jank the navigation transition. The
+        // short timeout keeps the wait bounded on a busy thread; the kickoff
+        // is cancelled on unmount so a screen the user has already left never
+        // starts a payment.
         CashuStore.resetPaymentState();
-        InteractionManager.runAfterInteractions(() => {
+        this.cancelPaymentKickoff = runWhenIdle(() => {
```
ajaysehwal: this callback doesn't return the promise from `payLnInvoiceFromEcash()`:

```ts
this.cancelPaymentKickoff = runWhenIdle(() => {
    CashuStore.payLnInvoiceFromEcash({ amount: paymentAmount });
}, 500);

```

So runWhenIdle can't observe or log a rejection from this call. It's fine in practice since payLnInvoiceFromEcash already handles its own errors, but returning the promise would make the new error-handling behavior consistent with the other call sites

utils/SchedulingUtils.ts
kaloudis: Good catch, and thanks for reading it closely.

You are right that `timeoutMs` never reaches the fallback, but passing it to `setTimeout` would be the wrong fix: `timeoutMs` is a deadline, not a delay. It is the `timeout` option for `requestIdleCallback`, which bounds how long the idle scheduler is allowed to hold the task back, not how long to wait before running it. On the fallback path there is no idle scheduler to wait on, so the next macrotask is already well inside that bound. `setTimeout(run, timeoutMs)` would make the fallback strictly slower than the deprecated `runAfterInteractions` stub it replaces, and with the 2000 ms default it would sit on a Cashu sweep for two seconds on a runtime that had nothing else to do.

So I took your second suggestion. Pushed d6d10f868, which spells it out in the docstring:

```
 * `timeoutMs` is a deadline, not a delay. It is handed to
 * `requestIdleCallback` as its `timeout` option, so it only bounds how long
 * the idle scheduler may hold the task back. It has no effect on the fallback
 * path, where there is no idle scheduler to wait on and the task runs on the
 * next macrotask, well inside the deadline.
```

and at the fallback itself:

```
    // No idle scheduler (Jest, legacy runtime). A macrotask at least gets the
    // work off the current call stack, which is all the deprecated
    // `runAfterInteractions` stub did anyway. Not `timeoutMs`: that is the
    // latest the task may run, and with nothing to wait for there is no
    // reason to hold it back.
```

Same commit adds a test so a future change cannot quietly turn the deadline into a delay:

```ts
it('does not hold the task back for the caller timeout', () => {
    jest.useFakeTimers();
    delete globalScope.requestIdleCallback;
    const task = jest.fn();

    runWhenIdle(task, 5000);
    jest.advanceTimersByTime(0);

    expect(task).toHaveBeenCalledTimes(1);
});
```

views/Cashu/CashuSendingLightning.tsx
kaloudis: Agreed, and there is a cause behind it worth fixing rather than just the instance. Pushed b62383634.

The reason the kickoff has a braced body that drops the promise is the task type. `payLnInvoiceFromEcash` resolves to `CashuPayment | undefined`, and against `task: () => void | Promise<void>` the obvious form does not compile:

```
error TS2322: Type 'Promise<CashuPayment | undefined>' is not assignable to type 'void | Promise<void>'.
  Type 'CashuPayment | undefined' is not assignable to type 'void'.
```

So the signature was quietly pushing every call site whose task resolves to a value into dropping the promise, which is exactly the inconsistency you spotted. `checkPendingItems` only escaped it because its `async` arrow resolves to `void`.

Widened the task type to `() => void | PromiseLike<unknown>`, so the value is still ignored but the promise is awaited and `runTask`'s catch sees the rejection:

```ts
// A task may resolve to anything: the value is ignored, but the promise is
// awaited so a rejection lands in the catch below. Typing this as
// `Promise<void>` would reject the common `() => store.doSomething()` shape
// and push callers into dropping the promise instead.
type IdleTask = () => void | PromiseLike<unknown>;
```

and the call site now hands the promise back:

```ts
this.cancelPaymentKickoff = runWhenIdle(
    () =>
        CashuStore.payLnInvoiceFromEcash({
            amount: paymentAmount
        }),
    500
);
```

You are right that there is no behaviour change today: `payLnInvoiceFromEcash` wraps its body in a try/catch, reports through `paymentError`/`paymentErrorMsg`, and returns rather than rethrowing, so nothing reaches the new path. The point is that it is now consistent and stays correct if that ever changes.

Also added a test so the widened type keeps its meaning, that a promise returned from a non-async task is awaited and its rejection logged. Suite is 12 green; tsc, lint and prettier clean. The 2026-08-31 device passes still stand, since the kickoff awaits the same call it already made.

ajaysehwal (approved):
tACK

utils/SchedulingUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+    let settled = false;
+    const run = () => {
+        if (settled) return;
+        settled = true;
+        runTask(task);
+    };
+
+    if (typeof requestIdleCallback === 'function') {
+        try {
+            const handle = requestIdleCallback(run, { timeout: timeoutMs });
```
shubhamkmr04: **Blocking: the `timeout` backstop does not work on bridgeless — RN never forwards it.**

In [`NativeIdleCallbacks.cpp` L84](https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/ReactCommon/react/nativemodule/idlecallbacks/NativeIdleCallbacks.cpp#L84) the `timeout` variable is declared and never assigned. `options.timeout` only feeds `expirationTime`, which just sets the `didTimeout` flag we ignore. So [L122](https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/ReactCommon/react/nativemodule/idlecallbacks/NativeIdleCallbacks.cpp#L122) always takes the else branch and the task gets the default idle deadline — **5 minutes**. That is the real ceiling at both call sites, not 500/2000 ms.

The device passes would not show this: on a quiet thread the scheduler picks the task up within a frame or two. Only the upper bound is missing, and that bites only when the thread is busy.

Worst case is the `CashuSendingLightning` kickoff — a busy thread can leave the user on the sending animation with the melt not started. That is step 6 of your test plan, capped at 5 minutes instead of 500 ms.

Fix: keep our own timer next to the idle callback. `settled` already makes first-to-fire safe.

```ts
const handle = requestIdleCallback(run, { timeout: timeoutMs });
timeoutId = setTimeout(run, timeoutMs); // the actual backstop
// run() clears timeoutId; cancel clears both the timer and the handle
```

This also answers your open question on tuning: as written, 500 and 2000 have no effect on scheduling.

*Nit:* the `typeof requestIdleCallback` check on L60 is outside the `try`. RN installs that global as a lazy getter over `getEnforcing('NativeIdleCallbacksCxx')`, so a missing module throws on the property read itself, skipping the fallback.

kaloudis: Confirmed, and fixed in 9103efd7a. Thank you for digging into the C++, this was a real hole: the backstop was the whole stated contract of the util, and it was inert.

I read the same path and agree on every step. `NativeIdleCallbacks.cpp` L84 declares `timeout` and never assigns it, L87-92 only compute `expirationTime` for the `didTimeout` flag we ignore, so L122 always takes the else branch and `scheduleIdleTask` falls back to `timeoutForSchedulerPriority(IdlePriority)` ([`SchedulerPriorityUtils.h` L52](https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/ReactCommon/react/renderer/runtimescheduler/SchedulerPriorityUtils.h#L52)), 5 minutes. Same code in 0.87.1, so #4546 does not make it go away.

Two things I found while checking, both of which strengthen your conclusion rather than soften it:

1. A forwarded `timeout` would never have meant 500 ms anyway. [`getResolvedTimeoutForIdleTask`](https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/ReactCommon/react/renderer/runtimescheduler/RuntimeScheduler_Modern.cpp#L21) adds LowPriority's 10 s to any custom timeout below 5 minutes, so even a working option would have given us 10.5 s.
2. The 5 minutes is not a ceiling in the timer sense either. Nothing fires a task when its expiration passes: expiration is only the min-heap key ([`Task.h` L48](https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/ReactCommon/react/renderer/runtimescheduler/Task.h#L48)) and the `didTimeout` flag. `runEventLoop` drains in expiration order, so an idle task runs as soon as the queue drains to it, which is why the device passes were clean, and it is held back only while higher priority work keeps sorting ahead of it. So there is no wall clock backstop at any duration, and no value of the option that would produce one.

Which makes a JS timer the only way to get one, and it does work: under bridgeless a `setTimeout` callback comes back through `RuntimeExecutor` to [`scheduleWork`](https://github.com/facebook/react-native/blob/v0.86.0/packages/react-native/ReactCommon/react/renderer/runtimescheduler/RuntimeScheduler_Modern.cpp#L45) at ImmediatePriority, front of the queue, driven by a platform timer off the JS thread.

What landed:

- a timer next to the idle callback as the actual deadline, first to fire drops the other, `settled` unchanged. The option is still passed so it becomes a priority hint if RN ever wires it up.
- the feature detect moved inside the `try`, per your nit. Agreed on the mechanism: `polyfillGlobal` installs the global through `defineLazyObjectProperty`, and the spec module is `TurboModuleRegistry.getEnforcing('NativeIdleCallbacksCxx')` at module scope, so a missing module throws on the property read and we would have thrown out of `componentDidMount` rather than falling back. Unlikely on bridgeless 0.86, free to guard.
- docstring and PR description corrected. Both claimed a backstop that did not exist.
- tests: the deadline fires when idle time never comes, the deadline path cancels the idle callback, the idle path cancels the timer, and a throwing global getter still falls back.

One judgement call worth your eyes, since it goes slightly beyond the fix. I raised the default from 2000 to 10000 ms. With the number inert, `checkPendingItems` effectively ran when the queue drained, which is exactly what a background mint sweep wants. Now that it is a real timer, 2 s would fire those round trips into a startup that is still busy, which is the thing the deferral was for. 10 s reads to me as "never, unless something is wrong". The 500 ms on the payment kickoff is unchanged, since there a bounded wait is the point. Happy to move either number.

That also answers the tuning question in the description: as written neither number did anything, and now they both do.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):

PR #4537, opened 2026-09-01
fix(nwc): return RATE_LIMITED instead of UNAUTHORIZED during connection updates

### LINKED ISSUE #4503: NWC: requests during update windows are rejected as UNAUTHORIZED, clients may treat a busy edit as revocation
### Summary

Follow-up to PR #4443 and PR #4500, split out from review discussion on #4500 so it does not get lost.

`withGlobalHandler` (`stores/NostrWalletConnectStore.ts`) rejects incoming NIP-47 requests with `UNAUTHORIZED` and a "connection not found" message while the connection id is in `pendingConnectionMutationIds`. For a delete window that is semantically right: the connection is going away. For an update window it is a transient busy condition, not a revocation.

### Problem

- Since #4500, budget/permission edits and relay changes guard `updateConnection` with the same pending set, so a client request that lands in the window is answered `UNAUTHORIZED`.
- The relay-change window spans `publishWalletServiceInfoWithRetry` (with backoff) plus a 500ms settle sleep, so it is not always tiny.
- #4500 also moved the unsubscribe to after the publish work (correctly, so a failed rotation no longer strands the connection). A side effect is that these rejections are now actually delivered over the still-live old subscription instead of being dropped, so the failure mode is more visible than it was pre-#4500.
- NIP-47 clients commonly treat `UNAUTHORIZED` as pairing revocation and may drop the connection over a routine edit.

### Proposal

Track the mutation kind, e.g. `Map<connectionId, 'delete' | 'update'>`, and return a retryable `INTERNAL_ERROR` for update windows while keeping `UNAUTHORIZED` for delete windows.

### Related hardening

`deleteConnection` and `updateConnection` share the single `pendingConnectionMutationIds` Set. If both run concurrently for the same id, whichever finishes first removes the id in its `finally` and unguards the other mid-mutation. The mutation-kind map above (or a refcount) would close this too.

ajaysehwal:
# Description

Fixes #4503

### Problem
`withGlobalHandler` rejected every NWC request while a connection was in `pendingConnectionMutationIds` with `UNAUTHORIZED` + "connection not found".

- For **delete** this is correct (the pairing is ending).
- For **update** (budget, permissions, name, expiry) it is only a short critical section. Many NWC clients treat `UNAUTHORIZED` as permanent revocation and drop the connection.

### Changes
- Track mutation kind (`'delete' | 'update' | 'rotate'`) instead of a plain set
- Keep returning `UNAUTHORIZED` for in-progress **deletes** and **relay rotations** (true revocation)
- Return `RATE_LIMITED` with a clear temporary message for in-progress **updates** (clients treat this as retryable, not as pairing revocation)
- Never downgrade a concurrent delete marker to `update`
- Only clear a mutation’s own marker in its `finally` block (prevents one mutation from unguarding another)
- Apply the same distinction both before and after `markConnectionUsed`
- Extract shared rejection logic into `mutationRejection()`

The race guard introduced in #4500 is preserved; only the error signal for the update window changes.

### Notes
- Relay change rotates the client pubkey and permanently invalidates the old pairing, so it is intentionally treated like a delete (`UNAUTHORIZED`).
- `RATE_LIMITED` matches existing precedent in this store (`paymentQueueWaitExceeded`, `makeInvoiceRateLimited`) and carries explicit retry semantics.

---

### Test plan

- [x] Start an NWC connection, then edit budget/permissions while a client is actively requesting (`get_info` / `get_balance` / `pay_invoice`)
- [x] Confirm the client receives `RATE_LIMITED` (not `UNAUTHORIZED`) and does **not** drop the pairing
- [x] Delete a connection while a request is in flight → still receives `UNAUTHORIZED`
- [x] Change relay / regenerate → still receives `UNAUTHORIZED` for the old pairing
- [x] Concurrent delete + update does not downgrade the delete marker
- [x] `yarn verify` (tsc / eslint / prettier) passes
- [x] Store unit tests pass

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -1776,7 +1776,7 @@
     "stores.NostrWalletConnectStore.error.failedToStartService": "Failed to start service",
     "stores.NostrWalletConnectStore.error.connectionNameExists": "Connection name already exists",
     "stores.NostrWalletConnectStore.error.connectionNotFound": "Connection not found",
-    "stores.NostrWalletConnectStore.error.connectionUpdating": "Connection is being updated, please try again",
+    "stores.NostrWalletConnectStore.error.connectionUpdating": "Connection is being updated. Please try again.",
     "stores.NostrWalletConnectStore.error.paymentQueueWaitExceeded": "Another payment was in progress and this request waited too long; it was not executed. Please try again.",
     "stores.NostrWalletConnectStore.error.makeInvoiceRateLimited": "Too many invoice requests from this connection. Please wait before trying again.",
     "stores.NostrWalletConnectStore.error.failedToConnectRelay": "Failed to Connect {{relayUrl}}",
--- a/locales/en.json
+++ b/locales/en.json
@@ -1776,6 +1776,7 @@
     "stores.NostrWalletConnectStore.error.failedToStartService": "Failed to start service",
     "stores.NostrWalletConnectStore.error.connectionNameExists": "Connection name already exists",
     "stores.NostrWalletConnectStore.error.connectionNotFound": "Connection not found",
+    "stores.NostrWalletConnectStore.error.connectionUpdating": "Connection is being updated, please try again",
     "stores.NostrWalletConnectStore.error.paymentQueueWaitExceeded": "Another payment was in progress and this request waited too long; it was not executed. Please try again.",
     "stores.NostrWalletConnectStore.error.makeInvoiceRateLimited": "Too many invoice requests from this connection. Please wait before trying again.",
     "stores.NostrWalletConnectStore.error.failedToConnectRelay": "Failed to Connect {{relayUrl}}",
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -1306,7 +1306,10 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
                 error: undefined
             })
         );
-        expect(lateResponse.error?.code).toBe('UNAUTHORIZED');
+        expect(lateResponse.error).toEqual({
+            code: 'INTERNAL_ERROR',
+            message: 'stores.NostrWalletConnectStore.error.connectionUpdating'
+        });
 
         releaseHandler({ result: { ok: true } });
         await handlerDone;
@@ -1314,6 +1317,35 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
 
         expect(updated).toBe(true);
         expect(connection.maxAmountSats).toBe(0);
+    });
+
+    it('returns an internal error when an update starts while marking a connection used', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store, {
+            expiresAt: new Date(Date.now() + 60_000)
+        });
+        const handler = jest.fn();
+
+        jest.spyOn(store as any, 'markConnectionUsed').mockImplementation(
+            async () => {
+                (store as any).pendingConnectionMutations.set(
+                    connection.id,
+                    'update'
+                );
+                return true;
+            }
+        );
+
+        const response = await (store as any).withGlobalHandler(
+            connection.id,
+            handler
+        );
+
+        expect(response.error).toEqual({
+            code: 'INTERNAL_ERROR',
+            message: 'stores.NostrWalletConnectStore.error.connectionUpdating'
+        });
+        expect(handler).not.toHaveBeenCalled();
     });
 
     it('defers relay release when updateConnection awaited in-flight handlers', async () => {
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -278,6 +278,54 @@ describe('NostrWalletConnectStore relay rotation', () => {
         ).toBe('delete');
     });
 
+    it('does not downgrade a pre-existing delete marker when an update starts', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store);
+        (store as any).pendingConnectionMutations.set(connection.id, 'delete');
+
+        let markerDuringUpdate: string | undefined;
+        jest.spyOn(store as any, 'subscribeToConnection').mockImplementation(
+            async () => {
+                markerDuringUpdate = (
+                    store as any
+                ).pendingConnectionMutations.get(connection.id);
+            }
+        );
+
+        await store.updateConnection(connection.id, {
+            name: 'Renamed App'
+        });
+
+        expect(markerDuringUpdate).toBe('delete');
+        expect(
+            (store as any).pendingConnectionMutations.get(connection.id)
+        ).toBe('delete');
+    });
+
+    it('does not downgrade a pre-existing rotate marker when a second update starts', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store);
+        (store as any).pendingConnectionMutations.set(connection.id, 'rotate');
+
+        let markerDuringUpdate: string | undefined;
+        jest.spyOn(store as any, 'subscribeToConnection').mockImplementation(
+            async () => {
+                markerDuringUpdate = (
+                    store as any
+                ).pendingConnectionMutations.get(connection.id);
+            }
+        );
+
+        await store.updateConnection(connection.id, {
+            name: 'Renamed App'
+        });
+
+        expect(markerDuringUpdate).toBe('rotate');
+        expect(
+            (store as any).pendingConnectionMutations.get(connection.id)
+        ).toBe('rotate');
+    });
+
     it('leaves relayUrl unchanged when storing the new key fails so retry can rotate', async () => {
         const store = buildStore();
         const connection = seedConnection(store);
@@ -1235,6 +1283,30 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         expect(deleted).toBe(true);
         expect(store.connections).toHaveLength(0);
         expect(store.waitingForInFlightHandlers).toBe(false);
+    });
+
+    it('does not clear a marker installed by a concurrent update when delete finishes', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store, {
+            expiresAt: new Date(Date.now() + 60_000)
+        });
+
+        jest.spyOn(store as any, 'deleteClientKeys').mockImplementation(
+            async () => {
+                // Simulate an update overwriting the marker after this
+                // delete already committed to removing the connection.
+                (store as any).pendingConnectionMutations.set(
+                    connection.id,
+                    'update'
+                );
+            }
+        );
+
+        await store.deleteConnection(connection.id);
+
+        expect(
+            (store as any).pendingConnectionMutations.get(connection.id)
+        ).toBe('update');
     });
 
     it('defers relay release when delete awaited in-flight handlers', async () => {
--- a/stores/NostrWalletConnectStore.test.ts
+++ b/stores/NostrWalletConnectStore.test.ts
@@ -231,6 +231,53 @@ describe('NostrWalletConnectStore relay rotation', () => {
         expect((store as any).publishedRelays.has(OLD_RELAY)).toBe(false);
     });
 
+    it('marks a relay-changing update as rotate, not update, while in flight', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store);
+        let markerDuringMutation: string | undefined;
+        jest.spyOn(store as any, 'subscribeToConnection').mockImplementation(
+            async () => {
+                markerDuringMutation = (
+                    store as any
+                ).pendingConnectionMutations.get(connection.id);
+            }
+        );
+
+        const result = await store.updateConnection(connection.id, {
+            relayUrl: NEW_RELAY
+        });
+
+        expect(result.success).toBe(true);
+        expect(markerDuringMutation).toBe('rotate');
+        expect(
+            (store as any).pendingConnectionMutations.has(connection.id)
+        ).toBe(false);
+    });
+
+    it('does not clear a concurrent delete marker when an update finishes', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store);
+
+        jest.spyOn(store as any, 'subscribeToConnection').mockImplementation(
+            async () => {
+                // Simulate a delete winning the race while this update is
+                // still in its critical section.
+                (store as any).pendingConnectionMutations.set(
+                    connection.id,
+                    'delete'
+                );
+            }
+        );
+
+        await store.updateConnection(connection.id, {
+            name: 'Renamed App'
+        });
+
+        expect(
+            (store as any).pendingConnectionMutations.get(connection.id)
+        ).toBe('delete');
+    });
+
     it('leaves relayUrl unchanged when storing the new key fails so retry can rotate', async () => {
         const store = buildStore();
         const connection = seedConnection(store);
@@ -1307,7 +1354,7 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
             })
         );
         expect(lateResponse.error).toEqual({
-            code: 'INTERNAL_ERROR',
+            code: 'RATE_LIMITED',
             message: 'stores.NostrWalletConnectStore.error.connectionUpdating'
         });
 
@@ -1319,7 +1366,7 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         expect(connection.maxAmountSats).toBe(0);
     });
 
-    it('returns an internal error when an update starts while marking a connection used', async () => {
+    it('returns a rate-limit error when an update starts while marking a connection used', async () => {
         const store = buildStore();
         const connection = seedConnection(store, {
             expiresAt: new Date(Date.now() + 60_000)
@@ -1342,8 +1389,66 @@ describe('NostrWalletConnectStore connection expiry enforcement', () => {
         );
 
         expect(response.error).toEqual({
-            code: 'INTERNAL_ERROR',
+            code: 'RATE_LIMITED',
             message: 'stores.NostrWalletConnectStore.error.connectionUpdating'
+        });
+        expect(handler).not.toHaveBeenCalled();
+    });
+
+    it('returns UNAUTHORIZED when a delete starts while marking a connection used', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store, {
+            expiresAt: new Date(Date.now() + 60_000)
+        });
+        const handler = jest.fn();
+
+        jest.spyOn(store as any, 'markConnectionUsed').mockImplementation(
+            async () => {
+                (store as any).pendingConnectionMutations.set(
+                    connection.id,
+                    'delete'
+                );
+                return true;
+            }
+        );
+
+        const response = await (store as any).withGlobalHandler(
+            connection.id,
+            handler
+        );
+
+        expect(response.error).toEqual({
+            code: 'UNAUTHORIZED',
+            message: 'stores.NostrWalletConnectStore.error.connectionNotFound'
+        });
+        expect(handler).not.toHaveBeenCalled();
+    });
+
+    it('returns UNAUTHORIZED when a relay rotation starts while marking a connection used', async () => {
+        const store = buildStore();
+        const connection = seedConnection(store, {
+            expiresAt: new Date(Date.now() + 60_000)
+        });
+        const handler = jest.fn();
+
+        jest.spyOn(store as any, 'markConnectionUsed').mockImplementation(
+            async () => {
+                (store as any).pendingConnectionMutations.set(
+                    connection.id,
+                    'rotate'
+                );
+                return true;
+            }
+        );
+
+        const response = await (store as any).withGlobalHandler(
+            connection.id,
+            handler
+        );
+
+        expect(response.error).toEqual({
+            code: 'UNAUTHORIZED',
+            message: 'stores.NostrWalletConnectStore.error.connectionNotFound'
         });
         expect(handler).not.toHaveBeenCalled();
     });
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -110,6 +110,8 @@ export const DEFAULT_NOSTR_RELAYS = [
 
 type AppStateSubscription = ReturnType<typeof AppState.addEventListener>;
 
+type PendingConnectionMutation = 'delete' | 'update' | 'rotate';
+
 interface ClientKeys {
     [pubkey: string]: string;
 }
@@ -187,7 +189,10 @@ export default class NostrWalletConnectStore {
         number
     >();
     private makeInvoiceTimestampsByConnection = new Map<string, number[]>();
-    private pendingConnectionMutations = new Map<string, 'delete' | 'update'>();
+    private pendingConnectionMutations = new Map<
+        string,
+        PendingConnectionMutation
+    >();
     private inFlightHandlersByConnection = new Map<
         string,
         Set<Promise<unknown>>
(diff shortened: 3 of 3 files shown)


kaloudis (commented):
Approach ACK. The core change is right and cleanly done: `pendingConnectionMutationIds` to `Map<string, 'delete' | 'update'>` is fully renamed with no stale references, and the mutation kind is exactly the shape #4503 asked for.

Verified locally rather than eyeballed, on the branch rebased onto current master (`2473e4fb2`):

- Master has not touched `stores/NostrWalletConnectStore.ts`, its test file, `utils/NostrConnectUtils.ts`, or `views/Settings/NostrWalletConnect/` since your base, so the 30-commit gap is textual only. Rebases clean.
- tsc, eslint, prettier clean; 61/61 store tests pass on the rebased branch.
- No deadlock introduced. `updateConnection` sets the marker synchronously and then awaits `awaitInFlightHandlers`; a request that hits either mutation branch returns immediately and releases its gate in the `finally`, so the guard cannot block the drain that installed it.
- The new test is non-vacuous: removing the post-mark `'update'` branch makes it fail.

Two items I would like to settle before merge, then some nits.

## 1. Relay change is a revocation, not a transient busy state

When `updates.relayUrl` differs, `updateConnection` rotates the credential: `generateConnectionSecret` -> `connection.pubkey = rotatedSecret.connectionPublicKey` -> `deleteClientKeys(oldPubkey)`. The old pairing URI is permanently dead and the user has to scan a new one. That is the case where `UNAUTHORIZED` is precisely correct: the enum's own comment reads "pubkey has no wallet connected", which is literally true for the old client after rotation.

So this PR swaps in a retryable error on the one update branch where revocation is real, and that is also the longest window (`publishWalletServiceInfoWithRetry` plus the 500ms settle), i.e. the one most likely to be hit. Practical effect on the success path: the old subscription and relay are torn down afterwards, so the client gets silence either way; `UNAUTHORIZED` just tells it immediately instead of leaving it retrying into a void.

`relayUrlChanged` is already computed above the `set` call, so this is cheap:

```ts
this.pendingConnectionMutations.set(
    connectionId,
    relayUrlChanged ? 'rotate' : 'update'
);
```

and treat `'rotate'` like `'delete'` in `withGlobalHandler`.

Counter-argument worth weighing: if the rotation throws, the old pairing survives and the `UNAUTHORIZED` was a false revocation. My read is that the success path dominates, but it is your call. Either way, the PR description's blanket "update is only a short critical section" does not hold for this branch, so the reasoning should be recorded in a comment.

## 2. "Fixes #4503" but the Related hardening half is not closed

#4503's second section is the concurrent delete plus update case, and a plain `Map` does not close it:

- `updateConnection` overwrites an existing `'delete'` with `'update'`, downgrading a live delete window to a retryable error.
- Whichever `finally` runs first calls `.delete(connectionId)` outright, unguarding the other mid-mutation. That is the original bug, unchanged.

Reachable in the UI: `AddOrEditNWCConnection.tsx` renders `leftComponent="Back"` unconditionally, so nothing stops you leaving the edit screen while the save spinner is up and then hitting Delete on the details screen.

Minimal patch that closes the downgrade and one of the two unguard directions:

```ts
// updateConnection
if (this.pendingConnectionMutations.get(connectionId) !== 'delete') {
    this.pendingConnectionMutations.set(connectionId, 'update');
}
...
} finally {
    if (this.pendingConnectionMutations.get(connectionId) === 'update') {
        this.pendingConnectionMutations.delete(connectionId);
    }
}
```

A refcount is what fully closes it. If you would rather not take that on here, change `Fixes #4503` to `Part of #4503` so the issue stays open.

## 3. Is `INTERNAL_ERROR` the code clients actually retry? (discussion)

The PR rests on "clients treat this as retryable", but NIP-47's retryable code is `RATE_LIMITED` ("The client is sending commands too fast. It should retry in a few seconds"). `INTERNAL` is a generic hard failure, and Zeus's enum value is the non-spec string `INTERNAL_ERROR` anyway (pre-existing, not your problem).

More to the point, this store already answers the identical class of condition with `RATE_LIMITED`: `paymentQueueWaitExceeded` in `enqueuePayment`, and `makeInvoiceRateLimited`. Both are "busy, come back shortly", same as an update window.

This matters because the window is not always small. `awaitInFlightHandlers` can block on an in-flight `pay_invoice` for up to `PAYMENT_TIMEOUT_SECONDS` (120), and every request that lands in that span is rejected. `RATE_LIMITED` is the code that carries explicit backoff semantics for exactly that.

Both codes meet the stated goal of not signalling revocation, so this is not a blocker, but I think `RATE_LIMITED` is the better signal and it matches existing precedent in this file.

## Nits

- **The post-mark `'delete'` branch has no test coverage.** Verified by mutation: deleting `if (mutationAfterMark === 'delete') return this.unauthorizedConnectionResponse<T>();` leaves 61/61 tests green. You added the symmetric test for the update branch; the same shape with `'delete'` pins the half that preserves the revocation signal across the await. (This was uncovered before your PR too, since the old combined `!marked || pending` check was only ever exercised through the pre-mark path, but this PR is the natural place to fix it.)
- The two mutation blocks are byte-identical. A small helper collapses about 20 lines to about 8:
  ```ts
  private mutationRejection<T>(connectionId: string): T | undefined {
      const mutation = this.pendingConnectionMutations.get(connectionId);
      if (mutation === 'delete') return this.unauthorizedConnectionResponse<T>();
      if (mutation === 'update')
          return NostrConnectUtils.createNip47Error(
              localeString('stores.NostrWalletConnectStore.error.connectionUpdating'),
              Nip47ErrorCode.INTERNAL_ERROR
          ) as T;
      return undefined;
  }
  ```
- Locale copy is a comma splice. "Connection is being updated. Please try again." matches the style of `paymentQueueWaitExceeded` right above it.
- Checklist: `yarn run test` is unchecked even though the PR adds tests, and Android is untested.

Two full-suite failures I saw locally (`PhotoUtils` relative-path assertion, an `LndUtils` worker SIGSEGV) are artifacts of running from a git worktree, not this PR. CI is green.

ajaysehwal:
> Approach ACK. The core change is right and cleanly done: `pendingConnectionMutationIds` to `Map<string, 'delete' | 'update'>` is fully renamed with no stale references, and the mutation kind is exactly the shape #4503 asked for.
> 
> Verified locally rather than eyeballed, on the branch rebased onto current master (`2473e4fb2`):
> 
> * Master has not touched `stores/NostrWalletConnectStore.ts`, its test file, `utils/NostrConnectUtils.ts`, or `views/Settings/NostrWalletConnect/` since your base, so the 30-commit gap is textual only. Rebases clean.
> * tsc, eslint, prettier clean; 61/61 store tests pass on the rebased branch.
> * No deadlock introduced. `updateConnection` sets the marker synchronously and then awaits `awaitInFlightHandlers`; a request that hits either mutation branch returns immediately and releases its gate in the `finally`, so the guard cannot block the drain that installed it.
> * The new test is non-vacuous: removing the post-mark `'update'` branch makes it fail.
> 
> Two items I would like to settle before merge, then some nits.
> 
> ## 1. Relay change is a revocation, not a transient busy state
> When `updates.relayUrl` differs, `updateConnection` rotates the credential: `generateConnectionSecret` -> `connection.pubkey = rotatedSecret.connectionPublicKey` -> `deleteClientKeys(oldPubkey)`. The old pairing URI is permanently dead and the user has to scan a new one. That is the case where `UNAUTHORIZED` is precisely correct: the enum's own comment reads "pubkey has no wallet connected", which is literally true for the old client after rotation.
> 
> So this PR swaps in a retryable error on the one update branch where revocation is real, and that is also the longest window (`publishWalletServiceInfoWithRetry` plus the 500ms settle), i.e. the one most likely to be hit. Practical effect on the success path: the old subscription and relay are torn down afterwards, so the client gets silence either way; `UNAUTHORIZED` just tells it immediately instead of leaving it retrying into a void.
> 
> `relayUrlChanged` is already computed above the `set` call, so this is cheap:
> 
> ```ts
> this.pendingConnectionMutations.set(
>     connectionId,
>     relayUrlChanged ? 'rotate' : 'update'
> );
> ```
> 
> and treat `'rotate'` like `'delete'` in `withGlobalHandler`.
> 
> Counter-argument worth weighing: if the rotation throws, the old pairing survives and the `UNAUTHORIZED` was a false revocation. My read is that the success path dominates, but it is your call. Either way, the PR description's blanket "update is only a short critical section" does not hold for this branch, so the reasoning should be recorded in a comment.
> 
> ## 2. "Fixes #4503" but the Related hardening half is not closed
> #4503's second section is the concurrent delete plus update case, and a plain `Map` does not close it:
> 
> * `updateConnection` overwrites an existing `'delete'` with `'update'`, downgrading a live delete window to a retryable error.
> * Whichever `finally` runs first calls `.delete(connectionId)` outright, unguarding the other mid-mutation. That is the original bug, unchanged.
> 
> Reachable in the UI: `AddOrEditNWCConnection.tsx` renders `leftComponent="Back"` unconditionally, so nothing stops you leaving the edit screen while the save spinner is up and then hitting Delete on the details screen.
> 
> Minimal patch that closes the downgrade and one of the two unguard directions:
> 
> ```ts
> // updateConnection
> if (this.pendingConnectionMutations.get(connectionId) !== 'delete') {
>     this.pendingConnectionMutations.set(connectionId, 'update');
> }
> ...
> } finally {
>     if (this.pendingConnectionMutations.get(connectionId) === 'update') {
>         this.pendingConnectionMutations.delete(connectionId);
>     }
> }
> ```
> 
> A refcount is what fully closes it. If you would rather not take that on here, change `Fixes #4503` to `Part of #4503` so the issue stays open.
> 
> ## 3. Is `INTERNAL_ERROR` the code clients actually retry? (discussion)
> The PR rests on "clients treat this as retryable", but NIP-47's retryable code is `RATE_LIMITED` ("The client is sending commands too fast. It should retry in a few seconds"). `INTERNAL` is a generic hard failure, and Zeus's enum value is the non-spec string `INTERNAL_ERROR` anyway (pre-existing, not your problem).
> 
> More to the point, this store already answers the identical class of condition with `RATE_LIMITED`: `paymentQueueWaitExceeded` in `enqueuePayment`, and `makeInvoiceRateLimited`. Both are "busy, come back shortly", same as an update window.
> 
> This matters because the window is not always small. `awaitInFlightHandlers` can block on an in-flight `pay_invoice` for up to `PAYMENT_TIMEOUT_SECONDS` (120), and every request that lands in that span is rejected. `RATE_LIMITED` is the code that carries explicit backoff semantics for exactly that.
> 
> Both codes meet the stated goal of not signalling revocation, so this is not a blocker, but I think `RATE_LIMITED` is the better signal and it matches existing precedent in this file.
> 
> ## Nits
> * **The post-mark `'delete'` branch has no test coverage.** Verified by mutation: deleting `if (mutationAfterMark === 'delete') return this.unauthorizedConnectionResponse<T>();` leaves 61/61 tests green. You added the symmetric test for the update branch; the same shape with `'delete'` pins the half that preserves the revocation signal across the await. (This was uncovered before your PR too, since the old combined `!marked || pending` check was only ever exercised through the pre-mark path, but this PR is the natural place to fix it.)
> * The two mutation blocks are byte-identical. A small helper collapses about 20 lines to about 8:
>   ```ts
>   private mutationRejection<T>(connectionId: string): T | undefined {
>       const mutation = this.pendingConnectionMutations.get(connectionId);
>       if (mutation === 'delete') return this.unauthorizedConnectionResponse<T>();
>       if (mutation === 'update')
>           return NostrConnectUtils.createNip47Error(
>               localeString('stores.NostrWalletConnectStore.error.connectionUpdating'),
>               Nip47ErrorCode.INTERNAL_ERROR
>           ) as T;
>       return undefined;
>   }
>   ```
> * Locale copy is a comma splice. "Connection is being updated. Please try again." matches the style of `paymentQueueWaitExceeded` right above it.
> * Checklist: `yarn run test` is unchecked even though the PR adds tests, and Android is untested.
> 
> Two full-suite failures I saw locally (`PhotoUtils` relative-path assertion, an `LndUtils` worker SIGSEGV) are artifacts of running from a git worktree, not this PR. CI is green.

fixed in 6855f309a7ec5411bb38be03b22e5a3674f7f4c8

kaloudis (commented):
Round 2. All three round-1 items are addressed and the `RATE_LIMITED` switch is the right call. One should-fix left: the no-downgrade rule got applied to `delete` but not to `rotate`, which leaves the rotate window worse off than a plain update window.

**Verified on the branch rebased onto current master (`2473e4fb2`), not eyeballed**

- Master still has not touched `stores/NostrWalletConnectStore.ts`, its test file, `utils/NostrConnectUtils.ts`, or `views/Settings/NostrWalletConnect/` since your base, so the rebase is textual only and these results apply to the PR as-is.
- tsc / eslint / prettier clean, 65/65 store tests pass.
- Round-1 coverage gap is closed: deleting the post-`markConnectionUsed` rejection block now fails 3 tests (it failed 0 before). Flipping `ownMutationKind` to always be `'update'`, and making `mutationRejection` treat `rotate` as retryable, each fail a test as well. The new tests are non-vacuous.

### Should fix: a plain update downgrades, then unguards, an in-flight rotate

`updateConnection` only refuses to downgrade `'delete'`:

```ts
if (this.pendingConnectionMutations.get(connectionId) !== 'delete') {
```

`'rotate'` is not in that check. So a second, non-relay `updateConnection` on the same connection overwrites the rotation's marker with `'update'`, and then its own `finally` clears it outright (`get() === ownMutationKind` is true) while the rotation is still inside its critical section.

Probe against this branch, with the rotation parked inside `storeClientKeys`:

```
{"markerDuringRotate":"rotate","respA":"UNAUTHORIZED","handlerACalled":0,
 "markerAfterPlainUpdate":null,"respB":null,"handlerBCalled":1}
```

Before the plain update: marker `rotate`, request correctly rejected with `UNAUTHORIZED`, handler not called. After it: no marker at all, and the request runs the handler while the pubkey rebind and `deleteClientKeys(oldPubkey)` are still pending. That is the exact hole #4500 was written to close, and it lands on the branch with the longest window, since `awaitInFlightHandlers` can block on an in-flight `pay_invoice` for up to `PAYMENT_TIMEOUT_SECONDS` (120).

Reachable by the same path as the delete race: `AddOrEditNWCConnection.tsx:859` renders `leftComponent="Back"` unconditionally. The Save button is correctly disabled on `loading`, but Back is not, so leaving mid-save and re-entering gives two concurrent `updateConnection` calls on one id.

Minimal fix, same shape as what you already wrote:

```ts
const existing = this.pendingConnectionMutations.get(connectionId);
if (existing !== 'delete' && existing !== 'rotate') {
    this.pendingConnectionMutations.set(connectionId, ownMutationKind);
}
```

That fixes the wrong signal and the unguarding for mixed kinds. Same-kind overlap (update + update, rotate + rotate) still has the first `finally` clear the second's marker. If you want "Fixes #4503" to be literally true, a small refcount closes the rest: `Map<string, PendingConnectionMutation[]>`, each call pushes its kind on entry and splices one occurrence out in `finally`, and `mutationRejection` picks the most severe kind present. That also covers the mirror case where a `deleteConnection` that throws clears the marker out from under a live update. Otherwise retitle to `Part of #4503` and I will track the remainder on the issue.

### Test gaps (mutation-verified)

Two guards you added have no test. Both mutations stay green at 65/65:

- Removing the `!== 'delete'` no-downgrade condition in `updateConnection` (always set). The test plan box "Concurrent delete + update does not downgrade the delete marker" is checked, but the test that was added (`does not clear a concurrent delete marker when an update finishes`) exercises the `finally` guard, not the downgrade guard. Worth one more test that seeds `'delete'` before calling `updateConnection`.
- Removing the `=== 'delete'` condition in `deleteConnection`'s `finally`.

### Consider: mark `rotate` at the commit point rather than at entry

Your comment says a relay change invalidates the old pairing "once it commits", but the marker is set before the fallible prefix: `awaitInFlightHandlers` (up to 120s), `loadWalletServiceKeys`, `generateConnectionSecret`, `storeClientKeys`. Through all of that the old pairing is still live and the rotation can still fail, and a `storeClientKeys` throw leaves `relayUrl` and the subscription unchanged by design (per the comment above it). Those clients get `UNAUTHORIZED` today and can permanently drop a pairing that ends up surviving. Starting at `'update'` and promoting to `'rotate'` where the pubkey is actually rebound would put the revocation signal exactly on the revocation. It is not a regression against master, which sends `UNAUTHORIZED` for the whole window, so I am fine either way. It would need `ownMutationKind` to become a `let` updated in tandem with the `finally` comparison.

### Non-blocking

`RATE_LIMITED` is the right pick: it is the only retryable code NIP-47 defines, and it matches `paymentQueueWaitExceeded` and `makeInvoiceRateLimited` precedent in this store. Locale change is `en.json` only, correct.

Happy to re-review quickly once the ranking fix is in. The rest is in good shape.

ajaysehwal:
> Round 2. All three round-1 items are addressed and the `RATE_LIMITED` switch is the right call. One should-fix left: the no-downgrade rule got applied to `delete` but not to `rotate`, which leaves the rotate window worse off than a plain update window.
> 
> **Verified on the branch rebased onto current master (`2473e4fb2`), not eyeballed**
> 
> * Master still has not touched `stores/NostrWalletConnectStore.ts`, its test file, `utils/NostrConnectUtils.ts`, or `views/Settings/NostrWalletConnect/` since your base, so the rebase is textual only and these results apply to the PR as-is.
> * tsc / eslint / prettier clean, 65/65 store tests pass.
> * Round-1 coverage gap is closed: deleting the post-`markConnectionUsed` rejection block now fails 3 tests (it failed 0 before). Flipping `ownMutationKind` to always be `'update'`, and making `mutationRejection` treat `rotate` as retryable, each fail a test as well. The new tests are non-vacuous.
> 
> ### Should fix: a plain update downgrades, then unguards, an in-flight rotate
> `updateConnection` only refuses to downgrade `'delete'`:
> 
> ```ts
> if (this.pendingConnectionMutations.get(connectionId) !== 'delete') {
> ```
> 
> `'rotate'` is not in that check. So a second, non-relay `updateConnection` on the same connection overwrites the rotation's marker with `'update'`, and then its own `finally` clears it outright (`get() === ownMutationKind` is true) while the rotation is still inside its critical section.
> 
> Probe against this branch, with the rotation parked inside `storeClientKeys`:
> 
> ```
> {"markerDuringRotate":"rotate","respA":"UNAUTHORIZED","handlerACalled":0,
>  "markerAfterPlainUpdate":null,"respB":null,"handlerBCalled":1}
> ```
> 
> Before the plain update: marker `rotate`, request correctly rejected with `UNAUTHORIZED`, handler not called. After it: no marker at all, and the request runs the handler while the pubkey rebind and `deleteClientKeys(oldPubkey)` are still pending. That is the exact hole #4500 was written to close, and it lands on the branch with the longest window, since `awaitInFlightHandlers` can block on an in-flight `pay_invoice` for up to `PAYMENT_TIMEOUT_SECONDS` (120).
> 
> Reachable by the same path as the delete race: `AddOrEditNWCConnection.tsx:859` renders `leftComponent="Back"` unconditionally. The Save button is correctly disabled on `loading`, but Back is not, so leaving mid-save and re-entering gives two concurrent `updateConnection` calls on one id.
> 
> Minimal fix, same shape as what you already wrote:
> 
> ```ts
> const existing = this.pendingConnectionMutations.get(connectionId);
> if (existing !== 'delete' && existing !== 'rotate') {
>     this.pendingConnectionMutations.set(connectionId, ownMutationKind);
> }
> ```
> 
> That fixes the wrong signal and the unguarding for mixed kinds. Same-kind overlap (update + update, rotate + rotate) still has the first `finally` clear the second's marker. If you want "Fixes #4503" to be literally true, a small refcount closes the rest: `Map<string, PendingConnectionMutation[]>`, each call pushes its kind on entry and splices one occurrence out in `finally`, and `mutationRejection` picks the most severe kind present. That also covers the mirror case where a `deleteConnection` that throws clears the marker out from under a live update. Otherwise retitle to `Part of #4503` and I will track the remainder on the issue.
> 
> ### Test gaps (mutation-verified)
> Two guards you added have no test. Both mutations stay green at 65/65:
> 
> * Removing the `!== 'delete'` no-downgrade condition in `updateConnection` (always set). The test plan box "Concurrent delete + update does not downgrade the delete marker" is checked, but the test that was added (`does not clear a concurrent delete marker when an update finishes`) exercises the `finally` guard, not the downgrade guard. Worth one more test that seeds `'delete'` before calling `updateConnection`.
> * Removing the `=== 'delete'` condition in `deleteConnection`'s `finally`.
> 
> ### Consider: mark `rotate` at the commit point rather than at entry
> Your comment says a relay change invalidates the old pairing "once it commits", but the marker is set before the fallible prefix: `awaitInFlightHandlers` (up to 120s), `loadWalletServiceKeys`, `generateConnectionSecret`, `storeClientKeys`. Through all of that the old pairing is still live and the rotation can still fail, and a `storeClientKeys` throw leaves `relayUrl` and the subscription unchanged by design (per the comment above it). Those clients get `UNAUTHORIZED` today and can permanently drop a pairing that ends up surviving. Starting at `'update'` and promoting to `'rotate'` where the pubkey is actually rebound would put the revocation signal exactly on the revocation. It is not a regression against master, which sends `UNAUTHORIZED` for the whole window, so I am fine either way. It would need `ownMutationKind` to become a `let` updated in tandem with the `finally` comparison.
> 
> ### Non-blocking
> `RATE_LIMITED` is the right pick: it is the only retryable code NIP-47 defines, and it matches `paymentQueueWaitExceeded` and `makeInvoiceRateLimited` precedent in this store. Locale change is `en.json` only, correct.
> 
> Happy to re-review quickly once the ranking fix is in. The rest is in good shape.

fixed in e2619b0924ec4de404ad5b718f3760da29060451

kaloudis (commented):
Round 3. The ranking fix landed exactly as discussed and closes the hole. utACK e2619b09.

**Verified on the branch rebased onto current master (`775c321d5`), not eyeballed**

- Master drift since round 2 is `locales/en.json` only (from #4547); `stores/NostrWalletConnectStore.ts`, its test file, and `utils/NostrConnectUtils.ts` are untouched, so these results apply to the PR as-is. Rebase is clean, CI is green on the head as well.
- tsc / eslint / prettier clean, 68/68 store tests pass.
- All three new tests are non-vacuous (mutation-checked, each mutant killed by exactly the intended test):
  - Dropping `existingMutation !== 'rotate'` from the set guard fails `does not downgrade a pre-existing rotate marker when a second update starts`. That was the round-2 probe hole.
  - Dropping the set guard entirely fails both downgrade tests. This mutant survived at round 2.
  - Making `deleteConnection`'s `finally` clear unconditionally fails `does not clear a marker installed by a concurrent update when delete finishes`. The other round-2 survivor.
- Re-ran the round-2 probe (rotation parked inside `storeClientKeys`, plain rename update runs to completion, then a request comes in):

  Round 2:
  ```
  {"markerDuringRotate":"rotate","respA":"UNAUTHORIZED","handlerACalled":0,
   "markerAfterPlainUpdate":null,"respB":null,"handlerBCalled":1}
  ```
  Round 3:
  ```
  {"markerDuringRotate":"rotate","respA":"UNAUTHORIZED","handlerACalled":0,
   "markerAfterPlainUpdate":"rotate","respB":"UNAUTHORIZED","handlerBCalled":0}
  ```

  The rotate window now stays guarded through a concurrent plain update, and the rotation's own `finally` still clears the marker once it completes.

**Residual, not blocking:** same-kind overlap (update+update, rotate+rotate) and the delete-throws case still let one mutation's `finally` unguard another's critical section. As discussed in round 2, that needs the refcount shape; since this PR closes #4503 on merge, I filed #4560 so it doesn't get lost. Strictly narrower than everything fixed here.

Nice work on this one.

kaloudis (approved):

PR #4534, opened 2026-08-31
chore(deps): bump the minor-and-patch group with 26 updates

kaloudis:
# Description

Recreation of dependabot PR #4532 (`chore(deps): bump the minor-and-patch group across 1 directory with 33 updates`), with the unsafe entries held back, dependabot's `resolutions` mistakes corrected, and `ios/Podfile.lock` regenerated.

26 of the 33 updates are included. Closes #4532 when merged.

This pull request is categorized as a:

- [x] Configuration change

## What dependabot got wrong

**1. It reverts the version bump.** The branch was cut from b413a8134, one commit before `Version bump: v13.2.2-alpha`, so its diff sets `package.json` back to `13.2.1-rc1` and drags `android/app/build.gradle`, `ios/zeus.xcodeproj/project.pbxproj` and `bug_report.yml` with it. This PR is based on current master and stays at `13.2.2-alpha`.

**2. It splits `@types/react` across the resolutions block.** `@types/react` moves to 19.2.18 but `**/@types/react` is left at 19.1.1. That is exactly the duplicate-types split that broke group PR #4247. Here both read 19.2.18 and `find node_modules -path '*@types/react/*'` resolves a single copy.

**3. It leaves six `@typescript-eslint/*` sibling pins behind.** `eslint-plugin` and `parser` move to 8.68.0 while `scope-manager`, `type-utils`, `types`, `typescript-estree`, `utils` and `visitor-keys` stay pinned at 8.65.0. `@typescript-eslint/eslint-plugin@8.68.0` depends on `^8.68.0` of those, so the pins force a skew. All ten now read 8.68.0.

**4. It never touches `ios/Podfile.lock`.** Five of the bumped packages are pods. Since `postinstall` runs `pod-install`, leaving the lock stale means every checkout goes dirty on the next `yarn`. Regenerated here; only the five bumped pods changed, no hermes/prebuilt checksum drift.

## Held back

[...]

PR #4538, opened 2026-09-01
chore(deps): bump js-lnurl to 0.6.0

kaloudis:
# Description

Bumps `js-lnurl` from 0.5.1 to 0.6.0, the latest published version. This was the one dependency held back from #4534 because it cannot be bumped without a code change.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [x] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## What 0.6.0 changes

0.6.0 is a pure repackaging. A tarball diff against 0.5.1 shows `getParams`, all four helper implementations and `lib/types.d.ts` are byte-identical. The real changes are:

- `helpers.ts` was split into `src/helpers/{decipherAES,decodelnurl,findlnurl,getDomain}.ts`, and `export * from './helpers'` was dropped from `lib/index.js` / `lib/index.d.ts`. The package root now exports **only** `getParams` and the types.
- `randomHex` was deleted (ZEUS never used it).
- The `@types/node ^17` dependency was dropped.
- `sideEffects: false` was added.

## Why it needs a code change

The removed re-export is a silent failure, not a build error. `import { findlnurl } from 'js-lnurl'` still *resolves* after the bump, the binding just comes back `undefined` at runtime:

```
> require('js-lnurl')
{ getParams: [Function] }
> require('js-lnurl').findlnurl
undefined
```

That would have disabled LNURL detection in `utils/handleAnything.ts` (`findlnurl`, `decodelnurl`) and AES successAction decryption in `views/LnurlPay/Success.tsx` (`decipherAES`) with no compile-time signal.

Both now import from `js-lnurl/lib/helpers`, the path the 0.6.0 README documents. The other nine files that import `js-lnurl` only pull types (`LNURLWithdrawParams`, `LNURLPaySuccessAction`), which are unchanged, so they are untouched.

## Verification

- `utils/handleAnything.test.ts` mocked `js-lnurl` wholesale, so it would have passed green even with the broken root imports. Split into a `js-lnurl` mock for `getParams` plus a `js-lnurl/lib/helpers` mock for the helpers.
- Added `utils/LnurlHelpers.test.ts` (12 tests, real module, no mock): asserts the root does *not* export the helpers, that the subpath does, and exercises `decodelnurl` with the LUD-01 vector and `decipherAES` with an AES-256-CBC vector. A future repackaging now fails the suite instead of quietly disabling LNURL.
- `tsc` and Jest do not prove Metro resolves a directory-index subpath, so I built a real Android bundle (`npx react-native bundle --platform android --dev true --entry-file index.js`, exit 0) and confirmed all four `js-lnurl/lib/helpers/*.js` modules plus `index.js` are present in the output.
- The lockfile diff is only the `js-lnurl` entry plus removal of the now-orphaned `@types/node@^17.0.0`. Hoisted `node_modules/@types/node` stays at 22.14.1, unchanged from master, so there is no global type surface drift.
- `js-lnurl` is pure JS with no pod, so `ios/Podfile.lock` is untouched.

## Checklist

Full `yarn verify`: 68 suites, 1473 tests passing.

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

**Draft until device tests are done.** The four paths that changed and need exercising on hardware:

1. LNURL-pay, including a payment whose `successAction` is `aes` (that is the `decipherAES` path)
2. LNURL-withdraw
3. An `lnurl1...` bech32 string pasted from the clipboard (the `findlnurl` path in `handleAnything`)
4. An lnurl scanned as a QR code

LNURL handling is backend-agnostic here, so one Android and one iOS pass against a single node should cover it.

### Locales

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "hash.js": "1.1.7",
     "humanize-duration": "3.34.0",
     "identicon.js": "2.3.3",
-    "js-lnurl": "0.5.1",
+    "js-lnurl": "0.6.0",
     "js-sha256": "0.12.0",
     "lodash": "4.18.1",
     "long": "5.3.2",
--- a/utils/LnurlHelpers.test.ts
+++ b/utils/LnurlHelpers.test.ts
@@ -1,0 +1,106 @@
+import { getParams } from 'js-lnurl';
+import {
+    decipherAES,
+    decodelnurl,
+    findlnurl,
+    getDomain
+} from 'js-lnurl/lib/helpers';
+
+// js-lnurl 0.6.0 stopped re-exporting the helpers from the package root, so
+// `import { findlnurl } from 'js-lnurl'` resolves fine but hands back
+// undefined at runtime. handleAnything and LnurlPay/Success both depend on
+// these helpers, so pin the import surface we actually rely on.
+describe('js-lnurl import surface', () => {
+    it('exposes getParams from the package root', () => {
+        expect(typeof getParams).toBe('function');
+    });
+
+    it('exposes the helpers from js-lnurl/lib/helpers', () => {
+        expect(typeof findlnurl).toBe('function');
+        expect(typeof decodelnurl).toBe('function');
+        expect(typeof decipherAES).toBe('function');
+        expect(typeof getDomain).toBe('function');
+    });
+
+    it('does not expose the helpers from the package root', () => {
+        // guards against a future version silently re-adding them, which would
+        // make the subpath imports above look optional again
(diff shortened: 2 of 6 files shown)


shubhamkmr04 (approved):
tACK

PR #4542, opened 2026-09-02
feat(swaps): add Coinos and SATS Routing providers, retire Eldamar Swaps

kaloudis:
# Description

Adds [Coinos](https://swap.coinos.io/swap) and [SATS Routing](https://satsrouting.exchange/) to the mainnet swap provider list, and removes Eldamar Swaps, which has shut down.

### Endpoints

The URLs above are the web frontends. Both deployments serve the Boltz v2 API from the same origin under `/v2`, so the hosts are `https://swap.coinos.io/v2` and `https://satsrouting.exchange/v2`. Neither has an `api.` subdomain (those fail to resolve), and Coinos' frontend bundle confirms it: `apiUrl: {normal: 'https://swap.coinos.io', tor: 'https://swap.coinos.io'}`. SATS Routing's is empty, i.e. same origin.

| | API base | Backend | Node | Limits |
|---|---|---|---|---|
| Coinos | `https://swap.coinos.io/v2` | Boltz 3.13.0 | CLN `030dfc53...` | 25k - 600k sats |
| SATS Routing | `https://satsrouting.exchange/v2` | Boltz 3.13.0 | LND `02b06bf3...` | 50k - 5M sats |

Verified live on both hosts, covering every route ZEUS calls:

- `GET /v2/swap/submarine` and `GET /v2/swap/reverse` return a `BTC.BTC` pair, which is what `SwapStore.getSwapFees` reads
- `POST /v2/swap/submarine` and `POST /v2/swap/reverse`
- `GET /v2/swap/{id}`, `GET /v2/swap/submarine/{id}/transaction`, `POST /v2/swap/submarine/{id}/claim`
- `POST /v2/swap/restore` answers with `[]` for an unknown xpub, so both are flagged `supportsRescue: true` and appear in the rescue-key restore picker in Seed Recovery
- `wss://<host>/v2/ws` returns `101 Switching Protocols`, the socket `SwapDetails` opens to track swap status
- `GET /v2/chain/BTC/height` matches mainnet, and both frontend bundles are configured `network: 'mainnet'`. Neither offers testnet, so `SWAP_HOST_KEYS_TESTNET` is unchanged.

Sending `Referral: pro` returns byte-identical pair hashes on both, so neither is marked `pro` and the Pro switch stays hidden for them, same as SwapMarket.

### Eldamar removal

`boltz-api.eldamar.icu` returns 404 on every path including `/v2/version`, and `eldamar.icu` itself returns 502.

Dropping the entry alone would strand anyone currently pinned to it: their persisted `swaps.hostMainnet` would match no entry in `SWAP_HOST_KEYS_MAINNET`, so the provider dropdown would render no selection while every swap request kept going to a dead endpoint. `MigrationUtils.migrateRetiredSwapHosts` moves those users back to `DEFAULT_SWAP_HOST_MAINNET`. It is guarded by MOD_KEY `swap-hosts-retired-eldamar`, runs on both the legacy and modern (`zeus-settings-v2`) settings paths, and leaves custom hosts alone, so an Eldamar URL typed in by hand is preserved rather than silently rewritten.

Swaps already created against Eldamar are unaffected by the roster change: `SwapStore` polls `swap.endpoint || getHost`, so they keep their original host and simply fail to refresh, as they would today.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [x] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Five new cases for `migrateRetiredSwapHosts`: rewrites a pinned retired host, leaves a hand-typed custom host alone, leaves still-operating providers alone, only sets the flag when there is no `swaps` block, and is a no-op once the MOD_KEY is set.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Opening as a draft: the endpoint probes prove the routes exist and answer, not that a swap settles end to end. Needs a real submarine and reverse swap against each new provider on device before this ships.

### Locales

Provider names render from the `key` field verbatim, so no locale strings are added.

### Other:

### DIFF
--- a/.claude/skills/zeus-config-and-flags/SKILL.md
+++ b/.claude/skills/zeus-config-and-flags/SKILL.md
@@ -216,8 +216,8 @@ Per-network TRIPLETS (mainnet/testnet/mutinynet). The canonical resolver is `get
 
 | Axis | Options | Default |
 |---|---|---|
-| `hostMainnet` | `SWAP_HOST_KEYS_MAINNET`: ZEUS (`https://swaps.zeuslsp.com/api/v2`), Boltz, SwapMarket, Eldamar Swaps, Custom | ZEUS host (`DEFAULT_SWAP_HOST_MAINNET`) |
-| `hostTestnet` | `SWAP_HOST_KEYS_TESTNET`: ZEUS, Boltz, Custom | `https://testnet-swaps.zeuslsp.com/api/v2` |
+| `hostMainnet` | `SWAP_HOST_KEYS_MAINNET`: Boltz, SwapMarket, Coinos, SATS Routing, Custom | `https://api.boltz.exchange/v2` (`DEFAULT_SWAP_HOST_MAINNET`) |
+| `hostTestnet` | `SWAP_HOST_KEYS_TESTNET`: Boltz, Custom | `https://api.testnet.boltz.exchange/v2` |
 | `customHost` | string | `''` |
 | `proEnabled` | bool — unlocks `pro: true` hosts (Boltz mainnet, Custom) | `false` |
 
--- a/stores/SettingsStore.ts
+++ b/stores/SettingsStore.ts
@@ -347,8 +347,14 @@ export const SWAP_HOST_KEYS_MAINNET = [
         supportsRescue: true
     },
     {
-        key: 'Eldamar Swaps',
-        value: 'https://boltz-api.eldamar.icu/v2',
+        key: 'Coinos',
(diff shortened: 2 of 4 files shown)


shubhamkmr04 (approved):
tACK

PR #4546, opened 2026-09-02
[rn-0.87.1] chore(deps): bump react-native to 0.87.1

kaloudis:
Bumps React Native from 0.86.0 to 0.87.1.

> [!NOTE]
> Rebased on master now that #4533 has merged, so the cherry-picked duplicate has dropped and all commits here are the bump itself. #4533 was a hard prerequisite rather than a nicety: RN 0.87 removes `InteractionManager` from core.

## Template changes

Everything in the [upgrade-helper diff](https://react-native-community.github.io/upgrade-helper/?from=0.86.0&to=0.87.1) is applied: Gradle wrapper 9.3.1 -> 9.4.1, `buildToolsVersion`/`compileSdkVersion` 36 -> 37, Kotlin 2.1.20 -> 2.2.0, `proguard-android-optimize.txt`, and the AGP 9 opt-outs.

Two deliberate deviations:

- **`edgeToEdgeEnabled` is flipped to `true`**, matching the template. This bump originally kept it `false` to avoid an app-wide UI migration, but review surfaced a status-bar-height gap above every header on Android 14 and below: RN 0.87 removed the `StatusBar` `translucent`/`backgroundColor` props that `@rneui`'s Header was using to draw under the status bar, so the old appearance cannot be reproduced on 0.87. Android 15+ already enforces edge-to-edge with our targetSdk 36, so the flip brings pre-15 devices in line with what Android 15/16 users already see. Verified before/after on an API 34 emulator.
- **TypeScript stays on 5.7.3** although the template moves to `^6.0.3`. Zeus does not extend `@react-native/typescript-config`, and `tsc` is clean on 5.7.3. Tracked in #4559.

## Breakages fixed here

None of these are in the template diff.

1. **MobX decorators (`babel.config.js`).** 0.87 changed the preset's default transform profile from `default` to `hermes-stable`. Hermes profiles set `preserveClasses`, which stops the preset shipping `@babel/plugin-transform-class-properties`. Legacy decorators need it, so every `@observable` property threw `Decorating class property failed` (116 test failures). Pinning the profile is deliberate: adding the plugin to `babel.config.js` instead makes it run *before* the preset strips TS types, so type-only declarations like `payment_hash: string;` become `this.payment_hash = void 0` after `super()` and silently wipe everything `BaseModel`'s constructor assigned.
2. **Type resolution (`tsconfig.json`).** 0.87 dropped react-native's top-level `types` field; types now resolve only via `exports`, which `moduleResolution: "node"` cannot read. This surfaced as TS7016 everywhere plus cascading nonsense in `@rneui` (`Property 'onPress' does not exist on ListItemProps`).
3. **Asset registry (`metro.config.js`).** `@react-native-vector-icons/common` and `react-native-svg` (`lib/*/lib/resolveAssetUri.js`) import `@react-native/assets-registry/registry`, which 0.87 removed from RN's dep tree. Aliased to `react-native/asset-registry` rather than installing the standalone package, which still exists but would register into a second Map so `getAssetByID` returns `undefined`. The alias is metro-only, so a jest `moduleNameMapper` entry points the same specifier at the same file (`react-native/src/asset-registry.js`), keeping the two resolvers in sync.
4. **AGP 9 proguard (`patches/`).** AGP 9 hard-errors at configuration time on `getDefaultProguardFile('proguard-android.txt')`. `react-native-notifications` 5.2.2 (latest) still calls it. Not fixable from the root `subprojects` block, since the error fires before `afterEvaluate`.
5. **AGP 9 `BuildConfig` (`android/build.gradle`).** No longer generated unless a module opts in, and the global `android.defaults.buildfeatures.buildconfig` property is gone. Re-enabled for all Android subprojects instead of per package.
6. **`ReactContext.getRuntimeExecutor()` (`WorkerReactContext.java`).** Newly abstract in 0.87.
7. **`StatusBar backgroundColor` (`views/Stealth/`).** Prop removed in 0.87.

`engines` and the README Node prerequisite now mirror RN's range, `^22.13.0 || ^24.3.0 || >= 26.0.0`. The previous `>= 22.13.0` admitted 23.x and 25.x, which React Native rejects.

## Testing

- `yarn verify` green: 70 suites, 1526 tests, plus tsc, prettier and lint
- `npx react-native bundle` succeeds for both android and ios with `--dev false`
- `./gradlew :app:assembleDebug` **BUILD SUCCESSFUL**, APKs produced for all ABIs
- `ios/Podfile.lock` regenerated: 129 dependencies, 130 pods, hermes-engine 250829098.0.17

## Still outstanding

- [ ] Device testing on both platforms, all backends
- [ ] **A release build specifically.** Zeus sets `enableProguardInReleaseBuilds = true`, so the switch to `proguard-android-optimize.txt` enables R8 optimization passes on a reflection-heavy app (protobuf, gRPC, LND JNI). Only a debug build has been exercised so far. R8 keep rules for all three JNA-backed uniffi packages (CDK, LDK Node, zeus_cashu_restore) are now in `proguard-rules.pro`.
- [ ] Reproducible-build check via `./build.sh`

## Follow-ups, not in this PR

All four are filed, so nothing here is lost when this merges:

- #4558: migrate the type-only class fields in `models/` and `stores/` to `declare` so the `hermes-stable` profile and its Hermes V1 optimisations can be restored.
- #4559: decide on TypeScript 6, per the deviation noted above.
- #4556: remaining edge-to-edge audit (POS, landscape, `views/Stealth/*` decoys, pre-15 devices); the flag itself is now flipped in this PR, see the deviation note above.
- #4557: `views/Stealth/NotepadApp.tsx` is `dark-content` on a light decoy. With `backgroundColor` gone, on pre-Android-15 the bar reverts to theme black and the icons may be hard to read. Left as-is rather than silently restyling a decoy screen.


shubhamkmr04 (commented):
<p align="center">
  
  
</p>

we are seeing a visible gap at the top

kaloudis:
Thanks for catching this @shubhamkmr04. Tracked it down; it's a real regression from the 0.87 bump, fixed in 7ff02becc.

Root cause: RN 0.87 removed the `StatusBar` `translucent` and `backgroundColor` props along with their native handlers. `@rneui/base`'s Header renders `<StatusBar translucent backgroundColor="transparent" />` on every screen, and that prop is what had been putting the window into draw-under-the-status-bar mode on Android 14 and below. On 0.87 those props are silently ignored, so the system now lays the app out below the status bar while our headers still add their safe-area top margin (`components/Header.tsx`, `components/WalletHeader.tsx`). The status bar height gets applied twice, which is the gap in your screenshots.

Android 15+ is unaffected because the OS enforces edge-to-edge for apps targeting API 35+ (RN aligns its internal flag with that in `updateEdgeToEdgeFeatureFlag`). I'm guessing your test device is on Android 14 or lower; can you confirm?

The fix flips `edgeToEdgeEnabled=true` in `android/gradle.properties`. That restores draw-under behavior on pre-15 devices via the supported path and matches what Android 15/16 users already see today. Keeping the old look wasn't an option: RN removed the translucent API that produced it, and Android 16 ignores the opt-out attribute anyway.

Verified on an API 34 (Android 14) emulator with before/after builds on the same screen: with the flag off, headers render a full status-bar height too low; with the flag on they sit directly below the status bar, matching what Android 15 devices already render. The broader edge-to-edge audit (POS, landscape, Stealth decoy screens) stays tracked in #4556.

shubhamkmr04 (commented):
Took another pass at `7ff02becc`. The edge-to-edge flip looks right to me, and it also resolves the concern I had about NotepadApp's status bar — with a transparent bar, `dark-content` on the yellow header is now correct rather than dark-on-black. Two things I'd still look at.

### 1. R8 keep rules for the uniffi bindings

`proguard-android-optimize.txt` drops the `-dontoptimize` that `proguard-android.txt` carried, so a release build runs R8 optimization for the first time. `proguard-rules.pro` keeps `org.cashudevkit.**` explicitly, but there's nothing for the other two JNA-backed uniffi packages:

```proguard
-keep class org.lightningdevkit.** { *; }
-keep class uniffi.** { *; }
-keepclassmembers class org.lightningdevkit.**, uniffi.** { native <methods>; }
```

Worth adding before the release-build test on your checklist — otherwise a break here surfaces at the FFI boundary and reads like an LDK or Cashu bug. If you'd rather keep 0.86 behaviour exactly for now, `-dontoptimize` in `proguard-rules.pro` also satisfies AGP 9.

### 2. Two small ones

- `react-native-svg@15.15.5` imports `@react-native/assets-registry/registry` as well (`lib/commonjs/lib/resolveAssetUri.js`), not just vector-icons. The alias covers both — might be worth naming in the description, since it strengthens the case for aliasing over installing the standalone package.
- That alias is Metro-only, and the lockfile drops the package outright, so Jest will fail the first time a suite touches an SVG or an icon. A `moduleNameMapper` entry pointing at `react-native/src/asset-registry` would keep the two resolvers in sync.

One note on the reproducible build: a quick `docker run --rm <builder image> node -v` is worth doing first. yarn 1 hard-errors on a root `engines` mismatch, and the new range excludes 23.x and 25.x, so if the pinned image ships either, the build dies at `yarn install` before it reaches Gradle.

shubhamkmr04:
iOS build never succeeds, it just goes on forever

kaloudis:
Thanks @shubhamkmr04, all taken. Pushed as 8b7491c98 + 739ca4048, and the description now names `react-native-svg` in the asset-registry bullet.

**R8 keeps:** added for `org.lightningdevkit.**` and `uniffi.**` (zeus_cashu_restore) in the same shape as the CDK block, including the `native <methods>` keepclassmembers. One clarification for the record: the JNA block in `proguard-rules.pro` was labeled "Required for CDK FFI" but is generic in effect, and `-keep class * implements com.sun.jna.** { *; }` already covered the reflective surface of all three bindings (`RustBuffer : Structure`, the `UniffiLib : Library` interfaces whose method names drive native symbol lookup, and the future-continuation `Callback`s). So the new rules are defense in depth against optimize-mode edge cases beyond the JNA surface rather than a fix for a proven hole. I kept optimization on rather than taking `-dontoptimize`; the release-build device test stays on the checklist as the real gate and will exercise LDK + Cashu paths specifically.

**Jest mapper:** added exactly as suggested, and your target turns out to be the only one that works. I tested three configs with a throwaway suite that requires `@react-native/assets-registry/registry`: no mapper fails with `Cannot find module` (confirming the latent break), mapping to `react-native/asset-registry` (the exports subpath metro resolves) fails under jest 29.7 because jest-resolve cannot use that subpath, and `react-native/src/asset-registry` passes. Both resolvers land on the same file, so they stay in sync. `yarn verify` still green: 70 suites / 1526 tests plus tsc, prettier, lint.

**Builder image node:** Docker daemon was down here so I checked the source instead: the `v18.0` tag's Dockerfile pins `ARG NODE_VERSION=22.14`, which satisfies `^22.13.0`. I'll still run your `node -v` one-liner against the pinned digest before the `./build.sh` check, since the digest is what build.sh actually uses. While I was in there: the image ships JDK 17, which meets AGP 9.x's minimum, so no hidden Gradle-side blocker either.

kaloudis:
> iOS build never succeeds, it just goes on forever

Moving this to a feature branch so that we can test against Xcode Cloud and get an exact err

PR #4547, opened 2026-09-02
fix(cashu): hide custom amount field for zero-amount invoices

ajaysehwal:
# Description

Cashu melts require a fixed-amount Lightning invoice — the mint has no way to settle an amount-less one. Two problems followed from that:

1. The Cashu payment-request screen still showed a "Custom Amount" field and an active pay flow for these invoices, leading to a dead end (or a failed payment attempt).
2. Even after fixing that screen, users could still land on it: the payment-method picker offered "pay with ecash" for an amount-less invoice in the first place.

This PR fixes both, addressing the root cause instead of only patching the screen.

### Changes

**Routing (`utils/handleAnything.ts`)**
- Decode a plain bolt11 invoice up front and skip `ChoosePaymentMethod` entirely when it carries no amount — go straight to the normal Lightning `PaymentRequest` screen, exactly as when Cashu is disabled. Ecash is never offered as a choice for an invoice it can't pay.

**Store (`stores/CashuStore.ts`)**
- `getPayReq` still guards the remaining path (a BIP21 URI with multiple destinations, where the picker is shown for reasons unrelated to Cashu): it detects a zero/no-amount invoice early and sets one friendly error (`stores.CashuStore.noAmountInvoiceNotSupported`) instead of round-tripping to the mint for a melt quote that would fail the same way.
- The fee estimate is left `undefined` (not `0`) in that case, so the screen no longer shows a stale "0 sats" fee for an invoice that can't be paid.

**View (`views/Cashu/CashuPaymentRequest.tsx`)**
- Removed the "Custom Amount" input; the screen now shows a single warning (the store error above) instead of a duplicate message.
- Hides the "Sending from" mint picker and pay button when the invoice has no amount.
- Hides the donation section too (it previously rendered `0 + 0 = 0 sats` for an unpayable invoice).
- Removed now-dead `customAmount`/`satAmount` state left over from the removed input.

### Test plan

- [x] Scan/paste a **zero-amount** Lightning invoice with Cashu enabled → skips the picker, goes straight to the normal Lightning `PaymentRequest` screen
- [x] From a BIP21 URI with multiple destinations (address + zero-amount invoice) → picker still shows, but selecting ecash for that invoice surfaces one clear warning, no custom-amount field, no dead pay controls, no stale fee estimate
- [x] Scan a **normal (fixed-amount)** invoice → payment flow completely unaffected, including via the picker
- [x] `yarn verify` (typecheck / lint / prettier) passes

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -2356,7 +2356,7 @@
     "stores.CashuStore.errorMintingToken": "Error minting token",
     "stores.CashuStore.insufficientBalance": "Insufficient balance",
     "stores.CashuStore.notEnoughFunds": "Not enough funds to cover payment",
-    "stores.CashuStore.noAmountInvoiceNotSupported": "This invoice has no amount. Cashu payments require a fixed amount invoice.",
+    "stores.CashuStore.noAmountInvoiceNotSupported": "This invoice has no amount, so it can't be paid with ecash. Pay it from a Lightning balance instead, or ask for an invoice with a fixed amount.",
     "stores.CashuStore.notEnoughFundsSelfSend": "Not enough funds to cover payment. {{mintName}} couldn't be used because it's this invoice's destination — pick a different mint to cover the gap.",
     "stores.CashuStore.meltFeeMismatch": "Multi-mint payment failed: {{mintName}} demanded a higher fee at melt time than it quoted ({{provided}} provided, {{needed}} needed). Top up {{mintName}}, pick a different mint, or lower the amount.",
     "stores.CashuStore.noMeltQuoteAvailable": "No melt quote available",
--- a/locales/en.json
+++ b/locales/en.json
@@ -2356,6 +2356,7 @@
     "stores.CashuStore.errorMintingToken": "Error minting token",
     "stores.CashuStore.insufficientBalance": "Insufficient balance",
     "stores.CashuStore.notEnoughFunds": "Not enough funds to cover payment",
+    "stores.CashuStore.noAmountInvoiceNotSupported": "This invoice has no amount. Cashu payments require a fixed amount invoice.",
     "stores.CashuStore.notEnoughFundsSelfSend": "Not enough funds to cover payment. {{mintName}} couldn't be used because it's this invoice's destination — pick a different mint to cover the gap.",
     "stores.CashuStore.meltFeeMismatch": "Multi-mint payment failed: {{mintName}} demanded a higher fee at melt time than it quoted ({{provided}} provided, {{needed}} needed). Top up {{mintName}}, pick a different mint, or lower the amount.",
     "stores.CashuStore.noMeltQuoteAvailable": "No melt quote available",
--- a/stores/CashuStore.ts
+++ b/stores/CashuStore.ts
@@ -4205,7 +4205,17 @@ export default class CashuStore {
 
             let totalFeeEstimate = 0;
 
-            if (isMultiMint) {
+            if (rawPaymentAmt <= 0) {
+                // Cashu melts (paying a Lightning invoice with ecash) always
+                // require a fixed amount — the mint has no notion of an
+                // amount the payer chooses. Surface this up front instead of
+                // round-tripping to the mint for a melt quote request that
+                // would fail the same way.
+                computedPayReqError = localeString(
+                    'stores.CashuStore.noAmountInvoiceNotSupported'
+                );
+                this.meltQuotes = [];
+            } else if (isMultiMint) {
                 if (paymentAmt > 0) {
                     if (__DEV__) {
                         console.log(
--- a/stores/CashuStore.ts
+++ b/stores/CashuStore.ts
@@ -4203,14 +4203,17 @@ export default class CashuStore {
             let singleMeltQuote: typeof this.meltQuote;
             let computedPayReqError: string | undefined;
 
-            let totalFeeEstimate = 0;
+            // Left undefined (rather than 0) in the no-amount branch below so
+            // the fee row doesn't render "0 sats" for an invoice that can't
+            // be paid; every branch here assigns it before use.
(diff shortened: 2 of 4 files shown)


kaloudis (commented):
Approach ACK on blocking the flow. Reviewed against the code, ran the checks, and traced the affected call paths.

`prettier` clean on both changed files, `eslint` clean, `jest` 1460/1461 (the single failure is an unrelated path artifact from my worktree). `tsc` fails on `NavigationService.ts:39`, but that is pre-existing and unrelated to this PR.

## The diagnosis is correct

Confirmed the dead end. `payLnInvoiceFromEcash` never uses its `amount` argument for the melt; it replays `this.meltQuote` (`stores/CashuStore.ts:4403-4437`), and the multimint leg replays `this.meltQuotes` (`:4512`, `:4533`). So the `AmountInput` was decorative.

Worse, in multimint mode the old code took the `else` branch at `:4247` and set no `computedPayReqError`, so the pay button was fully enabled and led to a "not enough funds" failure at the end of the flow. That is a real bug this fixes.

Also verified no collateral damage:

- NWC `handleCashuPayInvoice` (`stores/NostrWalletConnectStore.ts:2628`) already rejected amountless invoices at the later `amountSats <= 0` check, so this just moves the failure earlier with a better message.
- The LnurlPay cashu branch (`views/LnurlPay/LnurlPay.tsx:348`) is unaffected, since LNURL invoices always carry an amount.
- The expired-invoice early return at `:4162` still takes precedence over the new branch.

## Should fix

### 1. "Fee estimate: 0 sats" now renders on the blocked screen (single-mint regression)

`getPayReq` clears `feeEstimate` at `:4120`, the new branch leaves `totalFeeEstimate = 0`, and `:4304` assigns that 0. The view renders the fee row on an explicit zero:

```js
{(!!feeEstimate || feeEstimate === 0) && (
```

(`views/Cashu/CashuPaymentRequest.tsx:833`)

Before this PR the single-mint path threw out of `createMeltQuoteCDK`, so `feeEstimate` stayed `undefined` and the row never rendered. Net effect after this change: the screen says the invoice cannot be paid, and then shows a 0-sat fee estimate for it.

`feeEstimate` is `number | undefined` (`:224`), so this is a one-liner: widen `totalFeeEstimate` and set it to `undefined` in the new branch.

### 2. The stated rationale is not accurate, and the comment will mislead the next reader

The comment says the mint "has no notion of an amount the payer chooses". NUT-05 does support amountless melts, and Zeus already has the plumbing for them:

- `parseMeltOptions` maps `{"amountless": {"amount_msat": N}}` to `MeltOptions.Amountless` on both platforms (`android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt:107-112`, `ios/CashuDevKit/CashuDevKitModule.swift:118-125`)
- `createMeltQuoteCDK` already accepts an `options` argument (`stores/CashuStore.ts:946-961`)

It was added in 83ac9c0db and never wired up from the JS side. The accurate statement is "Zeus does not pass the amountless melt option yet", not "Cashu cannot do this". Worth rewording the code comment and softening the locale string so neither one forecloses the real fix, plus a follow-up issue to wire it.

One caveat if that follow-up happens: gating on mint capability is not possible today. `CDKMintInfo.nuts` is declared (`cashu-cdk/types.ts:56`) but `encodeMintInfo` never populates it (Kotlin `:335-344`), which means `mintSupportsMpp` (`stores/CashuStore.ts:993`) and the view's `supportsMultiMint` NUT-15 check (`views/Cashu/CashuPaymentRequest.tsx:324-340`) already always evaluate to false. That is a separate pre-existing bug and deserves its own issue.

## Nits

- **Dead state left behind.** `customAmount` (`:74`, `:107`) now has no reader and no writer. `satAmount` has no writer, so `triggerPayment`'s `satAmount ? satAmount.toString() : ...` (`:283`) and `sendPayment`'s `amount // used only for no-amount invoices` (`:230`) are now dead. Worth removing, otherwise the next reader will assume the custom-amount path still exists.
- **"Only one warning" holds only when there is a balance.** `noBalance` (`:601`) and the new store error (`:616`) both render when the ecash balance is 0.
- **Android was not tested**, and `enableDonations` is `Platform.OS !== 'ios'`, so this could not have shown up in the iOS pass: the donation section (`:846`) still renders on an unpayable invoice, and when expanded it shows `0 + 0 = 0 sats` (`:1022-1031`). Consider folding `isNoAmountInvoice` into that gate too.
- **`ChoosePaymentMethod` still offers ecash** for amountless invoices. It calls `CashuStore.getPayReq` purely to prefetch fee estimates and ignores `getPayReqError` (`views/ChoosePaymentMethod.tsx:174`), so users will keep landing on this dead-end screen from the picker. Follow-up candidate.
- **Wording:** "a fixed amount invoice" reads better as "a fixed-amount invoice", and the message could tell the user what to do next (pay from the Lightning balance, or ask for an invoice with an amount).
- **PR body:** "Yes" is checked for new unit tests but none were added. There is no `stores/CashuStore.test.ts`, so N/A is the honest answer here. Also, "Locales update" is checked while "I've added new locale text that requires translations" is not.

shubhamkmr04 (approved):
tACK

kaloudis (approved):

Issue #4222, opened 2026-07-13
Remove reanimated shared element transition patches once fixed upstream


kaloudis:
\`patches/patch-reanimated-set-fixes.mjs\` (added on the rn-v0.86.0 branch) applies two source fixes to reanimated's experimental SET proxy (\`LayoutAnimationsProxy_Experimental.cpp\`):

1. **Pop fallback** — return transitions raced the revealed screen's \`isActive\` update and were usually skipped on Android. Upstream: software-mansion/react-native-reanimated#9944
2. **Resync fix** — a cancelled screen close (e.g. aborted back-swipe) permanently disabled all shared element transitions, primarily on iOS. Upstream: software-mansion/react-native-reanimated#9945

Remove each edit when the corresponding upstream fix ships in a release. The patch warns loudly at postinstall if its anchors no longer match (e.g. after a reanimated bump) — treat that as the signal to re-check both upstream issues. Verify by exercising Contacts → ContactDetails push/pop repeatedly on both platforms, including after a cancelled back-swipe.

kaloudis:
Status check (2026-09-02): both upstream issues still open, and 4.7.0 will break this patch silently.

**Upstream state.** [#9944](https://github.com/software-mansion/react-native-reanimated/issues/9944) (pop race) and [#9945](https://github.com/software-mansion/react-native-reanimated/issues/9945) (cancelled-close wedge) are both still OPEN, still labeled "Missing repro", with no maintainer comment beyond the repro bot. Confirmed in source rather than by issue state: `4.7.0-nightly-20260901` still does `auto afterTopScreen = findActiveBoundary(root);` with no fallback, and the resync branch still gates on `if (!lightNodes_.contains(closingScreenTag_))`. Both edits are still required.

**The new problem.** 4.7.0 refactors `LayoutAnimationsProxy_Experimental` to one proxy instance per surface:

| 4.6.x and earlier | 4.7.x |
| --- | --- |
| `topScreen[surfaceId]` | `topScreen_` |
| `surfaceId` parameter | `surfaceId_` member |
| `filteredMutations` arg to `updateLightTree()` | `TransactionMeta transaction` |

[...]

Issue #4221, opened 2026-07-13
Remove reanimated packaging patch once react-native.config.js ships in the npm tarball


kaloudis:
\`patches/patch-reanimated-boundary-android.mjs\` (added on the rn-v0.86.0 branch) restores \`react-native.config.js\` inside \`node_modules/react-native-reanimated\`. The published 4.5.x tarballs omit the file, which leaves the Android \`SharedTransitionBoundary\` descriptor unregistered and silently disables all shared element transitions on Android.

Upstream: software-mansion/react-native-reanimated#9943

Remove the patch when a reanimated release includes the file (check with \`tar -tzf\` on the new tarball before deleting). Note: existing checkouts must clear \`android/build/generated/autolinking/\` for the change to take effect either way.

kaloudis:
Status check (2026-09-02): fixed upstream, not yet shipped in any release. Keeping this open.

Upstream [software-mansion/react-native-reanimated#9943](https://github.com/software-mansion/react-native-reanimated/issues/9943) was closed as completed on 2026-08-25 by [#10375](https://github.com/software-mansion/react-native-reanimated/pull/10375), which adds `react-native.config.js` to the `files` whitelist in `package.json`.

No published release carries it yet. Verified by unpacking the tarballs rather than trusting the changelog:

| version | published | `package/react-native.config.js` |
| --- | --- | --- |
| 4.6.0 (`latest`, pinned by Zeus) | 2026-08-21 | missing |
| 4.5.4 | 2026-08-25 | missing |
| 4.5.5 | 2026-08-27 | missing |
| 4.7.0-nightly-20260824 | 2026-08-24 | missing |
| 4.7.0-nightly-20260825 and later | 2026-08-25 onward | present |

4.5.4 and 4.5.5 were cut from the 4.5 maintenance line after the fix merged and did not pick it up, so a 4.6.x backport is not guaranteed either. First tarball containing the file is `4.7.0-nightly-20260825-404c5649d`, so in practice this reaches us with 4.7.0 stable.

Zeus pins `react-native-reanimated: 4.6.0` on master and on both open bump branches (#4546 for RN 0.87.1, #4534 for the dependabot batch), so `patches/patch-reanimated-boundary-android.mjs` is still doing real work on every install. It already no-ops when the file is present, so it is safe to carry through the 4.7.0 bump; remove it plus the registration in `patches/index.mjs` once the pinned version ships the file, then `rm -rf android/build/generated/autolinking`.

[...]

PR #4549, opened 2026-09-02
fix(patches): re-anchor reanimated SET fixes for 4.7 and fail loudly on a miss

kaloudis:
# Description

Relates to issue: **[#4222](https://github.com/ZeusLN/zeus/issues/4222)** (also relevant to [#4221](https://github.com/ZeusLN/zeus/issues/4221))

reanimated 4.7.0 refactors `LayoutAnimationsProxy_Experimental` to one proxy instance per surface, which breaks both anchors in `patches/patch-reanimated-set-fixes.mjs`:

| 4.6.x and earlier | 4.7.x |
| --- | --- |
| `topScreen[surfaceId]` | `topScreen_` |
| `surfaceId` parameter | `surfaceId_` member |
| `filteredMutations` arg to `updateLightTree()` | `TransactionMeta transaction` |

Today a missed anchor only warns, and `patches/index.mjs` then prints "All patches applied successfully" anyway. So a bump to 4.7.0 would produce a build with neither SET fix applied, silently reintroducing the two bugs tracked in #4222: return transitions racing the boundary's `isActive` update on pop (Android), and a cancelled screen close permanently wedging resynchronization (iOS). Both failure modes are invisible, with no error, no log, and nothing a test would catch.

This is not hypothetical timing. 4.7.0 is also the release that finally ships `react-native.config.js` (upstream [#10375](https://github.com/software-mansion/react-native-reanimated/pull/10375), merged 2026-08-25, still unreleased), so the bump that lets us delete the packaging patch from #4221 is the same bump that would disable this one.

Changes:

- each fix carries an anchor per source layout (`4.6.x` and `4.7.x`), selected automatically
- the patch throws when no anchor matches, so the install fails loudly instead of producing a quietly broken build. `ZEUS_ALLOW_UNPATCHED_REANIMATED=1` overrides for anyone who needs to install anyway
- `patches/index.mjs` catches patch failures, prints the message, and exits non-zero instead of reporting success

The member names the edits depend on (`lightNodes_`, `closingScreenTag_`, `findActiveBoundary`, `findBoundaryGuess`, `LightNode::parent`/`children`) are unchanged in 4.7, so the two edits port over as-is. Upstream [#9944](https://github.com/software-mansion/react-native-reanimated/issues/9944) and [#9945](https://github.com/software-mansion/react-native-reanimated/issues/9945) are both still open and still unfixed in 4.7.0 nightlies (verified in the nightly source, not just by issue state), so both edits remain necessary.

No behavior change on the currently pinned 4.6.0: the same two edits are applied by the same anchors as before.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

All four pass: `tsc` clean, `eslint` clean, `prettier --check "**/*.ts*"` clean, jest 68 suites and 1498 tests passing, plus `check-styles.test.ts` passing. Note that `prettier` and `tsc` only match `**/*.ts*`, so neither actually covers these two `.mjs` files; `eslint` does cover them.

I had to exclude `.claude/**` locally to run `lint` and `test` at all, since this checkout has git worktrees nested under `.claude/worktrees/` and neither `eslint.config.js` nor the jest `testPathIgnorePatterns` excludes them. `eslint .` then lints roughly thirty full copies of the repo (over 25 minutes at 100% CPU before I killed it, versus 17 seconds when excluded), and jest picks up the stale test copies (56 spurious suite failures). This is a local-only artifact, CI checkouts have no such directory, and it is unrelated to this PR.

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

Verified by running the patch against pristine upstream tarball sources rather than by inspection:

| source | result |
| --- | --- |
| 4.6.0 pristine (currently pinned) | both edits apply via the `4.6.x` anchors |
| 4.7.0-nightly-20260901 pristine | both edits apply via the `4.7.x` anchors |
| already-patched file (current `node_modules`) | detected, skipped, no double apply |
| re-run over any of the above | no-op |
| unrecognized layout | throws, install fails |
| unrecognized layout with `ZEUS_ALLOW_UNPATCHED_REANIMATED=1` | warns, install proceeds |

The resulting 4.7 diff was reviewed for correctness against the refactored source, and `<algorithm>` is already included there for the `std::find` call.

- [ ] Android
- [ ] iOS

This is a build-time patch script change with no app code in it, and the 4.7.x path cannot be exercised on device until we actually bump reanimated. Device re-testing of the transitions themselves (Contacts to ContactDetails, push and pop, plus a cancelled back swipe, on both platforms) is owed as part of that bump, which is why this is a draft.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

N/A, no node interaction.

### Locales

No dependency change. Existing checkouts do not need to re-run `yarn`, since the pinned 4.6.0 result is byte-identical to what is already patched in `node_modules`.

### Other:

Neither #4221 nor #4222 can be closed by this PR. Both stay open until a reanimated release actually ships the fixes.

### DIFF
--- a/patches/index.mjs
+++ b/patches/index.mjs
@@ -11,12 +11,17 @@ import { patchReanimatedSetFixes } from './patch-reanimated-set-fixes.mjs';
 console.log('Running postinstall patches...\n');
 
 (async () => {
-    patchJcenter();
-    patchNativeEventEmitter();
-    patchReactNativeNotifications();
-    patchNobleHashes();
-    patchReanimatedBoundaryAndroid();
-    patchReanimatedSetFixes();
+    try {
+        patchJcenter();
+        patchNativeEventEmitter();
+        patchReactNativeNotifications();
+        patchNobleHashes();
+        patchReanimatedBoundaryAndroid();
+        patchReanimatedSetFixes();
+    } catch (error) {
+        console.error(`\nPostinstall patch failed:\n\n${error.message}\n`);
+        process.exit(1);
+    }
 
     console.log('\nAll patches applied successfully.');
 })();
--- a/patches/patch-reanimated-set-fixes.mjs
+++ b/patches/patch-reanimated-set-fixes.mjs
@@ -1,7 +1,7 @@
 // Two behavioral fixes to react-native-reanimated's experimental shared
 // element transition proxy (LayoutAnimationsProxy_Experimental.cpp), applied
 // as source patches. Both were root-caused on-device in ZEUS; neither is
-// fixed upstream as of 4.5.1 (checked 2026-07-13).
+// fixed upstream as of 4.7.0-nightly-20260901 (re-checked 2026-09-02).
 //
 // Tracking: https://github.com/ZeusLN/zeus/issues/4222
 // Remove each edit once the corresponding upstream fix ships in a release.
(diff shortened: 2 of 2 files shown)


kaloudis:
Superseded by #4689, closing.

4.7.0 stable shipped a later refactor than the nightly this was anchored against: upstream #10371 renamed `LayoutAnimationsProxy_Experimental.cpp` to `LayoutAnimationsProxy.cpp` and removed `closingScreenTag_` and `synchronized_` altogether, which fixes software-mansion/react-native-reanimated#9945. So the resync edit is gone rather than re-anchored, the 4.6.x anchors are no longer needed, and the file path itself had to change.

That last part also turned out to be a hole in this PR: the fail-loud `throw` sat after the `fs.existsSync` early return, so on a real 4.7.0 install the patch would have skipped silently anyway. #4689 makes a missing proxy source fatal too, and keeps the `patches/index.mjs` exit-1 wrapper from here.

Issue #4138, opened 2026-06-04
Payments: factor in donation amount in 'Not enough funds to cover payment' display


kaloudis:
See https://github.com/ZeusLN/zeus/pull/3773#pullrequestreview-4427593898 for context

PR #4555, opened 2026-09-03
cashu: count the donation in the funds check

TheSeydiCharyyev:
# Description

Relates to issue: #4138

The donation is a second payment, drawn from the same balance right after the invoice is paid, but the funds check only covered the invoice and its fee reserve. A balance that cleared that check could still be too small for the donation, so the payment went out and the donation failed afterwards.

That failure is invisible. Every insufficient-funds branch that would report it is guarded by `if (!isDonationPayment)`, so when the donation is the payment that falls short, nothing is set and nothing is shown. The warning before paying is the only place a user can learn about it, which is what this PR fixes.

Both paths that send a donation now count it:

- Single mint. `CashuStore.getPayReq()` compared the mint balance against the melt quote amount plus its fee reserve. It now also exposes the two numbers it compared, and `CashuPaymentRequest` extends the check with the donation. The store's own verdict is untouched, so nothing that reads `getPayReqError` today behaves differently.
- Multi mint. `MultimintPayment` computes its own coverage check in the constructor and sets `step: FAILED` when the selected mints fall short, which is what stops `componentDidMount()` from starting the payment. That check now includes the donation.

A donation is only counted under the conditions the sending views apply before they send one: mainnet, donations enabled, and a non-zero amount. Those conditions used to be spelled out in each view; they are now resolved in one place by `getDonationToSend`. On testnet, on iOS, or with donations off, the check is unchanged.

One timing detail worth calling out. `NodeInfoStore.nodeInfo` starts as an empty object and is replaced once the node responds, so `isMainNet` can still be undefined while the `MultimintPayment` constructor runs, and the donation would be left out of a check made seconds before the donation is actually sent. `componentDidMount()` therefore redoes the check with the values available at that point and refuses to start the payment when the mints cannot cover both. `CashuPaymentRequest` does not need this: its check runs in `render()` and the component observes `nodeInfo`, so it re-renders once the value arrives.

The warning reuses `stores.CashuStore.notEnoughFunds`, so there is no new locale text. It also gates the pay button, and since the donation slider is on the same screen, lowering the donation clears it.

### What this does not cover

- The donation's own fee reserve. The full requirement is invoice + fee reserve + donation + the donation's fee reserve. The last one needs a separate melt quote, which needs an LNURL round trip before the user has paid anything, so this PR counts the donation only and can still be one fee reserve short.
- iOS. Only Android was tested; the donation paths are gated to non-iOS anyway, but the code was not exercised there.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

`getDonationToSend` and `calculateTotalWithDonation` are covered in `utils/DonationUtils.test.ts`. The two call sites are view code and there are no component tests in the repo, so they are not covered by tests.

- [x] Android
- [ ] iOS

Pixel 9 Pro emulator, x86_64, Android 16. Both flows verified: the single-mint warning appearing with a donation and clearing from the slider, and the multimint refuse-to-start. Details and screenshots in a comment below.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

None of these: the flow this PR touches is Cashu ecash. The wallet was an LDK Node mainnet wallet, used only so `isMainNet` is genuine; no LDK payment was made.

### Locales

### Other:

### DIFF
--- a/stores/CashuStore.ts
+++ b/stores/CashuStore.ts
@@ -1845,6 +1845,8 @@ export default class CashuStore {
         this.paymentPreimage = undefined;
         this.getPayReqError = undefined;
         this.feeEstimate = undefined;
+        this.payReqMintBalance = 0;
+        this.payReqAmount = 0;
         this.meltQuotes = [];
         this.meltQuote = undefined;
         this.noteKey = undefined;
@@ -4132,6 +4134,8 @@ export default class CashuStore {
         this.paymentRequest = bolt11Invoice;
         this.feeEstimate = undefined;
         this.getPayReqError = undefined;
+        this.payReqMintBalance = 0;
+        this.payReqAmount = 0;
 
         try {
             if (__DEV__) {
--- a/stores/CashuStore.ts
+++ b/stores/CashuStore.ts
@@ -223,6 +223,13 @@ export default class CashuStore {
     @observable public paymentError?: boolean;
     @observable public paymentErrorMsg?: string;
     @observable public feeEstimate?: number;
+    // Inputs of the single-mint sufficiency check run by getPayReq(), exposed
+    // so the payment view can extend it with the donation that is sent from
+    // the same mint right after the invoice is paid. Both stay 0 when that
+    // check did not run: multi-mint plans its own coverage in
+    // MultimintPayment, which refuses to start when the mints fall short.
+    @observable public payReqMintBalance: number = 0;
+    @observable public payReqAmount: number = 0;
     @observable public meltQuotes: {
         mintUrl: string;
         meltQuote: CDKMeltQuote;
@@ -4213,6 +4220,8 @@ export default class CashuStore {
             // the fee row doesn't render "0 sats" for an invoice that can't
             // be paid; every branch here assigns it before use.
             let totalFeeEstimate: number | undefined;
+            let mintBalanceChecked = 0;
+            let amountChecked = 0;
 
             if (rawPaymentAmt <= 0) {
                 // Zeus doesn't yet wire up CDK's amountless-melt option
@@ -4300,6 +4309,8 @@ export default class CashuStore {
                         : undefined;
 
                 totalFeeEstimate = Number(singleMeltQuote.fee_reserve) || 0;
+                mintBalanceChecked = mintBalance;
+                amountChecked = Number(singleMeltQuote.amount) || 0;
                 this.meltQuotes = [];
             }
 
@@ -4312,6 +4323,8 @@ export default class CashuStore {
                 this.meltQuote = singleMeltQuote;
                 this.feeEstimate = totalFeeEstimate;
                 this.getPayReqError = computedPayReqError;
+                this.payReqMintBalance = mintBalanceChecked;
+                this.payReqAmount = amountChecked;
             });
             if (__DEV__) {
                 console.log('getPayReq: Success, setting loading = false');
--- a/views/Cashu/MultimintPayment.tsx
+++ b/views/Cashu/MultimintPayment.tsx
@@ -33,6 +33,7 @@ import {
 } from '../../utils/CashuUtils';
 import {
     calculateTotalWithDonation,
+    getDonationToSend,
     loadDonationLnurl
 } from '../../utils/DonationUtils';
 import { localeString } from '../../utils/LocaleUtils';
@@ -111,19 +112,15 @@ export default class MultimintPayment extends React.Component<
             Number(route.params?.paymentAmount || 0);
         const feeEstimate = Number(CashuStore?.feeEstimate) || 0;
         // The donation is drawn from the same mints right after the invoice is
-        // paid, so it has to be covered before the payment starts. Mirrors the
-        // conditions componentDidMount checks before it sends the donation:
-        // anything else means no donation leaves the wallet.
-        const donationAmount =
-            NodeInfoStore?.nodeInfo?.isMainNet &&
-            route.params?.enableDonations &&
-            route.params?.donationAmount
-                ? Number(route.params.donationAmount) || 0
-                : 0;
+        // paid, so it has to be covered before the payment starts.
         const totalNeeded = calculateTotalWithDonation(
             requestAmount,
             feeEstimate,
-            donationAmount
+            getDonationToSend(
+                NodeInfoStore?.nodeInfo?.isMainNet,
+                route.params?.enableDonations,
+                route.params?.donationAmount
+            )
         );
 
         const hasNoMintsSelected = effectiveMints.length === 0;
@@ -193,6 +190,34 @@ export default class MultimintPayment extends React.Component<
         });
 
         if (this.state.step !== MultinutPaymentStep.FAILED) {
+            // nodeInfo starts out empty and is filled in once the node
+            // responds, so the constructor may have run before isMainNet was
+            // known and left the donation out of the coverage check. Redo it
+            // here, where the donation is read anyway, so a payment never
+            // starts against a balance that cannot also cover the donation.
+            const { NodeInfoStore, route } = this.props;
+            const totalNeeded = calculateTotalWithDonation(
+                this.paymentAmountSat,
+                Number(CashuStore?.feeEstimate) || 0,
+                getDonationToSend(
+                    NodeInfoStore?.nodeInfo?.isMainNet,
+                    route.params?.enableDonations,
+                    route.params?.donationAmount
+                )
+            );
+            const { totalSelectedBalance } = this.state;
+
+            if (
+                totalSelectedBalance > 0 &&
+                totalNeeded > totalSelectedBalance
+            ) {
+                this.setState({
+                    step: MultinutPaymentStep.FAILED,
+                    error: localeString('stores.CashuStore.notEnoughFunds')
+                });
+                return;
+            }
+
             this.executePayment();
         }
     }
--- a/views/Cashu/MultimintPayment.tsx
+++ b/views/Cashu/MultimintPayment.tsx
@@ -24,7 +24,6 @@ import Screen from '../../components/Screen';
 import { sendingStyles } from '../../components/sendingStyles';
 
 import CashuStore from '../../stores/CashuStore';
-import NodeInfoStore from '../../stores/NodeInfoStore';
 
 import {
     MintPaymentStatus,
@@ -33,7 +32,6 @@ import {
 } from '../../utils/CashuUtils';
 import {
     calculateTotalWithDonation,
-    getDonationToSend,
     loadDonationLnurl
 } from '../../utils/DonationUtils';
 import { localeString } from '../../utils/LocaleUtils';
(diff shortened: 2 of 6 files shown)


shubhamkmr04:
@TheSeydiCharyyev Please resolve conflicts

TheSeydiCharyyev:
Rebased on master and force-pushed as 55ff435f. Conflicts resolved, all four checks green.

The conflict was with #4547, on the one line where this PR declares its locals next to `totalFeeEstimate`. Master's `number | undefined` version is kept as is.

Two notes for review. `getPayReq` now has a leading `rawPaymentAmt <= 0` branch; the two numbers this PR exposes from the store stay 0 there, and the view check is guarded on `payReqMintBalance > 0`, so it does not fire for no-amount invoices. And `enableDonations` in `CashuPaymentRequest` now also requires `!isNoAmountInvoice` — the donation check reads that same flag, so it picks up the new condition without a change here.

kaloudis (commented):
Approach ACK 55ff435

The diagnosis is right and the scoping is careful: the store's own verdict is untouched and the views extend it, so nothing that reads `getPayReqError` today changes. The single/multi consistency also holds up under the edge cases I could construct. `payReqMintBalance` is only non-zero when `getPayReq` ran the single-mint branch, and `handleMultiMintToggle` re-runs `getPayReq` after changing the selection, so the single-mint donation warning can never wrongly block a payment that multimint could cover (including the toggle-on-with-one-eligible-mint case, where the view state and `triggerPayment`'s `length > 1` check stay consistent with the store branch). The iOS and testnet gates are preserved through `enableDonations` and `getDonationToSend`, the warning gates both the swipe control and the pay button, and lowering the slider clears it since the check runs in render. Helper tests are solid.

Two should-fixes inline: the `componentDidMount` recheck does not actually close the race it describes (the donation send needs to consume the same snapshot the coverage check used), and the two new observables are never reset on the paths that skip the success `runInAction`.

Since this touches the payment flow it will also need an Android device pass before merge: single-mint warning appearing and clearing via the donation slider, and the multimint refuse-to-start path.

views/Cashu/MultimintPayment.tsx
```diff
@@ -176,6 +190,34 @@ export default class MultimintPayment extends React.Component<
...
+                    NodeInfoStore?.nodeInfo?.isMainNet,
+                    route.params?.enableDonations,
+                    route.params?.donationAmount
+                )
+            );
+            const { totalSelectedBalance } = this.state;
+
+            if (
+                totalSelectedBalance > 0 &&
+                totalNeeded > totalSelectedBalance
```
kaloudis: This recheck can never observe a different `isMainNet` than the constructor did: constructor, render, and `componentDidMount` run synchronously in one React commit, and the MobX update from the network callback cannot interleave between them. The window that actually matters is the payment itself. `executePayment` re-reads `NodeInfoStore?.nodeInfo?.isMainNet` live when it decides whether to send the donation, so a donation this check never counted (because `isMainNet` was still undefined here) can still be sent once nodeInfo arrives mid-payment, reproducing exactly the silent failure this PR targets. The same live re-read exists in `CashuSendingLightning.componentDidUpdate` for the single-mint path.

Suggestion: snapshot the decision once where the check runs, e.g. `this.donationToSend = getDonationToSend(...)`, and have `executePayment` send the donation based on that snapshot instead of re-reading the store. That makes the check and the send consistent by construction, and also removes the mismatch where the send condition uses string truthiness (`donationAmount`, where `'0'` is truthy) while the helper uses `Number(x) || 0`.

Minor, while you are here: this failure path always uses the generic `notEnoughFunds`, dropping the `notEnoughFundsSelfSend` variant the constructor's identical check produces. Extracting the shared check into a method would fix that and remove the duplication.

stores/CashuStore.ts
```diff
@@ -223,6 +223,13 @@ export default class CashuStore {
     @observable public paymentError?: boolean;
     @observable public paymentErrorMsg?: string;
     @observable public feeEstimate?: number;
+    // Inputs of the single-mint sufficiency check run by getPayReq(), exposed
+    // so the payment view can extend it with the donation that is sent from
+    // the same mint right after the invoice is paid. Both stay 0 when that
+    // check did not run: multi-mint plans its own coverage in
+    // MultimintPayment, which refuses to start when the mints fall short.
+    @observable public payReqMintBalance: number = 0;
```
kaloudis, line 231: These two are only assigned in the success `runInAction`, so they go stale on every path that skips it: the expired-invoice early return, the catch block, and `clearPayReq`.

Concrete symptom: review invoice A on a single mint (fields get set), then open an expired invoice B. The expiry guard returns before the `runInAction` and leaves `getPayReqError` undefined, so the render computes `donationLeavesTooLittle` from A's balance and amount against B's donation and can stack a spurious not-enough-funds warning under the invoice-expired error. Cosmetic today, since the pay controls are already hidden by `!isPayReqExpired`, but two lines close it: zero both fields at the top of `getPayReq` next to the `feeEstimate` reset, and add them to `clearPayReq`.

TheSeydiCharyyev:
Both should-fixes are in, as c98560b6. Checks are green.

41fefeb5 resets `payReqMintBalance` and `payReqAmount` on every path out of `getPayReq`, and in `clearPayReq`. c98560b6 moves the donation decision into one place and drops the duplicated gate. Details are in the two threads.

No new tests here: the donation resolution is view code and the other change is a store field reset, and there are no component tests in the repo. `getDonationToSend` and `calculateTotalWithDonation` stay covered in `utils/DonationUtils.test.ts`.

The Android pass is still outstanding. I will report the two flows you named — the single-mint warning appearing and clearing from the donation slider, and the multimint refuse-to-start path — once I have a native build.

views/Cashu/MultimintPayment.tsx
TheSeydiCharyyev: You are right, and the recheck is gone. Constructor, render and componentDidMount are one commit, so it could never see a different `isMainNet`; the send was the real window.

The decision is now made once, in `CashuPaymentRequest.resolveDonationToSend`, and the sending views are handed the resolved amount instead of the inputs. `MultimintPayment` snapshots it in the constructor for the coverage check and `executePayment` sends that same number. `CashuSendingLightning` sends what it was given, so the single-mint path is consistent in the same way.

That let the duplicated gate go: the `enableDonations` route param and the `NodeInfoStore` injection are removed from both sending views, and the string-truthiness mismatch goes with them, since what is passed now is a resolved number.

On the `notEnoughFundsSelfSend` variant: with the recheck deleted there is only one coverage check left, the one in the constructor, so nothing drops that message anymore.

One behaviour change worth naming. If `nodeInfo` has not arrived when the user presses pay, the donation is now not sent at all, where before it could still go out once `nodeInfo` landed mid-payment. That is what makes the check and the send agree by construction, but it does mean a donation can be skipped on a slow node connection. Say the word if you would rather have it the other way and I will make the check wait for `nodeInfo` instead.

stores/CashuStore.ts
TheSeydiCharyyev: Fixed at both places you named: both fields are zeroed at the top of `getPayReq`, next to the `feeEstimate` reset, and added to `clearPayReq`. That also covers the catch block and the expired-invoice early return, since every `getPayReq` call passes the reset before it can reach either.

TheSeydiCharyyev:
Android pass done, on both flows you named. Everything below ran on the branch build with no local changes: a mainnet LDK Node wallet, so `isMainNet` is real, Cashu enabled, and the Testnut mint holding 1,000 sats. Donations on at the default 5%. The invoice is a real 960 sat bolt11.

Single mint. 960 + 19 fee = 979, which fits in 1,000, so before this PR there would have been no warning at all. With the 48 sat donation the total is 1,027, and the screen shows `Not enough funds to cover payment` with the pay button disabled. Dragging the donation slider to 0% clears the warning in place and enables the button, without leaving the screen.

<table>
<tr>
<td align="center">warning with the donation<br>with the 5% donation</td>
<td align="center">warning cleared<br>slider at 0%</td>
</tr>
</table>

Multi mint. With a second mint added at 0 sats, so the combined balance stays 1,000, and Multi-mint switched on, the review screen shows no warning. That is what the `payReqMintBalance > 0` guard is for: the store's single-mint check did not run on that path. Pressing pay lands on the sending screen in the FAILED state with `Not enough funds to cover payment`, and the payment never starts.

multimint refuses to start


Setup: Pixel 9 Pro emulator, x86_64, Android 16. The mint is Testnut, which runs a FakeWallet, so the sats are not real; the rest of the path is (mainnet wallet, Cashu protocol, real bolt11). iOS is still not covered, and the donation's own fee reserve is still not counted, as the PR body says.

TheSeydiCharyyev:
The one thing still open here is the question at the end of my last reply: https://github.com/ZeusLN/zeus/pull/4555#discussion_r3939357056

With this PR, if `nodeInfo` has not arrived when the user presses pay, the donation is skipped. Before, it could still go out once `nodeInfo` landed mid-payment. That is the price of making the coverage check and the send agree by construction.

Say which way you want it. Keeping the current shape needs nothing from me. The other way means the check waits for `nodeInfo` instead, which is a small change.

Separately, master has eight commits on `stores/CashuStore.ts` and `views/Cashu/CashuSendingLightning.tsx` since this branch was cut. GitHub still calls the PR mergeable, but say the word and I will rebase so review happens against current code.

kaloudis:
> The one thing still open here is the question at the end of my last reply: [#4555 (comment)](https://github.com/ZeusLN/zeus/pull/4555#discussion_r3939357056)
> 
> With this PR, if `nodeInfo` has not arrived when the user presses pay, the donation is skipped. Before, it could still go out once `nodeInfo` landed mid-payment. That is the price of making the coverage check and the send agree by construction.
> 
> Say which way you want it. Keeping the current shape needs nothing from me. The other way means the check waits for `nodeInfo` instead, which is a small change.
> 
> Separately, master has eight commits on `stores/CashuStore.ts` and `views/Cashu/CashuSendingLightning.tsx` since this branch was cut. GitHub still calls the PR mergeable, but say the word and I will rebase so review happens against current code.

Keep the shape. It's fine. Waiting is worse.

kaloudis:
One small related item, optional here or as a follow-up. The Cashu donation slider is not gated on mainnet: `enableDonations` in `CashuPaymentRequest.tsx` render (around line 440) checks platform, amountless invoice, and the setting, but not `isMainNet`. The Lightning screen does gate it (`PaymentRequest.tsx:1370`).

So before `nodeInfo` loads, and always on testnet, the user can pick a donation that `resolveDonationToSend()` will resolve to 0 and never send. This predates the PR, but now that the send is decided in one place it would be good for the UI to match. Adding `NodeInfoStore.nodeInfo?.isMainNet` to that `enableDonations` condition should be enough. The view is an observer, so the slider appears once `nodeInfo` arrives.

PR #4565, opened 2026-09-04
feat(nwc): allow viewing connection secret after creation

### LINKED ISSUE #4525: NWC: a connection's pairing URI is viewable only once, which blocks guarding regenerate's late navigate
Split out of the review of #4520, which fixed the delete half of this and deliberately left the regenerate half alone.

## The immediate problem

`NWCConnectionQR` takes `nostrUrl` as a route param:

```ts
route: Route<'NWCConnectionQR', { connectionId: string; nostrUrl: string }>;
```

and is only navigated to from two places, both immediately after the URI is minted: `AddOrEditNWCConnection` on create, and `NWCConnectionDetails.regenerateConnection` on regenerate. There is no path back to it. If the user dismisses that screen before pairing, or misses it, the pairing URI is gone from their point of view and the only recovery is to regenerate again.

## Why it blocks a fix in #4520

#4443 made `deleteConnection` wait for in-flight NIP-47 handlers, which can be ~150s behind a slow `pay_invoice`. #4487 covered the missing UI for that wait, and #4520 shipped it.

That long wait makes it natural for the user to leave the screen. React Navigation does not no-op a late `goBack()` or `navigate()` from a screen that has already been popped (`useNavigationCache` sets `source` but never `target`, and `StackRouter`'s `POP` only honors `source` when `action.target === state.key`), so the action lands on whatever is currently on top.

#4520 fixed the delete path with the `isUnmounted` guard already used in `views/LSPS/LSPPaymentAwait.tsx`. The same guard is not safe on the regenerate path:

```ts
await NostrWalletConnectStore.deleteConnection(connection.id);
const nostrUrl = await NostrWalletConnectStore.createConnection(params);
if (nostrUrl) {
    navigation.navigate('NWCConnectionQR', { connectionId, nostrUrl });
}
```

Skipping that `navigate` when the screen is gone would strand a freshly regenerated connection whose URI the user can never see again, which is worse than the current behavior of yanking them to the QR screen late. So regenerate is left with the stale-navigate behavior until the URI is recoverable.

## Everything needed to rebuild the URI is already persisted

The URI is a pure function of state we keep:

- `NWC_CLIENT_KEYS` maps `connection.pubkey` to the client private key (`storeClientKeys` / `deleteClientKeys` in `NostrWalletConnectStore`)
- the wallet service pubkey, `relayUrl`, and `lud16` are all available
- `NostrConnectUtils.buildWalletConnectConnectionUrl(walletPubKeyHex, relayUrl, connectionSecretKeyHex, lud16)` does the rest

`generateConnectionSecret` is just that builder plus a fresh key, so nothing about the URI is ephemeral. It is only the route param that is.

## Proposed

1. Let a connection's URI be viewed again from `NWCConnectionDetails`, either by adding an entry point that rebuilds the URI from persisted state, or by having `NWCConnectionQR` rebuild it when `nostrUrl` is absent from the route params.
2. Once that exists, guard `regenerateConnection`'s `navigate` with the same `isUnmounted` check #4520 added to delete, so a regenerate that finishes after the user has moved on stops hijacking their stack.

## Open design question

Re-displaying the URI re-exposes a bearer credential that grants the connection's permissions, so it is not the same as re-displaying a read-only detail. The exposure is no worse than the create-time screen that already shows it, but showing it on demand at any later time is a different threat model than showing it once at pairing. Worth deciding whether this sits behind the same authentication gate we use for other secret reveals, and whether it should be suppressed for connections with payment permissions.

ajaysehwal:
# Description

Relates to issue: [**ZEUS-4525**](https://github.com/ZeusLN/zeus/issues/4525)

## Summary

- Previously, the NWC pairing URL (connection secret) was only ever shown once, at connection creation/regeneration time, via a route param passed into `NWCConnectionQR`. Once that screen was dismissed there was no way to see it again.
- Adds a "View Connection Secret" button on `NWCConnectionDetails` that navigates to `NWCConnectionQR` with just a `connectionId` (no `nostrUrl`), gated behind the app's existing re-auth prompt (same pattern used for other secret reveals like the seed).

Simulator Screenshot - iPhone 17 Pro - 2026-09-04 at 16 02 49

- Adds `NostrWalletConnectStore.getConnectionUrl(connectionId)`, which rebuilds the pairing URL from persisted state — the stored client private key, the wallet service pubkey, and the connection's relay/lud16 — since nothing about the URL is actually ephemeral except the route param that used to carry it.
- `NWCConnectionQR` now supports loading the URL asynchronously when it's not passed via route params: shows a loading indicator while fetching, and a fallback message if the URL can't be rebuilt (e.g. connections created before client keys were persisted).
- Minor: guard `createConnection`'s post-navigation state update with `!this.isUnmounted` to avoid a set-state-after-unmount warning.

## Test plan
- [x] Create a new NWC connection, confirm the QR/secret still displays immediately as before.
- [x] From the connection details screen, tap "View Connection Secret", complete re-auth, and confirm the same pairing URL/QR is shown.
- [x] Verify the fallback "connection secret unavailable" message appears for a connection whose client key can't be found.
- [x] Regenerate a connection and confirm the regenerate flow is unaffected.

- [x] New feature
- [ ] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1066,11 +1066,15 @@ export default class NostrWalletConnectStore {
     /**
      * Rebuilds a connection's pairing URL from persisted state (the stored
      * client private key, the wallet service pubkey, and the connection's
-     * relay/lud16) so it can be viewed again after the QR screen shown at
+     * relay) so it can be viewed again after the QR screen shown at
      * create/regenerate time has been dismissed. Nothing about the URL is
-     * ephemeral except the route param that originally carried it.
-     * Returns null if the connection or its client key can no longer be
-     * found (e.g. it was created before client keys were persisted).
+     * ephemeral except the route param that originally carried it. The
+     * lud16 param reflects the current lud16Enabled setting and lightning
+     * address, not whatever was in effect when the connection was created.
+     * Returns null if the connection, its client key, or the wallet
+     * service keys can no longer be found (e.g. it was created before
+     * client keys were persisted) rather than regenerating anything, since
+     * this is a read-only path.
      */
     public getConnectionUrl = async (
         connectionId: string
@@ -1078,9 +1082,6 @@ export default class NostrWalletConnectStore {
         const { connection } = this.getConnection({ connectionId });
         if (!connection) return null;
 
-        if (!this.walletServiceKeys?.publicKey) {
-            await this.loadWalletServiceKeys();
-        }
         if (!this.walletServiceKeys?.publicKey) return null;
 
         const clientPrivateKey = await this.getClientPrivateKey(
--- a/stores/NostrWalletConnectStore.ts
+++ b/stores/NostrWalletConnectStore.ts
@@ -1063,6 +1063,41 @@ export default class NostrWalletConnectStore {
         return { connection, index };
     };
 
+    /**
+     * Rebuilds a connection's pairing URL from persisted state (the stored
+     * client private key, the wallet service pubkey, and the connection's
+     * relay/lud16) so it can be viewed again after the QR screen shown at
+     * create/regenerate time has been dismissed. Nothing about the URL is
+     * ephemeral except the route param that originally carried it.
+     * Returns null if the connection or its client key can no longer be
+     * found (e.g. it was created before client keys were persisted).
+     */
+    public getConnectionUrl = async (
+        connectionId: string
+    ): Promise<string | null> => {
+        const { connection } = this.getConnection({ connectionId });
+        if (!connection) return null;
+
+        if (!this.walletServiceKeys?.publicKey) {
+            await this.loadWalletServiceKeys();
+        }
+        if (!this.walletServiceKeys?.publicKey) return null;
+
+        const clientPrivateKey = await this.getClientPrivateKey(
+            connection.pubkey
+        );
+        if (!clientPrivateKey) return null;
+
+        const lud16 = this.lud16Enabled ? this.getLightningAddress() : null;
+
+        return NostrConnectUtils.buildWalletConnectConnectionUrl(
+            this.walletServiceKeys.publicKey,
+            connection.relayUrl,
+            clientPrivateKey,
+            lud16
+        );
+    };
+
     public getActivities = async (
         connectionId: string
     ): Promise<{ name: string; activity: ConnectionActivity[] }> => {
@@ -3539,6 +3574,17 @@ export default class NostrWalletConnectStore {
                     'stores.NostrWalletConnectStore.error.failedToStorePrivateKey'
                 )
             );
+        }
+    }
+    private async getClientPrivateKey(pubkey: string): Promise<string | null> {
+        try {
+            const storedKeys = await Storage.getItem(NWC_CLIENT_KEYS);
+            if (!storedKeys) return null;
+            const keys: ClientKeys = JSON.parse(storedKeys);
+            return keys[pubkey] || null;
+        } catch (error) {
+            console.error('Failed to read client keys:', error);
+            return null;
         }
     }
     private async deleteClientKeys(pubkey: string): Promise<void> {
--- a/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
+++ b/views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
@@ -24,6 +24,7 @@ import ModalStore from '../../../stores/ModalStore';
 import { confirmAction } from '../../../utils/ActionUtils';
 import DateTimeUtils from '../../../utils/DateTimeUtils';
 import NostrConnectUtils from '../../../utils/NostrConnectUtils';
+import { reAuthNavigation } from '../../../utils/NavigationUtils';
 import { themeColor } from '../../../utils/ThemeUtils';
 import { localeString } from '../../../utils/LocaleUtils';
 
@@ -183,6 +184,15 @@ export default class NWCConnectionDetails extends React.Component<
         });
     };
 
+    viewConnectionSecret = (connectionId: string) => {
+        // The pairing URL is a bearer credential for this connection's
+        // permissions, so re-displaying it on demand is gated behind the
+        // same re-auth prompt used for other secret reveals (e.g. the seed).
+        reAuthNavigation(this.props.navigation, 'NWCConnectionQR', {
+            connectionId
+        });
+    };
+
     buildConnectionParams = (connection: NWCConnection): any => {
         const params: any = {
             id: connection.id,
@@ -262,7 +272,7 @@ export default class NWCConnectionDetails extends React.Component<
             const nostrUrl = await NostrWalletConnectStore.createConnection(
                 params
             );
-            if (nostrUrl) {
+            if (nostrUrl && !this.isUnmounted) {
                 const createdConnection =
                     NostrWalletConnectStore.connections[0];
                 navigation.navigate('NWCConnectionQR', {
@@ -719,6 +729,17 @@ export default class NWCConnectionDetails extends React.Component<
                                     </Text>
                                 </View>
                             )}
+                            <Button
+                                title={localeString(
+                                    'views.Settings.NostrWalletConnect.viewConnectionSecret'
+                                )}
+                                onPress={() =>
+                                    this.viewConnectionSecret(connection.id)
+                                }
+                                disabled={regenerating || deleting}
+                                secondary
+                                noUppercase
+                            />
                             <Button
                                 title={localeString(
                                     'views.Settings.NostrWalletConnect.regenerateConnection'
--- a/views/Settings/NostrWalletConnect/NWCConnectionQR.tsx
+++ b/views/Settings/NostrWalletConnect/NWCConnectionQR.tsx
@@ -56,9 +56,16 @@ export default class NWCConnectionQR extends React.Component<
     connectionTimeout: ReturnType<typeof setTimeout> | null = null;
     navigateBackTimeout: ReturnType<typeof setTimeout> | null = null;
     isUnmounted = false;
+    // True when this screen was opened to re-display an existing
+    // connection's secret (no nostrUrl route param) rather than right
+    // after creating/regenerating one. In this mode the connection is
+    // already paired, so the waiting-for-first-connection flow (spinner,
+    // timeout modal, success redirect) does not apply.
+    isViewOnly: boolean;
 
     constructor(props: NWCConnectionQRProps) {
         super(props);
+        this.isViewOnly = !props.route.params.nostrUrl;
         this.state = {
             isConnected: false,
             appState: AppState.currentState,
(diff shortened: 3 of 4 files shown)


views/Settings/NostrWalletConnect/NWCConnectionQR.tsx
```diff
@@ -75,6 +85,19 @@ export default class NWCConnectionQR extends React.Component<
                 this.setState({ showTimeoutMessage: true });
             }
         }, CONNECTION_TIMEOUT_MS);
+
+        if (!this.state.nostrUrl) {
```
kaloudis, line 99: **Should fix:** the waiting-for-connection flow still runs unconditionally in view-only mode (`componentDidMount` calls `startWaitingForConnection` above), which misbehaves for an existing connection:

1. Every handled NIP-47 request goes through `withGlobalHandler` -> `markConnectionUsed` -> `stopWaitingForConnection`, which sets `connectionJustSucceeded`. `componentDidUpdate` then flips to the "Connected successfully" animation and auto-navigates to the connections list after 2s. Since NWC clients poll constantly (get_balance/get_info), a background request from the already-paired app can dismiss the QR seconds after it appears, before the user can scan or copy it. That undermines the feature's main use case.
2. The 20s "no event received" timeout modal will fire on essentially every view-only visit (and also appears over the `connectionSecretUnavailable` fallback).
3. "Waiting for app to connect..." shows under the QR for a connection that is already connected.

Suggestion: derive a view-only flag from the absence of `route.params.nostrUrl` (the same condition used here for the async load) and skip `startWaitingForConnection`, the timeout, and the success redirect in that mode.

views/Settings/NostrWalletConnect/NWCConnectionQR.tsx
```diff
@@ -75,6 +85,19 @@ export default class NWCConnectionQR extends React.Component<
...
+            NostrWalletConnectStore.getConnectionUrl(connectionId).then(
```
kaloudis: **Should fix:** no `.catch` here. `getConnectionUrl` can reject: `loadWalletServiceKeys` catches read errors, but its fallback calls `saveWalletServiceKeys`, which throws on storage failure. On rejection this is an unhandled promise rejection and the screen sits on the loading spinner forever. A `.catch` that sets `loadingUrl: false, urlUnavailable: true` covers it.

stores/NostrWalletConnectStore.ts
```diff
@@ -1063,6 +1063,41 @@ export default class NostrWalletConnectStore {
...
+     * found (e.g. it was created before client keys were persisted).
+     */
+    public getConnectionUrl = async (
+        connectionId: string
+    ): Promise<string | null> => {
+        const { connection } = this.getConnection({ connectionId });
+        if (!connection) return null;
+
+        if (!this.walletServiceKeys?.publicKey) {
+            await this.loadWalletServiceKeys();
```
kaloudis: Note: if `walletServiceKeys` is unset here, `loadWalletServiceKeys()` will **generate and persist a brand-new keypair** when storage is empty or corrupt. In that state this method silently returns a URL whose wallet pubkey no relay subscription is listening on until the next restart, with no error surfaced. Narrow edge since the store loads keys at init, but for a read path it would be safer to return `null` when keys are absent instead of regenerating.

stores/NostrWalletConnectStore.ts
```diff
@@ -1063,6 +1063,41 @@ export default class NostrWalletConnectStore {
...
+        }
+        if (!this.walletServiceKeys?.publicKey) return null;
+
+        const clientPrivateKey = await this.getClientPrivateKey(
+            connection.pubkey
+        );
+        if (!clientPrivateKey) return null;
+
+        const lud16 = this.lud16Enabled ? this.getLightningAddress() : null;
```
kaloudis, line 1092: Nit: the doc comment above says the URL is rebuilt from "the connection's relay/lud16", but `lud16` comes from the current global `lud16Enabled` setting and the current lightning address, not from anything stored per connection. If the address changed since creation, the rebuilt URL differs in that param. That's harmless metadata and matches what a regenerate would produce, but the comment should say the lud16 reflects current settings rather than the connection.

views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
```diff
@@ -183,6 +184,15 @@ export default class NWCConnectionDetails extends React.Component<
         });
     };
 
+    viewConnectionSecret = (connectionId: string) => {
+        // The pairing URL is a bearer credential for this connection's
+        // permissions, so re-displaying it on demand is gated behind the
+        // same re-auth prompt used for other secret reveals (e.g. the seed).
+        reAuthNavigation(this.props.navigation, 'NWCConnectionQR', {
```
kaloudis, line 191: Nit: after viewing the secret, both the back header and the Close button in `NWCConnectionQR` do `popTo('NostrWalletConnect')`, so the user lands on the connections list instead of back on this details screen they came from. Worth routing back to details in view-only mode if you add the flag suggested in the QR screen comment.

kaloudis (commented):
Approach ACK. This is the right shape for #4525: rebuilding the pairing URL from persisted state (stored client private key, wallet service pubkey, connection relay) is sound, since nothing about the URL was actually ephemeral except the route param that carried it. Gating the reveal behind `reAuthNavigation`, consistent with the seed reveal, is the right call for a bearer credential, and the widened `nostrUrl?` route param, loading/fallback states, and `isUnmounted` guards all check out.

I verified the plumbing: `buildWalletConnectConnectionUrl` argument order matches the call site, `getClientPrivateKey` mirrors the existing `deleteClientKeys` read logic, and the TS narrowing in the render ternary is sound.

One functional blocker before this is merge-ready, detailed inline: the waiting-for-connection flow (spinner, 20s timeout modal, and the success redirect) still runs when the screen is opened in view-only mode, and an incoming request from the already-paired client will dismiss the QR mid-scan. There is also a missing `.catch` on the async URL load, plus a few smaller notes inline.

No new storage or network egress; client private keys were already persisted in `NWC_CLIENT_KEYS`, this only adds a read path.

views/Settings/NostrWalletConnect/NWCConnectionQR.tsx
ajaysehwal: fixed in  559102859f27b5fdd81f9f6a737efa1d0f680087

views/Settings/NostrWalletConnect/NWCConnectionQR.tsx
ajaysehwal: fixed in 559102859f27b5fdd81f9f6a737efa1d0f680087

stores/NostrWalletConnectStore.ts
ajaysehwal: fixed in 559102859f27b5fdd81f9f6a737efa1d0f680087

stores/NostrWalletConnectStore.ts
ajaysehwal: fixed in 559102859f27b5fdd81f9f6a737efa1d0f680087

views/Settings/NostrWalletConnect/NWCConnectionDetails.tsx
ajaysehwal: fixed in 559102859f27b5fdd81f9f6a737efa1d0f680087

kaloudis (commented):
Round 2, reviewing 559102859 and 787284a41: all five round-1 comments are addressed. utACK 787284a41.

Verified:
- `isViewOnly`, derived from the absent `nostrUrl` route param, now gates `startWaitingForConnection`, the 20s timeout, and the success flip in `componentDidUpdate`, and the timeout modal can no longer open in view-only mode since `showTimeoutMessage` is never set there. I also checked the trickier store path: if a connection that was never used receives its first NIP-47 request while the view-only QR is open, `markConnectionUsed` starts and stops the waiting state internally and briefly sets `connectionJustSucceeded`, but the gated `componentDidUpdate` ignores it, so the screen can no longer be dismissed mid-scan.
- The unconditional `cancelWaitingForConnection()` in `componentWillUnmount` is just an observable reset when nothing was started, so leaving it ungated is fine.
- `getConnectionUrl` is now a pure read path: it returns null when `walletServiceKeys` is unset instead of regenerating a keypair, and the doc comment now correctly describes the lud16 behavior.
- The `.catch` covers the rejection path and lands on the `connectionSecretUnavailable` fallback instead of an eternal spinner.
- Back navigation works on both re-auth paths: with a login method configured, `Lockscreen.proceed` uses `popTo('NWCConnectionQR')`, which replaces the Lockscreen in the stack, so `goBack()` lands back on the details screen; without one, the direct navigate does the same.

One optional polish, non-blocking: in view-only mode, when the URL cannot be rebuilt, the fallback text renders with no button beneath it, since the footer Close is hidden in view-only mode. The header back arrow still works, but a Go Back button under the fallback text would match the happy path.

CI is green. Noting for the merge decision that testing so far is iOS + LDK Node per the PR body; an Android pass has not been claimed.

ajaysehwal:
> Round 2, reviewing [5591028](https://github.com/ZeusLN/zeus/commit/559102859f27b5fdd81f9f6a737efa1d0f680087) and [787284a](https://github.com/ZeusLN/zeus/commit/787284a41065e5c44c2c897741d94a9a9c058b0c): all five round-1 comments are addressed. utACK [787284a](https://github.com/ZeusLN/zeus/commit/787284a41065e5c44c2c897741d94a9a9c058b0c).
> 
> Verified:
> 
> * `isViewOnly`, derived from the absent `nostrUrl` route param, now gates `startWaitingForConnection`, the 20s timeout, and the success flip in `componentDidUpdate`, and the timeout modal can no longer open in view-only mode since `showTimeoutMessage` is never set there. I also checked the trickier store path: if a connection that was never used receives its first NIP-47 request while the view-only QR is open, `markConnectionUsed` starts and stops the waiting state internally and briefly sets `connectionJustSucceeded`, but the gated `componentDidUpdate` ignores it, so the screen can no longer be dismissed mid-scan.
> * The unconditional `cancelWaitingForConnection()` in `componentWillUnmount` is just an observable reset when nothing was started, so leaving it ungated is fine.
> * `getConnectionUrl` is now a pure read path: it returns null when `walletServiceKeys` is unset instead of regenerating a keypair, and the doc comment now correctly describes the lud16 behavior.
> * The `.catch` covers the rejection path and lands on the `connectionSecretUnavailable` fallback instead of an eternal spinner.
> * Back navigation works on both re-auth paths: with a login method configured, `Lockscreen.proceed` uses `popTo('NWCConnectionQR')`, which replaces the Lockscreen in the stack, so `goBack()` lands back on the details screen; without one, the direct navigate does the same.
> 
> One optional polish, non-blocking: in view-only mode, when the URL cannot be rebuilt, the fallback text renders with no button beneath it, since the footer Close is hidden in view-only mode. The header back arrow still works, but a Go Back button under the fallback text would match the happy path.
> 
> CI is green. Noting for the merge decision that testing so far is iOS + LDK Node per the PR body; an Android pass has not been claimed.

addressed in 2cd9d16dd371e97d569270b3bbedeed80d5a18c1

kaloudis (approved):

Issue #4559, opened 2026-09-03
Decide on TypeScript 6 now that the React Native 0.87 template ships it


kaloudis:
Split out of #4546 (React Native 0.86.0 -> 0.87.1).

The RN 0.87 upgrade-helper diff moves `typescript` to `^6.0.3`. #4546 deliberately stays on 5.7.3.

### Why the RN bump did not force it

The template bumps TypeScript because the template's `tsconfig.json` is essentially `{"extends": "@react-native/typescript-config"}`, and the 0.87 version of that shared config is authored against TS 6.

Zeus does not extend it. `tsconfig.json` is fully standalone with no `extends` at all, and `@react-native/typescript-config` is declared in `package.json` purely to keep the `@react-native/*` packages in version lockstep. Nothing outside `package.json` and `yarn.lock` references it. `yarn tsc` is clean on 5.7.3 against the RN 0.87 type surface.

So the upgrade is an independent toolchain change. Landing a major compiler version inside an already large RN bump would leave any resulting type error ambiguous between the two.

### To decide

- [ ] Actually try `typescript@6` and see whether `yarn tsc` stays clean. **This has not been attempted**, so there is no known incompatibility, only untested territory.
- [ ] Check `@typescript-eslint` compatibility, including the `resolutions` entries
- [ ] Pin exactly rather than using the template's `^6.0.3` caret, per the repo's convention
- [ ] Decide what to do with the unused `@react-native/typescript-config` devDep: drop it, or start extending it

That last point is not free. The shared config sets `customConditions: ["react-native"]`, which opts into the Strict TypeScript API. #4546 sets `customConditions: ["react-native-legacy-deep-imports"]` to preserve the pre-0.87 type surface, since 0.87 dropped react-native's top-level `types` field and types now resolve only through `exports`. Adopting the shared config means taking the Strict API migration at the same time.

Issue #4558, opened 2026-09-03
Migrate type-only class fields to declare so the Hermes babel transform profile can be restored


kaloudis:
Split out of #4546 (React Native 0.86.0 -> 0.87.1).

### Background

RN 0.87's `@react-native/babel-preset` changed `getTransformProfile`'s default from `default` to `hermes-stable`. Hermes profiles set `preserveClasses`, which stops the preset from shipping `@babel/plugin-transform-class-properties`. MobX 5 legacy decorators require that plugin, so on a stock 0.87 config every `@observable` field throws `Decorating class property failed`, which showed up as 116 test failures.

#4546 works around it by pinning the old profile:

```js
presets: [['module:@react-native/babel-preset', { unstable_transformProfile: 'default' }]]
```

That is correct and green, but it gives up the Hermes V1 optimisations that `hermes-stable` exists to enable.

### The trap, for whoever picks this up

Do **not** instead add `@babel/plugin-transform-class-properties` to the top-level `plugins` array. Top-level plugins run *before* the preset strips TypeScript types, so type-only declarations like `payment_hash: string;` are treated as real class fields and emit `this.payment_hash = void 0` after `super()`. That silently wipes everything `BaseModel`'s constructor assigned, and produces a different and considerably worse set of 84 failures. This was confirmed by diffing `babel.transformSync` output between the two configs.

### The actual fix

Mark type-only class fields with `declare`, so they emit nothing regardless of transform profile:

```ts
declare payment_hash: string;
```

The scale is why it is not in the bump: roughly 900 to 1150 fields across `models/` and `stores/`, depending on how strictly you match them. For contrast there are ~541 `@observable` occurrences in those two directories.

### To do

- [ ] Migrate type-only class fields in `models/` and `stores/` to `declare`
- [ ] Remove `unstable_transformProfile` from `babel.config.js`
- [ ] Confirm the suite stays green, and specifically that `BaseModel` subclasses still hydrate (the failure mode above is silent at runtime, not a crash)
- [ ] Sanity check bundle size and startup with the Hermes profile restored, so we know the migration bought something

Issue #4557, opened 2026-09-03
Stealth: NotepadApp status bar may be illegible on pre-Android-15 after the RN 0.87 StatusBar prop removal


kaloudis:
Split out of #4546 (React Native 0.86.0 -> 0.87.1).

RN 0.87 removed the `backgroundColor`, `translucent` and `networkActivityIndicatorVisible` props from `StatusBar`. The four `views/Stealth/*` decoy screens were the only places in the app setting `backgroundColor`, so #4546 drops it, matching `App.tsx`, which is already `barStyle`-only.

Three of the four are unaffected: `CalculatorApp`, `VPNApp` and `QRScannerApp` are all `barStyle="light-content"` over dark decoys.

`views/Stealth/NotepadApp.tsx:170` is the exception. It is `barStyle="dark-content"` over a light notepad decoy, and it previously pinned the bar to a matching light yellow:

```diff
-<StatusBar barStyle="dark-content" backgroundColor="#FFF9C4" />
+<StatusBar barStyle="dark-content" />
```

With no way to set the background, on pre-Android-15 (where the bar is not forced transparent) the bar can fall back to the theme's dark background, leaving dark icons on a dark bar. Since this is a decoy screen whose entire purpose is to look unremarkable, a status bar that looks broken is worth more than a cosmetic nit.

It was left alone in the bump rather than silently restyling a decoy screen as a side effect of a version bump.

### To do

- [ ] Check `NotepadApp` on a pre-Android-15 device to confirm the icons are actually illegible rather than just theoretically so
- [ ] Decide the fix: restyle the decoy to a dark header, drive the bar from the decoy's own surface colour some other way, or let it resolve as part of #4556

Note that #4556 (edge-to-edge) likely changes the answer here, since a transparent bar over the decoy's own background would make this correct by construction. Worth sequencing after it.

Issue #4556, opened 2026-09-03
Android: decide on edgeToEdgeEnabled=true as its own UI migration


kaloudis:
Split out of #4546 (React Native 0.86.0 -> 0.87.1).

The RN 0.87 template flips `edgeToEdgeEnabled` from `false` to `true` in `android/gradle.properties`. #4546 deliberately leaves it at `false` (`android/gradle.properties:47`).

Leaving it false is still a supported configuration rather than a deprecated path: the flag is read by the 0.87 Gradle plugin (`EDGE_TO_EDGE_ENABLED` in `PropertyUtils.kt`, alongside the newer scoped `react.edgeToEdgeEnabled`). But flipping it draws content under the system bars app-wide, which is an inset and padding migration across every screen. That is its own piece of UI work, not part of a version bump, so it was kept out.

Worth scheduling rather than deferring indefinitely, since Zeus is on `targetSdkVersion = 36` and Android 15 enforces edge-to-edge for apps targeting API 35 and above. The per-app opt-out is itself on a deprecation path, so the current appearance is more likely being produced by the compatibility path than by our own layout choices.

Related prior work: #2908, the Play Console deprecation warnings for `setDecorFitsSystemWindows` / `setStatusBarColor` / `setNavigationBarColor`, closed as completed.

### To do

- [ ] Audit safe-area and inset handling across screens
- [ ] Flip the flag and test on Android 15+ and on pre-15
- [ ] Pay particular attention to the POS and landscape layouts, and to the `views/Stealth/*` decoy screens

kaloudis:
Update: the flag flip itself has moved into #4546 (commit 7ff02becc) rather than waiting for this issue.

Review on #4546 surfaced a status-bar-height gap above every header on Android 14 and below. Root cause: RN 0.87 removed the `StatusBar` `translucent`/`backgroundColor` props, and `@rneui/base`'s Header had been rendering `<StatusBar translucent backgroundColor="transparent" />` on every screen, which is what put the window into draw-under-the-status-bar mode on pre-15 devices. With those props gone there is no supported way to reproduce the old appearance on 0.87, so "leave the flag false" stopped being a viable holding position: the choice was between edge-to-edge and a doubled top inset.

Since Android 15+ already enforces edge-to-edge for apps targeting API 35+, the flip mostly extends what Android 15/16 users have been seeing all along to older devices. Verified before/after on an API 34 emulator: with the flag on, headers sit directly below the status bar and match Android 15 rendering.

[...]

PR #4566, opened 2026-09-04
[feat] chore(deps): bump react-native to 0.87.1

kaloudis:
Bumps React Native from 0.86.0 to 0.87.1.

> [!NOTE]
> Rebased on master now that #4533 has merged, so the cherry-picked duplicate has dropped and all commits here are the bump itself. #4533 was a hard prerequisite rather than a nicety: RN 0.87 removes `InteractionManager` from core.

## Template changes

Everything in the [upgrade-helper diff](https://react-native-community.github.io/upgrade-helper/?from=0.86.0&to=0.87.1) is applied: Gradle wrapper 9.3.1 -> 9.4.1, `buildToolsVersion`/`compileSdkVersion` 36 -> 37, Kotlin 2.1.20 -> 2.2.0, `proguard-android-optimize.txt`, and the AGP 9 opt-outs.

Two deliberate deviations:

- **`edgeToEdgeEnabled` is flipped to `true`**, matching the template. This bump originally kept it `false` to avoid an app-wide UI migration, but review surfaced a status-bar-height gap above every header on Android 14 and below: RN 0.87 removed the `StatusBar` `translucent`/`backgroundColor` props that `@rneui`'s Header was using to draw under the status bar, so the old appearance cannot be reproduced on 0.87. Android 15+ already enforces edge-to-edge with our targetSdk 36, so the flip brings pre-15 devices in line with what Android 15/16 users already see. Verified before/after on an API 34 emulator.
- **TypeScript stays on 5.7.3** although the template moves to `^6.0.3`. Zeus does not extend `@react-native/typescript-config`, and `tsc` is clean on 5.7.3. Tracked in #4559.

## Breakages fixed here

None of these are in the template diff.

1. **MobX decorators (`babel.config.js`).** 0.87 changed the preset's default transform profile from `default` to `hermes-stable`. Hermes profiles set `preserveClasses`, which stops the preset shipping `@babel/plugin-transform-class-properties`. Legacy decorators need it, so every `@observable` property threw `Decorating class property failed` (116 test failures). Pinning the profile is deliberate: adding the plugin to `babel.config.js` instead makes it run *before* the preset strips TS types, so type-only declarations like `payment_hash: string;` become `this.payment_hash = void 0` after `super()` and silently wipe everything `BaseModel`'s constructor assigned.
2. **Type resolution (`tsconfig.json`).** 0.87 dropped react-native's top-level `types` field; types now resolve only via `exports`, which `moduleResolution: "node"` cannot read. This surfaced as TS7016 everywhere plus cascading nonsense in `@rneui` (`Property 'onPress' does not exist on ListItemProps`).
3. **Asset registry (`metro.config.js`).** `@react-native-vector-icons/common` and `react-native-svg` (`lib/*/lib/resolveAssetUri.js`) import `@react-native/assets-registry/registry`, which 0.87 removed from RN's dep tree. Aliased to `react-native/asset-registry` rather than installing the standalone package, which still exists but would register into a second Map so `getAssetByID` returns `undefined`. The alias is metro-only, so a jest `moduleNameMapper` entry points the same specifier at the same file (`react-native/src/asset-registry.js`), keeping the two resolvers in sync.
4. **AGP 9 proguard (`patches/`).** AGP 9 hard-errors at configuration time on `getDefaultProguardFile('proguard-android.txt')`. `react-native-notifications` 5.2.2 (latest) still calls it. Not fixable from the root `subprojects` block, since the error fires before `afterEvaluate`.
5. **AGP 9 `BuildConfig` (`android/build.gradle`).** No longer generated unless a module opts in, and the global `android.defaults.buildfeatures.buildconfig` property is gone. Re-enabled for all Android subprojects instead of per package.
6. **`ReactContext.getRuntimeExecutor()` (`WorkerReactContext.java`).** Newly abstract in 0.87.
7. **`StatusBar backgroundColor` (`views/Stealth/`).** Prop removed in 0.87.

`engines` and the README Node prerequisite now mirror RN's range, `^22.13.0 || ^24.3.0 || >= 26.0.0`. The previous `>= 22.13.0` admitted 23.x and 25.x, which React Native rejects.

## Testing

- `yarn verify` green: 70 suites, 1526 tests, plus tsc, prettier and lint
- `npx react-native bundle` succeeds for both android and ios with `--dev false`
- `./gradlew :app:assembleDebug` **BUILD SUCCESSFUL**, APKs produced for all ABIs
- `ios/Podfile.lock` regenerated: 129 dependencies, 130 pods, hermes-engine 250829098.0.17

## Still outstanding

- [ ] Device testing on both platforms, all backends
- [ ] **A release build specifically.** Zeus sets `enableProguardInReleaseBuilds = true`, so the switch to `proguard-android-optimize.txt` enables R8 optimization passes on a reflection-heavy app (protobuf, gRPC, LND JNI). Only a debug build has been exercised so far. R8 keep rules for all three JNA-backed uniffi packages (CDK, LDK Node, zeus_cashu_restore) are now in `proguard-rules.pro`.
- [ ] Reproducible-build check via `./build.sh`

## Follow-ups, not in this PR

All four are filed, so nothing here is lost when this merges:

- #4558: migrate the type-only class fields in `models/` and `stores/` to `declare` so the `hermes-stable` profile and its Hermes V1 optimisations can be restored.
- #4559: decide on TypeScript 6, per the deviation noted above.
- #4556: remaining edge-to-edge audit (POS, landscape, `views/Stealth/*` decoys, pre-15 devices); the flag itself is now flipped in this PR, see the deviation note above.
- #4557: `views/Stealth/NotepadApp.tsx` is `dark-content` on a light decoy. With `backgroundColor` gone, on pre-Android-15 the bar reverts to theme black and the icons may be hard to read. Left as-is rather than silently restyling a decoy screen.


kaloudis:
Pushed 5e2d76dc1 to fix the Xcode Cloud iOS build failure (non-modular header errors in module React from LncModule.h / RCTCallInvoker.h).

Root cause: RN 0.87 activates a framework module map on the prebuilt React.xcframework (0.86's prebuilt shipped without one), and our Objective-C++ app-target code (LncMobile, AppDelegate.mm via the Fabric renderer headers) cannot compile against it. This went unnoticed locally because CocoaPods kept the cached 0.86.0 React-Core-prebuilt podspec, so local builds were silently mixing the 0.86 prebuilt binary with 0.87.1 pods; the committed lockfile still said React-Core-prebuilt (0.86.0). Fresh installs on Xcode Cloud resolved the real 0.87.1 artifact and broke.

Fix, kept minimal:
- Pin `RCT_USE_PREBUILT_RNCORE=0` in the Podfile so every pod install (local and CI) builds React core from source with the textual header semantics we shipped through 0.86, and regenerate the lockfile (React-Core-prebuilt is gone from it).
- One pre-include of `ReactCommon/CallInvoker.h` in AppDelegate.h: the 0.87 source layout modularizes the ReactCommon pod but its umbrella does not cover CallInvoker.h, leaving `NativeMethodCallInvoker` invisible at its use site in RCTTurboModule.h.

I tried the prebuilt-friendly route first (`CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES=YES`); it fixes LncModule but then AppDelegate.mm fails on module/textual header visibility conflicts deeper in the renderer headers, so source builds are the safer path until upstream supports Objective-C++ app code against the prebuilt core. Cost is longer iOS build times.

Verified with a clean Debug simulator build of the zeus scheme (BUILD SUCCEEDED, arm64 + x86_64). Release/archive on Xcode Cloud still needs a green run to confirm.

ajaysehwal (commented):
@kaloudis, we’re getting this warning during the build:

`Attempted to import ... ReactNativeFeatureFlags which is not listed in the "exports" of react-native ... Falling back to file-based resolution.`

It looks like something is importing `ReactNativeFeatureFlags` through an internal React Native path that isn’t exposed in the package `exports`.

Could we check which dependency/call site is importing this and update it to use a supported API/path?

kaloudis:
@ajaysehwal Tracked it down: there's no dependency or call site on our side to update. The importer is React Native's own `@react-native/virtualized-lists@0.87.1` (version-locked by `react-native` itself), which deep-imports `react-native/src/private/featureflags/ReactNativeFeatureFlags` at `VirtualizedList.js:67` and `VirtualizeUtils.js:16`. RN 0.86 exported `./src/*`; react/react-native#57484 removed that mapping for 0.87.0 while the imports stayed behind, so Metro warns and then falls back to file-based resolution, which resolves the real file.

The warning is cosmetic by construction: VirtualizedList backs every FlatList in the app, so if the fallback ever failed, no list would render anywhere.

Upstream tracks this as react/react-native#57933, closed 2026-09-08 by commit react/react-native@a506ed66cc5744ec017556f49a2c296fb4bad318, which re-routes the import through the exported `react-native/react-private-interface` entry point. The fix landed after 0.88.0-rc.0 was cut, so it ships in a later 0.88 release; there is no 0.87.x pick as of today.

Given that, I'd rather not alias `src/private/*` in `metro.config.js` just to silence it. The one resolver alias this PR carries (asset registry) exists because resolution actually failed there; this one resolves fine and the warning goes away on our next RN bump.

patches/patch-react-native-notifications.mjs
```diff
@@ -158,4 +158,24 @@ public class FcmToken implements IFcmToken {
             console.log('  - Patched react-native.config.js for RN 0.84 compatibility');
         }
     }
+
+    // Fix the release buildType for AGP 9 (RN 0.87+)
```
ajaysehwal: @kaloudis, one issue with the new AGP 9 ProGuard patch: it sits after the early return for a missing `FcmToken.java`, so the Gradle patch is skipped whenever that unrelated file isn’t present.

Also, if a future `react-native-notifications` release changes the `getDefaultProguardFile` line (e.g. double quotes or a small formatting change), the exact-string replacement silently does nothing. The Android build then fails later during AGP 9 configuration without pointing back to this patch.

This differs from the fail-loudly behavior of the Reanimated patch. Could we move the Gradle patch before the `FcmToken.java` early return and throw/log when the expected anchor isn’t found?

kaloudis: fixed in 068105b56

I left the config.js step non-throwing because its patched output can't be distinguished from a future upstream change. For context, react-native-notifications is pinned exactly at 5.2.2, so the skip could only happen on a version bump.

android/build.gradle
```diff
@@ -1,11 +1,11 @@
 buildscript {
     ext {
-        buildToolsVersion = "36.0.0"
+        buildToolsVersion = "37.0.0"
```
ajaysehwal, line 3: One concern: The PR updates buildToolsVersion/compileSdkVersion to 37 and moves to Kotlin 2.2.0, AGP 9, and Gradle 9.4.1, but build.sh and docs/ReproducibleBuilds.md are unchanged. The Docker image in build.sh is pinned by digest, so I’m not sure whether that image includes Android SDK/platform 37 and build-tools 37.0.0.

If those components aren’t available in the pinned image, ./build.sh could either fail or cause AGP to download them at build time. The latter would undermine the digest-pinned, reproducible-build setup.

Since the reproducible-build checklist item is currently unchecked and the Android workflow is workflow_dispatch-only, this may not be caught by PR CI.

Could you verify this by running ./build.sh --no-tty against the PR head? If the image doesn’t contain the required SDK/build-tools, I think we should update the pinned image digest and the reproducible-build docs in this PR.

kaloudis: Confirmed. The pinned 18.0 image has platform 36 and build-tools 36.0.0, so AGP was downloading "Android SDK Platform 37.0 (revision 2)" and Build-Tools 37.0.0 at build time ([run 36042387524](https://github.com/ZeusLN/zeus/actions/runs/36042387524), 068105b56).

Fixed in 43d94b38f: `build.sh` now pins `react-native-android` 21.1 (`sha256:d4b8ea07…`), which includes both. On that head ([run 36058508505](https://github.com/ZeusLN/zeus/actions/runs/36058508505)) the two downloads are gone. NDK 28.0.13004108 and CMake 3.22.1 are still fetched at build time. That was already the case on master, since the image ships NDK 27.1, so I've left it out of this PR.

The four APK hashes are identical across both runs, so the image change does not affect the output:
```
325cdc4a295bad1ba7b6c13021c722be766c6a1c0f03e453a9b4b15fd07a0ed3  zeus-arm64-v8a.apk
6b773bdbc896f871cebea4a8b2e89c80c348c4b8d8864f69535cb17cb02e0bd2  zeus-armeabi-v7a.apk
1893d85ee67838c39e830dafc42229856048c3a8759b49ea85a22a56e08de7a7  zeus-universal.apk
3ddadda99c8a6bd1fc9b3f21ad59a486b062d3afdefac808ea0ee68ed828a702  zeus-x86_64.apk
```
`docs/ReproducibleBuilds.md` doesn't name the image version, so it needed no change.

PR #4572, opened 2026-09-05
tests: add BIP39 wordlist invariants for Bip39Utils

kunalKumar-13:
### What

Adds tests for `BIP39_WORD_LIST` in `utils/Bip39Utils.ts`. The file is 2052 lines and had no test coverage.

It is wallet-critical data — it builds and restores seed phrases — so a single altered, missing or misordered word would produce mnemonics that other wallets reject, or silently break restores of existing wallets. Nothing currently guards it.

CONTRIBUTING suggests improving test coverage as a first contribution, so I started there.

### Approach

Asserts the properties BIP39 requires rather than pinning all 2048 words, so a legitimate future edit still fails loudly while the test stays readable:

- exactly 2048 entries (2^11 — each word encodes 11 bits, so any other length makes the encoding wrong rather than merely unusual)
- sorted alphabetically
- no duplicates
- **unique first four characters** — the spec guarantees this so a phrase stays unambiguous when words are truncated to four characters, which hardware wallets rely on for entry
- lowercase `a-z` only, no whitespace or accents
- word lengths between 3 and 8
- first word `abandon`, last word `zoo`

I also checked the list against [`bitcoin/bips/bip-0039/english.txt`](https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt): it is identical, so this locks in a list that is already correct rather than fixing anything.

### Verifying the tests actually catch things

A passing test on correct data proves little, so I mutated the wordlist four ways and confirmed each is caught:

| mutation | tests that failed |
|---|---|
| `'zoo'` → `'zoO'` | sorted, lowercase-only, first/last |
| deleted `'ability'` | length 2048 |
| `'absurd'` → `'abandoned'` | sorted, 4-char prefixes, word length |
| `'acid'` → `'able'` (duplicate) | sorted, no duplicates, 4-char prefixes |

The wordlist was restored afterwards and re-verified byte-for-byte against the canonical BIP39 list.

### Testing

`yarn jest` — **71 suites, 1547 tests passing**, including the 7 added here. `eslint` and `prettier --check` clean on the new file.

### On the proof-of-work requirement

I have read the first-time contributor rule about screenshot or video evidence. This is a test-only change — no feature and no UI — so there is nothing to screenshot; the mutation table above is the equivalent evidence that it was actually run rather than assumed. Happy to provide anything else that would help.

### AI

I used Claude Code for this change and this description. The test counts, the mutation results and the comparison against the canonical wordlist are all from running the code.

### DIFF
--- a/utils/Bip39Utils.test.ts
+++ b/utils/Bip39Utils.test.ts
@@ -1,0 +1,59 @@
+import { BIP39_WORD_LIST } from '.././utils/Bip39Utils';
+
+/**
+ * The BIP39 English wordlist is wallet-critical data: it is used to build and
+ * to restore seed phrases. A single altered, missing or misordered word would
+ * produce mnemonics that other wallets reject, or silently break restores of
+ * existing wallets, and nothing in the codebase currently guards it.
+ *
+ * These tests assert the properties BIP39 requires of the list, rather than
+ * pinning all 2048 words, so a legitimate edit still fails loudly while the
+ * test stays readable.
+ *
+ * Reference: https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md
+ */
+describe('BIP39_WORD_LIST', () => {
+    it('contains exactly 2048 words', () => {
+        // 2048 = 2^11, so each word encodes exactly 11 bits of entropy. Any
+        // other length makes the mnemonic encoding wrong, not merely unusual.
+        expect(BIP39_WORD_LIST).toHaveLength(2048);
+    });
+
+    it('is sorted alphabetically', () => {
+        // BIP39 requires this so wallets can binary-search the list, and it is
+        // what makes a word's index — and therefore the entropy it encodes —
+        // well defined.
+        expect(BIP39_WORD_LIST).toEqual([...BIP39_WORD_LIST].sort());
+    });
+
+    it('has no duplicate words', () => {
+        expect(new Set(BIP39_WORD_LIST).size).toBe(BIP39_WORD_LIST.length);
+    });
+
+    it('has a unique first four letters for every word', () => {
+        // The spec guarantees this so a phrase stays unambiguous when words are
+        // truncated to four characters, which hardware wallets rely on for
+        // entry. A duplicate prefix would make two different seeds collide.
+        const prefixes = BIP39_WORD_LIST.map((word) => word.slice(0, 4));
+        expect(new Set(prefixes).size).toBe(prefixes.length);
+    });
+
+    it('contains only lowercase a-z, with no whitespace or accents', () => {
+        const offenders = BIP39_WORD_LIST.filter(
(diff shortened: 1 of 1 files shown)


shubhamkmr04 (commented):
One gap worth closing before merge: the tests don't catch the failure the commit message is about. A single altered word passes all seven, so long as it keeps its sort position, stays lowercase, is 3–8 characters, and doesn't collide on a 4-character prefix.

I ran your test file against mutated copies of the wordlist:

| mutation | your 7 tests |
|---|---|
| `absurd` → `absurdly` | all pass |
| `kingdom` → `kingdoms` | all pass |
| `giraffe` → `giraffs` | all pass |
| `zoo` → `zoO` (control) | 3 fail |
| delete `ability` (control) | 1 fail |

The mutations in your table all break ordering, case, or length, so they exercise the structural checks — but not word *identity*. Substitution is the case that would silently break restores.

One test closes it:

```ts
import { createHash } from 'crypto';

it('matches the canonical BIP39 English wordlist', () => {
    // sha256 of bitcoin/bips bip-0039/english.txt. Pins word identity, which
    // the structural checks cannot: a same-shape substitution passes all of them.
    const digest = createHash('sha256')
        .update(BIP39_WORD_LIST.join('\n') + '\n')
        .digest('hex');
    expect(digest).toBe(
        '2f5eed53a4727b4bf8880d8f3f199efc90e58503646d9ff8eff3a2ed3b24dbda'
    );
});
```

I checked that digest against the upstream `english.txt`, not the copy in this repo — they match word for word. The test passes as written.

Worth keeping the other seven: they turn a checksum failure into a readable diagnosis.

kunalKumar-13:
You're right, and thank you — that is a real gap and I should have caught it.

I confirmed your point before fixing it. With only the original seven tests:

```
absurd -> absurdly    7 passed, 0 failed
```

A corrupted wordlist passes cleanly. My mutation table exercised ordering, case and length, so every mutation in it broke *structure* — none tested word **identity**, which is the one that would silently break restores. The table looked thorough and was not.

Added your test in ba2ffeda. I verified the digest independently against upstream `english.txt` rather than taking it from the review — `shasum -a 256` on the raw file and a `join('\n') + '\n'` in Python both give `2f5eed53…`, so it pins the list as published.

Re-running your three substitutions against the new set:

| mutation | result |
|---|---|
| `absurd` → `absurdly` | 1 failed — checksum |
| `kingdom` → `kingdoms` | 1 failed — checksum |
| `giraffe` → `giraffs` | 1 failed — checksum |

and the wordlist is restored and re-verified as canonical afterwards.

Kept the other seven for the reason you gave: a bare checksum failure says "the list changed", while these say *how*. 71 suites / 1548 tests passing, prettier and eslint clean.

shubhamkmr04 (approved):
tACK

kaloudis (approved):

PR #4576, opened 2026-09-06
fix(security): bind LNURL-auth callback host to signing domain

kaloudis:
# Description

LnurlAuth displayed and derived its service-specific linking key from js-lnurl's `getDomain` while independently re-parsing the callback with `url.parse` for transmission. The two parsers disagree on URLs containing backslashes: `getDomain` splits on the last `@` but not on `\`, while `url.parse` (like WHATWG parsers and the native HTTP stacks that ultimately dial the socket) treats `\` as a path delimiter.

A crafted `https://attacker.example\@victim.example/?tag=login&k1=...` therefore showed **victim.example** to the user, derived victim.example's linking key, and sent the resulting signature to **attacker.example** — handing the attacker a valid login response for the victim's account on the target service. Reproduced offline against the installed parsers.

Changes:
- New `verifyLnurlAuthCallback` in `LnurlPayUtils`: rejects backslashes and userinfo outright, applies the same https-only / private-address callback policy LNURL-pay already enforces (`isLnurlCallbackAllowed`), and requires the parsed callback host to equal the displayed/derived domain (case-insensitive)
- Enforced at view construction (invalid-params alert) **and** again immediately before the signature leaves the device
- Key derivation is skipped entirely when constructor validation fails (previously `componentDidMount` would still sign)
- Regression tests including the parser-confusion PoC URL

`keyauth://` URLs are unaffected: `decodelnurl` rewrites them to `https://` (or `http://` for .onion) before they reach this view.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

`utils/LnurlPayUtils.test.ts` gains 9 cases for `verifyLnurlAuthCallback`: matching host, case-insensitivity, onion-over-http, the backslash PoC, host mismatch, userinfo, cleartext http, empty domain, missing callback.

- [ ] Android
- [ ] iOS

Device tests pending (will update): normal LNURL-auth login against a real service on both platforms; crafted backslash URL rejected with invalid-params alert.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

The validation is backend-independent (it runs before `BackendUtils.lnurlAuth`); a login smoke test on one lnurlAuth-capable backend plus lndhub's auth-mode path should suffice.

### Locales

### Other:

### DIFF
--- a/utils/LnurlPayUtils.ts
+++ b/utils/LnurlPayUtils.ts
@@ -256,3 +256,47 @@ export const isLnurlCallbackAllowed = (callback: string): LnurlCheckResult => {
 
     return { ok: true };
 };
+
+/**
+ * Bind an LNURL-auth callback to the domain that is displayed to the user
+ * and used to derive the service-specific linking key. js-lnurl's getDomain
+ * and the url parser used for transmission disagree on URLs containing
+ * backslashes or userinfo: getDomain splits on the last '@' but not on '\',
+ * while url.parse (like WHATWG parsers and native HTTP stacks) treats '\'
+ * as a path delimiter. A crafted
+ * "https://attacker.example\@victim.example/?tag=login&k1=..." therefore
+ * displays and signs for victim.example while transmitting the signature to
+ * attacker.example, handing the attacker a valid login for the victim's
+ * account. Reject the ambiguous constructs outright and require the host
+ * that will actually be contacted to equal the displayed domain.
+ */
+export const verifyLnurlAuthCallback = (
+    callback: string,
+    expectedDomain: string
+): LnurlCheckResult => {
+    if (typeof callback === 'string' && callback.includes('\\')) {
+        return { ok: false, reason: 'callback URL contains a backslash' };
+    }
+
+    const allowed = isLnurlCallbackAllowed(callback);
+    if (!allowed.ok) return allowed;
+
+    const parsed = url.parse(callback);
+    if (parsed.auth) {
+        return { ok: false, reason: 'callback URL contains userinfo' };
+    }
+
+    const hostname = (parsed.hostname || '').toLowerCase();
+    if (
+        typeof expectedDomain !== 'string' ||
+        expectedDomain.length === 0 ||
+        hostname !== expectedDomain.toLowerCase()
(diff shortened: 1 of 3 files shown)


utils/LnurlPayUtils.ts
```diff
@@ -256,3 +256,47 @@ export const isLnurlCallbackAllowed = (callback: string): LnurlCheckResult => {
...
+ * attacker.example, handing the attacker a valid login for the victim's
+ * account. Reject the ambiguous constructs outright and require the host
+ * that will actually be contacted to equal the displayed domain.
+ */
+export const verifyLnurlAuthCallback = (
+    callback: string,
+    expectedDomain: string
+): LnurlCheckResult => {
+    if (typeof callback === 'string' && callback.includes('\\')) {
+        return { ok: false, reason: 'callback URL contains a backslash' };
```
ajaysehwal, line 283: nit: localize these error messages

kaloudis: Good catch, fixed in ea550e7a8.

The `reason` strings stay as developer-facing diagnostics (the tests assert on them, and `LnurlCheckResult.reason` is now commented as such), but nothing untranslated reaches the user anymore. Both surfaces now show a new localized key, following the `views.LnurlPay.LnurlPay.unsafeCallback` precedent:

- constructor rejection: the invalid-params alert body
- pre-send rejection in `sendValues`: `errorMsgAuth`

`views.LnurlAuth.unsafeCallback`: "This service asked the wallet to send your login signature to an address that doesn't match the domain you're logging in to. Login cancelled to protect your account."

kaloudis: Added a comment right above these lines in a317ff27a explaining why they stay untranslated, so the next reader does not have to re-derive it.

ajaysehwal (approved):
tACK

PR #4577, opened 2026-09-06
fix(security): require confirmation before LNURL-channel peer connect

kaloudis:
# Description

`LnurlChannel` called `connectPeer` in `componentDidMount`, so scanning an attacker-controlled channel-request QR immediately opened an outbound Lightning transport connection to a service-chosen host — disclosing the node's identity pubkey and network address (IP, unless the node is Tor-configured) before any visible confirmation. On ldk-node the peer was additionally persisted (`persist` defaults on when `perm` is omitted) and reconnected across restarts.

Changes:
- The peer connection now happens when the user presses the Connect button; on success the channel request proceeds in the same press (already-connected peers short-circuit as before via the "already connected" error path)
- `perm: false` is passed so the peer is not persisted before a channel actually exists (LDK Node maps `perm` to `persist`; Embedded LND forwards it; remote LND already defaults to non-persistent; CLNRest ignores it)
- The peer error message initializes empty instead of a stale hardcoded 'Error' string that previously flashed before the mount-time connect resolved

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

Device tests pending (will update): scan an lnurl-channel QR and verify no peer connection occurs until Connect is pressed (node peer list / service logs); full happy-path channel open via an LSP offering lnurl-channel; ldk-node restart to confirm the peer is not persisted after backing out.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Reaches `connectPeer` on peer-capable backends: LDK Node (persist behavior), Embedded LND, LND REST, LNC, CLNRest. NWC/LndHub don't support lnurl-channel.

### Locales

### Other:

### DIFF
--- a/views/LnurlChannel.tsx
+++ b/views/LnurlChannel.tsx
@@ -93,11 +93,11 @@ export default class LnurlChannel extends React.Component<
             connectingToPeer: false,
             peerSuccess: false,
             lnurlChannelSuccess: false,
-            errorMsgPeer: 'Error'
+            errorMsgPeer: ''
         };
     }
 
-    triggerConnect() {
+    triggerConnect(onSuccess?: () => void) {
         const { node_pubkey_string, host } = this.state;
         this.setState({ connectingToPeer: true });
 
@@ -105,11 +105,14 @@ export default class LnurlChannel extends React.Component<
             addr: {
                 pubkey: node_pubkey_string,
                 host
-            }
+            },
+            // don't persist a peer the user hasn't opened a channel with
+            perm: false
         })
             .then(() => {
                 this.setState({ connectingToPeer: false });
                 this.setState({ peerSuccess: true });
+                if (onSuccess) onSuccess();
             })
             .catch((error: any) => {
                 // handle error
@@ -121,11 +124,9 @@ export default class LnurlChannel extends React.Component<
                     this.state.errorMsgPeer.includes('already')
                 ) {
                     this.setState({ peerSuccess: true });
+                    if (onSuccess) onSuccess();
                 }
             });
-    }
-    componentDidMount() {
-        this.triggerConnect();
     }
 
     sendValues() {
@@ -276,9 +277,17 @@ export default class LnurlChannel extends React.Component<
                                 color: themeColor('background')
                             }}
                             onPress={() => {
-                                this.sendValues();
+                                // only contact the service-supplied peer
+                                // after explicit user confirmation
+                                if (this.state.peerSuccess) {
+                                    this.sendValues();
+                                } else {
+                                    this.triggerConnect(() =>
+                                        this.sendValues()
+                                    );
+                                }
                             }}
-                            disabled={!peerSuccess}
+                            disabled={this.state.connectingToPeer}
                         />
                     </View>
 


views/LnurlChannel.tsx
```diff
@@ -93,23 +93,26 @@ export default class LnurlChannel extends React.Component<
...
+            },
+            // don't persist a peer the user hasn't opened a channel with
+            perm: false
         })
             .then(() => {
                 this.setState({ connectingToPeer: false });
                 this.setState({ peerSuccess: true });
+                if (onSuccess) onSuccess();
             })
             .catch((error: any) => {
```
ajaysehwal, line 135: @kaloudis, I think there’s a bug in the “peer already connected” recovery branch. The code calls:

```ts
this.setState({ errorMsgPeer: error.toString() });
```

and then immediately checks `this.state.errorMsgPeer.includes('already')`. With React 18+ batching (Zeus is on React 19), `this.state` can still contain the previous value at that point, so the check won’t reliably see the error just set.

That means when `connectPeer` returns an “already connected” error, `peerSuccess` may never be set and the new `onSuccess()` path won’t fire, leaving the user stuck retrying the same connection.

Could we check `error.toString()` directly (or use the `setState` callback) instead of reading `this.state.errorMsgPeer` immediately after the update?

views/LnurlChannel.tsx
```diff
@@ -276,9 +277,17 @@ export default class LnurlChannel extends React.Component<
...
+                                if (this.state.peerSuccess) {
+                                    this.sendValues();
+                                } else {
+                                    this.triggerConnect(() =>
+                                        this.sendValues()
+                                    );
+                                }
                             }}
-                            disabled={!peerSuccess}
+                            disabled={this.state.connectingToPeer}
```
ajaysehwal: @kaloudis, one small race on the Connect button: it’s only guarded by `connectingToPeer`, so two rapid taps before the first `setState({ connectingToPeer: true })` commits can both pass the check and trigger concurrent `connectPeer` calls.

This isn’t a fund-safety issue since peer connection is effectively idempotent, but the duplicate requests could race to update `peerSuccess` / `errorMsgPeer`.

Could we add an imperative guard (e.g. a ref) or otherwise ensure `triggerConnect()` can’t start while another connection attempt is already in flight?

views/LnurlChannel.tsx
kaloudis: Good catch, confirmed and fixed in 9a19ddaa3.

You are right that the read is unsafe: with React 19 and automatic batching in the new architecture, `this.state` still holds the old value inside that promise callback. The practical symptom was slightly different from "stuck retrying": the first press showed the error and did nothing, and the *second* press worked, because by then `errorMsgPeer` happened to hold the previous "already connected" string. Either way it broke the most common case, which is scanning a second channel request from an LSP you are already connected to.

`triggerConnect()` now branches on the error itself and resolves a boolean instead of taking an `onSuccess` callback:

```ts
.catch((error: any) => {
    const errorMsgPeer = error.toString();
    this.setState({ errorMsgPeer });

    // we're already connected to the peer, so we can proceed
    if (errorMsgPeer.includes('already')) {
        this.setState({ peerSuccess: true });
        return true;
    }

    return false;
});
```

views/LnurlChannel.tsx
kaloudis: Agreed, guarded in 9a19ddaa3. Digging into it, the window was wider than two taps inside one batch:

1. `connectingToPeer` only covered the connect phase, so the button stayed live for the entire `sendValues()` round trip to the service, which is seconds over Tor.
2. Worse, my `disabled={connectingToPeer}` dropped a guard master had. Master used `disabled={!peerSuccess}`, and `sendValues()` sets `peerSuccess: false` on both of its branches, so the button re-disabled after a submission. With my version it stayed enabled after `lnurlChannelSuccess`, so further presses could replay the callback with the same `k1`.

The fix covers all three variants:

- One `loading` state flag (replacing `connectingToPeer`) spans connect plus callback, and also drives the `LightningIndicator`, so there is now feedback during the callback fetch too.
- `disabled={loading || lnurlChannelSuccess}`, restoring the post-success lock.
- An instance field `requestInFlight` as the synchronous authority, for exactly the case you describe where the press lands before the state update commits.

```ts
private handleConnectPress = async () => {
    if (this.requestInFlight) return;
    this.requestInFlight = true;
    this.setState({ loading: true, errorMsgPeer: '' });

    try {
        const connected =
            this.state.peerSuccess || (await this.triggerConnect());
        if (connected) await this.sendValues();
    } finally {
        this.requestInFlight = false;
        this.setState({ loading: false });
    }
};
```

Stale peer errors are also cleared when a new attempt starts. Device tests still owed; I will squash before merge.

views/LnurlChannel.tsx
```diff
@@ -90,43 +94,60 @@ export default class LnurlChannel extends React.Component<
...
 
-        BackendUtils.connectPeer({
+        return BackendUtils.connectPeer({
             addr: {
                 pubkey: node_pubkey_string,
                 host
-            }
+            },
+            // don't persist a peer the user hasn't opened a channel with
+            perm: false
```
ajaysehwal: @kaloudis, the deferred connect looks good. One concern with `perm: false` on LDK Node (`persist: false`): for an inbound LNURL-channel connection, ldk-node 0.7.0 only persists the peer from its `ChannelPending` handler when the peer’s node announcement/address is present in the local graph.

With an unannounced LNURL-channel service, an RGS snapshot without its address, or before the graph has synced, the peer may not be persisted. After a restart, LDK then won’t reconnect to it, leaving the channel offline — especially problematic for a phone behind NAT where the LSP can’t dial back in.

Since the connection now only happens after the user explicitly presses Connect, could we persist the peer again, or re-persist it with `perm: true` once the callback succeeds? Also worth adding a device check for: open an LNURL-channel with LDK Node → restart → confirm the peer reconnects.

views/LnurlChannel.tsx
```diff
@@ -49,6 +49,10 @@ export default class LnurlChannel extends React.Component<
     LnurlChannelProps,
     LnurlChannelState
 > {
+    // synchronous guard - state updates are batched, so a second press can
+    // land before a `loading` update commits
+    private requestInFlight = false;
+
     constructor(props: LnurlChannelProps) {
```
ajaysehwal: nit: when processNodeUri fails in the constructor, the state has an empty pubkey and host, and Connect is now pressable (before, it stayed disabled). Adding || !node_pubkey_string to disabled would avoid sending an empty connect request to the backend.

views/LnurlChannel.tsx
```diff
@@ -90,43 +94,60 @@ export default class LnurlChannel extends React.Component<
...
+    private handleConnectPress = async () => {
+        if (this.requestInFlight) return;
+        this.requestInFlight = true;
+        this.setState({ loading: true, errorMsgPeer: '' });
+
+        try {
+            const connected =
+                this.state.peerSuccess || (await this.triggerConnect());
+            if (connected) await this.sendValues();
+        } finally {
```
shubhamkmr04: There's no `catch` here, so a throw from `triggerConnect` just becomes an unhandled rejection and the user sees nothing. It's easy to hit on NWC/LndHub: `connectPeer` doesn't exist there, so `BackendUtils.call` returns `false` and `.then` throws. `handleAnything` also routes `channelRequest` here without checking `supportsChannelManagement()`. Maybe gate it there and add a catch that sets `errorMsgPeer`?

views/LnurlChannel.tsx
```diff
@@ -90,43 +94,60 @@ export default class LnurlChannel extends React.Component<
...
-    triggerConnect() {
+    // only contact the service-supplied peer after explicit user confirmation
+    private handleConnectPress = async () => {
+        if (this.requestInFlight) return;
+        this.requestInFlight = true;
+        this.setState({ loading: true, errorMsgPeer: '' });
+
+        try {
+            const connected =
+                this.state.peerSuccess || (await this.triggerConnect());
```
shubhamkmr04: nit: `sendValues` resets `peerSuccess` to false on both success and error, so this short-circuit almost never kicks in. Every retry dials the peer again. Could just drop it.

views/LnurlChannel.tsx
```diff
@@ -90,43 +94,60 @@ export default class LnurlChannel extends React.Component<
...
+            if (connected) await this.sendValues();
```
shubhamkmr04: nit: if the user backs out while the connect is still pending (slow over Tor), `sendValues()` still fires once it resolves. Might be worth an unmounted check.

views/LnurlChannel.tsx
```diff
@@ -144,7 +165,7 @@ export default class LnurlChannel extends React.Component<
         u.search = querystring.stringify(qs);
         u.query = querystring.stringify(qs);
 
-        ReactNativeBlobUtil.fetch('get', url.format(u))
+        return ReactNativeBlobUtil.fetch('get', url.format(u))
```
shubhamkmr04: Small note on the description: the service still sees our IP at scan time, since `getlnurlParams` hits it over clearnet, and this callback also goes out over clearnet even when the node is on Tor. So what this PR fixes is linking the node pubkey to that IP before the user confirms, not the IP leak itself. Might be worth rewording.

views/LnurlChannel.tsx
kaloudis: Confirmed, thanks. In our fork (v0.7.0-zeus-rl-0.2.5) `ChannelPending` only persists an inbound channel peer when the graph has an announced address for it, and master had been persisting on scan, so `perm: false` alone regressed this.

Fixed in 41ea36acc: the first connect still uses `perm: false`, and once the service accepts the callback we call `connectPeer` again with `perm: true`. ldk-node `connect()` returns Ok for an already-connected peer and still writes it to the peer store. On LND the call returns "already connected" before `perm` is read, so it is ignored there (LND reconnects channel peers itself). Persisting only after acceptance keeps a refused request from leaving a peer that reconnects on every start.

Added your restart check to the device test list.

views/LnurlChannel.tsx
kaloudis: Fixed in 41ea36acc: `disabled` now includes `!node_pubkey_string`.

views/LnurlChannel.tsx
kaloudis: Confirmed: on NWC/LndHub `call()` returns `false` and `false.then` throws synchronously. Fixed both ways in 41ea36acc. `handleConnectPress` now has a `catch` that shows the error, and `handleAnything` only routes `channelRequest` when `supportsChannelManagement()` is true, with an alert otherwise (same pattern as `login`). Added a test for the unsupported case.

views/LnurlChannel.tsx
kaloudis: Agreed, `sendValues()` clears it on every path. Removed in 41ea36acc, so every press dials the peer, which is also what we want after a service error in case the peer dropped.

views/LnurlChannel.tsx
kaloudis: Added in 41ea36acc: an `unmounted` flag set in `componentWillUnmount`, checked after the connect resolves and before `sendValues()`.

views/LnurlChannel.tsx
kaloudis: Fair point. Both the params fetch and this callback go over clearnet, so the service sees the phone IP regardless. I reworded the description: the pre-confirmation leak is the node pubkey linked to the requester, plus the node's own address on remote backends (LND REST, LNC, CLNRest), where the connection comes from the node host and not the phone.

PR #4578, opened 2026-09-06
fix(security): dispose CDK handles before full-wipe database sweep

kaloudis:
# Description

`clearCDKDatabase` (used by `clearAllData`, including the duress-PIN and failed-authentication wipes) unlinked Cashu database files without first closing native CDK handles. POSIX unlink succeeds under an open handle, but the proof data stays reachable through the still-open connection until the app restarts, and the tracked db path could re-materialize files in a warm session.

This applies the same dispose-then-unlink ordering `deleteCashuData` already uses (#4376): `CashuDevKit.deleteWalletDatabase()` is called best-effort (guarded by `isAvailable()`) before the filename sweep. The prefix sweep remains the primary mechanism since it covers every wallet's hashed database plus WAL/SHM sidecars, not just the currently tracked one. Errors remain swallowed by design on the duress path (surfacing them would disclose the duress mechanism).

Round 3, per @shubhamkmr04's review: the dispose was not actually closing anything on Android. The uniffi bindings are `Disposable`, and `CashuDevKitModule.kt` only nulled the Kotlin references, so the Rust object and the SQLite connection it owns survived until the Cleaner next ran after a GC. That was bounded on the full-wipe path (every `clearAllData()` caller restarts in a `finally`, and Android's `restartApp()` is a ProcessPhoenix rebirth that closes every fd) but unbounded on the single-wallet path, where switching wallets leaves the outgoing wallet's connection open with no restart to close it. The handles are now destroyed, dependents first (prepared sends and per-mint wallets before the repository and database), a prepared melt or a confirmed/cancelled send is destroyed at the end of its operation for the same reason, and `currentDbPath` is published only once the handles it describes are the ones actually open, so the guard above cannot point at a database the module never opened. iOS was already correct under ARC and picks up the `currentDbPath` ordering plus a `preparedSends` clear on re-init.

Round 4, per @ajaysehwal's review: the JS-side guard in `clearCDKDatabaseForNode` was itself a check-then-act race across two bridge calls, so the basename compare and the handle teardown moved into a single native critical section, `closeWalletDatabase(dbFileName)` (`synchronized(handleLock)` on Android, `walletQueue.sync` on iOS). It unlinks nothing; file deletion stays with the JS caller.

Behavior note: because a wipe or wallet switch now destroys handles immediately instead of leaving them to the GC Cleaner, an in-flight Cashu operation that loses the race fails with a promise rejection instead of continuing against stale state. The vendored uniffi bindings make this safe (every call runs on a cloned Arc and the Rust free is deferred until the last in-flight call completes), so it is an error-surface change only, and deliberate: an operation surviving a wallet switch would otherwise mix two wallets' state.

Follow-up to #4328 / #4376; closes the last lifecycle gap noted in the internal assessment for the full-wipe path.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

Device tests pending (will update): initialize a Cashu wallet, run Clear All Data with the wallet warm, verify `cashu_wallet*` files (including `-wal`/`-shm`) are gone from the app container and no Cashu state resurfaces after restart; duress-PIN wipe smoke test; and for the native handle change, switch from wallet A to wallet B without restarting, delete A, and confirm no process still holds A's database open (`ls -l /proc/<pid>/fd` on Android) plus a send/melt on B afterwards to check the disposal did not take the live wallet's handles with it.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Backend-independent: affects the wipe path for Cashu native state only.

### Locales

### Other:

### DIFF
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -485,10 +485,10 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                     repo = newRepo
                     walletUnit = currencyUnit
                     // Publish the path only now that these are the handles
-                    // actually open: clearCDKDatabaseForNode keys its
-                    // dispose-before-unlink decision off getDatabasePath(),
-                    // so a path published before construction succeeded would
-                    // point the guard at a database the module never opened
+                    // actually open: closeWalletDatabase keys its
+                    // dispose-before-unlink decision off this path, so one
+                    // published before construction succeeded would point
+                    // the guard at a database the module never opened
                     currentDbPath = dbPath
                     isInitialized = true
                     previous
@@ -1899,8 +1899,14 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
      */
     @ReactMethod
     fun deleteWalletDatabase(promise: Promise) {
-        val path = currentDbPath
-        disposeHandles()
+        // Read the path in the same critical section that takes the handles:
+        // read outside it, a concurrent initializeWallet could swap wallets
+        // in between, disposing one wallet's handles and unlinking another's
+        // files
+        val (path, taken) = synchronized(handleLock) {
+            Pair(currentDbPath, takeHandlesLocked())
+        }
+        destroyHandles(taken)
         if (path == null) {
             promise.resolve(false)
             return
@@ -1912,6 +1918,36 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                 Log.w(TAG, "deleteWalletDatabase: failed to delete $p", e)
             }
         }
+        promise.resolve(true)
+    }
+
+    /**
+     * Disposes the open CDK handles if, and only if, the database currently
+     * open is [dbFileName], compared by basename because the JS side cannot
+     * reconstruct the absolute path this module resolves. The compare and the
+     * handle swap happen under one handleLock acquisition, so a wallet switch
+     * cannot land between the check and the teardown - the check-then-act gap
+     * a JS-side getDatabasePath() + deleteWalletDatabase() pair has, where a
+     * switch in the gap retargets the dispose (and its unlink) at the newly
+     * opened wallet's database. Unlinks nothing: the caller owns file
+     * deletion. Resolves true if the handles were disposed, false when no
+     * database or a different wallet's is open.
+     */
+    @ReactMethod
+    fun closeWalletDatabase(dbFileName: String, promise: Promise) {
+        val taken = synchronized(handleLock) {
+            val open = currentDbPath
+            if (open != null && File(open).name == dbFileName) {
+                takeHandlesLocked()
+            } else {
+                null
+            }
+        }
+        if (taken == null) {
+            promise.resolve(false)
+            return
+        }
+        destroyHandles(taken)
         promise.resolve(true)
     }
 
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -254,7 +254,25 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
             currentRepo.createWallet(url, walletUnit, null)
             currentRepo.getWallet(url, walletUnit)
         }
-        wallets[normalized] = wallet
+        // Insert under handleLock with a staleness check: this is the one
+        // map write that can land after a teardown drained the map (the repo
+        // was read before the create above, outside the lock), and an
+        // unconditional insert would park an undestroyed Wallet - holding
+        // its own reference to the closed database - in the map until the
+        // next initializeWallet drains it, briefly reopening the
+        // reachability window the wipe paths exist to close
+        val inserted = synchronized(handleLock) {
+            if (repo === currentRepo) {
+                wallets[normalized] = wallet
+                true
+            } else {
+                false
+            }
+        }
+        if (!inserted) {
+            destroyHandles(listOf(wallet))
+            throw FfiException.Internal("Wallet not initialized")
+        }
         return wallet
     }
 
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -40,11 +40,101 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
     private var isInitialized = false
     private val scope = CoroutineScope(Dispatchers.IO + SupervisorJob())
 
+    // Guards the swap of the handle set above. destroy() is idempotent, so
+    // this only has to keep two teardowns from taking the same handles, and
+    // a teardown from racing an initialization
+    private val handleLock = Any()
+
     companion object {
         private const val TAG = "CashuDevKitModule"
     }
 
     override fun getName(): String = "CashuDevKitModule"
+
+    // ========================================================================
+    // Handle Lifecycle
+    // ========================================================================
+
+    /**
+     * Removes every CDK handle from the module's state and returns them for
+     * destruction. Call under handleLock.
+     *
+     * Order matters: the per-mint Wallet handles and any outstanding
+     * PreparedSend hold their own references to the repository and database,
+     * so the SQLite connection is not released until those are destroyed
+     * first.
+     */
+    private fun takeHandlesLocked(): List<Disposable> {
+        val taken = mutableListOf<Disposable>()
+        taken.addAll(preparedSends.values)
+        preparedSends.clear()
+        taken.addAll(wallets.values)
+        wallets.clear()
+        repo?.let { taken.add(it) }
+        repo = null
+        db?.let { taken.add(it) }
+        db = null
+        isInitialized = false
+        currentDbPath = null
+        return taken
+    }
+
+    private fun destroyHandles(handles: List<Disposable>) {
+        for (handle in handles) {
+            try {
+                handle.destroy()
+            } catch (e: Exception) {
+                // Best effort: one handle failing to release must not strand
+                // the others, and destroy() is a no-op if it already ran
+                Log.w(
+                    TAG,
+                    "destroyHandles: failed to destroy ${handle.javaClass.simpleName}",
+                    e
+                )
+            }
+        }
+    }
+
+    /**
+     * Runs [block] on a transient CDK handle and destroys the handle after,
+     * success or not.
+     *
+     * Same reasoning as disposeHandles: a prepared melt or send holds its own
+     * reference to the wallet's database, so one left for the Cleaner keeps
+     * the SQLite connection alive past a teardown that destroyed everything
+     * the module tracks. Inline so suspending calls inside the block stay
+     * legal.
+     */
+    private inline fun <T : Disposable, R> T.useHandle(block: (T) -> R): R {
+        try {
+            return block(this)
+        } finally {
+            destroyHandles(listOf(this))
+        }
+    }
+
+    /**
+     * Destroys every live CDK handle and resets the module to uninitialized.
+     *
+     * The uniffi bindings are Disposable: dropping the Kotlin reference alone
+     * leaves the Rust object - and the SQLite connection it owns - alive
+     * until the Cleaner next runs after a GC. An unlinked database file
+     * therefore keeps its blocks allocated for an unbounded time, which is
+     * the reachability the wipe paths exist to close. iOS has no equivalent
+     * gap: ARC frees the handle as soon as the last reference goes.
+     *
+     * In-flight work is aborted rather than left to finish. uniffi keeps an
+     * already-started call alive (the Rust side holds its own reference), but
+     * the next call on a destroyed handle throws IllegalStateException, which
+     * the per-method catch turns into a promise rejection. That is the
+     * intended outcome: every caller here has either torn the wallet down or
+     * switched to a different one, and the methods below re-read repo/db
+     * between steps, so letting a half-finished flow continue would mean
+     * operating on a mix of the outgoing and incoming wallet's state.
+     */
+    private fun disposeHandles() {
+        destroyHandles(synchronized(handleLock) { takeHandlesLocked() })
+    }
 
     // ========================================================================
     // Helper Methods
@@ -168,6 +258,7 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         return wallet
     }
 
+    @Volatile
     private var currentDbPath: String? = null
 
     private fun getDatabasePath(mnemonic: String): String {
@@ -370,7 +461,6 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         scope.launch {
             try {
                 val dbPath = getDatabasePath(mnemonic)
-                currentDbPath = dbPath
                 val database = WalletSqliteDatabase(dbPath)
 
                 val currencyUnit = parseCurrencyUnit(unit)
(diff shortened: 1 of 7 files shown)


ajaysehwal (commented):
@kaloudis, I noticed `clearCDKDatabaseForNode` has the same unlink-before-dispose issue that this PR fixes in `clearCDKDatabase`. It can be called for the active node, but currently doesn't call `CashuDevKit.deleteWalletDatabase()` before deleting the files. I think the same `isAvailable()`-gated dispose-then-unlink handling should be applied here as well.

kaloudis:
Good catch, thanks @ajaysehwal. Confirmed and addressed in d28091994.

You are right that the path is reachable for the active wallet: `deleteNodeConfig` calls `clearCDKDatabaseForNode` at `views/Settings/WalletConfiguration.tsx:850`, after `stopLnd()` / `stopLdkNode()`, and neither of those touches CDK, so the SQLite connection is still live across the unlink loop.

One scope note: the last-wallet case was already covered, since `remainingNodes.length === 0` falls through to `clearCDKDatabase()` a few lines later, which this PR now disposes. The genuinely uncovered case is deleting the active wallet while other wallets remain, where nothing ever closed the handle and it stayed open for the rest of the session.

The one thing I could not do is apply the dispose unconditionally. `deleteWalletDatabase()` takes no argument: it drops `repo`/`db`/`wallets` and unlinks whatever single database the native module currently has open (`currentDbPath`, `CashuDevKitModule.kt:1785` / `.swift:1648`). Calling it from `clearCDKDatabaseForNode` while a *different* wallet is warm would delete that wallet's proof database, so deleting wallet B would destroy wallet A's ecash. That would trade a residual-reachability window for actual bearer-proof loss.

So the dispose is gated on `getDatabasePath()` matching the wallet being deleted. CDK holds one database open at a time, so a non-match proves nothing is open on that file and the plain unlink is already safe. The comparison is on basename rather than full path: the JS side builds `${DocumentDir}/../files` on Android while native reports the resolved absolute `filesDir`, so full-string equality would silently never match and the dispose would never fire.

Worth noting an `active` flag would have been the wrong discriminator here as well. If the user opens ecash on wallet A and then switches to B, `currentDbPath` still points at A even though B is the active wallet.

Also added the `../cashu-cdk` jest mock that was missing from `DataClearUtils.test.ts`. Without it `isAvailable()` is false in every test, so neither dispose branch executed and the suite passed while proving nothing. New cases cover dispose-before-unlink ordering, the different-wallet-open regression (which fails against the unconditional version), the no-database-open and module-absent paths, and a dispose failure not aborting the sweep.

Device tests for both wipe paths are still pending and I will update the checklist when they are done.

shubhamkmr04:
ACK — the change is correct, and the conditional guard is the right call. An unconditional dispose really would delete a warm sibling wallet's proofs.

Two non-blocking notes, both Android-only and both the same root cause.

**1. On Android, the "dispose" drops references but doesn't actually close anything.**

`deleteWalletDatabase()` sets `repo`/`db` to null, but these are uniffi objects (`WalletSqliteDatabase`, `WalletRepository`, `Wallet` are all `Disposable`/`AutoCloseable`). The Rust-side SQLite connection is only freed by an explicit `destroy()`, or whenever the uniffi Cleaner next runs. There is no `destroy()` or `close()` call anywhere in `CashuDevKitModule.kt`. On iOS, ARC frees it immediately, so that platform behaves exactly as the description says.

This doesn't undermine the PR. All three `clearAllData()` callers restart in a `finally`, and on Android `restartApp()` is a ProcessPhoenix process rebirth (`RestartUtils.ts:43`), which closes every fd and releases the unlinked inodes. And the part that actually mattered — CDK staying initialized so app code could keep reading the deleted database, or a pooled connection recreating the file at `currentDbPath` — is fixed on both platforms by the reference drop alone, since every read path in the module guards on `db == null || !isInitialized`.

**2. Worth a follow-up: the same gap in `clearCDKDatabaseForNode`, which has no restart to bound it.**

`initializeWallet` replaces `db`/`repo` without destroying the previous handles (`CashuDevKitModule.kt:386`). So on Android, CDK holds one *tracked* database but can have several *open* ones. Concretely:

1. Open ecash on wallet A → `currentDbPath = A`.
2. Switch to wallet B. No restart happens — `Wallets.tsx` just updates settings, and `Wallet.tsx` re-inits CDK. A's connection is now unreferenced but still open.
3. Delete wallet A. `getDatabasePath()` returns B, so the guard correctly skips the dispose, and A's database is unlinked under A's still-live leaked connection.

That is the exact window this PR closes elsewhere, and single-wallet deletion never restarts, so it lasts the rest of the session. Your point in the thread that `active` would be the wrong discriminator holds here too — this is just the mirror image of it: `currentDbPath` is necessary but not sufficient on Android.

Not urgent: the leaked inode has no filesystem path, `/proc/<pid>/fd` is uid-scoped, and recovering it needs root or a forensic image of the live process — at which point the plaintext database was readable before the wipe anyway. The fix is small: destroy the outgoing handles in `initializeWallet`, and `destroy()` rather than only null them in `deleteWalletDatabase`.

Only doc nit on the PR itself: worth noting that on Android the dispose relies on the restart that follows, rather than being an immediate close.

shubhamkmr04 (commented):
Two non-blocking Android notes, same root cause.

**The dispose drops references but doesn't close anything.** `WalletSqliteDatabase`/`WalletRepository`/`Wallet` are uniffi `Disposable`s, so the Rust SQLite connection is freed only by an explicit `destroy()` or whenever the Cleaner next runs — and there's no `destroy()`/`close()` anywhere in `CashuDevKitModule.kt`. iOS is fine, ARC frees it immediately. Harmless on this path: every `clearAllData()` caller restarts in a `finally`, and Android's `restartApp()` is a ProcessPhoenix process rebirth that closes every fd. The part that actually mattered — app code still reading the deleted db — is fixed by the reference drop alone.

**Follow-up: the same gap in `clearCDKDatabaseForNode`, which has no restart to bound it.** `initializeWallet` replaces `db`/`repo` without destroying the old handles (`CashuDevKitModule.kt:386`), so Android tracks one database but can hold several open. Open ecash on wallet A → switch to B (no restart) → delete A: `getDatabasePath()` returns B, the guard correctly skips, and A's db is unlinked under A's still-live connection. The mirror image of your `active`-flag point — `currentDbPath` is necessary but not sufficient on Android.

Low urgency: exploiting it needs root or a forensic image of the live process. Fix is small — destroy the outgoing handles in `initializeWallet`, and `destroy()` rather than only null them in `deleteWalletDatabase`.

kaloudis:
Both notes hold up, thanks @shubhamkmr04. Verified against `cdk-jvm.jar` rather than assumed, and fixed here in 14b47b6 rather than deferred, since the second one is a genuinely unbounded window.

**What I confirmed.** `WalletSqliteDatabase`, `WalletRepository`, `Wallet` and `PreparedSend` all `implements Disposable, AutoCloseable` with `destroy()`, a `wasDestroyed` AtomicBoolean and a `callCounter` AtomicLong, and the jar ships `UniffiCleaner` / `JavaLangRefCleaner` / `UniffiJnaCleaner`. There was no `destroy()` or `close()` on a CDK object anywhere in `CashuDevKitModule.kt`. iOS is fine as you say: `CashuDevKit.swift` has `deinit { ... uniffi_cdk_ffi_fn_free_walletsqlitedatabase(...) }`.

Your bound on the full-wipe path holds too: all three `clearAllData()` callers restart in a `finally`, and Android's `restartApp()` is `LndMobileTools.restartApp()` -> `ProcessPhoenix.triggerRebirth`. The single-wallet path has no such bound, and the switch is in-process as you describe (`Wallets.tsx` updates `selectedNode` and navigates, `Wallet.tsx` calls `CashuStore.initializeWallets()`, which reaches `CashuDevKit.initializeWallet` at `stores/CashuStore.ts:391`), so the scenario reproduces exactly.

**What the fix does.** Two things beyond the sketch, both load-bearing:

1. Destroying `repo` and `db` is not sufficient. The per-mint `Wallet` handles and any outstanding `PreparedSend` hold their own reference to the repository and database, so disposal goes dependents first: prepared sends, then wallets, then repo, then db, snapshotted and swapped under a lock so two teardowns cannot take the same handles.

2. Transient handles pin the connection the same way. A `PreparedMelt` from `prepareMelt`/`prepareMeltProofs`, and a `PreparedSend` once confirmed or cancelled, were dropped without being destroyed, so a single melt earlier in the session would have kept the connection alive past an otherwise complete disposal. Those are now destroyed at the end of their operation. `Token` handles are left to the Cleaner, since they do not hold the database.

On `initializeWallet`: destroying the outgoing handles there is what closes your scenario, because by the time A is deleted its connection went away at the switch. In-flight work is aborted rather than left to run, since the next call on a destroyed handle throws `IllegalStateException` and the per-method `catch (e: Exception)` turns that into a promise rejection. That is deliberate: these methods re-read `repo`/`db` between steps, so a flow that survived a switch was already operating on a mix of the two wallets' state (`meltPartial` would gather proofs from B's database and melt them through A's wallet handle).

One thing that fell out along the way: `currentDbPath` was published before `WalletRepository` construction could fail, so a failed init left `getDatabasePath()` pointing at a database the module never opened, which is exactly the signal the guard in `clearCDKDatabaseForNode` keys off. It is now published with the handles it describes, on both platforms. iOS also clears `preparedSends` on re-init, which it was not doing.

Verified with `:app:compileDebugKotlin` plus tsc/lint/prettier and the `DataClearUtils` suite. Device tests are still owed, and the switch-then-delete case is now on the list along with a send on the incoming wallet to confirm the disposal does not take the live handles with it.

ajaysehwal (commented):
@kaloudis, one narrow thing to fully close the loop: `clearCDKDatabaseForNode` currently has a check-then-act gap between `getDatabasePath()` and `deleteWalletDatabase()`. A wallet switch landing between those calls could mean the dispose happens against whatever wallet is open at that point, rather than the path that was checked.

Since `deleteWalletDatabase()` already handles `wallets` / `preparedSends` / `repo` / `db` under `handleLock`, an alternative would be a native method that takes the target path and checks + disposes it atomically inside the same `synchronized` block.

Feels like a nice-to-have rather than blocking given how narrow the window is, but flagging it since it’s the same kind of race the other fixes here are addressing.

kaloudis:
Good catch again @ajaysehwal. Took your suggestion rather than deferring, in 36621d1ac. The window is one bridge round trip wide, but a hit is the exact bearer-proof loss this PR is about: `deleteWalletDatabase()` re-reads `currentDbPath` at call time, so a switch landing in the gap would retarget both the dispose and its unlink at the newly opened wallet's database.

`closeWalletDatabase(dbFileName)` now does the basename compare and the handle teardown in one `synchronized(handleLock)` block (`walletQueue.sync` on iOS), resolves whether it disposed, and unlinks nothing: file deletion stays with the JS caller, which already removes the db and its WAL/SHM sidecars. `clearCDKDatabaseForNode` calls that instead of the `getDatabasePath()` pair.

The same shape turned out to exist inside `deleteWalletDatabase` itself on both platforms: the path was read outside the critical section that takes the handles, so a switch in between could dispose one wallet's handles and unlink another wallet's files. The path read moved into the section.

Jest covers the call shape: the basename argument, dispose-before-unlink ordering, and the false/throw/module-absent fallthroughs, plus lockstep source scans that fail if the native method disappears, loses its bridge declaration, or the compare drifts outside the lock. The atomicity itself is native, so it stays on the device-test checklist with the rest.

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
ajaysehwal, line 1: @kaloudis, one thing I want to flag rather than block on: `getWallet()` (and a few other read sites) still access `repo`/`wallets` outside `handleLock`, so a full wipe or wallet delete racing an in-flight Cashu operation can now return a handle that gets `destroy()`’d moments later. Previously, that stale handle would just keep working until GC.

Given the `disposeHandles()` doc comment, this looks intentional — fail fast rather than continue operating on mixed wallet state. And if uniffi’s generated `destroy()` is safe to race per object, this should be a rejection/error-surface change rather than a memory-safety issue.

Could you confirm that this concurrency guarantee also holds for the CDK bindings we’re vendoring, and maybe mention in the PR description that in-flight Cashu operations can now fail during a wipe instead of continuing with stale state?

Not a blocker — just worth documenting the tradeoff.

kaloudis: Confirmed on both counts, @ajaysehwal. Verified against the vendored jar rather than assumed, and your read is right: it is an error-surface change, not a memory-safety one.

**What javap on `android/cdk/cdk-jvm.jar` shows** (`org.cashudevkit.Wallet`; same shape on `WalletRepository`, `WalletSqliteDatabase`, `PreparedSend`): `callCounter` starts at 1, every method CAS-increments it and throws `IllegalStateException` if the handle is already destroyed, and each call then runs against a cloned Arc (`uniffi_cdk_ffi_fn_clone_wallet`). `destroy()` only flips `wasDestroyed` and decrements the counter; the Rust-side free runs when the counter reaches 0. So an in-flight call always completes against a live object, and the loser of the race gets a rejection on its next call, never a use-after-free. Suspend methods release the counter once the future is created, but the Rust future holds its own Arc, so the guarantee holds there too. iOS never destroys explicitly (ARC deinit), so the race cannot exist on that side.

The fail-fast is deliberate, as you inferred from the `disposeHandles()` comment: these methods re-read `repo`/`db` between steps, so an operation that survived a wallet switch was already mixing two wallets' state (`meltPartial` would gather the new wallet's proofs and melt them through the old wallet's handle). A rejection is strictly better than letting that complete. The tradeoff is now noted in the PR description.

**One residual your question surfaced**, disclosing it rather than sitting on it: `getWallet()`'s lazy-create path can insert a fresh `Wallet` into the `wallets` map after a concurrent teardown has drained it (the repo read wins, teardown runs, then the map insert lands). That orphan keeps the unlinked database open until the next `initializeWallet()` swap drains the map, so the window is bounded by the next init or restart, versus unbounded before this PR. Android only; iOS serializes on `walletQueue`. The fix is small (take `handleLock` around the insert with a staleness check on the repo identity), happy to fold it in here if you would rather see it closed in this PR than left as a noted residual.

kaloudis: Went ahead and closed the residual rather than leaving it noted, in b5fee5871. `getWallet()`'s map insert now happens inside `handleLock` (`walletQueue.sync` on iOS) with a staleness check on the repository identity; on a mismatch the fresh handle is destroyed on Android (released by ARC on iOS) and the call rejects with the same not-initialized error as any other post-teardown call. Applied to both platforms since the iOS insert had the same shape: the repo read and the map write are separate `walletQueue.sync` blocks with awaits between them, so the queue alone does not close the window.

The lockstep source scans in `DataClearUtils.test.ts` now also assert the compare sits inside the critical section on both platforms. Verified with `:app:compileDebugKotlin`, tsc/lint/prettier, and the full `DataClearUtils` suite.

ajaysehwal:
@kaloudis, great fix — this closes the teardown race cleanly on both platforms, and the tests pin the ordering down well.

One adjacent case I noticed: the staleness check only catches `repo` changes. If two callers race on the very first access to the same mint (with no teardown), both can pass the check and write to `wallets[normalized]`. The second write overwrites the first `Wallet` handle without destroying it, leaving the losing handle to eventually rely on the Cleaner — the same class of leak this PR is addressing.

This predates the commit, so not asking you to fix it here. Just flagging it as a possible follow-up for symmetry, e.g., destroying the newly created wallet if the map already contains a different entry for the same key.

kaloudis:
Good catch @ajaysehwal. Took this one in-PR too rather than deferring, in b168673fb, since it is the same orphaned-handle class the PR exists to close: the losing wrapper holds its own reference to the repository and database, sits outside the map, and so survives `disposeHandles` until the Cleaner runs.

The locked insert is now first-wins: it re-checks the map inside the critical section, a loser destroys its fresh handle on Android (released by ARC on iOS, where the window was already bounded by the racing caller's scope) and the existing entry is returned. Returning the winner rather than rejecting is deliberate: unlike the repo-mismatch case there is no teardown in flight, both wrappers are Arc clones of the same underlying wallet, and the caller's operation can proceed on the surviving handle.

The DataClearUtils source scans now pin the re-check inside the lock on both platforms. Verified with `:app:compileDebugKotlin` plus jest/tsc/lint/prettier. Device tests remain on the checklist.

ajaysehwal (approved):
tACK

PR #4580, opened 2026-09-06
views: use sending capacity for payment selection

kaloudis:
# Description

|Before|After|
|-|-|
|Simulator Screenshot - iPhone 17 Pro - 2026-09-06 at 10 50 33|Simulator Screenshot - iPhone 17 Pro - 2026-09-06 at 10 49 54|

Keep the Home/BalancePane Lightning balance unchanged, but show reserve-adjusted sending capacity on Select Payment Method when channel data is available. For example, a wallet with 12,291 sats and 11,243 sats outbound should not offer a 12,000-sat Lightning payment based on its headline balance.

- Display sending capacity with an "Available to send" label for Lightning, Lightning address, and Offer rows.
- Use the same amount for insufficient-funds checks and payment-row eligibility, including the Lightning contribution to CLINK eligibility.
- Track channel-capacity availability separately from zero capacity; fall back to the reported balance when channel data is unavailable or unsupported.
- Preserve LNURL-withdraw behavior and leave Home balances unchanged.
- Add nine regression cases covering the reserve gap, capacity boundaries, zero capacity, fallback behavior, withdrawals, and the balance/label passed to the payment list.

Capacity is before routing fees and is not a guarantee of route availability. Screenshots and hands-on iOS/Android testing are pending; this PR is a draft.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I have run `yarn run tsc` and made sure my code compiles correctly
- [x] I have run `yarn run lint` and made sure my code did not contain any problematic patterns
- [x] I have run `yarn run prettier` and made sure my code is formatted correctly
- [x] I have run `yarn run test` and made sure all of the tests pass

## Testing

All 71 Jest suites pass (1,549 tests), including nine new regression cases. TypeScript passes. The initial `yarn verify` found one formatting issue; after fixing it, both `yarn lint` (including the stylesheet test) and `yarn prettier` passed. `git diff --check` also passed.

- [ ] No, I am a fool
- [ ] Yes
- [x] N/A

No utility files changed. Regression tests were added in `views/ChoosePaymentMethod.test.ts`.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Manual platform/backend testing and screenshots remain pending. Unit tests use mocked component and backend boundaries.

### Locales
- [x] I have added new locale text that requires translations
- [x] I am aware that new translations should be made on the ZEUS [Transfix page](https://app.transifex.com/ZeusLN/zeus/) and not directly to this repo

Only `locales/en.json` was changed: "Available to send".

No dependencies changed. Dependency checks, if applicable: verify that `package.json` and `yarn.lock` have been properly updated; verify that dependencies are installed for both iOS and Android platforms.

### Other:

### DIFF
--- a/stores/ChannelsStore.ts
+++ b/stores/ChannelsStore.ts
@@ -83,7 +83,15 @@ export default class ChannelsStore {
     // redesign
     @observable public largestChannelSats = 0;
     @observable public totalOutbound = 0;
-    @observable public hasSendingCapacity = false;
+    // Local balance sitting in channels that are currently inactive. These
+    // sats cannot be routed right now but are not lost: the peer may be
+    // momentarily disconnected. Tracked separately so callers can tell
+    // "no money" apart from "money you cannot reach at this instant".
+    @observable public totalOutboundOffline = 0;
+    // Whether a channel fetch has ever succeeded for this node. Callers use
+    // it to decide if the totals above are meaningful; it says nothing about
+    // whether there is any capacity.
+    @observable public hasChannelData = false;
     @observable public totalInbound = 0;
     @observable public totalOffline = 0;
     @observable public chanInfo: ChannelInfoIndex = {};
@@ -199,7 +207,7 @@ export default class ChannelsStore {
     @action
     public reset = () => {
         this.resetOpenChannel();
-        this.hasSendingCapacity = false;
+        this.hasChannelData = false;
         this.haveAnnouncedChannels = false;
         this.nodes = {};
         this.channels = [];
@@ -214,6 +222,7 @@ export default class ChannelsStore {
         this.filteredPeers = [];
         this.largestChannelSats = 0;
         this.totalOutbound = 0;
+        this.totalOutboundOffline = 0;
         this.totalInbound = 0;
         this.totalOffline = 0;
         this.channelsType = ChannelsType.Open;
@@ -504,8 +513,13 @@ export default class ChannelsStore {
 
     @action
     private getChannelsError = () => {
-        this.hasSendingCapacity = false;
+        this.hasChannelData = false;
         this.channels = [];
+        this.largestChannelSats = 0;
+        this.totalOutbound = 0;
+        this.totalOutboundOffline = 0;
+        this.totalInbound = 0;
+        this.totalOffline = 0;
         this.error = true;
         this.loading = false;
         this.closingChannel = false;
@@ -513,19 +527,23 @@ export default class ChannelsStore {
 
     @action
     public getChannels = () => {
-        this.hasSendingCapacity = false;
         this.loading = true;
-        this.channels = [];
-        this.largestChannelSats = 0;
-        this.totalOutbound = 0;
-        this.totalInbound = 0;
-        this.totalOffline = 0;
 
         const loadPromises = [
+            // The channel list and its totals are accumulated into locals and
+            // published in one go once the fetch succeeds. Zeroing them up
+            // front would make every refresh briefly report an empty wallet to
+            // observers (payment method selection, Receive's inbound checks,
+            // the channels header), and refreshes are frequent.
             BackendUtils.getChannels().then((data: any) => {
                 const channels = data.channels.map(
                     (channel: any) => new Channel(channel)
                 );
+                let largestChannelSats = new BigNumber(0);
+                let totalInbound = new BigNumber(0);
+                let totalOutbound = new BigNumber(0);
+                let totalOutboundOffline = new BigNumber(0);
+                let totalOffline = new BigNumber(0);
                 channels.forEach((channel: Channel) => {
                     const channelRemoteBalance = new BigNumber(
                         channel.receivingCapacity
@@ -535,23 +553,26 @@ export default class ChannelsStore {
                     );
                     const channelTotal =
                         channelRemoteBalance.plus(channelLocalBalance);
-                    if (channelTotal.gt(this.largestChannelSats))
-                        this.largestChannelSats = channelTotal.toNumber();
+                    if (channelTotal.gt(largestChannelSats))
+                        largestChannelSats = channelTotal;
                     if (!channel.isActive) {
-                        this.totalOffline = new BigNumber(this.totalOffline)
-                            .plus(channelTotal)
-                            .toNumber();
+                        totalOffline = totalOffline.plus(channelTotal);
+                        totalOutboundOffline =
+                            totalOutboundOffline.plus(channelLocalBalance);
                     } else {
-                        this.totalInbound = new BigNumber(this.totalInbound)
-                            .plus(channelRemoteBalance)
-                            .toNumber();
-                        this.totalOutbound = new BigNumber(this.totalOutbound)
-                            .plus(channelLocalBalance)
-                            .toNumber();
+                        totalInbound = totalInbound.plus(channelRemoteBalance);
+                        totalOutbound = totalOutbound.plus(channelLocalBalance);
                     }
                 });
-                this.channels = channels;
-                this.hasSendingCapacity = true;
+                runInAction(() => {
+                    this.largestChannelSats = largestChannelSats.toNumber();
+                    this.totalInbound = totalInbound.toNumber();
+                    this.totalOutbound = totalOutbound.toNumber();
+                    this.totalOutboundOffline = totalOutboundOffline.toNumber();
+                    this.totalOffline = totalOffline.toNumber();
+                    this.channels = channels;
+                    this.hasChannelData = true;
+                });
             })
         ];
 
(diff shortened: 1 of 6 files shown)


kaloudis:
Screenshots added. Open to discussion about how to convey this better.

views/ChoosePaymentMethod.tsx
```diff
@@ -176,9 +184,23 @@ export default class ChoosePaymentMethod extends React.Component<
...
+            BackendUtils.supportsChannelManagement() &&
+            !!this.props.ChannelsStore?.hasSendingCapacity
+        );
+    }
+
+    get lightningPaymentBalance() {
+        return this.usesSendingCapacity
+            ? this.props.ChannelsStore!.totalOutbound
+            : this.props.BalanceStore!.lightningBalance;
+    }
```
shubhamkmr04, line 207: **Offline channels now read as "Not enough funds".**

`totalOutbound` only sums channels where `isActive` is true — inactive ones go to `totalOffline` instead (`ChannelsStore.getChannels`, the `if (!channel.isActive)` branch). On LDK `active` maps to `isUsable`, which is false while the peer is disconnected: routine for the first seconds after foregrounding, and normal during the `getChannelsWithPolling` window on embedded LND / LDK.

In that window `hasSendingCapacity` is already `true` and `totalOutbound` is `0`, so:
- the Lightning row is disabled, and
- the banner says `stores.CashuStore.notEnoughFunds`

…for a wallet that does have the funds and whose peer reconnects a moment later. Before this PR the user could at least attempt the payment.

Suggestion: treat "channels exist but none active" as *no data* rather than *no money* — e.g. fall back to `lightningBalance` when `totalOutbound === 0 && totalOffline > 0`, or keep the row disabled but show a channels-offline message instead of an insufficient-funds one.

stores/ChannelsStore.ts
```diff
@@ -510,6 +513,7 @@ export default class ChannelsStore {
 
     @action
     public getChannels = () => {
+        this.hasSendingCapacity = false;
```
shubhamkmr04: **This reset re-opens the bug for the duration of every refresh.**

This line (and `totalOutbound = 0` four lines down) runs synchronously *before* the network call, so while any `getChannels()` is in flight the screen falls back to `BalanceStore.lightningBalance`.

`ChoosePaymentMethod` is an `@observer`, so this happens live while the user is looking at it. Refreshes are frequent: `getChannelsWithPolling` runs up to 20 × 3s after connect, and `Receive`, `SendingLightning` (after a payment), `Rebalance` and ChannelsPane pull-to-refresh all call `getChannels()`. Scan an invoice inside one of those windows and the 12,000-sat payment is offered again on the 11,243-sat-outbound wallet — the exact case this PR fixes. The "Available to send" label also blinks in and out.

Making the successful fetch atomic fixes it (and removes the existing "channel list momentarily empty on every refresh" flicker). Keep the resets in `reset()` and `getChannelsError()`, drop them here, and publish at the end:

```ts
BackendUtils.getChannels().then((data: any) => {
    let largest = 0, inbound = 0, outbound = 0, offline = 0;
    const channels = data.channels.map((c: any) => new Channel(c));
    channels.forEach((channel: Channel) => {
        // accumulate into the locals
    });
    this.largestChannelSats = largest;
    this.totalInbound = inbound;
    this.totalOutbound = outbound;
    this.totalOffline = offline;
    this.channels = channels;
    this.hasSendingCapacity = true;
})
```

(Partially dropping just this line would be worse — the flag would stay `true` while the totals are zeroed, showing 0 capacity mid-refresh — so the totals need to move with it.)

shubhamkmr04 (commented):
Two cases where I think the capacity number can be wrong in the other direction

views/ChoosePaymentMethod.tsx
kaloudis: Confirmed, and thanks for catching this. `active` maps to `isUsable` on LDK (`backends/LdkNode.ts:2000`), so a disconnected peer sends the whole channel to `totalOffline` and `totalOutbound` reads 0. `hasInsufficientBalance` in `PaymentMethodList` treats a 0 balance as insufficient regardless of the amount, so the row went disabled and the banner fired for a wallet that does have the funds. It did not happen before this PR because `BalanceStore.lightningBalance` does not exclude offline channels. Worth noting it also does not reproduce on CLN, where `Channel.isActive` accepts `CHANNELD_NORMAL` regardless of connectivity, which is probably why it got past me.

On the suggested heuristic: `totalOffline` is total channel capacity rather than local balance, so it cannot say how much of the offline capacity is actually ours, and the mixed case (one drained active channel plus one funded offline channel) would still under-report. I added a `totalOutboundOffline` accumulator instead:

- display stays `totalOutbound` under the "Available to send" label, which is honestly what can be routed at this instant
- eligibility (row enabled state, red amount, insufficient-funds banner) uses `totalOutbound + totalOutboundOffline`, so a reconnecting peer never blocks a payment the wallet can afford
- when the amount only fits with the offline capacity, a `WarningMessage` says some channels are offline and the payment may fail until they reconnect, rather than the insufficient-funds error

To let display and eligibility differ, `PaymentMethodList` rows take an optional `eligibilityBalance` that defaults to `balance`. The CLINK row uses it too, so its `disabled` check does not regress the same way.

stores/ChannelsStore.ts
kaloudis: Agreed on all of it, including the warning about dropping only the flag line. Fixed by accumulating into locals and publishing in a single `runInAction` at the end of the successful fetch, with the resets kept in `reset()` and `getChannelsError()`. `getChannelsError` now also zeroes the totals, so the error path keeps the semantics it had when they were cleared up front.

This also removes the mid-refresh flicker for the other observers of these fields: the channels header, and `Receive`, which compares requested amounts against `totalInbound` to decide the LSP flow.

Added `stores/ChannelsStore.test.ts`. Two cases target this directly: one asserts the previous totals stay visible while a fetch is outstanding, and one uses a mobx `autorun` to assert no observer ever sees `hasChannelData` true against zeroed totals. I checked both fail against the three variants: the original reset, a flag-only reset, and a totals-only reset.

Also renamed `hasSendingCapacity` to `hasChannelData`. It records that a fetch succeeded, not that there is capacity, and reading it the other way is what produced your first finding.

kaloudis:
Both review findings are addressed locally and I will push shortly.

One caveat neither comment covers, worth stating for the record: `ChoosePaymentMethod` never calls `getChannels()` itself, so `totalOutbound` is only as fresh as the last refresh from elsewhere, and if that fetch errored the view silently falls back to the headline balance. That fallback is deliberate, but it means the guard is best effort rather than a guarantee. I will note it in the description alongside the existing routing-fee caveat, along with the fact that `totalOutbound` is a cross-channel sum, so a single non-MPP payment cannot necessarily use all of it.

Manual iOS and Android testing is still pending, in particular the foregrounding window on embedded LND and LDK that @shubhamkmr04 identified.

kaloudis:
Pushed as c81647830. Both round-1 findings are addressed:

- `getChannels()` now publishes the channel list and its totals atomically, in a single `runInAction` at the end of a successful fetch, instead of zeroing them synchronously up front. The resets stay in `reset()` and `getChannelsError()`, and the latter now zeroes the totals too.
- New `totalOutboundOffline` tracks the local balance of inactive channels. Display still uses `totalOutbound`, so the row shows what is spendable right now, but eligibility uses `totalOutbound + totalOutboundOffline`. A payment that only fits once peers reconnect now gets a channels-offline warning rather than an insufficient-funds error.
- `hasSendingCapacity` is renamed `hasChannelData`. It records that a fetch succeeded, not that there is capacity, and reading it the other way is what produced the false "Not enough funds".

`stores/ChannelsStore.test.ts` is new (5 cases) and was verified to fail against the original reset, a flag-only reset, and a totals-only reset.

Two caveats for the record, neither covered by the review comments:

1. `ChoosePaymentMethod` never calls `getChannels()` itself, so `totalOutbound` is only as fresh as the last refresh from elsewhere, and if that fetch errored the view falls back to the headline balance. That fallback is deliberate, but it makes the guard best effort rather than a guarantee.
2. `totalOutbound` is a cross-channel sum, so a single non-MPP payment cannot necessarily use all of it.

One correction to the description: this now adds a second string to `locales/en.json`, `views.ChoosePaymentMethod.channelsOffline`, alongside `availableToSend`.

Manual iOS and Android testing is still pending, in particular the foregrounding window on embedded LND and LDK that @shubhamkmr04 identified.

shubhamkmr04 (approved):
tACK

ajaysehwal (approved):
tACK

PR #4585, opened 2026-09-06
feat: link the local wallet FAQs from Troubleshooting and Help

kaloudis:
## Summary

The [Embedded LND](https://docs.zeusln.app/for-users/local-wallets/embedded-lnd/faq) and [LDK Node](https://docs.zeusln.app/for-users/local-wallets/ldk-node/faq) FAQs cover most of what the Troubleshooting screen does — express graph sync resets, Neutrino peers, `Sync wallets`, network graph resets — with the surrounding context on when and why to reach for each. Nothing in the app pointed at them.

- **Troubleshooting**: a `Read the FAQ` button at the top of the screen, above the implementation-specific controls, so it's seen before the reset buttons get pressed.
- **Help**: an `Embedded node FAQ` entry directly under the general documentation entry.

Both resolve to the FAQ for the implementation in use.

## Notes

`UrlUtils.getLocalWalletFaqUrl` keeps the implementation → URL mapping in one place. It returns `undefined` for anything that isn't `embedded-lnd` or `ldk-node`, so the Help entry is simply hidden for remote nodes rather than sending those users to a local wallet page. The Troubleshooting screen is already local-wallet-only, but it's guarded the same way.

## Testing

- `npx tsc --noEmit` — clean
- `npx eslint` on the changed files — clean
- `npx prettier --check` on the changed files — clean
- `npx jest utils/UrlUtils.test.ts` — 35 passing, including new coverage for `getLocalWalletFaqUrl`


utils/UrlUtils.ts
```diff
@@ -177,6 +177,19 @@ const goToBlockExplorerChannelId = (channelId: string, testnet?: boolean) =>
...
+    'embedded-lnd':
+        'https://docs.zeusln.app/for-users/local-wallets/embedded-lnd/faq',
+    'ldk-node': 'https://docs.zeusln.app/for-users/local-wallets/ldk-node/faq'
+};
+
+/**
+ * FAQ page for the local wallet implementation in use. Returns undefined for
+ * remote nodes, where neither local wallet FAQ applies.
+ */
+const getLocalWalletFaqUrl = (implementation?: string): string | undefined =>
```
ajaysehwal: nit: Could we use a static type for `implementation` here instead of `string`, e.g. `'embedded-lnd' | 'ldk-node'` or maybe Implementations (type from SettingsStore)?

kaloudis: Good catch, took the `Implementations` variant in 6a67c64.

I went with `Implementations` rather than `'embedded-lnd' | 'ldk-node'` for the parameter because both call sites pass whatever implementation is currently active, and returning `undefined` for the other five is the point of the helper. A two-member parameter would make the `Help.tsx` call site a type error and push the check back onto callers.

The bigger win turned out to be the map itself: `Partial<Record<Implementations, string>>` (same shape as `NODE_NAMES` in `ClientInfoUtils`) means a typo'd or renamed implementation key now fails to compile, where the string index signature would have quietly resolved to `undefined`. The `undefined` guard stays, since `implementation` can be unset before settings load.

ajaysehwal (approved):
tACK

PR #4588, opened 2026-09-07
build: bump CashuDevKit (CDK) to 0.18.1

kaloudis:
# Description

Bumps the vendored CashuDevKit (CDK) artifacts from 0.17.4 to 0.18.1.

**Android packaging rework (upstream, landed in 0.18.0):** the `cdk-jvm` bindings jar was discontinued after 0.17.5; the compiled Kotlin bindings now ship inside the `cdk-android` aar again (package unchanged: `org.cashudevkit`), and the native library was renamed from `libcdk_ffi.so` to `libcdk_ffi_kotlin.so`. `fetch-libraries.sh` drops the jar fetch (and removes stale local copies so an old-API jar cannot shadow the bundled bindings), and gradle drops the jar dependency and updates `packagingOptions`.

**API changes absorbed in the native bridges (both platforms kept in lockstep):**

* `Wallet.getMintKeysets(KeysetFilter)` was removed in favor of `keysets(policy:)`, which returns the new `KeySet` type. Both Zeus call sites (token decode and the `getMintKeysets` RN method) were local-store reads in 0.17.x, so they now read cached keysets straight from `WalletSqliteDatabase.getMintKeysets`, which still returns `[KeySetInfo]`. The token-decode path keeps its no-network guarantee for foreign tokens.
* `MintQuote` gained an `updatedAt` field. External quotes are inserted with 0 (oldest possible) so any later saga update, which drops stale responses by timestamp, supersedes the externally sourced row.
* `FfiException`/`FfiError` gained a `PaymentRequestDeliveryFailed` variant (NUT-18 transport failure). Mapped explicitly to `PAYMENT_FAILED`; the path is unreachable until the bridge exposes `preparePayRequest`.
* `KeySetInfo.inputFeePpk` is no longer optional on Android.

**What 0.18.1 adds over 0.18.0 (this PR was previously targeting 0.18.0):**

* **The 32-bit blocker is gone.** 0.18.0 shipped the `cdk-android` aar for arm64-v8a and x86_64 only; 0.18.1 restores `armeabi-v7a` (and `x86`, which we no longer ship since #4596) via cashubtc/cdk#2489, so the aar again covers every ABI we publish. See the discussion in #4589.
* Wallet-side fixes: receive and restore responses now validate signature counts, requested amounts and keysets, and supplied DLEQ proofs before storing spendable proofs; swap, issuance and receive recovery retain pending state until every expected output is restored (cashubtc/cdk#2547, #2548). The rest of the 0.18.1 changelog is mint-side.
* The generated bindings are byte-identical between 0.18.0 and 0.18.1: the `cdk-android` `classes.jar` hashes to `3bb5916f…` in both, and `cdk-swift`'s `Sources/Cdk/CashuDevKit.swift` to `336d3fdf…` (which is why `iosBindingsSha256` is unchanged from the 0.18.0 revision of this PR). No bridge changes were needed for the retarget.

**Still draft** because device testing is pending (see below). Note for testers: **0.18.x runs one-way wallet DB migrations automatically on open** (transaction status, quote timestamps, derivation counters). A build rolled back to 0.17.x cannot reopen a migrated Cashu DB, so test with a throwaway wallet or be ready to restore from seed. Upgrade-in-place with existing balance and history is the key scenario. 0.18.0 also added default per-mint request pacing (persistent token bucket); multimint balance refresh timing should be sanity checked on device.

**Follow-up unlocked (not in this PR):** `Transaction` now carries `status` (pending/completed/failed), which #4493 can surface to drive pending/failed activity rows from CDK as the source of truth. New 0.18 surface worth exploring later: NUT-16 animated QR token encoding, `getOrCreateWallet`, cross-mint transfer quotes.

Verification done so far, all on 0.18.1 artifacts:

* `fetch-libraries.sh` runs end to end with every checksum verified; the xcframework still ships a macOS slice and still needs (and passes) the strip workaround.
* `:app:compileDebugKotlin` compiles clean against the 0.18.1 bindings (the exhaustive-when check on `FfiException` is what originally surfaced the new error variant).
* `:app:assembleDebug` packages `libcdk_ffi_kotlin.so` into each per-ABI APK: `lib/armeabi-v7a` (13.7 MB), `lib/arm64-v8a` (18.5 MB), `lib/x86_64` (20.9 MB), with no `lib/x86` anywhere. All three `LOAD` segments are 16 KB aligned, so the restored publish pipeline keeps the Android 15 page-size requirement.
* `pod install` regenerates the `CashuDevKit` lock entry; `ios/Podfile.lock` is updated to `0.18.1` with the matching spec checksum.
* `yarn verify` is green (`utils/DataClearUtils.test.ts`, which parses both native sources for the DB filename scheme, included).

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

Contributors will need to re-run `bash fetch-libraries.sh` after this PR is merged in (`package.json` and `yarn.lock` are untouched; the CDK artifacts and their pinned hashes changed).

### Other:

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -166,11 +166,13 @@ android {
         pickFirst "lib/x86/libgojni.so"
         pickFirst "lib/x86_64/libgojni.so"
 
-        // CashuDevKit native libraries (Rust FFI)
-        pickFirst "lib/armeabi-v7a/libcdk_ffi.so"
-        pickFirst "lib/arm64-v8a/libcdk_ffi.so"
-        pickFirst "lib/x86/libcdk_ffi.so"
-        pickFirst "lib/x86_64/libcdk_ffi.so"
+        // CashuDevKit native libraries (Rust FFI; renamed from libcdk_ffi.so
+        // in 0.18.0, which upstream currently only ships for arm64-v8a and
+        // x86_64)
+        pickFirst "lib/armeabi-v7a/libcdk_ffi_kotlin.so"
+        pickFirst "lib/arm64-v8a/libcdk_ffi_kotlin.so"
+        pickFirst "lib/x86/libcdk_ffi_kotlin.so"
+        pickFirst "lib/x86_64/libcdk_ffi_kotlin.so"
 
         // Zeus Cashu Restore native libraries (Rust FFI)
         pickFirst "lib/armeabi-v7a/libuniffi_zeus_cashu_restore.so"
@@ -204,10 +206,9 @@ dependencies {
 
     implementation fileTree(dir: "libs", include: ["*.jar", "*.aar"])
 
-    // CashuDevKit - native libs (cdk-android) + Kotlin bindings (cdk-jvm),
+    // CashuDevKit - native libs + Kotlin bindings in one aar (cdk-android),
     // vendored from Maven Central by fetch-libraries.sh
     implementation files("../cdk/cashudevkit.aar")
-    implementation files("../cdk/cdk-jvm.jar")
     // Zeus Cashu Restore - standalone NUT-13 restore for v1 legacy seeds
     implementation files("../zeus-restore/zeus-cashu-restore.aar")
     // CDK / LDK Node dependencies
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -222,6 +222,10 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
                 legacyErrorCode(e.code, e.errorMessage) to e.errorMessage
             is FfiException.Internal ->
                 legacyErrorCode(null, e.errorMessage) to e.errorMessage
+            // 0.18.0 NUT-18 transport failure; unreachable until the bridge
+            // exposes preparePayRequest
+            is FfiException.PaymentRequestDeliveryFailed ->
+                "PAYMENT_FAILED" to e.errorMessage
         }
     }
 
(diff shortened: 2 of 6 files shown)


kaloudis:
Opened #4589 to discuss dropping the 32-bit release builds entirely, which is the alternative path to unblocking this. Key input needed there: Google Play Console active device counts by ABI.

kaloudis:
Rebased onto master (was conflicting after #4596 dropped the x86 ABI) and retargeted from 0.18.0 to 0.18.1.

0.18.1 restores the `armeabi-v7a` native library through cashubtc/cdk#2489, which removes the blocker this PR was parked on: the aar again covers all three ABIs we publish. Verified by building the debug splits, `libcdk_ffi_kotlin.so` lands in `lib/armeabi-v7a`, `lib/arm64-v8a` and `lib/x86_64`, no `lib/x86`, all LOAD segments 16 KB aligned.

No bridge changes were needed for the retarget: the generated bindings are byte-identical between 0.18.0 and 0.18.1 on both platforms. Also picked up the `ios/Podfile.lock` entry that the 0.18.0 revision missed.

Still draft pending device tests on both platforms (one-way Cashu DB migration, upgrade-in-place with balance and history, multimint refresh under the new request pacing).

Issue #4612, opened 2026-09-08
Storage: a failed settings read is indistinguishable from “no settings”, and the launch write then persists wallet-less defaults over the real blob


kaloudis:
## Context

Surfaced while investigating #4610 (multi-device iOS user, wallet list replaced by an outdated one after updating to v13.2.1; same reporter as #3625).

The headline cause of that report is the iCloud keychain leak tracked in #4403 / PR #4385 (`cloudSync: false` is treated as `true` by react-native-keychain 10.0.0, upstream oblador/react-native-keychain#800). This issue is about a **separate, self-contained defect in the settings load/write path** that turns a single transient storage read failure into a permanent wipe of the persisted wallet list. It needs no iCloud involvement, applies to Android as well as iOS, and can be fixed independently of (and sooner than) the keychain work — while on iOS the sync bug also broadcasts the damage to every other device the user owns.

## Mechanism

**1. `getSettings` cannot distinguish "no settings" from "read failed".**

`stores/SettingsStore.ts:1962` wraps the whole load in one try/catch whose handler (`:2009`) only logs. On iOS, `getInternetCredentials` resolves `false` for `errSecItemNotFound` but **rejects** for every other `OSStatus` — `errSecInteractionNotAllowed`, `errSecAuthFailed`, `errSecMissingEntitlement` (`RNKeychainManager.m`, `getInternetCredentialsForServer`); Android throws on decryption failures. A corrupt or truncated blob fails the same way via `JSON.parse`. In all of those cases `getSettings` swallows the error and returns `this.settings`, which is still the default object declared at `stores/SettingsStore.ts:1488` — and that object has **no `nodes` key**.

So a caller cannot tell "this user has no wallets" from "we could not read the wallets".

**2. Every launch unconditionally writes the whole blob back.**

`views/Wallet/Wallet.tsx:333`, in `componentDidMount` (`:317`):

```ts
const supportedBiometryType = await getSupportedBiometryType();
await updateSettings({ supportedBiometryType });
```

No diff check, no guard, and it runs before `handleFocus` has decided whether the app should redirect to `IntroSplash`. `updateSettings` → `applySettingsUpdate` (`:2053`) re-reads via `getSettings` (`:2058`), spreads onto it, and calls `setSettings` (`:2023`) → `Storage.setItem`.

**3. The write is destructive.**

[...]

Issue #4610, opened 2026-09-08
All wallets lost again after update to v13.2.1 on iOS/iPadOS


LOCHER-21:
### Describe the bug

All Wallets deleted AGAIN after update to v13.2.1 on iOS/iPadOS. On one or more of my device was a very outdated stand of my wallets loaded and synced to all my other devices. This was not the first time that an update deleted or changed all my existing working wallets.

### Reproduce

Update to v13.2.1

### ZEUS version

v13.2.1

### Node interface

LND (REST)

### Network

Clearnet

### Device

iPhone 15 Pro

### Device operating system

iOS 26.6.1

### Log output

```shell

```

LOCHER-21:
After setting up my wallets again all of my wallets are lost again after tired using a second device with ZEUS and the same Apple ID. The problem have to be related to the iCloud sync between devices. All of my setting are lost to. The app looks like a fresh install.

kaloudis:
@LOCHER-21 I'm sorry. This is the second time this has hit you, and you were right in #3625 that users shouldn't have to check after every update whether their wallets survived. That's on us.

You're on LND (REST), so what was lost is the wallet connection configuration (host, macaroon, settings). Your node and your funds were never touched.

Do this before anything else. If any device still shows the correct wallet list, don't open Zeus on the others yet. On the good device go to Menu > Tools > Export/Import Wallet Configurations > Export wallet configurations, set a password, and save the encrypted file off the phone. You can re-import it on every other device from the same screen.

If no device has the right list, try Menu > Tools > Export/Import Wallet Configurations > Restore legacy wallet configurations, which scans for older copies left by previous versions. It may come up empty here: the current version replaces the keychain entry outright rather than versioning it, so the immediately previous copy isn't retained. Your own backups may again be the fastest route.

Optional, and only after you've exported: turning off iCloud Keychain (Settings > your name > iCloud > Passwords) on your secondary devices stops them sharing one copy of the list. If you do, choose "Keep on My iPhone" when iOS asks, since Zeus's entries are currently part of what iCloud Keychain manages.

Mail us if you want help reconstructing anything.

kaloudis:
Your description ("a very outdated state of my wallets loaded and synced to all my other devices") was the decisive clue, and it matches what we found. Three things compound:

[...]

Issue #4403, opened 2026-08-13
Remove react-native-keychain cloudSync patch once oblador/react-native-keychain#800 ships a fix (KEY-004)


kaloudis:
## Context

PR #4385 (KEY-004) works around a bug in react-native-keychain 10.0.0. Filed upstream as https://github.com/oblador/react-native-keychain/issues/800.

On iOS, the library's `cloudSyncValue()` (`RNKeychainManager.m`) checks the `cloudSync` option for **presence**, not value. A JS `false` crosses the bridge as a non-nil `@NO` NSNumber, which is truthy as an ObjC pointer, so `{ cloudSync: false }` writes `kSecAttrSynchronizable = true`. Every keychain write in `storage/index.ts` passed `cloudSync: false`, so **all `zeus:*` keychain data — including wallet seed phrases (LND aezeed, LDK and Cashu mnemonics), macaroons, runes, and the settings blob — was synced to iCloud Keychain** for every iOS user with iCloud Keychain enabled, contrary to the device-local storage the code intended. This also retroactively explains a family of "wallet data appears on a new iPhone" reports (e.g. #3625) and the duress-wipe-doesn't-fully-wipe behavior.

## What PR #4385 ships

1. `patches/patch-keychain-cloudsync.mjs` — postinstall patch making `cloudSyncValue` respect the boolean (`[options[@"cloudSync"] boolValue]`); throws if the pattern is missing so a dependency bump can't silently reintroduce the leak.
2. `keychainDesyncMigration` — copies existing synced entries into the device-local partition (must run before `keychainCloudSyncMigration`).
3. Consent-gated purge (`utils/KeychainPurgeUtils.ts` + Tools > Keychain Cleanup + one-time Wallet offer) — deletes the synchronizable copies from iCloud, gated on an explicit "all my devices are updated" checkbox because deleting a synced item propagates to every device in the circle.
4. Native `ios/KeychainAudit.m` internet-password enumeration to drive the cleanup screen.

## Why this needs tracking

- The patch is a monkey-patch of a third-party pod at postinstall. Any `react-native-keychain` bump must re-verify the patch still applies (the patch throws on mismatch, which is the intended tripwire, but it needs a human to re-point it).
- The desync migration and purge tooling are transitional. Once every reasonably-current install has run them, the one-time offer and possibly the standalone cleanup screen can be retired.

## Exit criteria

- [ ] oblador/react-native-keychain#800 is fixed and released
[...]

PR #4617, opened 2026-09-08
fix: never persist a wallet-less settings blob over a populated one

kaloudis:
# Description

Relates to issue: **#4612** (surfaced while investigating #4610)

## Problem

`getSettings` could not distinguish a settings read that **failed** from a user who has **no** settings. On iOS `getInternetCredentials` resolves `false` only for `errSecItemNotFound`; every other `OSStatus` (locked item, missing entitlement, auth failure) rejects, and Android throws on decryption failures. A truncated or corrupt blob fails the same way in `JSON.parse`. All of those landed in the one `catch`, which only logged, so `getSettings` returned `this.settings`: the store's default object, which has no `nodes`.

The next write then persisted it. `Wallet.componentDidMount` refreshed `supportedBiometryType` through `updateSettings` on **every launch**, before anything had checked whether the settings actually loaded, so one transient read failure at startup was enough to replace the user's wallet list with the defaults. `Storage.setItem` is a delete-then-add of a single keychain item, so the previous blob is not retained, and while #4403 / #4385 are unfixed that item is synchronizable, which propagates the loss to the user's other devices.

Nothing here depends on iCloud: the same sequence loses the wallet list on Android. The iCloud bug is what turns a one-device loss into an all-device loss.

## Fix

Four guards, all in the write path:

1. **Distinguish error from absence.** `getSettings` latches `settingsLoadFailed` when the modern `Storage.getItem`, the `JSON.parse`, or the legacy `EncryptedStorage.getItem` throws, and clears it on the next successful load. `Storage.getItem` returning `false` still means "no settings" and takes the legacy / fresh-install path unchanged.
2. **Refuse writes while latched.** `setSettings` returns `false` without writing. Every `updateSettings` re-reads first, so this self-heals as soon as storage recovers; `applySettingsUpdate` already skips its derived-state updates when a write is refused.
3. **Never drop the wallet list.** `setSettings` refuses to persist a blob with no `nodes` over one that has them. `this.settings` is the comparison base rather than a fresh read, so no extra keychain round trip per write. Deleting the last wallet is the only legitimate caller and opts in via the new `SettingsWriteOptions.allowEmptyNodes`, threaded through `updateSettings`.
4. **Stop the unconditional launch write.** `applySettingsUpdate` skips the write entirely when the merged result deep-equals what is loaded, and `Wallet.componentDidMount` only calls `updateSettings` when `supportedBiometryType` actually changed. That removes one full read-merge-write of the blob from every cold start (helps #4470, and on iOS removes one delete-then-add of the shared keychain item per launch).

## Call-site audit for the empty-`nodes` guard

- `views/Settings/WalletConfiguration.tsx` deletion path: passes `{ allowEmptyNodes: true }` (the only opt-in).
- Duress/lockout wipe: no longer writes settings at all, it goes through `clearAllData()` + restart, so it is unaffected.
- `views/Settings/Wallets.tsx:78` `nodes: []` is component state, not a settings write.
- `CustodialWalletWarning`, the `EmbeddedNode/*` server pickers, and the import flow all write a non-empty `nodes` array.
- `MigrationUtils` writes go through `setSettings` with the loaded blob (non-empty when the user has wallets) or `Storage.setItem` directly, which is unchanged.

## Audit for the equality short-circuit

Skipping equal writes is only safe if no caller mutates settings in place and relies on the write. Checked: only `utils/MigrationUtils.ts` mutates in place, and it calls `setSettings` directly, which has no short-circuit. The `EmbeddedNode/*` pickers copy both the array and the node object. No caller uses `updateSettings` purely to trigger a refresh (`views/Lockscreen.tsx` and `views/Settings/SeedRecovery.tsx` set `triggerSettingsRefresh` directly).

## Migration plan

There is none to run, and that is deliberate:

- **No new persisted keys**, no change to the `zeus-settings-v2` shape, no `MOD_KEY` flag. Nothing to register in `STORAGE_KEYS` / `OTHER_KEYS` / `migrationKeys`.
- **Cohorts.** Fresh install: no persisted blob, `Storage.getItem` returns `false`, latch stays clear, writes proceed (covered by a test). Legacy blob user: unchanged path, plus the legacy read now latches instead of silently returning defaults. Modern v2 user: unchanged unless the read or parse fails, which previously destroyed the wallet list and now does nothing. User whose blob is already corrupt: writes are refused rather than overwriting it, which keeps the data available to the Keychain Recovery tool.
- **Idempotency / crash safety.** The latch is in-memory only and re-derived on every load, so a crash cannot leave it stuck.
- **Revert story.** Revert-safe. Nothing is written, deleted, or re-keyed by this change; reverting restores the old (unguarded) write behavior with no counter-migration.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

7 new tests in `stores/SettingsStore.test.ts` (`SettingsStore write guards`): failed read does not persist defaults over the wallet list, corrupt blob is not overwritten, writes resume after a later successful load, empty wallet list is refused over an existing one, the deletion opt-in still works, a fresh install with no wallets still persists, and an unchanged update performs no write. 6 of the 7 fail against the unmodified store; the seventh is the opt-in path, whose behavior is unchanged by design.

Full suite: 72 suites / 1581 tests pass, plus tsc, lint and prettier. The `MigrationUtils` mock in that file was also completed (`migrateRgsDefaultsToV2`, `migrateSwapHostsToBoltz`, `migrateRetiredSwapHosts`, `migrateOlympusHostsToZeusLsp`) so `getSettings` exercises its real path instead of throwing on an undefined mock.

Manual device matrix (pending, to be filled before merge):

- [ ] Fresh install: onboarding works, settings persist
- [ ] Upgrade from a release build with several wallets: all wallets intact, settings still writable
- [ ] Legacy blob upgrade
- [ ] Delete a wallet, and delete the last remaining wallet (the `allowEmptyNodes` path)
- [ ] Simulated read failure at launch: wallet list survives and writes resume on the next launch
- [ ] Corrupt blob: app does not overwrite it and the Keychain Recovery tool still sees the data

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Settings-layer change; applies to all backends. Device matrix above.

### Locales

No new user-facing strings; the refusals log to the console only.

### Other:

### DIFF
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -317,6 +317,53 @@ describe('SettingsStore write guards', () => {
         expect(store.settingsLoadFailed).toEqual(false);
     });
 
+    it('keeps writing when the legacy read fails on a fresh install', async () => {
+        // EncryptedStorage rejects persistently when its Android
+        // keystore-backed store cannot initialize, so refusing writes
+        // here would block onboarding forever.
+        const EncryptedStorage = require('react-native-encrypted-storage');
+        EncryptedStorage.getItem.mockRejectedValueOnce(
+            new Error('Could not initialize SharedPreferences')
+        );
+        const store = new SettingsStore();
+
+        const result = await store.updateSettings({ fiat: 'EUR' });
+
+        expect(result.fiat).toEqual('EUR');
+        expect(persistedSettings().fiat).toEqual('EUR');
+        expect(store.settingsLoadFailed).toEqual(false);
+    });
+
+    it('loads the settings blob even when a pre-read migration throws', async () => {
+        seedSettings({ nodes: [nodeA, nodeB], fiat: 'USD' });
+        const MigrationUtils = require('../utils/MigrationUtils');
+        MigrationUtils.purgeRescueKeyFiles.mockRejectedValueOnce(
+            new Error('EncryptedStorage unavailable')
+        );
+        const store = new SettingsStore();
+
+        const result = await store.updateSettings({ fiat: 'EUR' });
+
+        // The blob lives in the keychain and is still readable: a failed
+        // migration flag lookup must not strand the store on defaults.
+        expect(result.nodes).toHaveLength(2);
+        expect(persistedSettings().nodes).toHaveLength(2);
+        expect(persistedSettings().fiat).toEqual('EUR');
+    });
+
+    it('persists an update whose updater mutates settings in place', async () => {
+        seedSettings({ nodes: [nodeA], fiat: 'USD' });
+        const store = new SettingsStore();
+        await store.getSettings();
+
+        await store.updateSettings((currentSettings: any) => {
+            currentSettings.nodes[0].macaroonHex = 'cccc';
+            return { nodes: currentSettings.nodes };
+        });
+
+        expect(persistedSettings().nodes[0].macaroonHex).toEqual('cccc');
+    });
+
     it('skips the write when nothing changed', async () => {
         seedSettings({ nodes: [nodeA], fiat: 'USD' });
         const store = new SettingsStore();
--- a/stores/SettingsStore.test.ts
+++ b/stores/SettingsStore.test.ts
@@ -32,8 +32,11 @@ jest.mock('../utils/LocaleUtils', () => ({
 jest.mock('../utils/MigrationUtils', () => ({
     keychainCloudSyncMigration: jest.fn().mockResolvedValue(undefined),
     purgeRescueKeyFiles: jest.fn().mockResolvedValue(undefined),
-    migrateRgsDefaultToZeus: jest.fn().mockResolvedValue(undefined),
+    migrateRgsDefaultsToV2: jest.fn().mockResolvedValue(undefined),
+    migrateSwapHostsToBoltz: jest.fn().mockResolvedValue(undefined),
+    migrateRetiredSwapHosts: jest.fn().mockResolvedValue(undefined),
     migrateInvoiceExpiryDisplay: jest.fn().mockResolvedValue(undefined),
+    migrateOlympusHostsToZeusLsp: jest.fn().mockResolvedValue(undefined),
     legacySettingsMigrations: jest.fn().mockResolvedValue({}),
     storageMigrationV2: jest.fn().mockResolvedValue(undefined)
 }));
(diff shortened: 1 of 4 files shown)


kaloudis:
Thanks — two of the three land, and chasing the first one surfaced a fourth issue that was not reported. Fixed in 51ae0972a (to be squashed before merge), each with a regression test that fails against 467302472.

**MEDIUM: legacy-read failure permanently latches writes — valid, fixed.** Confirmed the premise: `react-native-encrypted-storage` rejects whenever its Android keystore-backed store fails to initialize (`RNEncryptedStorageModule.java`, `sharedPreferences == null`) and on any non-`errSecItemNotFound` status on iOS (`RNEncryptedStorage.m`). The Android case is persistent, not transient, so the latch would refuse every settings write forever and block onboarding on a fresh install. The observation that the branch is only reached when the modern read succeeded and returned nothing is correct, so there is no populated blob to protect there. The legacy read now logs and is treated as absent data.

**MEDIUM: equality short-circuit vs in-place mutation — valid, fixed.** No current caller mutates in place (only `utils/MigrationUtils.ts` does, and it calls `setSettings` directly, which has no short-circuit), but the comparison base being the live object is a real footgun in exactly the file meant to prevent silent data loss. The comparison is now against a `JSON.stringify` snapshot taken before the updater runs, which fails in the safe direction: a stale snapshot costs a redundant write, never a dropped one.

**MEDIUM: biometry equality check can skip the initial settings load — not applicable, no change.** The load does not ride on that write. `componentDidMount` calls `handleFocus()` immediately after the biometry branch, `initialLoad` is `true` in the initial state (`views/Wallet/Wallet.tsx:203`), and `handleFocus` therefore triggers `getSettingsAndNavigate`, whose first step is `await SettingsStore.getSettings()` (`views/Wallet/Wallet.tsx:430`). Skipping the write skips a write, not a read. This matches the limitation the review noted about only having diff hunks.

**Unreported, found while verifying the first finding:** `MigrationsUtils.purgeRescueKeyFiles()` runs before the settings read and has no error handling of its own, so a rejected `EncryptedStorage.getItem` or a failed `RNFS` unlink threw into `getSettings`'s outer catch and aborted the load **before the keychain was read at all**. The store was then left on its defaults with no latch set, and the next write persisted them: the exact wipe this PR exists to prevent, on every launch of an affected device. (`keychainCloudSyncMigration` is safe; it swallows its own errors.) The pre-load migrations are now non-fatal to the load, which also removes the last scenario where the first finding could have bitten.

PR #4618, opened 2026-09-09
backends: type the CLN responses CLNRest reads

### LINKED ISSUE #3925: types: flesh out CLNRest
To improve type safety and code clarity, it's best to avoid using `any`. Let's interfaces for the `channel` and `peers`, objects based on the expected structure from the CLN API response.

For example:
```typescript
interface IClnClosedChannel {
  peer_id: string;
  total_msat: number;
  channel_id: string;
  short_channel_id: string;
  last_commitment_txid: string;
  opener: 'local' | 'remote';
  closer: 'local' | 'remote';
  private: boolean;
  // ... other properties
}

// ...

const formattedClosedChannels = data.closedchannels.map((channel: IClnClosedChannel) => ({
  // ...
}));
```
This leverages TypeScript's strengths to catch potential errors during development and makes the code easier to understand and maintain.

TheSeydiCharyyev:
# Description

Relates to issue: #3925

`listPeerChannels`, `listClosedChannels` and `listPeers` took `any` and read fields straight off it. This adds interfaces for the three Core Lightning responses this handler consumes.

The types are not guessed. They come from the schemas shipped with CLN in `doc/schemas`, cross-checked between **v24.11** - the oldest release `CLNRest.supports()` gates on - and **v26.06.7**. A field is only non-optional here when both schemas mark it required, so everything optional in the interfaces can really be missing from a node ZEUS supports. `*_msat` values are typed as numbers because that is what the schemas' own recorded examples show on the wire, and because the mapper already does arithmetic on them.

## What the types found

Making the optional fields optional turned eleven reads into type errors. Every one of them was arithmetic on a value CLN may not send, and every one produced the string `NaN`:

- peer channels: `capacity`, `local_balance`, `remote_balance`, `total_satoshis_sent`, `total_satoshis_received`, `num_updates`, `local_chan_reserve_sat`, `remote_chan_reserve_sat`
- closed channels: `last_commitment_fee_satoshis`

They now fall back to 0. The peer-channel ones are reachable in practice: `OPENINGD` is not among the three states the mapper filters out, and a channel in that state carries no amounts yet.

## Left alone, for you to judge

Three things the schemas show that I did not change, because they are behaviour on a backend path rather than typing:

- The mapper skips `state === 'CLOSED'`, but `CLOSED` is not one of the 14 values in the `state` enum in either v24.11 or v26.06.7, so that branch looks unreachable. I typed `state` as `string` rather than the enum so the check still compiles.
- On closed channels `peer_id` is optional ("can be missing with pre-v23.05 closes", per the schema) and so is `last_commitment_txid`. Those are the two fields #3924 mapped to `remote_pubkey` and `closing_txid` to get the peer name and the close-tx link into the UI, so on an old close both quietly fall back.
- `last_commitment_fee_satoshis` is written here and read nowhere else in the repo.

## Scope

The issue asks for interfaces for the CLN API response, so that is what this covers. The `any` on the other side of the mappers - the LND-shaped objects the handlers return - is untouched.

- [ ] New feature
- [x] Bug fix
- [x] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Nine new cases in `backends/CoreLightningRequestHandler.test.ts`, extending the harness already there: the absent-amount cases on both channel mappers, the msat-to-sats conversion, the state filter, the alias lookup, and a disconnected peer that carries neither `netaddr` nor `features`. Both absent-amount cases fail without the fallbacks in the first commit; I checked by removing them.

- [ ] Android
- [ ] iOS

Neither. The changed code path only runs against a Core Lightning node, which I do not have, so a device pass would not have exercised any of it. What the change does is type-level plus the 0 fallbacks above, and that part is covered by the unit tests.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

None. I could not test against a CLN node. The field-by-field basis for the types is CLN's published schemas at the two versions named above rather than one node's output, which also covers the states a single node would not have shown me.

### Locales

### Other:

### DIFF
--- a/backends/CoreLightningRequestHandler.test.ts
+++ b/backends/CoreLightningRequestHandler.test.ts
@@ -1,4 +1,5 @@
 const mockPostRequest = jest.fn();
+const mockGetNode = jest.fn();
 
 // the handler constructs its CLNRest instance at import time, so postRequest
 // has to resolve the mock lazily
@@ -8,13 +9,21 @@ jest.mock('./CLNRest', () => ({
         postRequest(...args: Array<any>) {
             return mockPostRequest(...args);
         }
+        getNode(...args: Array<any>) {
+            return mockGetNode(...args);
+        }
     }
 }));
 // only used by getChainTransactions, but importing it pulls in the whole
 // store graph
 jest.mock('../utils/AddressUtils', () => ({ __esModule: true, default: {} }));
 
-import { getUTXOs } from './CoreLightningRequestHandler';
+import {
+    getUTXOs,
+    listClosedChannels,
+    listPeerChannels,
+    listPeers
+} from './CoreLightningRequestHandler';
 
 const listfunds = (outputs: Array<any>) => ({ outputs, channels: [] });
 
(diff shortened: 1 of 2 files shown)


TheSeydiCharyyev:
The bot found nothing to fix, so this only answers what it listed as unverifiable.

`tsconfig.json` sets `"strict": true`, and that is exactly how the eleven sites in the first commit were found: once the optional fields are marked optional, the checker rejects every expression that does arithmetic on them, and the fallbacks are what makes it compile again. `private: peer.private` is fine because the object the mapper builds has no declared type, so `boolean | undefined` is accepted there.

The three functions have a single caller, `backends/CLNRest.ts`, in its `listClosedChannels`, `listPeerChannels` and `listPeers` wrappers. It passes the raw `postRequest` result, so narrowing the parameters compiles. `yarn tsc` is green across the project, which is what the tsc job checks here too.

The schemas are `doc/schemas/listpeerchannels.json`, `listclosedchannels.json` and `listpeers.json`. One thing worth knowing if you re-check: the path changed between the two versions I used. At v24.11 they are still named `doc/schemas/lightning-listpeerchannels.json` and so on; the `lightning-` prefix is gone by v26.06.7. Requiredness comes from the `required` array on the item schema plus the conditional `allOf` / `then` blocks - `netaddr` and `features` are required only when `connected` is true, which is why they are optional here.

`channel_type` was already read as `channel_type?.names`, and `peer_id` and `closer` are passed straight through into the mapped object, so nothing dereferences them.

kaloudis (approved):
Well done.

PR #4621, opened 2026-09-09
tests: add BBQr split/join invariants for BbqrUtils

kunalKumar-13:
## Description

`utils/BbqrUtils.ts` had no tests.

It is what moves a PSBT across the air gap — `splitQRs` in
`hooks/useQRAnimation.ts` when showing, `joinQRs` in
`views/HandleAnythingQRScanner.tsx` when reading. Both ends are governed by the
[BBQr spec](https://bbqr.org/) rather than by our own choice, so an off-by-one
in the header or the base36 counter produces parts that other implementations
reject — or, worse, that reassemble into the wrong bytes.

## Changes

Seventeen tests asserting the properties the spec fixes, rather than pinning
encoded output a legitimate change may alter:

- **`intToBase36`** renders the counter as exactly two base36 digits, `00` to
  `ZZ`, and rejects `1296`, negatives and non-integers. 1295 *is* `ZZ`, so
  accepting 1296 would silently widen the field and shift the payload for every
  reader.
- **`isValidSplit` agrees with that ceiling** — otherwise a split could be
  accepted and then fail to render its own header.
- **`isValidVersion`** accepts only real QR versions, and `versionToChars`
  refuses to report a capacity for one that does not exist.
- **`validateSplitOptions`** fills in the documented defaults and keeps what the
  caller passed.
- **`encodeData`**: hex is two chars per byte with `splitMod` 2, base32 splits
  on 8, and asking for Zlib on incompressible input comes back as base32 — the
  header must report what was *actually* used, or the reader will try to inflate
  raw base32.
- **`splitQRs`** builds the 8-character `B$<enc><type><count><index>` header,
  numbers parts sequentially with the same total in each, and rejects a file
  type outside the spec.
- **Round trip**: split then join returns byte-identical data through all three
  encodings, and through a 5000-byte payload needing more than five parts.

| | |
|---|---|
| Statements | **90.51%** (105/116) |
| Branches | **78.94%** (60/76) |
| Functions | **100%** (20/20) |

## Testing

`npx jest utils/BbqrUtils.test.ts` — 17 passed. Prettier clean.

Checked against mutations rather than only run:

| mutation | result |
|---|---|
| widen the base36 ceiling to 1296 | fails `rejects values two base36 digits cannot express` |
| let `isValidSplit` allow 1296 | fails `bounds the split count at the two-digit base36 ceiling` |
| swap the count and index fields in the header | fails **four** tests, including both round trips |

### DIFF
--- a/utils/BbqrUtils.test.ts
+++ b/utils/BbqrUtils.test.ts
@@ -1,0 +1,188 @@
+import {
+    encodeData,
+    HEADER_LEN,
+    intToBase36,
+    isValidSplit,
+    isValidVersion,
+    joinQRs,
+    splitQRs,
+    validateSplitOptions,
+    versionToChars
+} from '.././utils/BbqrUtils';
+
+/**
+ * BBQr is the format Zeus uses to move a PSBT (or any payload too large for one
+ * QR) across an air gap, in `hooks/useQRAnimation.ts` when showing and in
+ * `views/HandleAnythingQRScanner.tsx` when reading. Both sides are governed by
+ * the spec rather than by our own choice, so an off-by-one in the header or the
+ * base36 counter produces parts that other BBQr implementations reject — or,
+ * worse, that reassemble into the wrong bytes.
+ *
+ * These tests assert the properties the spec fixes, and a split/join round trip,
+ * rather than pinning encoded output that a legitimate change may alter.
+ *
+ * Reference: https://bbqr.org/
+ */
+describe('BbqrUtils', () => {
+    describe('intToBase36', () => {
+        it('renders the part counter as exactly two base36 digits', () => {
+            // The header gives each counter two characters, so every value has
+            // to be zero-padded to two and never wider.
+            expect(intToBase36(0)).toBe('00');
+            expect(intToBase36(36)).toBe('10');
+            expect(intToBase36(1295)).toBe('ZZ');
+        });
+
+        it('rejects values two base36 digits cannot express', () => {
+            // 1295 is ZZ, the largest two-digit base36 number. Allowing 1296
+            // would silently produce a three-character field and shift the
+            // payload for every reader.
+            expect(() => intToBase36(1296)).toThrow('Out of range');
+            expect(() => intToBase36(-1)).toThrow('Out of range');
+            expect(() => intToBase36(1.5)).toThrow('Out of range');
+        });
+    });
+
+    describe('validity checks', () => {
+        it('bounds the split count at the two-digit base36 ceiling', () => {
+            // isValidSplit and intToBase36 have to agree, or a split can be
+            // accepted and then fail to render its own header.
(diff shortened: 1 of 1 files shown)


kaloudis:
HA BBQr might be a cooked format post Coldcard catastrophe, but we appreciate the tests nonetheless.

utils/BbqrUtils.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    encodeData,
+    HEADER_LEN,
+    intToBase36,
+    isValidSplit,
+    isValidVersion,
+    joinQRs,
+    splitQRs,
+    validateSplitOptions,
+    versionToChars
+} from '.././utils/BbqrUtils';
```
kaloudis: Nit: the sibling tests in `utils/` all use the plain relative form.

```suggestion
} from './BbqrUtils';
```

utils/BbqrUtils.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        ])('returns the original bytes through %s', (_label, encoding) => {
+            // This is the property that actually matters at the air gap: what
+            // the scanner reassembles has to be byte-identical to what was
+            // shown, whichever encoding was negotiated.
+            const original = new Uint8Array(600).map((_, i) => (i * 7) % 256);
+
+            const { parts } = splitQRs(original, 'P', {
+                encoding: encoding as any,
+                maxVersion: 10
+            });
```
kaloudis: Worth pinning the negotiated encoding here. If this payload ever stopped compressing, `encodeData` would silently fall back from 'Z' to '2', the round trip would still pass, and `pako.inflate` would no longer be exercised by anything. Asserting the split result's encoding matches the requested one holds for all three rows ('H' and '2' pass through, and this payload does compress under 'Z') and makes the test unable to degrade:

```suggestion
            const { parts, encoding: used } = splitQRs(original, 'P', {
                encoding: encoding as any,
                maxVersion: 10
            });

            // A payload that stopped compressing would silently fall back
            // from 'Z' to '2' and leave the inflate path untested.
            expect(used).toBe(encoding);
```

kaloudis (commented):
Nice work, these are the right invariants to pin, and property-style assertions instead of golden output is the right call for this module. Ran the suite locally against master: 17/17 pass, Prettier clean.

Three suggestions, none blocking. Two are inline; the third is about coverage:

The biggest remaining gap is the scan side. `joinQRs` only has happy-path coverage here, and it is the one function in this file that consumes attacker-controlled input (whatever the camera scans). The uncovered branches are its error paths: a missing part, conflicting headers across parts, a duplicate index with different content, a part index >= the declared count, an unknown encoding or filetype byte, and a header not starting with `B$`. A small negative-test block for those would be the highest-value follow-up, either in this PR or a later one.

kaloudis (approved):

PR #4628, opened 2026-09-10
fix(cashu): omit zero-valued NUT-11 tags so P2PK tokens are redeemable

### LINKED ISSUE #4627: Invalid P2PK tokens generated
### Describe the bug

Got report from a user that he could not redeem a token generated with Zeus wallet, with a P2PK lock.
After decoding the token, the lock includes:

`["n_sigs", "0"],["n_sigs_refund","0"]`

NUT-11 assumes that if the tag is set, it is positive integer. CDK mint implementation refuses to accept token with tags present and set to "0".

CDK validation fails fast, so expired lock does not make the token redeemable either.

### Reproduce

Try to swap such token against CDK mint.

### ZEUS version

v13.0.1

### Node interface

Embedded LND

### Network

Tor

### Device

iPhone 11

### Device operating system

iOS 8.1

### Log output

```shell

```

ajaysehwal:
# Description

Fixes #4627 — P2PK-locked Cashu tokens could not be redeemed against a CDK mint.

## Root cause

`parseP2PKConditions` on both native bridges hardcoded `numSigs`/`numSigsRefund` to `0` and defaulted `locktime` to `0` when unset. NUT-11 requires these optional tags to be *absent* when they don't apply — a mint receiving `["n_sigs", "0"]` or `["n_sigs_refund", "0"]` rejects the token outright (validation fails fast), and `["locktime", "0"]` marks the lock as already expired.

- `ios/CashuDevKit/CashuDevKitModule.swift`, `android/.../CashuDevKitModule.kt`: only emit `locktime`/`num_sigs`/`num_sigs_refund` when positive; also stop silently dropping `pubkeys` and always-`SigInputs` `sig_flag`.
- `cashu-cdk/types.ts`: extend `CDKP2PKCondition` with `pubkeys`/`num_sigs_refund`.
- `stores/CashuStore.ts`: never forward `locktime: 0` from the send flow.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -59,28 +59,48 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         val data = json.optJSONObject("data") ?: return null
         val pubkeyHex = data.optString("pubkey").takeIf { it.isNotEmpty() } ?: return null
 
-        val locktime = if (data.has("locktime") && !data.isNull("locktime")) {
-            data.optLong("locktime").toULong()
-        } else {
-            0UL
-        }
+        // NUT-11 tags must be omitted entirely when they don't apply. Emitting
+        // them with a value of 0 (["locktime", "0"], ["n_sigs", "0"],
+        // ["n_sigs_refund", "0"]) produces tokens that mints such as CDK refuse
+        // to redeem, and a zero locktime marks the lock as already expired.
+        val locktime = optionalPositiveULong(data, "locktime")
+        val numSigs = optionalPositiveULong(data, "num_sigs")
+        val numSigsRefund = optionalPositiveULong(data, "num_sigs_refund")
+
+        val pubkeys = data.optJSONArray("pubkeys")?.let { arr ->
+            (0 until arr.length()).mapNotNull { i ->
+                arr.optString(i).takeIf { it.isNotEmpty() }
+            }
+        } ?: emptyList()
+
         val refundKeys = data.optJSONArray("refund_keys")?.let { arr ->
             (0 until arr.length()).mapNotNull { i ->
                 arr.optString(i).takeIf { it.isNotEmpty() }
             }
         } ?: emptyList()
 
+        val sigFlag: UByte = if (data.optString("sig_flag") == "SigAll") 1.toUByte() else 0.toUByte()
+
         return SpendingConditions.P2pk(
             pubkey = pubkeyHex,
             conditions = Conditions(
                 locktime = locktime,
-                pubkeys = emptyList(),
+                pubkeys = pubkeys,
                 refundKeys = refundKeys,
-                numSigs = 0UL,
-                sigFlag = 0.toUByte(),
-                numSigsRefund = 0UL
+                numSigs = numSigs,
+                sigFlag = sigFlag,
+                numSigsRefund = numSigsRefund
             )
         )
+    }
+
+    /**
+     * Returns the value at [key] only when it is a positive integer, so that
+     * optional NUT-11 tags stay absent instead of being serialized as 0.
+     */
+    private fun optionalPositiveULong(json: JSONObject, key: String): ULong? {
+        if (!json.has(key) || json.isNull(key)) return null
+        return readPositiveLong(json, key).takeIf { it > 0L }?.toULong()
     }
 
     private fun readPositiveLong(json: JSONObject, key: String): Long {
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -81,6 +81,21 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
 
         val sigFlag: UByte = if (data.optString("sig_flag") == "SigAll") 1.toUByte() else 0.toUByte()
 
+        // NUT-11: "If n_sigs or n_sigs_refund ... exceeds the total number of
+        // keys in its pathway, the P2PK secret is malformed and the Proof
+        // MUST be rejected as unspendable." The mint only enforces this at
+        // redemption time, so an out-of-range value here would otherwise
+        // silently mint a token nobody can ever fully sign for. Main pathway
+        // has `1 (pubkey) + pubkeys.size` possible signers; refund pathway
+        // has `refundKeys.size`.
+        val maxMainSigs = 1 + pubkeys.size
+        require(numSigs == null || numSigs.toInt() <= maxMainSigs) {
+            "num_sigs ($numSigs) exceeds the number of available pubkeys ($maxMainSigs)"
+        }
+        require(numSigsRefund == null || numSigsRefund.toInt() <= refundKeys.size) {
+            "num_sigs_refund ($numSigsRefund) exceeds the number of available refund_keys (${refundKeys.size})"
+        }
+
         return SpendingConditions.P2pk(
             pubkey = pubkeyHex,
             conditions = Conditions(
@@ -956,11 +971,15 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
 
         scope.launch {
             try {
-                // Parse spending conditions if provided
+                // Parse spending conditions if provided. Only tolerate a
+                // malformed JSON *blob* silently (treat as "no conditions");
+                // a validation failure inside well-formed conditions (e.g. an
+                // impossible num_sigs) must fail loudly instead of quietly
+                // minting an unlocked proof the caller never asked for.
                 val conditions = conditionsJson?.let { json ->
-                    runCatching {
-                        parseP2PKConditions(JSONObject(json))
-                    }.getOrNull()
+                    val parsed = runCatching { JSONObject(json) }.getOrNull()
+                        ?: return@let null
+                    parseP2PKConditions(parsed)
                 }
                 val wallet = getWallet(mintUrl)
                 val proofs = wallet.mint(
--- a/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
+++ b/android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
@@ -87,12 +87,15 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
         // redemption time, so an out-of-range value here would otherwise
         // silently mint a token nobody can ever fully sign for. Main pathway
         // has `1 (pubkey) + pubkeys.size` possible signers; refund pathway
-        // has `refundKeys.size`.
+        // has `refundKeys.size`. Compared in the unsigned domain rather than
+        // via toInt(), which truncates to the low 32 bits (e.g. 4294967296
+        // -> 0) and could let a huge num_sigs wrap into a small, passing
+        // value instead of being rejected.
         val maxMainSigs = 1 + pubkeys.size
-        require(numSigs == null || numSigs.toInt() <= maxMainSigs) {
+        require(numSigs == null || numSigs <= maxMainSigs.toULong()) {
             "num_sigs ($numSigs) exceeds the number of available pubkeys ($maxMainSigs)"
         }
-        require(numSigsRefund == null || numSigsRefund.toInt() <= refundKeys.size) {
+        require(numSigsRefund == null || numSigsRefund <= refundKeys.size.toULong()) {
             "num_sigs_refund ($numSigsRefund) exceeds the number of available refund_keys (${refundKeys.size})"
         }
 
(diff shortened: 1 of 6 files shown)


ios/CashuDevKit/CashuDevKitModule.swift
```diff
@@ -42,39 +42,57 @@ class CashuDevKitModule: RCTEventEmitter {
...
+        // them with a value of 0 (`["locktime", "0"]`, `["n_sigs", "0"]`,
+        // `["n_sigs_refund", "0"]`) produces tokens that mints such as CDK
+        // refuse to redeem, and a zero locktime marks the lock as already
+        // expired.
+        let locktime = optionalPositiveUInt64(from: condData, key: "locktime")
+        let numSigs = optionalPositiveUInt64(from: condData, key: "num_sigs")
+        let numSigsRefund = optionalPositiveUInt64(from: condData, key: "num_sigs_refund")
+
+        let pubkeys: [String] = {
+            if let keys = condData["pubkeys"] as? [String] {
```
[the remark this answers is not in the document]
ajaysehwal: fixed in [c97a51f](https://github.com/ZeusLN/zeus/pull/4628/commits/c97a51ff2abdde877d1474f04749faa1e5f0d983)

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
```diff
@@ -59,30 +59,50 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
...
-            data.optLong("locktime").toULong()
-        } else {
-            0UL
-        }
+        // NUT-11 tags must be omitted entirely when they don't apply. Emitting
+        // them with a value of 0 (["locktime", "0"], ["n_sigs", "0"],
+        // ["n_sigs_refund", "0"]) produces tokens that mints such as CDK refuse
+        // to redeem, and a zero locktime marks the lock as already expired.
+        val locktime = optionalPositiveULong(data, "locktime")
+        val numSigs = optionalPositiveULong(data, "num_sigs")
```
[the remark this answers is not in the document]
ajaysehwal: fixed in [c97a51f](https://github.com/ZeusLN/zeus/pull/4628/commits/c97a51ff2abdde877d1474f04749faa1e5f0d983)

ios/CashuDevKit/CashuDevKitModule.swift
```diff
@@ -42,39 +42,57 @@ class CashuDevKitModule: RCTEventEmitter {
...
+        }()
+
+        let refundKeys: [String] = {
+            if let keys = condData["refund_keys"] as? [String] {
+                return keys.filter { !$0.isEmpty }
+            }
+            return []
+        }()
+
+        let sigFlag: UInt8 = (condData["sig_flag"] as? String) == "SigAll" ? 1 : 0
```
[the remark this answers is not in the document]
ajaysehwal: verified against the actual upstream cdk-ffi source that the sig_flag 0/1 mapping is already correct, so there was nothing to change there

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
```diff
@@ -59,30 +59,65 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
...
+
+        // NUT-11: "If n_sigs or n_sigs_refund ... exceeds the total number of
+        // keys in its pathway, the P2PK secret is malformed and the Proof
+        // MUST be rejected as unspendable." The mint only enforces this at
+        // redemption time, so an out-of-range value here would otherwise
+        // silently mint a token nobody can ever fully sign for. Main pathway
+        // has `1 (pubkey) + pubkeys.size` possible signers; refund pathway
+        // has `refundKeys.size`.
+        val maxMainSigs = 1 + pubkeys.size
+        require(numSigs == null || numSigs.toInt() <= maxMainSigs) {
```
kaloudis: `toInt()` wraps for values above `Int.MAX_VALUE`: `readPositiveLong` allows anything up to `Long.MAX_VALUE`, so e.g. `num_sigs = 4294967296` becomes `0` here and passes the check, minting exactly the unspendable token this guard was added to block. Values in the 2^31..2^32 range wrap negative and also pass. Comparing in the unsigned domain avoids it:

```suggestion
        require(numSigs == null || numSigs <= maxMainSigs.toULong()) {
```

Same for the `numSigsRefund` check below (`numSigsRefund <= refundKeys.size.toULong()`). Not reachable from our own JS today (we only send `pubkey` and `locktime`), but this code exists purely as defensive validation, so it should be watertight.

ios/CashuDevKit/CashuDevKitModule.swift
```diff
@@ -41,40 +41,90 @@ class CashuDevKitModule: RCTEventEmitter {
...
+
+        // NUT-11: "If n_sigs or n_sigs_refund ... exceeds the total number of
+        // keys in its pathway, the P2PK secret is malformed and the Proof
+        // MUST be rejected as unspendable." The mint only enforces this at
+        // redemption time, so an out-of-range value here would otherwise
+        // silently mint a token nobody can ever fully sign for. Main pathway
+        // has `1 (pubkey) + pubkeys.count` possible signers; refund pathway
+        // has `refundKeys.count`.
+        let maxMainSigs = 1 + pubkeys.count
+        if let numSigs, Int(numSigs) > maxMainSigs {
```
kaloudis: `Int(numSigs)` traps at runtime when the value exceeds `Int64.max`, which is reachable through `readPositiveUInt64`'s string path (`UInt64("18446744073709551615")` parses successfully). That turns a malformed conditions payload into a hard crash instead of a rejected promise. Comparing in the unsigned domain avoids the conversion:

```suggestion
        if let numSigs, numSigs > UInt64(maxMainSigs) {
```

Same for the `numSigsRefund` check below (`numSigsRefund > UInt64(refundKeys.count)`).

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
```diff
@@ -936,11 +971,15 @@ class CashuDevKitModule(private val reactContext: ReactApplicationContext) :
...
+                // Parse spending conditions if provided. Only tolerate a
+                // malformed JSON *blob* silently (treat as "no conditions");
+                // a validation failure inside well-formed conditions (e.g. an
+                // impossible num_sigs) must fail loudly instead of quietly
+                // minting an unlocked proof the caller never asked for.
                 val conditions = conditionsJson?.let { json ->
-                    runCatching {
-                        parseP2PKConditions(JSONObject(json))
-                    }.getOrNull()
+                    val parsed = runCatching { JSONObject(json) }.getOrNull()
```
kaloudis, line 983: Non-blocking, pre-existing: an unparseable blob (or `kind != "P2PK"`) still silently degrades to "no conditions", so a caller that asked for a locked mint would get unlocked proofs. Fine for now since the JS side builds this JSON itself via `JSON.stringify` (and currently never passes conditions to `mint()` at all), but a stricter follow-up could reject here too, matching the loud-failure policy this PR introduces for validation errors. Same applies to the iOS mint path.

utils/CashuUtils.ts
```diff
@@ -277,6 +277,33 @@ class CashuUtils {
...
+            typeof (error as { message?: unknown }).message === 'string'
+                ? (error as { message: string }).message.toLowerCase()
+                : '';
+
+        return (
+            message.includes('p2pk') &&
+            message.includes('signature') &&
+            (message.includes('missing') ||
+                message.includes('not provided') ||
+                message.includes('required'))
```
kaloudis: Nit: this breadth means a partial-signature style error (e.g. "2 of 3 signatures required") would also classify as locked-to-another-wallet. Acceptable since Zeus never creates multisig locks and the union is needed to cover CDK's wording variance across versions; just flagging that it's intentionally broad in case a future multisig feature needs to tighten it.

kaloudis (commented):
Took a close look at the native parsing changes. Verified the `Conditions` bindings accept nulls on both platforms (boxed `kotlin.ULong` params in cdk-jvm.jar, `UInt64?` fields in CashuDevKit.swift), that `readPositiveLong`/`readPositiveUInt64` clamp negatives so the `optionalPositive*` helpers behave, and that all four `parseP2PKConditions` call sites sit inside catch blocks that reject the promise. Approach is right and matches the evidence in #4627.

Two small hardening fixes requested on the new range validation, plus two non-blocking notes below.

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
ajaysehwal: fixed in [897ce6a](https://github.com/ZeusLN/zeus/pull/4628/commits/897ce6a7bd48c852d13c4c4c435738c2cb79215d)

android/app/src/main/java/com/zeus/cashudevkit/CashuDevKitModule.kt
ajaysehwal: fixed in [897ce6a](https://github.com/ZeusLN/zeus/pull/4628/commits/897ce6a7bd48c852d13c4c4c435738c2cb79215d)

kaloudis (commented):
Round 2 verified: 897ce6a applies both requested changes exactly.

- Kotlin now compares in the unsigned domain (`numSigs <= maxMainSigs.toULong()`, `numSigsRefund <= refundKeys.size.toULong()`), closing the `toInt()` wrap bypass for values at or above 2^31.
- Swift now compares `numSigs > UInt64(maxMainSigs)` / `numSigsRefund > UInt64(refundKeys.count)`, avoiding the `Int(numSigs)` runtime trap reachable through `readPositiveUInt64`'s string path.

CI is green across tsc/lint/prettier/test.

utACK 897ce6a

One clarification on the mint-path thread: the loud-failure split for validation errors was already in place before 897ce6a; the residual there (a malformed JSON blob or non-P2PK kind still silently degrades to no conditions) is unchanged, and that remains fine as a potential follow-up rather than a blocker.

Last thing before merge: the closing evidence for #4627 is an end-to-end redeem against a CDK mint. Can you confirm your Android/iOS testing included (1) locking a token to another wallet's pubkey and redeeming it there, and (2) attempting redeem from a wallet without the key and seeing the new locked-token error surface?

kaloudis (commented):
Device tests done on both bridges, end to end against a local cdk-mintd 0.18.0 (fakewallet backend): iPhone 16 Pro simulator (iOS 18.1) and Pixel 9 Pro emulator (API 35), both running LDK Node wallets with the mint added over localhost.

**Fix verified on both platforms:**

1. iOS bridge: locked 100 sats to the Android wallet's Cashu pubkey (Forever). Decoding the resulting cashuB token shows every proof secret carrying `[["sigflag","SIG_INPUTS"]]` only, with no zero-valued `locktime`/`n_sigs`/`n_sigs_refund` tags. The Android wallet redeemed it against the mint successfully.
2. Android bridge: same flow in reverse. Token decodes to the identical clean tag set; the iOS wallet redeemed it successfully.
3. Before/after control: through a stale-install accident I also exercised the pre-fix Kotlin bridge (old APK, new JS) against the same mint. Its token decoded to `[["locktime","0"],["n_sigs","0"],["n_sigs_refund","0"],["sigflag","SIG_INPUTS"]]` and could not be redeemed even by the wallet holding the correct key (CDK: "Signature missing or invalid"), which is #4627 exactly. Rebuilding from this branch and repeating the identical flow produces the clean token that redeems fine, so the tag omission is confirmed as the fix.
4. Cosmetic confirmation: the zero-locktime token renders "Locked until Wed, Dec 31 '69" in Cashu Token Info, while fixed tokens correctly show "Forever".

**One more requested change surfaced by testing:** `isTokenLockedError` never fires on CDK 0.18. Redeeming a token locked to a different wallet's key surfaces the raw error, because cdk 0.18's wallet-side NUT-11 error message is exactly `Signature missing or invalid` (nut11 `SignatureMissingOrInvalid`), which contains no `p2pk` substring, so the classifier's `message.includes('p2pk')` gate rejects it and users see the raw string instead of `stores.CashuStore.claimError.lockedToWallet`. Suggest adding the exact phrase as a disjunct:

```ts
return (
    message.includes('signature missing or invalid') ||
    (message.includes('p2pk') &&
        message.includes('signature') &&
        (message.includes('missing') ||
            message.includes('not provided') ||
            message.includes('required')))
);
```

plus a unit case pinning the 0.18 wording. Scoping stays safe since the helper is only consulted in the claimToken catch path.

Pre-existing and non-blocking, noting for a follow-up: READ TOKEN FROM CLIPBOARD with non-token clipboard content surfaces an uncaught promise rejection toast ("Error: Value provided was not a valid Bitcoin address or Lightning Invoice") rather than a handled error.

With the wording added to `isTokenLockedError` (and a test), this gets my tACK.

kaloudis (approved):
tACK

PR #4629, opened 2026-09-10
fix(cashu): stop amount from being clobbered by lock-settings round trip

ajaysehwal:
# Description

Fixes a UI bug in the Send Ecash flow: after locking a token to a pubkey, tapping the amount field to open the keypad and confirming a new amount would not reliably show up in the field.

## Root cause

`SendEcash.handleScreenFocus` restores `value`/`satAmount` from `route.params` back into component state on every return from the lock-settings sub-flow (`SendEcash` → `CashuLockSettings` → possibly `Contacts`/`ContactDetails` → back). But `SendEcash` never unmounts while that sub-flow is open (single native stack navigator), so its own component state is already the live, correct amount the entire time — nothing needs restoring. The restore-from-params instead reapplies a stale copy, captured back when the "Lock to Pubkey" button was first pressed, over whatever the user has since typed into the keypad.

## Fix

In `views/Cashu/SendEcash.tsx`'s `handleScreenFocus`:
- Stop restoring `value`/`satAmount` from `route.params` into state — the two lines that caused the clobbering.
- Fix an adjacent bug in the same block: `navigation.setParams()` merges with the route's *existing* params rather than replacing them, so `delete updatedParams.fromLockSettings` before calling `setParams` never actually cleared it from the route - the merge just left the old `fromLockSettings: true` in place. Left unfixed, this whole restoration block would keep re-firing on *every* future focus of the screen (not just returns from locking), reapplying stale `pubkey`/`duration`/`memo`/`contactName` each time. Explicitly setting `fromLockSettings: undefined` does survive the merge and actually clears it.

Single file, single method, ~20-line diff — no other screens need touching, since the amount never needs to leave `SendEcash`'s own state in the first place.

## Testing

- Manual verification recommended: lock to a pubkey, tap the amount field, enter a new amount via the keypad, confirm, and verify it's reflected correctly; also verify returning to Send Ecash from elsewhere in the app afterward doesn't silently re-apply an old lock pubkey/duration

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Cashu/SendEcash.tsx
+++ b/views/Cashu/SendEcash.tsx
@@ -124,13 +124,18 @@ export default class SendEcash extends React.Component<
         const params: SendEcashParams = route.params || {};
 
         if (params.fromLockSettings) {
+            // `value`/`satAmount` are intentionally left out of this
+            // restoration: this screen is never unmounted while the
+            // lock-settings/contact-picker sub-flow is open underneath it,
+            // so its own state is always the live, up-to-date amount.
+            // Restoring it from params here would instead reapply a stale
+            // copy - captured back when the lock button was first pressed -
+            // over whatever the user has since typed into the keypad.
             const stateUpdate: Partial<SendEcashState> = {
                 pubkey: params.pubkey ?? this.state.pubkey,
                 duration: params.duration ?? this.state.duration,
                 locktime: this.convertDurationToSeconds(params.duration),
                 memo: params.memo ?? this.state.memo,
-                value: params.value ?? this.state.value,
-                satAmount: params.satAmount ?? this.state.satAmount,
                 contactName: params.contactName ?? this.state.contactName,
                 showCustomDuration:
                     params.showCustomDuration ?? this.state.showCustomDuration,
@@ -141,9 +146,14 @@ export default class SendEcash extends React.Component<
                     params.customDurationUnit ?? this.state.customDurationUnit
             };
             this.setState(stateUpdate as SendEcashState, () => {
-                const updatedParams = { ...params };
-                delete updatedParams.fromLockSettings;
-                this.props.navigation.setParams(updatedParams);
+                // `setParams` merges with the route's existing params rather
+                // than replacing them, so `delete`-ing a key here wouldn't
+                // actually clear it from the route - only assigning
+                // `undefined` does.
+                this.props.navigation.setParams({
+                    ...params,
+                    fromLockSettings: undefined
+                });
             });
         }
     };


kaloudis (commented):
utACK 33dbd7d90b20

Verified both root-cause claims against the sources rather than taking the description on faith:

- `SET_PARAMS` merges (`{ ...r.params, ...action.payload.params }` in `@react-navigation/routers` BaseRouter), so the old delete-then-`setParams` really never cleared `fromLockSettings`, and assigning `undefined` explicitly does survive the merge.
- The repro path checks out: `AmountInput` pushes an `AmountKeypad` screen, so confirming an amount pops back to `SendEcash` and re-fires the focus listener with the stuck flag, which then re-applied the stale `value`/`satAmount` captured when the lock button was first pressed. `SendEcash` stays mounted under the whole sub-flow, so its own state is the live amount and the removed restore lines were never needed.

Both halves of the fix are necessary: clearing the flag alone would still allow one stale restore, and removing the amount lines alone would leave memo/pubkey/duration re-clobbering on every future focus.

Nits, non-blocking:

- The `...params` spread in the new `setParams` call is redundant: `setParams` merges, and `params` already is the route's params, so `setParams({ fromLockSettings: undefined })` alone is equivalent. Skipping the spread also avoids re-affirming the now-unused stale `value`/`satAmount` keys that linger in `route.params`.
- The 9-line comment records a real invariant (the screen never unmounts, its state is authoritative) and is worth keeping, but could be trimmed a bit.

The same `setParams`-merge misunderstanding exists in two other spots (`resetForm` here, and `CashuLockSettings.handleFocus`, where it plausibly resurrects a cleared contact). Filed #4633 and #4634 as follow-ups rather than expanding this PR.

Will still need an Android device test before merge.

kaloudis (commented):
Android device test done (Pixel 9 Pro emulator, API 35, LDK Node ecash wallet against a local cdk-mintd 0.18 fakewallet mint), covering the platform the earlier testing missed:

- Control on pre-fix JS: set 100 sats, do the lock round trip (pubkey + 1 Day), reopen the amount keypad, enter a new amount, confirm: the display snaps back to 100 sats, reproducing the reported clobber.
- On 33dbd7d90: the same flow keeps the newly entered amount (20 sats) after the round trip, the lock chip still shows the restored pubkey and duration, and SEND ECASH produces a 20 sat locked token with the expected P2PK secret.

Both halves of the fix behave as intended on device: stale value/satAmount no longer restored, and the fromLockSettings flag actually clears so later focuses don't re-apply anything.

tACK 33dbd7d90

kaloudis (approved):

PR #4638, opened 2026-09-11
fix(cashu): make Lock Settings Back cancel instead of dropping the lock

### LINKED ISSUE #4637: Cashu: Back from Lock Settings drops the lock but keeps the contact, leaving LOCK disabled
### Steps to reproduce
1. Send ecash → Lock to Pubkey → pick a contact → LOCK. SendEcash shows the lock.
2. Tap the lock button again to reopen Lock Settings.
3. Tap the header Back button.
4. SendEcash no longer shows the lock ("Lock to Pubkey").
5. Tap the lock button again: Lock Settings shows the contact chip, but LOCK is disabled.

### Expected
Back cancels the change and the existing lock stays. Or, if Back is meant to remove the lock, it removes all of it, including the contact.

### Root cause
`CashuLockSettings.onBack` sends `popTo('SendEcash', { fromLockSettings: true, pubkey: '', duration: '', showCustomDuration: false })`. It clears the pubkey but not `contactName`. `SendEcash.handleScreenFocus` restores with `??`, so `''` wipes the pubkey while the old `contactName` stays. SendEcash then holds a contact name with no pubkey. The lock button on SendEcash only checks `pubkey`, and when you open Lock Settings again the chip only checks `contactName` while LOCK needs a pubkey.

Two related points:
- The hardware/gesture back keeps the lock. Only the header Back clears it, so the two behave differently.
- The header Back is currently the only way to remove a lock (there's no remove control, and LOCK is disabled without a pubkey), so a fix has to keep some way to unlock.

### Suggested fix
Make the header Back cancel with `popTo('SendEcash')` and no params, the same as the hardware back. Add an explicit "Remove lock" button in Lock Settings when it's opened with an existing lock, which sends the full clearing params (pubkey, contactName, duration and the custom duration fields). Separately, never let a contact name exist without a pubkey in either screen.

### ZEUS version

13.2.2-alpha

### Node interface

LDK Node

### Device

iphone 17

### Device operating system

iOS 26.6.2

### Log output

``` shell

```

ajaysehwal:
# Description

Fixes #4637.

## Root cause

`CashuLockSettings.onBack` sent SendEcash `{ fromLockSettings: true, pubkey: '', duration: '', showCustomDuration: false }` — clearing the pubkey but never sending `contactName`. SendEcash's focus handler restores each field with `params.field ?? this.state.field`, so the explicit `''` wiped the pubkey while the omitted `contactName` fell through to the old value and survived.

Repro: lock a token to a contact, open Lock Settings again, tap the header Back button. SendEcash loses the lock (shows "Lock to Pubkey" again). Reopening Lock Settings shows the contact chip with no pubkey behind it, and LOCK stays disabled since it requires a valid pubkey — there was no way out of that state except restarting the flow.

The header Back also behaved differently from the hardware/gesture back, which left `route.params` alone and kept the lock.

## Fix

- `CashuLockSettings.onBack` now pops back to SendEcash with no params at all, so the existing lock (if any) is left untouched — the same behavior as the hardware/gesture back. `popTo` (not `goBack`) also correctly unwinds a second Lock Settings screen that the contact picker can push onto the stack.
- Removing a lock is now an explicit **Remove lock** button, shown only when Lock Settings opens with an existing lock (`hasExistingLock`). It clears every lock field, including `contactName`.
- SendEcash restores the lock through a new `resolveLockTarget` helper in `utils/CashuUtils.ts`, which never keeps a contact name without the pubkey it was chosen with. This closes the underlying invariant, not just this one code path, so the broken state can't come back through a different route later.

## Testing

`resolveLockTarget` is a pure function extracted specifically so this fix is unit-tested; `utils/CashuUtils.test.ts` covers it directly, including the exact case from #4637. Verified the regression tests fail against the old `??`-only logic and pass with the fix.

Manually verified: lock a token to a contact → reopen Lock Settings → header Back → the lock is still shown on SendEcash. Reopening Lock Settings again shows the same contact with LOCK enabled. Remove lock clears the contact chip and returns SendEcash to "Lock to Pubkey".

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

### DIFF
--- a/views/Cashu/CashuLockSettings.tsx
+++ b/views/Cashu/CashuLockSettings.tsx
@@ -33,8 +33,6 @@ export interface CashuLockSettingsParams extends SendEcashParams {
     // Forwarded through the contact-picker round trip (CashuLockSettings ->
     // Contacts/ContactDetails -> CashuLockSettings) since that hop pushes a
     // second instance whose own currentLockPubkey param is empty - without
-    // this, "Remove lock" would wrongly disappear on that instance (#4637
-    // follow-up flagged in review).
     hasExistingLock?: boolean;
 }
 interface CashuLockSettingsProps {
@@ -550,7 +548,8 @@ export default class CashuLockSettings extends React.Component<
                                     showCustomDuration,
                                     customDurationValue,
                                     customDurationUnit,
-                                    selectedDurationIndex
+                                    selectedDurationIndex,
+                                    hasExistingLock
                                 } = this.state;
                                 navigation.navigate('Contacts', {
                                     SendScreen: true,
@@ -563,7 +562,7 @@ export default class CashuLockSettings extends React.Component<
                                     customDurationValue,
                                     customDurationUnit,
                                     selectedDurationIndex,
-                                    hasExistingLock: this.state.hasExistingLock
+                                    hasExistingLock
                                 });
                             }}
                             style={{ position: 'absolute', right: 10 }}
--- a/views/Cashu/CashuLockSettings.tsx
+++ b/views/Cashu/CashuLockSettings.tsx
@@ -29,6 +29,13 @@ export interface CashuLockSettingsParams extends SendEcashParams {
     destination?: string | null;
     hasCashuPubkey?: boolean;
     selectedDurationIndex?: number;
+    // Whether the flow that opened this screen already had a saved lock.
+    // Forwarded through the contact-picker round trip (CashuLockSettings ->
+    // Contacts/ContactDetails -> CashuLockSettings) since that hop pushes a
+    // second instance whose own currentLockPubkey param is empty - without
+    // this, "Remove lock" would wrongly disappear on that instance (#4637
+    // follow-up flagged in review).
+    hasExistingLock?: boolean;
 }
 interface CashuLockSettingsProps {
     navigation: NativeStackNavigationProp<any, any>;
@@ -50,6 +57,7 @@ interface CashuLockSettingsState {
     hasClipboardContent: boolean;
     isPubkeyValid: boolean;
     selectedDurationIndex: number;
+    hasExistingLock: boolean;
 }
 
 const TIME_UNITS: string[] = [
@@ -87,7 +95,8 @@ export default class CashuLockSettings extends React.Component<
             customDurationValue,
             customDurationUnit,
             duration,
-            selectedDurationIndex
+            selectedDurationIndex,
+            hasExistingLock
         } = route.params || {};
 
         this.state = {
@@ -111,7 +120,11 @@ export default class CashuLockSettings extends React.Component<
                     ? selectedDurationIndex
                     : currentDuration
                     ? DURATION_OPTIONS.indexOf(currentDuration)
-                    : 0
+                    : 0,
+            // A lock exists either if this instance was opened with one
+            // directly (currentLockPubkey) or if the contact-picker round
+            // trip forwarded that an earlier instance had one.
+            hasExistingLock: !!currentLockPubkey || !!hasExistingLock
         };
         this.handleContactSelection = this.handleContactSelection.bind(this);
     }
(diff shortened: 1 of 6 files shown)


views/Cashu/CashuLockSettings.tsx
```diff
@@ -358,12 +361,26 @@ export default class CashuLockSettings extends React.Component<
         return isPubkeyValid && selectedDurationIndex !== undefined;
     };
 
+    // Back cancels: return without params so SendEcash keeps whatever lock
+    // it already has, the same as the hardware/gesture back. popTo (rather
+    // than goBack) also unwinds a second Lock Settings screen pushed by the
+    // contact picker. Removing a lock is the explicit removeLock action.
     onBack = () => {
+        this.props.navigation.popTo('SendEcash');
```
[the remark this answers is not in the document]
ajaysehwal: This isn't correct for the version in this repo — @react-navigation/routers 7.6.4's `POP_TO` case does `params = createParamsFromAction(...)` when no `merge` option is given, which replaces the target route's params, not "leaves them untouched." I ran the actual `StackActions.popTo('SendEcash')` through the installed reducer to confirm:

    dispatched action: {"type":"POP_TO","payload":{"name":"SendEcash"}}
    SendEcash route.params after Back: undefined

So the stale-fromLockSettings-reapplied scenario can't happen. The "conversely" branch doesn't apply either — `handleScreenFocus` already does `route.params || {}` before anything touches `params`, and `resolveLockTarget` is only called inside `if (params.fromLockSettings)`, so it's never reached with an undefined `params`. No change needed here.

views/Cashu/CashuLockSettings.tsx
```diff
@@ -111,7 +112,9 @@ export default class CashuLockSettings extends React.Component<
                     ? selectedDurationIndex
                     : currentDuration
                     ? DURATION_OPTIONS.indexOf(currentDuration)
-                    : 0
+                    : 0,
+            // only a lock that already exists can be removed
+            hasExistingLock: !!currentLockPubkey
```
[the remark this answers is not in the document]
ajaysehwal: Good catch; this one's real. Neither `CashuLockSettings.tsx`'s navigate-to-Contacts call nor `Contacts.tsx`'s two navigate-to-CashuLockSettings calls forwarded anything about an existing lock, so the picker's second CashuLockSettings instance always computed `hasExistingLock: false` from its empty `currentLockPubkey`. Fixed by threading a `hasExistingLock` param through both hops (CashuLockSettings.tsx, Contacts.tsx); ContactDetails.tsx needed no change since it already spreads `...cashuLockData`. Verified end-to-end with a router simulation that the " Remove lock now stays visible on the second screen after picking a new contact.

kaloudis:
Can you rebase as we just merged #4636?

views/Cashu/CashuLockSettings.tsx
```diff
@@ -29,6 +29,13 @@ export interface CashuLockSettingsParams extends SendEcashParams {
     destination?: string | null;
     hasCashuPubkey?: boolean;
     selectedDurationIndex?: number;
+    // Whether the flow that opened this screen already had a saved lock.
+    // Forwarded through the contact-picker round trip (CashuLockSettings ->
+    // Contacts/ContactDetails -> CashuLockSettings) since that hop pushes a
+    // second instance whose own currentLockPubkey param is empty - without
+    // this, "Remove lock" would wrongly disappear on that instance (#4637
+    // follow-up flagged in review).
```
kaloudis: nit: the closing "(#4637 follow-up flagged in review)" is review-process narration that becomes noise once merged. The first sentence plus the second-instance explanation already carry everything the next reader needs; suggest ending the comment at "...would wrongly disappear on that instance (#4637)."

views/Cashu/CashuLockSettings.tsx
```diff
@@ -534,7 +562,8 @@ export default class CashuLockSettings extends React.Component<
                                     showCustomDuration,
                                     customDurationValue,
                                     customDurationUnit,
-                                    selectedDurationIndex
+                                    selectedDurationIndex,
+                                    hasExistingLock: this.state.hasExistingLock
```
kaloudis: nit: the nine sibling fields here are destructured from `this.state` a few lines up; adding `hasExistingLock` to that destructure and passing it bare would match the surrounding style.

kaloudis (commented):
Two nits, both non-blocking. Verified the popTo-without-params semantics independently in @react-navigation/routers 7.6.4 (POP_TO with no merge replaces the target route's params via createParamsFromAction), so the Back-as-cancel approach is sound, and the hasExistingLock threading reaches the picker-pushed second instance through both Contacts branches and the ...cashuLockData spread in ContactDetails. New resolveLockTarget tests pass at this head.

views/Cashu/CashuLockSettings.tsx
ajaysehwal: fixed in [6a2487a](https://github.com/ZeusLN/zeus/pull/4638/commits/6a2487ac771328489cd63a7d4e72f57570375bc4)

views/Cashu/CashuLockSettings.tsx
ajaysehwal: fixed in [6a2487a](https://github.com/ZeusLN/zeus/pull/4638/commits/6a2487ac771328489cd63a7d4e72f57570375bc4)

kaloudis (approved):
tACK

PR #4644, opened 2026-09-12
fix(channels): require a host when connecting a peer

kaloudis:
# Description

|Before|After (backstop)|After (with guard)|
|-|-|-|
|Simulator Screenshot - iPhone 17 Pro Max - 2026-09-12 at 07 33 09|Simulator Screenshot - iPhone 17 Pro Max - 2026-09-12 at 07 35 47|Simulator Screenshot - iPhone 17 Pro Max - 2026-09-12 at 07 37 00|

Reported via user feedback: pressing **Connect Peer** with a valid pubkey but an empty Host field surfaced LND's raw proto error instead of a helpful message:

```
proto: (line 1:180): invalid value for uint64 type: ""
```

Two compounding bugs:

1. **Stale validity flags in the view.** Switching the Peer dropdown to Custom clears `node_pubkey_string` and `host` to `''` but left `isNodePubkeyValid` / `isNodeHostValid` at `true` (validated against the Olympus defaults). After pasting a pubkey and leaving Host empty, the button stayed enabled. Now the flags are reset whenever the peer destination changes.

2. **Empty-host fallthrough in the store.** `ChannelsStore.connectPeer` treats an empty host as "peer already connected" and queues a `channelRequest`, which the store reaction turns into an `openChannelSync` call, even in connect-peer-only mode, where there is no funding amount. The empty `local_funding_amount` is the uint64 field LND choked on. In connect-peer-only mode the store now surfaces a friendly, localized "host is required" error instead of hitting the backend. The empty-host fallthrough for the channel-open flow is preserved.

Added regression tests for both the guard and the preserved fallthrough; the empty-host guard test fails against the previous store code.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

New unit tests cover `ChannelsStore.connectPeer` (empty-host guard in connect-peer-only mode, preserved channel-open fallthrough, and the connected-peer happy path).

- [ ] Android
- [x] iOS (iPhone 17 Pro Max simulator, iOS 26.5)

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

New string: `views.OpenChannel.hostRequired`.

### Other:


stores/ChannelsStore.ts
```diff
@@ -788,6 +788,20 @@ export default class ChannelsStore {
         if (!silent) this.connectingToPeer = true;
 
         if (!request.host) {
+            if (connectPeerOnly) {
+                // there is no address to dial, so surface an error
+                // instead of sending the backend a malformed request
+                if (!silent) {
```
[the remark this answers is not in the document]
kaloudis: Declining this one as unreachable. The only `silent: true` caller in the codebase is `LSPStore.getLSPInfo`, and its host is built as `` `${method.address}:${method.port}` ``, a template literal that can never be an empty string. It also ignores the return value entirely (`try { await ... } catch {}`, a best-effort pre-connect before the LSP flow). The other three callers (OpenChannel, LSPS1, InvoicesStore) are all non-silent, so they receive the error state.

Setting `errorPeerConnect` regardless of `silent` would be a regression: `silent` exists specifically so background LSP pre-connects don't raise the error banner in whatever view is mounted. And note that the previous behavior in this hypothetical was strictly worse: it queued a channel request with an empty funding amount and produced the proto error this PR fixes.

views/OpenChannel.tsx
```diff
@@ -682,13 +682,23 @@ export default class OpenChannel extends React.Component<
...
+                                                )
                                         });
                                     } else {
                                         this.setState({
                                             channelDestination: 'Custom',
                                             node_pubkey_string: '',
-                                            host: ''
+                                            host: '',
+                                            isNodePubkeyValid: false,
+                                            isNodeHostValid: false
```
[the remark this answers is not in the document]
kaloudis: Two points here:

1. The suggested fix is a no-op: `ValidationUtils.validateNodeHost('')` returns `false`, which is exactly the value being set.

2. The empty-host channel-open path this would "block" only existed through the stale-flag bug this PR fixes. In released code, an empty host is already blocked the moment the field is touched (`onChangeText` validates `''` to false and disables the button) and via route params (`initFromProps` validates the resolved host). This change just makes the dropdown path consistent with every other path. There's also no visible error on the untouched form, only the disabled button, same as the equally-empty pubkey field.

Deliberately supporting empty-host opens to already-connected peers (which LND allows, and which the store fallthrough exists for) would be a product decision for a follow-up, e.g. relaxing `isInvalidPeer` to ignore host validity when `!connectPeerOnly && host === ''`.

kaloudis:
Pushed 6ed5feff6: on regtest, an empty host is now allowed for channel opens, since peers there are local and typically already connected; the store's existing empty-host fallthrough opens directly to the connected peer. Connect-peer-only mode still requires a host on every network, as there is no address to dial without one.

kaloudis:
Amended the regtest commit (6ed5feff6 -> 0d7236d83) to extend the bypass to the Connect Peer tab. On regtest, an empty host is now accepted in both modes: channel opens use the store's empty-host fallthrough to the already-connected peer, and peer connects pass the empty host through to the backend, which reports the actual peer state (CLN connects via gossip since CLNRest omits the empty host and port; LND returns already-connected or a dial failure). All other networks keep the host requirement.

kaloudis:
Dropped the regtest exemption commit entirely; the branch is back to 810619e83. Polar already connects every Lightning node to every other one automatically, both at network start and when a single new node is started into a running network (monitorStartup runs connectAllPeers over the whole network), so peers on a regtest network are pre-meshed and the exemption served no real workflow. Host is required in all modes on all networks.

stores/ChannelsStore.ts
```diff
@@ -788,6 +788,20 @@ export default class ChannelsStore {
         if (!silent) this.connectingToPeer = true;
 
         if (!request.host) {
```
ajaysehwal, line 790: @kaloudis, small follow-up worth considering: if (!request.host) here only catches an empty host — it doesn’t validate the host’s actual format.

ValidationUtils.validateNodeHost() already exists and is used for the manual “Custom” host input in OpenChannel.tsx, but connectPeer itself doesn’t call it. This matters because some callers build host programmatically without UI validation — e.g. LSPStore.ts:304 and InvoicesStore.ts:390 construct it as ${method.address}:${method.port} from server-supplied data.

If method.address or method.port is missing/malformed, that produces a non-empty but invalid host that bypasses this guard and falls through to BackendUtils.connectPeer, potentially surfacing the same raw backend error this PR is trying to avoid.

kaloudis: Thanks for digging into the call sites. Host format validation is already handled at the view layer: every intake path in OpenChannel runs `ValidationUtils.validateNodeHost()` (manual entry, clipboard import, route params, LSP config selection), invalid input renders in the error color, and the connect button is disabled while the host is invalid. So for any user-driven flow, the validation and the user-facing error already exist.

The store isn't the right layer to repeat it:

1. The callers you found are headless and deliberately silent. The LSPStore call runs with `silent = true` inside an empty catch, and the InvoicesStore call's error flags only render on the OpenChannel screen, which resets them on mount. There's no user to surface a validation error to, so a garbage host would just fail silently at validation instead of silently at the dial.
2. `validateNodeHost()` would miss the most likely bad shape anyway. A missing `method.address` produces `undefined:9735`, and `"undefined"` is a syntactically valid hostname, so it passes. The missing-field information is destroyed by the template string before any validator sees it.
3. Strict format checks in the store would add false-rejection risk for unusual-but-dialable hosts on every caller, which the empty-host guard doesn't have.

The residual issue you identified is real though, and the right fix is at the construction sites: check `method?.address && method?.port` before building the string. Happy to add that here or as a follow-up.

ajaysehwal:
https://github.com/user-attachments/assets/b833566f-9ddb-4871-8d1a-8195093efe1d

@kaloudis, we’re not showing a loading indicator after tapping the “Connect Peer” button.

kaloudis:
Good catch, and thanks for the recording. Stepping through it frame by frame: the loading state is wired up correctly, since the button greys out on tap and stays disabled for the whole attempt, and the layout reserves the indicator's 100pt box between the tab bar and the form. The box just paints nothing.

The culprit is `LightningIndicator`, which renders `assets/images/GIF/loading.gif` through `Image`. That GIF's first frame is fully transparent, so anything that shows it statically instead of animating gives you an empty box. It only has 5 call sites, all in screens that are rarely exercised, which is why it went unnoticed. Everything else in the app (119 files) uses `LoadingIndicator`, the themed Lottie loader. Fixed app-wide in #4678 rather than here, since it is unrelated to the store guard.

Two changes here from your feedback:

- 62d9c869e: your recording also caught an "Uncaught (in promise)" LogBox toast. `connectPeer` rejects on backend failures, and the Connect Peer button was the only caller not wrapping the call. The store already records the failure in `errorMsgPeer`, so the view now swallows the rejection.
- d9f961013: the residual issue from your earlier review, guarding the two LSP pre-connect sites with `method?.address && method?.port` so an incomplete connection method can never build `undefined:9735`.

stores/ChannelsStore.ts
```diff
@@ -788,6 +788,20 @@ export default class ChannelsStore {
...
+                if (!silent) {
+                    this.connectingToPeer = false;
+                    this.errorMsgPeer = localeString(
+                        'views.OpenChannel.hostRequired'
+                    );
+                    this.errorPeerConnect = true;
+                }
+                return;
+            }
+            // no host means the peer is assumed to be connected already;
```
ajaysehwal: One very small nit: the code comment in ChannelsStore.ts:800 ("no host means the peer is assumed to be connected already") is placed right after the new if (connectPeerOnly) {...return;} block rather than directly above the if (!request.host) check — slightly easy to misread as describing the new branch instead of the existing fallthrough.

kaloudis: Good catch on the readability. Moving it above `if (!request.host)` would make it describe both branches, and it doesn't hold for the new one (no host there is an error, not an assumed connection), so I reworded it in place instead: it now opens with "otherwise (channel-open flow), ..." to make explicit which branch it covers. Folded into the original commit in 19fd1ceed.

ajaysehwal (approved):
tACK

shubhamkmr04 (approved):
tACK

PR #4652, opened 2026-09-13
fix(receive): don't show LSP connection error on non-LSP receive tabs

### LINKED ISSUE #4646: Receive screen shows "Could not connect to LSP" error on the on-chain / Lightning Address tabs where no LSP is needed
## Description
The Receive screen shows a "Could not connect to LSP. Please check your LSP settings or try again later." error banner with a "GO TO LSP SETTINGS" button even when the user is viewing a plain on-chain (or Lightning Address) receive address, which does not require an LSP at all. This is misleading — it makes a perfectly working on-chain address generation look broken/blocked.

## Screenshot

Image

## Steps to reproduce
1. Enable LSP in settings, with an LSP that is unreachable/misconfigured (or simulate a network failure to the LSP).
2. Open Receive.
3. Switch to (or land directly on) the On-chain tab.
4. Observe the LSP connection error banner and "GO TO LSP SETTINGS" button above a valid on-chain QR code / address.

## Expected behavior
The LSP connection error and "GO TO LSP SETTINGS" button should only appear when the user is on a tab that actually needs the LSP (Unified / Lightning), matching the gating already used by other LSP-specific UI in the same screen.

## Actual behavior
The error banner block has no tab-based gate, so a stale LSP connection failure — triggered by a background `getLSPInfo()` call on wallet load, unrelated to the user's tab selection — is shown regardless of which receive tab is active, including On-chain and Lightning Address.

### ZEUS version

v13.2.2

### Node interface

LDK Node

### Network

Tor

### Device

Iphone 17

### Device operating system

iOS 26.2.2

### Log output

```shell

```

ajaysehwal:
# Description

Relates to issue: **[ZEUS-4646](https://github.com/ZeusLN/zeus/issues/4646)**

## Summary
- The LSP connection error banner ("Could not connect to LSP...") and its "Go to LSP Settings" button had no `selectedIndex` gate, so a stale LSP failure (from the background `getLSPInfo()` call on wallet load) was shown regardless of which Receive tab was active, including On-chain and Lightning Address, where no LSP is involved.
- The previously combined `error_msg` (`LSPStore.flow_error_msg || InvoicesStore.error_msg`) is split into `lspErrorMsg` and `invoiceErrorMsg`. **This is an intentional behavior change**: the two sources can now render stacked on Unified/Lightning instead of one silently hiding the other — the old fallback could mask a real invoice/on-chain-address error behind an irrelevant LSP error.
- `lspErrorMsg` (and the LSP settings shortcut) is gated to the Unified/Lightning tabs (`selectedIndex` 0/1) **and** `lspIsActive`, so the banner also disappears if the user disables LSP in settings after a background failure. `invoiceErrorMsg` (covers `getNewAddress`/on-chain and invoice-creation failures) stays visible on every tab, unconditionally, as it always was.

## Changes
- `views/Receive.tsx`: split `error_msg` into `lspErrorMsg`/`invoiceErrorMsg`; gate `lspErrorMsg` and the LSP-settings button on `lspIsActive && (selectedIndex === 0 || selectedIndex === 1)`.

## Known follow-up (not in this PR)
  The root cause — `getLSPInfo()`'s error persisting in the shared `LSPStore` singleton with nothing clearing it when Receive mounts — is untouched, so the banner can still reappear on Unified/Lightning on a later visit after a background failure. Clearing `flow_error_msg`/`showLspSettings` on Receive unmount would close that out; flagging as a separate follow-up.

## Test plan
- [x] LSP unreachable, on Unified/Lightning: banner + settings button show.
- [x] Switch to On-chain/Lightning Address: banner and button disappear.
- [x] Disable "Enable LSP" in settings after a background failure; stay on Unified/Lightning: banner and button no longer show.
- [x] Trigger an on-chain address generation failure: `invoiceErrorMsg` shows regardless of tab.
- [x] `yarn verify` passes.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


views/Receive.tsx
```diff
@@ -1526,22 +1526,31 @@ export default class Receive extends React.Component<
...
-                            {showLspSettings && (
-                                <View style={{ margin: 10 }}>
-                                    <Button
-                                        title={localeString(
-                                            'views.Receive.goToLspSettings'
-                                        )}
-                                        onPress={() =>
-                                            navigation.navigate('LSPSettings')
-                                        }
+                            {error_msg &&
```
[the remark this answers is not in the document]
ajaysehwal: Good catch, thanks — confirmed this was a real bug. `InvoicesStore.error_msg` is also set by `getNewAddress()` failures (on-chain address generation), so gating the combined `error_msg` to selectedIndex 0/1 would have silently swallowed genuine on-chain errors.

Fixed in [cdc8d53](https://github.com/ZeusLN/zeus/commit/cdc8d538d0db26f82724822006061e5f2713394a) : split the combined value into `lspErrorMsg` (`LSPStore.flow_error_msg`) and `invoiceErrorMsg` (`InvoicesStore.error_msg`). Only `lspErrorMsg` is now gated to the Unified/Lightning tabs; `invoiceErrorMsg` renders unconditionally on every tab, the same as before this PR.

views/Receive.tsx
```diff
@@ -1526,22 +1526,31 @@ export default class Receive extends React.Component<
...
+                                    selectedIndex === 1) && (
+                                    <ErrorMessage
+                                        message={error_msg}
+                                        dismissable
                                     />
-                                </View>
-                            )}
+                                )}
+
+                            {showLspSettings &&
```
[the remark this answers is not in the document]
ajaysehwal: Checked this against the rest of the file. `selectedIndex` is a fixed, stable mapping used identically throughout Receive.tsx (0=Unified, 1=Lightning, 2=On-chain, 3=Lightning Address) — including in pre-existing, untouched code at lines 1584-1727 that already gates other LSP-only UI (the zero-amount notice, zero-conf fee banner) the same way.

When `onChainOnly`/`lnOnly` modes are active, the tab bar is hidden entirely (`!lnOnly && !onChainOnly` gates the ButtonGroup) and `selectedIndex` is forced to a fixed value (1 or 2) rather than remapped — so there's no scenario in this codebase where index 0/1 stops corresponding to Unified/Lightning. This diff just extends an existing, already-relied-upon convention rather than introducing a new risk, so no change made here.

views/Receive.tsx
```diff
@@ -1526,22 +1527,38 @@ export default class Receive extends React.Component<
...
+                            {lspErrorMsg &&
+                                (selectedIndex === 0 ||
+                                    selectedIndex === 1) && (
+                                    <ErrorMessage
+                                        message={lspErrorMsg}
+                                        dismissable
                                     />
-                                </View>
-                            )}
+                                )}
```
kaloudis, line 1545: **nit:** consider including `lspIsActive` (already in component state) in this gate and the settings-button gate below. If the user toggles "Enable LSP" off in the advanced sheet after a background failure, the stale "Could not connect to LSP" banner and settings button still show on Unified/Lightning even though the invoice being created won't touch the LSP. Not a blocker, that stale state predates this PR.

views/Receive.tsx
```diff
@@ -1023,7 +1023,8 @@ export default class Receive extends React.Component<
         const { lightningAddress } = LightningAddressStore;
         const lightningAddressLoading = LightningAddressStore.loading;
 
-        const error_msg = LSPStore.flow_error_msg || InvoicesStore.error_msg;
+        const lspErrorMsg = LSPStore.flow_error_msg;
+        const invoiceErrorMsg = InvoicesStore.error_msg;
```
kaloudis, line 1027: **nit:** worth a sentence in the PR body: previously the two error sources shared one banner with the LSP error taking precedence; now both can render stacked on Unified/Lightning. I think that's strictly better, since the old `flow_error_msg || error_msg` fallback could hide a real invoice/address error behind an irrelevant LSP error, but it should be called out as intentional.

kaloudis (commented):
Clean, minimal fix. Verified the load-bearing assumptions: tab indices 0/1 are stable (only the Lightning Address tab at index 3 is conditional in the `buttons` array), `showLspSettings` is only ever written alongside `flow_error_msg` in the Flow paths, and `InvoicesStore.error_msg` also carries on-chain address-generation failures, so leaving it un-gated is correct. Two nits inline.

One follow-up thought, not for this PR: the root cause (the wallet-load `getLSPInfo()` error persisting in the `LSPStore` singleton with nothing clearing it when Receive mounts) is untouched, so the stale banner will still reappear on Unified/Lightning on later visits. Clearing `flow_error_msg`/`showLspSettings` on Receive unmount would finish the job.

views/Receive.tsx
ajaysehwal: Agreed, added to the commit message: this is an intentional behavior change — the old `flow_error_msg || error_msg` fallback could mask a real invoice/address error behind an irrelevant LSP error, so splitting them means both can now render stacked on Unified/Lightning instead of one hiding the other. Will carry the same note into the PR description.

views/Receive.tsx
ajaysehwal: Good point, done in [ce06d41](https://github.com/ZeusLN/zeus/commit/ce06d41a8afed2365f7c10c8035da6cde670eac7) — added `lspIsActive &&` to both the error banner and the settings-button gate. Now if the user disables LSP after a background failure, neither the stale banner nor the settings shortcut shows on Unified/Lightning, since the invoice being created won't touch the LSP anyway.

shubhamkmr04 (approved):
tACK

shubhamkmr04 (commented):
https://github.com/user-attachments/assets/9fb3e307-7c21-4a29-b538-99902be9fee4

bug: reset btn keeps disappearing again and again,

ajaysehwal:
> Simulator.Screen.Recording.-.iPhone.16.Pro.-.2026-09-14.at.19.02.52.mov 
> bug: reset btn keeps disappearing again and again,

@shubhamkmr04, this is a pre-existing bug in master. I think we should fix this in a separate PR

shubhamkmr04 (approved):
tACK

kaloudis:
> > Simulator.Screen.Recording.-.iPhone.16.Pro.-.2026-09-14.at.19.02.52.mov
> > bug: reset btn keeps disappearing again and again,
> 
> @shubhamkmr04, this is a pre-existing bug in master. I think we should fix this in a separate PR

Please file an issue for this before we merge this

shubhamkmr04:
> 
> Please file an issue for this before we merge this

https://github.com/ZeusLN/zeus/issues/4656

kaloudis (approved):

PR #4653, opened 2026-09-13
fix(lsps1): hide order form and enable retry when get_info fails

### LINKED ISSUE #4645: LSPS1 create order screen: form and CREATE ORDER button remain fully interactive after get_info fetch fails
## Description
When the LSPS1 "Create Order" screen's `get_info` request fails (network error, non-200 response, timeout), the screen shows an error banner ("Error fetching get_info data") but still renders the full input form with placeholder/default values (`0`, `N/A`) and a fully active, tappable CREATE ORDER button. There's no clear indication the form is unusable, and nothing stops the user from submitting an order based on data the app never actually received.

## Screenshot

Image

## Steps to reproduce
1. Navigate to LSP → Create Order (LSPS1 flow) with a broken/unreachable LSP endpoint or no network.
2. Observe the "Error fetching get_info data" banner.
3. Note that "Initial LSP balance (sats)" shows `0`, "Channel expiry blocks" shows `N/A`, the duration selector (1/3/6/12 mo) is active, and CREATE ORDER is enabled and pressable.

## Expected behavior
- When `get_info` fails, the form inputs should be hidden or disabled, not shown with meaningless defaults.
- CREATE ORDER should be disabled (or hidden) until valid LSP info has been fetched.
- The existing retry UI (Retry / Go to Settings) should also appear for this error, not only for the timeout case.
- Ideally surface a more actionable error message (e.g. distinguishing "no network," "LSP unreachable," "unexpected response") rather than one generic string for every failure path.

## Actual behavior
- The form renders unconditionally except when `error_msg` matches the `timeoutError` string specifically.
- Inputs default to `0` / `N/A` with no visual indication they're inert/meaningless.
- The min/max slider is correctly hidden when `getInfoData` is empty, but the two main text inputs and duration selector are not — inconsistent gating within the same form.
- CREATE ORDER has no `disabled` state tied to the error or empty `getInfoData`, so users can attempt to submit an order that has no chance of succeeding.

## Suggested fix
- Gate the whole form (not just the accordions) on `getInfoData` having loaded successfully, mirroring the slider's existing gating.
- Extend the retry-block condition to include this error path (or generalize it to trigger on `LSPStore.error` regardless of message).
- Add `disabled={LSPStore.error || !hasLoadedInfo}` (or similar) to the CREATE ORDER button.
- Route the error string through `localeString` for consistency/i18n, and consider differentiating timeout vs. generic fetch failure messaging.

### ZEUS version

v13.2.2

### Node interface

LDK Node

### Network

Clearnet

### Device

iphone 17

### Device operating system

iOS 26.2.2

### Log output

```shell

```

ajaysehwal:
# Description

Relates to issue: **[ZEUS-4645](https://github.com/ZeusLN/zeus/issues/4645)**

## Summary
When the LSPS1 "Create Order" screen's `get_info` request fails (network error, non-200 response, or a custom-message timeout), the screen still rendered the full order form — "Initial LSP balance" defaulting to `0`, "Channel expiry blocks" defaulting to `N/A`, and a fully active, tappable CREATE ORDER button — alongside a generic, unhelpful "Error fetching get_info data" banner. A retry mechanism already existed but was scoped only to the timeout case, so the more common REST/network failure left users stuck with a live-looking but non-functional form and no way to retry from the UI.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td align="center">
      Before
    </td>
    <td align="center">
      After
    </td>
  </tr>
</table>

This PR:
1. Hides the order form and CREATE ORDER button whenever `get_info` hasn't successfully loaded, for any failure path (native, REST, custom message) — not just the timeout case.
2. Extends the existing Retry / "Go to Settings" UI to appear for all of those failure paths.
3. Replaces the generic `'Error fetching get_info data'` string with two distinct, localized, actionable messages: one for a network/connection failure and one for an unexpected LSP response.

## Changes
- `views/LSPS1/index.tsx`: introduce `isGetInfoError` (`LSPStore.error` with no `getInfoData` loaded) and use it in place of the `error_msg === timeoutError` checks that previously gated the error view and the retry block. Order-creation errors (which occur after info has already loaded) still render inline above the form as before — unaffected.
- `stores/LSPStore.ts`: `lsps1GetInfoREST` now sets `stores.LSPStore.getInfoNetworkError` on a fetch failure (`.catch`) and `stores.LSPStore.getInfoResponseError` on a non-200 response, instead of one hardcoded string.
- `locales/en.json`: add the two new message strings (source locale only, per contribution guidelines).

## Before / After
- Before: form + enabled CREATE ORDER button shown under a generic error, with no retry for this case.
- After: form is hidden, and the user sees a specific message plus Retry / Go to Settings buttons.

## Test plan
- [x] Point LSPS1 at an unreachable REST endpoint (or disable network) — confirm the form and CREATE ORDER button are hidden, the network-error message is shown, and Retry/Go to Settings work.
- [x] Simulate a non-200 `get_info` response — confirm the response-error message is shown with the same retry UI.
- [x] Confirm the custom-message timeout path still hides the form and shows retry (regression check).
- [x] Successfully load `get_info`, then trigger a create-order failure — confirm the form stays visible with the inline error banner (unaffected by this change).
- [x] `yarn verify` (lint, prettier, typecheck) passes.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [ ] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


stores/LSPStore.ts
```diff
@@ -753,15 +753,19 @@ export default class LSPStore {
...
                         this.loadingLSPS1 = false;
                     }
                 });
             })
             .catch(() => {
                 runInAction(() => {
                     this.error = true;
-                    this.error_msg = 'Error fetching get_info data';
+                    this.error_msg = localeString(
+                        'stores.LSPStore.getInfoNetworkError'
```
kaloudis, line 786: **should fix:** `lsps1GetInfoREST` never clears `error`/`error_msg`, neither at entry nor on the 200 path. The native and custom-message siblings both reset them at entry (`lsps1GetInfoNative`, `lsps1GetInfoCustomMessage`). Before this PR the asymmetry was unreachable because REST failures had no retry button; now: get_info fails, user taps Retry, fetch succeeds, `getInfoData` populates, but the error flags are still set, so `isGetInfoError` goes false and the top inline block renders the stale "Could not connect to the LSP..." banner above a fully working form until unmount.

Fix: add `this.error = false; this.error_msg = '';` at the top of `lsps1GetInfoREST`, matching the other two paths.

kaloudis (commented):
Concept ACK, the `isGetInfoError` consolidation is the right refactor and the create-order inline path is correctly preserved. Two should-fix items and a few nits inline.

views/LSPS1/index.tsx
```diff
@@ -342,6 +342,10 @@ export default class LSPS1 extends React.Component<LSPS1Props, LSPS1State> {
         const result = createOrderResponse?.result || createOrderResponse;
         const payment = result?.payment;
         const isFreeOrder = isOrderFree(payment);
+        const isGetInfoError =
+            LSPStore.error &&
+            !!LSPStore.error_msg &&
+            Object.keys(getInfoData ?? {}).length === 0;
```
kaloudis, line 348: **should fix:** the custom-message path is only covered for the timeout case, not for an error *response*. In `handleCustomMessages`, the `getInfoId` branch assigns `this.getInfoData = data` unconditionally, with no `data.error` check, unlike every other branch in that handler. If the LSP returns a JSON-RPC error for `lsps1.get_info`, the error envelope lands in `getInfoData`: the error flags are never set, `Object.keys(getInfoData).length` is nonzero, so `isGetInfoError` is false and the user gets the exact bug this PR fixes (form with `0`/`N/A` defaults and an active CREATE ORDER). It also defeats the focus-listener refetch guard, which keys on `getInfoData` being empty.

Fix: mirror the `createOrderId` branch, on `data.error` set `error`/`error_msg` (the new `getInfoResponseError` string is a natural fallback) instead of storing the envelope.

views/LSPS1/index.tsx
```diff
@@ -1320,48 +1322,41 @@ export default class LSPS1 extends React.Component<LSPS1Props, LSPS1State> {
...
+                                if (BackendUtils.supportsLSPS1native()) {
+                                    LSPStore.lsps1GetInfoNative();
+                                } else if (BackendUtils.supportsLSPS1rest()) {
+                                    LSPStore.lsps1GetInfoREST();
+                                } else if (
+                                    BackendUtils.supportsLSPScustomMessage()
+                                ) {
+                                    await this.connectPeer();
+                                    await this.subscribeToCustomMessages();
+                                    LSPStore.lsps1GetInfoCustomMessage();
```
kaloudis, line 1345: **note (pre-existing, not blocking):** retry never re-arms the custom-message timeout. `subscribeCustomMessages` early-returns when `customMessagesSubscriber` exists, so the 7s timeout is only armed once. After a timeout error, this handler sets `loadingLSPS1 = true` and clears the error, and if the LSP stays silent the spinner runs forever with no second timeout. This predates the PR, but since it's about making retry work everywhere it's worth a follow-up.

stores/LSPStore.ts
```diff
@@ -753,15 +753,19 @@ export default class LSPStore {
                         this.loadingLSPS1 = false;
                     } else {
                         this.error = true;
-                        this.error_msg = 'Error fetching get_info data';
+                        this.error_msg = localeString(
+                            'stores.LSPStore.getInfoResponseError'
```
kaloudis: **nit:** the non-200 path discards `response.data`, which for a spec-following LSPS1 server often carries a usable error message. Consider try-parsing it through `errorToUserFriendly` and falling back to `getInfoResponseError`.

views/LSPS1/index.tsx
kaloudis, line 348: **nit:** this is a heuristic ("any store error while info is empty"). Free-order status polling resumed by the focus listener also writes `LSPStore.error`, so in the narrow window where a stacked duplicate has wiped `getInfoData`, a poll failure would present as a get_info failure with a Retry button. Low likelihood, fine to ship, but a dedicated `getInfoError` flag on the store would be sturdier than inferring from message state.

locales/en.json
```diff
@@ -1766,6 +1766,8 @@
     "stores.SettingsStore.lncConnectError": "Failed to connect the LNC client to the proxy server",
     "stores.LSPStore.error": "LSP error",
     "stores.LSPStore.connectionError": "Could not connect to LSP. Please check your LSP settings or try again later.",
+    "stores.LSPStore.getInfoNetworkError": "Could not connect to the LSP. Please check your internet connection and LSP settings, then try again.",
```
kaloudis, line 1769: **nit:** `getInfoNetworkError` near-duplicates the existing `stores.LSPStore.connectionError` string. The new wording is more actionable, so keeping both is defensible, just flagging the overlap.

stores/LSPStore.ts
ajaysehwal: confirmed the asymmetry with the native/custom-message siblings. Fixed in [e6f7676](https://github.com/ZeusLN/zeus/pull/4653/commits/e6f7676244834dc5dad26f7fd922a161745ff7c9) : `lsps1GetInfoREST` now resets `error`/`error_msg` at entry, matching the other two get_info paths.

views/LSPS1/index.tsx
ajaysehwal: Confirmed — this was the one branch in handleCustomMessages that didn't check `data.error` before use. Fixed in [e6f7676](https://github.com/ZeusLN/zeus/pull/4653/commits/e6f7676244834dc5dad26f7fd922a161745ff7c9) , mirroring the createOrderId branch: on `data.error`, sets `error`/`error_msg` (falling back to the new `getInfoResponseError` string when there's no `data.error.data.message`) instead of storing the error envelope into `getInfoData

views/LSPS1/index.tsx
ajaysehwal: Agreed this is out of scope here — filing it as a follow-up rather than folding it into this PR, since fixing retry-rearm for the custom-message timeout touches subscribeCustomMessages's subscriber-reuse logic, which is unrelated to the get_info error-display bug this PR addresses

stores/LSPStore.ts
ajaysehwal: Done in [e6f7676](https://github.com/ZeusLN/zeus/pull/4653/commits/e6f7676244834dc5dad26f7fd922a161745ff7c9) — now try-parses `response.data` for `message`/`error.message` and uses it as `error_msg` when present, falling back to the generic `getInfoResponseError` string otherwise. Kept it a plain try/catch rather than `errorToUserFriendly` since that helper matches against known Error.toString() patterns, not arbitrary server-supplied strings — same approach the sibling `lsps1GetOrderREST` already uses for its response body.

views/LSPS1/index.tsx
ajaysehwal: Fair point, and I agree a dedicated `getInfoError` flag would be sturdier than inferring from shared `error`/`error_msg` state. Given the low likelihood and that it'd mean threading a new observable through several call sites (native/REST/custom-message/free-order polling), I'd rather do that as a follow-up than widen this PR's scope — happy to pick it up separately if you'd like it tracked.

locales/en.json
ajaysehwal: Yeah, there's overlap — kept both since `connectionError` is used for the wallet-load LSPStore.getLSPInfo() flow error shown on Receive, while `getInfoNetworkError` is specific to the LSPS1 create-order screen's get_info fetch. Could consolidate them, but figured scoping the message to its screen was clearer than reusing a string written for a different context. Let me know if you'd rather I just reuse `connectionError` here.

shubhamkmr04 (commented):
Tested on Embedded LND with a wrong LSPS1 pubkey: the screen keeps showing the loading spinner. No error message, Retry or Go to Settings button ever appears.

ajaysehwal:
> Tested on Embedded LND with a wrong LSPS1 pubkey: the screen keeps showing the loading spinner. No error message, Retry or Go to Settings button ever appears.

Confirmed and root-caused. `subscribeCustomMessages()` returned early whenever `customMessagesSubscriber` was already set — and that field is never cleared on unmount (only the separate `reset()` clears it, which the LSPS1 screen doesn't call). So once any custom-message subscription had succeeded once in the app session, the 7s timeout was permanently skipped for every later get_info call, including a fresh visit's first request. Wrong pubkey → no response → no timeout armed → spinner forever, exactly as reproduced.

Fixed in 10a92f1 moved the timer arm + `resolvedCustomMessage` reset above the early-return, so a fresh timeout is always scheduled per call regardless of whether the listener itself needs re-registering. This also resolves the retry-never-rearms-the-timeout note flagged earlier — same root cause, worse than I'd scoped it (reachable on first load, not just retry).

shubhamkmr04 (approved):
tACK

stores/LSPStore.ts
```diff
@@ -686,6 +697,8 @@ export default class LSPStore {
             }
         }, timer);
 
+        if (this.customMessagesSubscriber) return;
```
kaloudis: **should fix:** re-arming the timeout on every call regresses in two ways.

(a) `views/Wallet/Wallet.tsx:1341` calls `LSPStore.subscribeCustomMessages()` from `fetchData`, which runs on connect, refresh, and focus, with no LSPS request in flight. Before this change, only the first-ever call armed the timer, so that was a one-shot latent no-op. Now every wallet refresh on a `supportsLSPScustomMessage` backend sets `error = true`, `error_msg = timeoutError`, and `loadingLSPS1/7 = false` seven seconds later, since no custom message is coming. That trips the polling check in `views/LSPS7/Order.tsx` (the `LSPStore.error && error_msg !== ''` branch around line 88), which falls back to the stale stored order, and it flashes phantom errors on any screen observing `LSPStore.error`.

(b) The `clearTimeout(timeoutId)` calls live inside the subscriber callback closures, which are created only on the first call, so they forever clear only the first timer. Every later timer is merely suppressed by `resolvedCustomMessage`, and each new subscribe call resets that flag to false, re-activating any still-pending stale timer. Concrete race: wallet `fetchData` arms a timer at t=0; the user enters LSPS1 at t=3 and `get_info` goes out; the wallet timer fires at t=7 before the response, showing a premature timeout error; when the response then lands, `getInfoData` populates while `error`/`error_msg` stay set, leaving a stale timeout banner over a working form. That's the same bug class e6f7676 just fixed.

Suggest reverting this commit and doing the re-arm as the follow-up you originally proposed. Done properly, this is a request timeout, not a subscription timeout: arm it in the request initiators (`lsps1GetInfoCustomMessage` and friends), or at minimum keep a single store-level timeout handle that's cleared on each re-arm and cleared from the message handler via the store field, and never armed from the Wallet call path.

stores/LSPStore.ts
```diff
@@ -605,7 +605,14 @@ export default class LSPStore {
         console.log('Received custom message', { peer, data });
 
         if (data.id === this.getInfoId) {
-            this.getInfoData = data;
+            if (data.error) {
+                this.error = true;
+                this.error_msg = data?.error?.data?.message
```
kaloudis: **nit:** JSON-RPC 2.0 puts the human-readable text in `error.message`; `error.data` is optional structured detail. As written, a spec-following LSP's actual message gets discarded in favor of the generic locale string. Consider `data?.error?.data?.message || data?.error?.message` (the `getOrderId` branch below already reads `error.message`).

kaloudis (commented):
Round 2. Both round-1 should-fixes are resolved in e6f7676 and verified: `lsps1GetInfoREST` now resets `error`/`error_msg` at entry, matching its native and custom-message siblings, and the `getInfoId` branch now handles `data.error` instead of storing the error envelope as service info. The non-200 `response.data` parsing nit is addressed too. CI is green.

One new should-fix on 10a92f18 (the timeout re-arm), detailed inline: it introduces phantom timeout errors via the `Wallet.tsx` subscribe call and a stale-timer race that can resurrect the exact stale-banner bug this PR fixes. I'd drop that commit from this PR and handle retry re-arm in the follow-up you originally proposed.

With that commit dropped this is a tACK candidate. It still needs a test pass on a REST backend (LND REST), since testing so far covered the LDK Node native path on iOS; the REST branch of `lsps1GetInfoREST` is one of the main paths this PR changes.

stores/LSPStore.ts
ajaysehwal: Good catch on both — the Wallet.tsx fetchData call path and the stale-closure/resolvedCustomMessage race are both real, and (b) is exactly the bug class this PR is trying to close, so reverting rather than patching further.

Dropped in [139af23](https://github.com/ZeusLN/zeus/pull/4653/commits/139af23ae020f415d3364f38cd3514b6cf64e599) (revert of 10a92f186). Agreed on scoping the real fix to the request initiators (or a store-level timeout handle cleared via `this` rather than a closure, never armed from Wallet's fetchData) — tracking that as the follow-up rather than reopening it here.

stores/LSPStore.ts
ajaysehwal: Fixed in  139af23 — now reads `data?.error?.data?.message || data?.error?.message`, matching the getOrderId branch below. Thanks for catching the spec detail.

kaloudis:
139af23ae verified: the revert is byte-identical to master and the error.message fallback looks good. CI green, nothing outstanding in the diff. One ask before merge: please file the follow-up issue for the custom-message timeout rework now, so the wrong-pubkey infinite-spinner bug you root-caused doesn't get lost. Link your analysis and the request-initiator design from the earlier thread.

ajaysehwal:
> [139af23](https://github.com/ZeusLN/zeus/commit/139af23ae020f415d3364f38cd3514b6cf64e599) verified: the revert is byte-identical to master and the error.message fallback looks good. CI green, nothing outstanding in the diff. One ask before merge: please file the follow-up issue for the custom-message timeout rework now, so the wrong-pubkey infinite-spinner bug you root-caused doesn't get lost. Link your analysis and the request-initiator design from the earlier thread.

https://github.com/ZeusLN/zeus/issues/4669

kaloudis (approved):

PR #4658, opened 2026-09-15
fix: pay BIP 353 offers from records with multiple params

kaloudis:
# Description

Reported by a user paying a BIP 353 address that pairs a silent payment address with a BOLT 12 offer. Sending from ZEUS on a Core Lightning backend failed with:

```
Offer: unparsable offer: invalid bech32 string: invalid token
```

The same address pays fine from Phoenix.

## Root cause

Two issues compound, and they only line up when a BIP 353 record has **more than one payment parameter** and **no on-chain address**.

`handleAnything` routes such a record straight to `Send`, because BOLT 12 is the only method ZEUS can use: `sp=` is not a parameter `bitcoinQrParser` handles, so `b12Value` is empty and the `hasMultipleB12` check that would otherwise route to `ChoosePaymentMethod` is false. That branch passed the **whole reassembled URI** as `bolt12` while correctly passing the extracted offer as `destination` — so the screen displayed the right offer but held the wrong string.

`Send.payBolt12` then recovered the offer with `bolt12.split('=')[1]`, which only works for a single-parameter URI. On `bitcoin:?sp=sp1q...&lno=lno1...` it returns everything between the first and second `=`:

```
sp1qqduw6gtpcuxc9mtx4m2hf0kvw3enekrd5gsseu8jwv555dd8v6lakq5vvnkhgctmutus3ufrtwz5w6np6hl7fx5s5lsjvh8m42tx7s8wmv87ertf&lno
```

`&` and `o` are both outside the bech32 data charset, hence CLN's `invalid token`. LDK Node fails the same way with a less legible message; this is not CLN-specific.

## Verified against the reporter's live record

| Step | Result |
| --- | --- |
| DoH TXT lookup | `Status 0, AD true` — DNSSEC fine, one `bitcoin:` record |
| TXT reassembly (3 character-strings) | 662-char URI, correct |
| `bitcoinQrParser` | params `sp`, `lno`; `b12Value = ''` |
| `hasMultipleB12` | **false** → `Send` branch, not `ChoosePaymentMethod` |
| `/.well-known/lnurlp/<user>` | **404** → `.catch()` fires, which is the buggy branch |
| chars outside bech32 charset, before | `["&", "o"]` |
| chars outside bech32 charset, after | none; `lno1…`, 529 chars |

## Fix

- `utils/handleAnything.ts` — pass the extracted offer as `bolt12`, not the raw URI, in both `.catch()` branches (Tor and non-Tor).
- `views/Send.tsx` — parse the `lno` parameter with `AddressUtils.processBIP21Uri` instead of splitting on `=`, so any URI shape resolves. Defence in depth for other callers.

## Why existing tests missed it

The BIP 353 vector already under test (`matt@mattcorallo.com`) includes an on-chain address, which populates `b12Value` and takes the healthy `ChoosePaymentMethod` path. This branch was never exercised.

## Known gap, not addressed here

`bitcoinQrParser` only handles `amount|label|message|lightning|lno|noffer`, so the `sp=` silent payment address is still silently dropped. Not a regression; out of scope for this fix.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Added:
- `utils/handleAnything.test.ts` — a BIP 353 record with no on-chain address must hand `Send` the bare offer. Confirmed failing on the old code with `Received: "bitcoin:?sp=...&lno=..."`.
- `utils/AddressUtils.test.ts` — `processBIP21Uri` extracts `lno` alongside a parameter ZEUS cannot pay.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

**Not yet device-tested.** Verification so far is the four CI gates plus a replay of the reporter's live DNS record through the real parsing code. Needs an on-device pass against CLNRest and LDK Node paying a multi-parameter BIP 353 address before merge.

### Locales

### Other:


ajaysehwal (commented):
@kaloudis, while reviewing this PR, I came across this view. I think we should prioritize the **"Backend does not support BOLT 12"** message and show it at the top first. Maybe not as part of this PR, but we could address this UI issue separately?

Image

kaloudis:
> @kaloudis, while reviewing this PR, I came across this view. I think we should prioritize the **"Backend does not support BOLT 12"** message and show it at the top first. Maybe not as part of this PR, but we could address this UI issue separately?
> Image

Room for improvement here for sure. Let's file this as a separate issue.

kaloudis:
@ajaysehwal agreed, and agreed it doesn't belong in this PR. Filed as #4715 with your screenshot and the code pointers.

Two things worth flagging while writing it up: the zero-balance banner at `views/Send.tsx:774-786` doesn't check `supportsOffers` at all, so on LND/LNC/Embedded/LndHub/NWC it fires alongside the capability error and is the only message with banner prominence, even though funding a channel wouldn't help. And the capability error itself is a bare `<Text>` rather than the `ErrorMessage` component the same screen already uses at the top. Keysend has the identical problem at `views/Send.tsx:1536-1552`, so both can be fixed in one pass.

This PR is unrelated to that: it only stops `payBolt12` from mangling the offer string on multi-parameter BIP 353 records.

utils/handleAnything.ts
ajaysehwal: Optional DRY cleanup: the two .catch fallbacks (Tor and non-Tor) are identical, which is why this bug needed fixing in both places. A small bip353Fallback() helper could keep the two paths in sync.

Also, the Tor branch only runs for .onion domains, and the BIP-353 DNS-over-HTTPS lookup won’t resolve onion names with DNSSEC, so b12Offer is effectively never set on that path

kaloudis: Good catch on the Tor path. Confirmed: Cloudflare DoH returns NXDOMAIN (Status 3, root-signed denial) for any `*.user._bitcoin-payment.<x>.onion` name, so `b12Offer` is never set there and the BOLT 12 handling in both Tor callbacks is unreachable. The fix here only matters on the non-Tor path; the Tor-side edit just keeps the two in step.

Related: that DoH `fetch` goes out over clearnet even for `.onion` addresses with Tor enabled, so the onion name leaks to Cloudflare. Filed #4727 to skip the BIP 353 lookup for `.onion` domains and remove the dead Tor-branch BOLT 12 code. That also takes care of the duplication, since only one fallback remains. Keeping this PR to the offer-mangling fix.

kaloudis: On the `bip353Fallback()` helper: skipping it here. The two `.catch()` blocks are only duplicated because the Tor branch carries a copy that can never run. Once #4727 removes that code, a single fallback remains in the non-Tor `.catch()`, so there is nothing left for a helper to keep in sync. Adding one in this PR would mean refactoring code that #4727 deletes.

Issue #4531, opened 2026-08-31
NWC: cannot create a connection with payment permissions when Lightning balance is zero, even with Cashu funds


kaloudis:
### Summary

If your Lightning balance is zero, you cannot create an NWC connection that has any payment permission, even when you are holding a meaningful Cashu balance in the same wallet. The budget slider is replaced by a "No balance available" box, the budget stays pinned at 0, and the save button is permanently disabled with no explanation of what would unblock it.

Since Cashu is supported on Embedded LND and LDK Node (`supportsCashuWallet()` returns true only for those two backends), the affected user is exactly the one who onboarded via ecash and has not opened a channel yet. That is a normal state for a new Zeus wallet, not an edge case.

### Steps to reproduce

1. Fresh Embedded LND (or LDK Node) wallet with no channels, so the Lightning balance is 0.
2. Enable ecash and mint or receive some Cashu, for example 20,000 sats.
3. Settings > Nostr Wallet Connect > add a connection.
4. Leave the default permission preset (Full Access, which includes `pay_invoice`).
5. Budget section shows "No balance available". There is no input and no slider.
6. Save is greyed out and stays greyed out no matter what else you change.

### Expected

Zeus should let you create the connection and fund it from ecash. The budget is a spending cap on a connection that may live for months, so it should not be clamped to whatever balance happens to exist at the moment the form is open.

### Root cause

The budget ceiling is sourced from a single balance, chosen globally, and then used as a hard form validation gate.

`stores/NostrWalletConnectStore.ts:3688` `loadMaxBudget()` picks one balance source and only one:

```ts
if (this.isCashuConfigured) {
    this.maxBudgetLimit = this.cashuStore.totalBalanceSats || 0;
} else {
    const balance = await this.balanceStore.getLightningBalance(true);
    this.maxBudgetLimit = Number(balance?.lightningBalance) || 0;
}
```

`isCashuConfigured` (`stores/NostrWalletConnectStore.ts:3683`) is `cashuEnabled && cashuStore.isProperlyConfigured()`, and `cashuEnabled` comes from the `zeus-nwc-cashu-enabled` key, which defaults to false (`loadCashuSetting`, `stores/NostrWalletConnectStore.ts:3557`). So by default the Cashu balance is invisible to NWC and `maxBudgetLimit` is 0.

That zero then closes every exit from the form:

[...]

PR #4662, opened 2026-09-15
fix(nwc): point empty NWC budget at Cashu settings when Lightning balance is zero

ajaysehwal:
# Description

## Summary

Related to [#4531](https://github.com/ZeusLN/zeus/issues/4531): a user with zero Lightning balance but a funded Cashu balance hits a dead end on the Add/Edit NWC Connection screen — the budget section says "No balance available" with no indication that enabling Cashu as the funding source (in NWC Settings) would unblock it.

- The empty-budget message now branches on `NostrWalletConnectStore.isCashuConfigured` — the same condition `loadMaxBudget` uses to pick the balance source — so it always names the balance that's actually empty: **"No Lightning balance available"** or **"No Cashu balance available"**, never the wrong one.
- When enabling Cashu would genuinely help — backend supports it, ecash is enabled, and the switch is currently off — a tappable **"Enable Cashu in NWC Settings"** link appears and navigates straight to NWC Settings.
- Added `isCashuAvailable` and `canOfferCashuSwitch` computeds to `NostrWalletConnectStore`, next to `isCashuConfigured`, so this predicate lives in the store instead of being duplicated in the view. `loadCashuSetting` now reuses `isCashuAvailable` instead of re-deriving the same expression.
- The view no longer reaches into `SettingsStore` at all — it reads the store's computed directly.

<table>
  <tr>
    <th>iOS</th>
    <th>Android</th>
  </tr>
  <tr>
    <td align="center">
      iOS
    </td>
    <td align="center">
      Android
    </td>
  </tr>
</table>

### Review history

An earlier version of this branch hardcoded "No Lightning balance available" regardless of funding source, which was wrong whenever the NWC Cashu switch was already on and Cashu itself was empty (or misconfigured, causing a silent fallback to Lightning). Branching the copy on `isCashuConfigured` — the same predicate `loadMaxBudget` already uses — fixes both that case and the pre-existing cold-start/misconfigured-mint edge case, since the message and the actual balance source can never disagree anymore.

## Test plan

- [x] Fresh Embedded LND/LDK Node wallet, 0 Lightning balance, ecash enabled, Cashu switch off: shows "No Lightning balance available" + "Enable Cashu in NWC Settings" link; tapping it opens NWC Settings
- [x] Cashu switch on and properly configured, Cashu balance 0 (Lightning balance irrelevant): shows "No Cashu balance available", link hidden
- [x] Cashu switch on but mint not configured (fallback to Lightning), Lightning balance 0: shows "No Lightning balance available", link hidden (switch is already on)
- [x] Backend without Cashu support (LND, CLN, LNC): link always hidden
- [x] Ecash disabled in wallet settings: link hidden

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [x] LDK Node
- [ ] Embedded LND

Remote
- [x] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


kaloudis:
Concept ACK. Pointing the empty-budget dead end at the one setting that resolves it is the right idea, and the navigation round trip works: `componentDidMount`'s focus listener calls `loadData`, which calls `loadMaxBudget`, so coming back from NWC Settings with the switch on re-reads the budget rather than leaving a stale zero on screen. `NWCSettings` is a valid route. CI is green.

One must-fix before this can merge, plus a related gap.

### Must fix: the reworded message is wrong whenever Cashu is the funding source

`maxBudgetLimit` is not always the Lightning balance. `NostrWalletConnectStore.loadMaxBudget` branches on `isCashuConfigured` (`NostrWalletConnectStore.ts:3829`):

```ts
if (this.isCashuConfigured) {
    this.maxBudgetLimit = this.cashuStore.totalBalanceSats || 0;
} else {
    const balance = await this.balanceStore.getLightningBalance(true);
    this.maxBudgetLimit = Number(balance?.lightningBalance) || 0;
}
```

So `maxBudgetLimit === 0` has two meanings, and this PR hardcodes one of them into the copy. A user who has already enabled the NWC Cashu switch, has a mint, and has spent their ecash down to zero now gets told "No Lightning balance available" on a connection that does not spend Lightning at all, and quite possibly while their Lightning balance is healthy. The old generic string was vague but true in both states; the new one is specific and false in one of them.

The link is correctly hidden there, so the only thing wrong is the text. Two ways out:

- Branch the copy on `NostrWalletConnectStore.isCashuConfigured`: the Lightning string when Lightning is the source, an ecash-specific string when Cashu is. This is the better outcome for the user, since "no ecash balance" tells them exactly what to top up.
- Or revert the rename, keep `noBalanceAvailable` generic, and ship only the link. Smaller diff, and the link already carries the actionable information.

I would take the first. Either is fine, but the current state is a regression for Cashu users.

### Should fix: the gate uses `cashuEnabled`, the balance uses `isCashuConfigured`

`canEnableCashu` checks `!NostrWalletConnectStore.cashuEnabled`, but what actually selects the balance source is `isCashuConfigured`, which is `cashuEnabled && cashuStore.isProperlyConfigured()` (`NostrWalletConnectStore.ts:3814`). Those come apart, and when they do the result is exactly the dead end this PR exists to remove:

`isProperlyConfigured()` (`CashuStore.ts:1946`) requires a non-empty `mintUrls`, a selected mint, and an initialized wallet for it. With the NWC switch on but any of those missing, `loadMaxBudget` silently falls back to the Lightning balance, and if that is zero the user sees the empty-budget box with no link, because `cashuEnabled` is true. Two ways to land there:

- The user removes their last mint after turning the NWC switch on.
- Cold start. `cashuWallets[selectedMintUrl]` is populated asynchronously, so a screen focused before CDK finishes reads `isCashuConfigured` as false. `maxBudgetLimit` is only recomputed on focus, so the stale Lightning-based zero sticks until the screen is focused again.

The second one is pre-existing, not something you introduced, but it is worth deciding what this screen should say in that state rather than leaving it as the one case the new affordance does not cover. At minimum the message should not claim the Lightning balance is the problem, which the first fix above would handle.

### Nits

1. **Reaching through the store for settings.** `NostrWalletConnectStore?.settingsStore?.settings?.ecash?.enableCashu` works, but `SettingsStore` is a registered store and this component just does not inject it. Either add it to `@inject` and read `this.props.SettingsStore.settings...`, or better, move the whole predicate onto `NostrWalletConnectStore` as a `@computed` next to `isCashuConfigured`. `loadCashuSetting` already computes `supportsCashuWallet() && settings.ecash.enableCashu` (`NostrWalletConnectStore.ts:3689`), so the view is duplicating store logic that the store should own.

2. **Locale checkboxes.** You added a new key (`enableCashuInSettings`), so "I've added new locale text that requires translations" applies and should be checked. Also worth saying in the description that rewording the existing `noBalanceAvailable` leaves every other locale on the old string until Transifex catches up. That is expected, not a problem, but a reviewer should not have to work it out.

3. **Android is unchecked.** This is a small view change and the screenshot is iOS, so an Android look at the box with the link present is cheap and would complete the matrix.

4. `textAlign: 'center'` on the message is a no-op while the string fits on one line, since the parent `View` already has `alignItems: 'center'`. It does start to matter now that the string is longer and can wrap, so keeping it is fine, just noting it is doing less than it looks like it is.

No tests owed: this is view-only and there are no component tests for this screen.

Happy to re-review quickly once the message branches on the actual balance source.

ajaysehwal:
> Concept ACK. Pointing the empty-budget dead end at the one setting that resolves it is the right idea, and the navigation round trip works: `componentDidMount`'s focus listener calls `loadData`, which calls `loadMaxBudget`, so coming back from NWC Settings with the switch on re-reads the budget rather than leaving a stale zero on screen. `NWCSettings` is a valid route. CI is green.
> 
> One must-fix before this can merge, plus a related gap.
> 
> ### Must fix: the reworded message is wrong whenever Cashu is the funding source
> `maxBudgetLimit` is not always the Lightning balance. `NostrWalletConnectStore.loadMaxBudget` branches on `isCashuConfigured` (`NostrWalletConnectStore.ts:3829`):
> 
> ```ts
> if (this.isCashuConfigured) {
>     this.maxBudgetLimit = this.cashuStore.totalBalanceSats || 0;
> } else {
>     const balance = await this.balanceStore.getLightningBalance(true);
>     this.maxBudgetLimit = Number(balance?.lightningBalance) || 0;
> }
> ```
> 
> So `maxBudgetLimit === 0` has two meanings, and this PR hardcodes one of them into the copy. A user who has already enabled the NWC Cashu switch, has a mint, and has spent their ecash down to zero now gets told "No Lightning balance available" on a connection that does not spend Lightning at all, and quite possibly while their Lightning balance is healthy. The old generic string was vague but true in both states; the new one is specific and false in one of them.
> 
> The link is correctly hidden there, so the only thing wrong is the text. Two ways out:
> 
> * Branch the copy on `NostrWalletConnectStore.isCashuConfigured`: the Lightning string when Lightning is the source, an ecash-specific string when Cashu is. This is the better outcome for the user, since "no ecash balance" tells them exactly what to top up.
> * Or revert the rename, keep `noBalanceAvailable` generic, and ship only the link. Smaller diff, and the link already carries the actionable information.
> 
> I would take the first. Either is fine, but the current state is a regression for Cashu users.
> 
> ### Should fix: the gate uses `cashuEnabled`, the balance uses `isCashuConfigured`
> `canEnableCashu` checks `!NostrWalletConnectStore.cashuEnabled`, but what actually selects the balance source is `isCashuConfigured`, which is `cashuEnabled && cashuStore.isProperlyConfigured()` (`NostrWalletConnectStore.ts:3814`). Those come apart, and when they do the result is exactly the dead end this PR exists to remove:
> 
> `isProperlyConfigured()` (`CashuStore.ts:1946`) requires a non-empty `mintUrls`, a selected mint, and an initialized wallet for it. With the NWC switch on but any of those missing, `loadMaxBudget` silently falls back to the Lightning balance, and if that is zero the user sees the empty-budget box with no link, because `cashuEnabled` is true. Two ways to land there:
> 
> * The user removes their last mint after turning the NWC switch on.
> * Cold start. `cashuWallets[selectedMintUrl]` is populated asynchronously, so a screen focused before CDK finishes reads `isCashuConfigured` as false. `maxBudgetLimit` is only recomputed on focus, so the stale Lightning-based zero sticks until the screen is focused again.
> 
> The second one is pre-existing, not something you introduced, but it is worth deciding what this screen should say in that state rather than leaving it as the one case the new affordance does not cover. At minimum the message should not claim the Lightning balance is the problem, which the first fix above would handle.
> 
> ### Nits
> 1. **Reaching through the store for settings.** `NostrWalletConnectStore?.settingsStore?.settings?.ecash?.enableCashu` works, but `SettingsStore` is a registered store and this component just does not inject it. Either add it to `@inject` and read `this.props.SettingsStore.settings...`, or better, move the whole predicate onto `NostrWalletConnectStore` as a `@computed` next to `isCashuConfigured`. `loadCashuSetting` already computes `supportsCashuWallet() && settings.ecash.enableCashu` (`NostrWalletConnectStore.ts:3689`), so the view is duplicating store logic that the store should own.
> 2. **Locale checkboxes.** You added a new key (`enableCashuInSettings`), so "I've added new locale text that requires translations" applies and should be checked. Also worth saying in the description that rewording the existing `noBalanceAvailable` leaves every other locale on the old string until Transifex catches up. That is expected, not a problem, but a reviewer should not have to work it out.
> 3. **Android is unchecked.** This is a small view change and the screenshot is iOS, so an Android look at the box with the link present is cheap and would complete the matrix.
> 4. `textAlign: 'center'` on the message is a no-op while the string fits on one line, since the parent `View` already has `alignItems: 'center'`. It does start to matter now that the string is longer and can wrap, so keeping it is fine, just noting it is doing less than it looks like it is.
> 
> No tests owed: this is view-only and there are no component tests for this screen.
> 
> Happy to re-review quickly once the message branches on the actual balance source.

Thanks for catching both — pushed 4d2f69d.

- Message now branches on `isCashuConfigured` (same predicate `loadMaxBudget` uses), with a new `noCashuBalanceAvailable` string. This also fixes the cold-start/misconfigured-mint case for free, since that state also reads `isCashuConfigured` as false.
- Moved the predicate onto the store as `isCashuAvailable` / `canOfferCashuSwitch` (`@computed`, next to `isCashuConfigured`), and `loadCashuSetting` now reuses `isCashuAvailable` instead of duplicating it. View no longer touches `SettingsStore` — `canEnableCashu()` is gone.
- Dropped `textAlign: center` — you're right it was doing nothing given the parent's `alignItems: center`.

kaloudis:
Thanks for the quick turnaround. Re-reviewed at 4d2f69d.

The round-1 must-fix is resolved: the copy now follows `isCashuConfigured`, so a user with the switch on and empty ecash is told the right thing. Moving the predicate into `isCashuAvailable` / `canOfferCashuSwitch` next to `isCashuConfigured` is exactly what I was after. Everything the computeds read is observable (`cashuEnabled`, the settings blob, `implementation` via `BackendUtils.call`, and CashuStore's `mintUrls` / `selectedMintUrl` / `cashuWallets`), so nothing goes stale. The `loadCashuSetting` change is behavior-neutral. The "switch on, mint not configured" case also behaves correctly: Lightning copy, link hidden.

### Should fix: the copy and the number can still disagree

The description says the message and the balance source can never disagree anymore. They can. The copy reads `isCashuConfigured` live on every render, but the zero on screen is `this.state.maxBudgetLimit`, a snapshot taken on focus (`loadData` -> `loadMaxBudget` -> `updateMaxBudgetLimit`).

Cold start: if `loadMaxBudget` runs before CDK has populated `cashuWallets`, `isCashuConfigured` is false, the Lightning balance is read, and 0 is stored. Once CDK finishes, `isCashuConfigured` flips to true and the observer re-renders "No Cashu balance available" over a Lightning zero, possibly for a user whose ecash is funded. So the cold-start case isn't fixed for free: it goes from a vague message to a wrong one.

Suggest recording the source inside `loadMaxBudget` (e.g. an `@observable` next to `maxBudgetLimit` that you set in each branch) and branching the copy on that rather than on the live predicate. That makes the copy describe the number actually on screen.

### Nits

- The category is ticked as "Code refactor". This is a bug fix / UX change.
- The test plan mentions Embedded LND but that box is unchecked. Please tick it if you ran it.
- Please squash 4d2f69d into the first commit before merge.

kaloudis (commented):
Re-reviewed at 5f51a7e.

The round-2 should-fix is resolved. `maxBudgetSource` is set in the same `runInAction` as `maxBudgetLimit` in every branch of `loadMaxBudget`, including the error path, and the copy branches on it. The cold-start case now reads "No Lightning balance available" over a Lightning zero. That's accurate, and the link is hidden because the switch is on. Code ACK.

### Optional nit

The number on screen is still `this.state.maxBudgetLimit`, copied in `updateMaxBudgetLimit` and `loadConnectionForEdit`. The copy reads the live store `maxBudgetSource`. The store can reload while this screen is focused: `initializeService` calls `loadMaxBudget`, and each NWC payment calls `scheduleMaxBudgetRefresh`. If `isCashuConfigured` flips between the two loads, the copy changes and the number doesn't. The window is small. Copying `maxBudgetSource` into state in the same two `setState` calls would close it.

### Before merge

- Please squash the three commits into one.
- The test plan mentions Embedded LND but that box is still unchecked.

ajaysehwal:
> Re-reviewed at [5f51a7e](https://github.com/ZeusLN/zeus/commit/5f51a7ef3c549dd66b2c31a84e546dac4184cd90).
> 
> The round-2 should-fix is resolved. `maxBudgetSource` is set in the same `runInAction` as `maxBudgetLimit` in every branch of `loadMaxBudget`, including the error path, and the copy branches on it. The cold-start case now reads "No Lightning balance available" over a Lightning zero. That's accurate, and the link is hidden because the switch is on. Code ACK.
> 
> ### Optional nit
> The number on screen is still `this.state.maxBudgetLimit`, copied in `updateMaxBudgetLimit` and `loadConnectionForEdit`. The copy reads the live store `maxBudgetSource`. The store can reload while this screen is focused: `initializeService` calls `loadMaxBudget`, and each NWC payment calls `scheduleMaxBudgetRefresh`. If `isCashuConfigured` flips between the two loads, the copy changes and the number doesn't. The window is small. Copying `maxBudgetSource` into state in the same two `setState` calls would close it.
> 
> ### Before merge
> * Please squash the three commits into one.
> * The test plan mentions Embedded LND but that box is still unchecked.

Thanks, squashed the three commits into one and tick the Embedded LND box once I've re-run it.

On the nit: agreed the window is small. I'll keep it as is, since `maxBudgetSource` is set in the same `runInAction` as `maxBudgetLimit`, and any reload re-renders the screen from the store. If you'd rather I copy the source into state alongside the limit,

kaloudis (approved):

PR #4664, opened 2026-09-16
fix: LSP and LSPS1 reset button visibility and Olympus detection

### LINKED ISSUE #4656: LSP: Reset button disappears after editing the LSP URL
### Describe the bug

On Settings → LSP, the Reset button appears while you edit the LSP URL, then disappears a moment later. If the LSP URL is the only setting that differs from the defaults, this screen gives no way to reset it.

### Reproduce

- Go to Settings → LSP, with Enable LSP on
- Change the LSP URL (e.g. https://invalid.example)
- Reset appears briefly, then disappears
- Leave and reopen the screen: there's still no Reset button, even though the URL isn't the default

### ZEUS version

v13.2.2

### Node interface

Embedded LND

### Network

Tor

### Device

iPhone 16 Pro

### Device operating system

18.6

### Log output

```shell

```

shubhamkmr04:
# Description

Relates to issue: https://github.com/ZeusLN/zeus/issues/4656

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [X] Yes
- [ ] N/A

- [X] Android
- [X] iOS

On-device
- [X] LDK Node
- [X] Embedded LND

Remote
- [X] LND (REST)
- [X] LND (Lightning Node Connect)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


ajaysehwal (commented):
@shubhamkmr04 , the diff itself looks correct — comparing against the hardcoded network defaults instead of the live settings value fixes the reset-button issue cleanly.

I did spot the same bug class in `LSPStore.isOlympus()` (`stores/LSPStore.ts:222-225`): `this.getLSPS1Rest() === config.lsps1Rest` compares the current value against itself, so it’s effectively always `true` for REST LSPS1.

That means the Olympus branding can still be shown on LSPS1/LSPS7 screens even when a user has configured a completely custom third-party REST LSP. This looks pre-existing — I traced it back to `6d01c05f1` — so it’s out of scope for this PR.

Since this PR now exposes `defaultLsps1Rest`, that could be used in a follow-up as `this.getLSPS1Rest() === config.defaultLsps1Rest`. Just flagging it as a related follow-up rather than something that needs to block this PR.

kaloudis (commented):
Reviewed at `ef7f36c1f`. Approach ACK: the diagnosis is right, and comparing against the network default constant matches how the sibling reset screens already work (`getDefaultEsploraServer(network)` in `views/Settings/EmbeddedNode/EsploraServer.tsx:96`, `getDefaultRgsServer` in `RapidGossipSync.tsx:94`). Verified on the PR head: `tsc`, `eslint` and `prettier` clean, and the jest suite passes.

Three things I'd like fixed before merge, plus one follow-up.

### 1. The other half of the same predicate is inverted (`views/Settings/LSP.tsx:72`)

`requestSimpleTaproot ||` treats `true` as "differs from default", but the default is `true` (`stores/SettingsStore.ts:1603`), and `MigrationUtils` forces it true for everyone once via `lsp-taproot-mod` (`utils/MigrationUtils.ts:344`). The reset handler then writes `false` (lines 333, 342).

Blame explains it: `showReset` is from 010a22eb39 (2023-11) and the reset handler from 727bbefa8e (2024-01), both predating 47ddd53fe (2024-02-24), which flipped the default to true. So on a default install today the Reset button is permanently visible with nothing customized, and tapping it silently disables simple taproot channel requests, writing a non-default value. Since this PR is about reset-button visibility, worth flipping here: `!requestSimpleTaproot ||`, and reset writing `requestSimpleTaproot: true`.

### 2. On LND and LNC the LSPS1 Reset button is still hidden (`views/LSPS1/Settings.tsx:191`)

`isOlympus` ORs the two transport checks, and both LND (`backends/LND.ts:1051-1052`) and LNC (`backends/LightningNodeConnect.ts:909-910`) report `supportsLSPScustomMessage()` and `supportsLSPS1rest()` true. Change only the REST host and `isOlympusCustomMessage` is still true (pubkey and host untouched), so `isOlympus` stays true: no Reset button, and the Olympus logo still renders. That is the #4656 symptom one screen over.

Suggest splitting the two concerns, keeping `isOlympus` for branding and gating the button on "every applicable field is at its default":

```ts
const usesCustomMessage = BackendUtils.supportsLSPScustomMessage();
const usesRest = BackendUtils.supportsLSPS1rest();
const isDefaultConfig =
    (!usesCustomMessage ||
        (pubkey === lspConfig.defaultPubkey &&
            host === lspConfig.defaultLsps1Host)) &&
    (!usesRest || restHost === lspConfig.defaultLsps1Rest) &&
    lsps1Token === '';
```

### 3. `lsps1Token` isn't part of the reset-visibility check

`handleReset` clears it (`views/LSPS1/Settings.tsx:161`), but nothing in the visibility condition looks at it, so setting only a discount code leaves the user with no Reset button. `LSP.tsx` handles the analogous `accessKey !== ''` correctly.

### 4. On @ajaysehwal's `LSPStore.isOlympus()` point: agreed, with one correction

The impact isn't only branding. `views/LSPS1/index.tsx:1190` hides the Announce Channel toggle when `isOlympus`, and line 447 labels the provider "Olympus by ZEUS", so someone pointed at a third-party LSPS1 REST provider gets Olympus branding on the purchase screen and loses the announce toggle.

The one-line fix isn't quite enough though: LSPS1 picks REST before custom message (`views/LSPS1/index.tsx:122-128`, `stores/LSPStore.ts:1376-1379`), while `isOlympus()` checks custom message first, so on LND and LNC it answers for a transport LSPS1 never uses. LSPS7 is custom-message only, so the check likely needs to be service-aware rather than a single boolean. Fine as a follow-up PR, but let's not lose it.

### Smaller notes, all pre-existing, fine to leave here

- Both reset handlers overwrite the mainnet, testnet and mutinynet keys, while the visibility checks only look at the current network. Resetting on testnet wipes a customized mainnet LSP.
- The custom-message branch in `views/LSPS1/Settings.tsx` writes `lsps1Pubkey/HostTestnet|Mainnet` keyed off `isTestNet`, but mutinynet is a separate flag (`models/NodeInfo.ts:63`) that `getLspConfigForNetwork` reads first. On Embedded LND + mutinynet those edits land in the mainnet keys and are lost on reopen. The native branch gets this right.
- `supportsLSPS1native()` is false on every backend (only `LdkNode` defines it, as `false`), so that whole branch, including `isOlympusNative`, is currently dead code.

### Testing

`stores/SettingsStore.test.ts` already exists and `getLspConfigForNetwork` is pure and exported, so roughly 20 lines pinning that `defaultFlowHost` / `defaultLsps1Rest` / `defaultLsps1Host` stay equal to the constants while the matching `settings.lsp*` keys hold custom values would lock in exactly this regression. The template's "N/A" isn't right when a utility's return shape changed.

The PR is iOS-only with LDK Node and Embedded LND, but the LSPS1 half mostly affects LND, LNC and CLNRest. Please add an Android pass and at least one LND (REST) run before merge.

shubhamkmr04:
@kaloudis thanks, addressed all four:

1. `showReset` now treats simple taproot as on by default, and reset writes `true` (5c5d897d8)
2. LSPS1 reset is gated on every field the backend uses being at its default, separate from the `isOlympus` branding (1a580bf61)
3. `lsps1Token` is part of that check (46791af17)
4. `isOlympus()` now takes the service and follows the transport each one actually uses (351061cc4)

Added `getLspConfigForNetwork` tests in `SettingsStore.test.ts` (ae9e6afbf). Rebased on master, tested on Android and iOS with LDK Node, Embedded LND and LND (REST).

kaloudis (commented):
Round 2 review of ae9e6afbf.

All four items from the last round are resolved:

- Flow screen reset visibility now compares against `defaultFlowHost` and treats `requestSimpleTaproot` as default-true, with Reset restoring `true` (10b8f3059, 5c5d897d8)
- LSPS1 Settings splits branding from `isDefaultConfig`, so on backends with both transports the Reset button stays until every field the backend actually uses is back at default (1a580bf61)
- `lsps1Token` is included in the visibility check (46791af17)
- `LSPStore.isOlympus` is now service- and transport-aware (351061cc4). I verified the priority (native, then REST, then custom message) matches the actual selection order in `views/LSPS1/index.tsx` (`fetchGetInfo` and the order-creation branch)
- The new `getLspConfigForNetwork` test pins the default fields against custom settings on all three networks

Verified locally on ae9e6afbf: tsc clean, `stores/SettingsStore.test.ts` 8/8 pass. CI green.

One residual item in scope for this PR:

1. `views/LSPS1/Settings.tsx:191`: the branding `isOlympus` still ORs the three transports, while the store now resolves by priority. On LND or LNC (which support both custom message and REST), a custom REST host with a default pubkey keeps `isOlympusCustomMessage` true, so the Olympus logo shows while purchases actually go through the custom REST endpoint. Mirroring the store's chain fixes it:

```ts
const isOlympus = BackendUtils.supportsLSPS1native()
    ? isOlympusCustom
    : BackendUtils.supportsLSPS1rest()
    ? isOlympusRestMatch
    : isOlympusCustomMessage;
```

Two optional nits, fine as follow-ups:

- `views/Settings/LSP.tsx:75`: `!requestSimpleTaproot` is not gated on `supportsSimpleTaprootChannels()`. LDK Node shows the Flow screen but hides the taproot toggle, so a stale `false` saved from an LND wallet makes Reset appear with no visible non-default field. `BackendUtils.supportsSimpleTaprootChannels() && !requestSimpleTaproot` would cover it.
- `stores/LSPStore.ts:216`: the comment says LSPS7 only speaks custom messages, but LDK Node uses native LSPS7 (`lsps7CreateOrder` branch at line 1279). The pubkey check is still correct for both paths since they key off the same setting; just worth rewording so the comment does not mislead.

With item 1 addressed this is ACK from me on the code side. The reported testing matrix (Android + iOS across LDK Node, Embedded LND, LND REST) covers the affected flag combinations.

shubhamkmr04:
Fixed in 39378a3a4 — `isOlympus` on the settings screen now mirrors the store's priority chain (native → REST → custom message), and the three per-transport consts are gone.

Tested on LNC, where a custom REST host with the default pubkey no longer shows the Olympus logo, plus LDK Node and Embedded LND for no regression. 

Taking the two nits as a follow-up PR.

views/LSPS1/Settings.tsx
```diff
@@ -177,18 +177,28 @@ export default class LSPS1Settings extends React.Component<
         const isTestNet = nodeInfo?.isTestNet;
 
         const isOlympusCustom =
```
ajaysehwal, line 180: @shubhamkmr04, one inconsistency: `LSPStore.isOlympus()` checks only the pubkey for the native transport, while `LSPS1/Settings.tsx` checks both the pubkey and the host.

So a native LDK Node user with the default pubkey but a custom host can see the Olympus logo on the order screen (`LSPS1/index.tsx`) but not in Settings.

Since the comment says this should match the store’s priority order, it looks like the native case should drop the host check here as well.

shubhamkmr04: Fixed, branding now keys off the pubkey alone on both peer transports, matching the store. The pubkey identifies the node; the host is only where it is reached, so the same LSP on a different address is still Olympus. The reset button keeps the host check, since host is an editable field that has to be resettable.

Note the same mismatch applied to the custom message branch, which is the one that actually runs today — `supportsLSPS1native()` is only defined on LdkNode, as `false`.

@kaloudis this reverses the native half of your snippet from the last round (`isOlympusCustom` -> pubkey only); flagging in case you prefer the host check kept and the store changed instead.

kaloudis:
Round 3 review of `51f062769`.

Item 1 from the last round is resolved, and the extra `51f062769` on top is the right call. `views/LSPS1/Settings.tsx:188-192` now mirrors the store's priority chain, and dropping the host from the branding test matches `LSPStore.isOlympus` (`stores/LSPStore.ts:220`): the pubkey is what identifies the node on a peer transport, the host is only where it is reached, and you cannot complete a Noise handshake with a custom host unless it really is that node. `isOlympusCustom` staying pubkey-and-host for `isDefaultConfig` is correct, since Reset has to stay up until every field is back at its default.

Verified on the PR head: `tsc` clean, `eslint` and `prettier` clean, `stores/SettingsStore.test.ts` 8/8, full `jest` 1790/1791 (the one failure is an unrelated path artifact of my review worktree). I also re-walked `isDefaultConfig` against the capability matrix: Embedded LND (custom message only) gates on pubkey, host and token; CLNRest and LDK Node (REST only) gate on REST host and token; LND and LNC (both) gate on all four. Each of those fields is rendered on the screen in every one of those cases, so the button never appears without a visible non-default field.

## One last item, same shape as the one you just fixed

`views/LSPS1/index.tsx:449-455`. `isOlympus` is now correct on that screen, but the fallback it falls through to still orders the transports custom message before REST, while everything else on the screen picks native, then REST, then custom message (`fetchGetInfo` at 122-128, the retry button at 1281-1293, the order-creation branch at 1336-1341, and `LSPStore.lsps1CheckOrderStatus` at 1447).

On LND or LNC with a custom LSPS1 REST host and the pubkey left at its default, `supportsLSPScustomMessage()` is true, so the `LSP` row in the service-info accordion (line 623, with a copy icon) renders `LSPStore.getLSPSPubkey()`: Olympus's pubkey, presented as the authoritative identifier for an LSP that is never contacted over that transport. `getInfoData` next to it came from `lsps1GetInfoREST()` against the custom host, so the two disagree.

This is pre-existing (`634c047b3`, `67738cf22`), but it only becomes reachable because of this PR: on master `isOlympus` returned true for that user and the row read "Olympus by ZEUS", so the fallback never ran. It is the same misbranding one line down from the one `51f062769` fixed, so I would rather it went in here than in a follow-up:

```ts
const lspDisplay = isOlympus
    ? 'Olympus by ZEUS'
    : BackendUtils.supportsLSPS1native()
    ? localeString('views.LSPS1.nativeLsps1Enabled')
    : BackendUtils.supportsLSPS1rest()
    ? LSPStore.getLSPS1Rest()
    : LSPStore.getLSPSPubkey();
```

`views/LSPS7/index.tsx:315-317` needs no change: LSPS7 keys off the pubkey on both its native and custom-message paths, so the pubkey fallback is already the right one there.

Taking the two nits as a follow-up is fine. Please open the issue or PR and link it here so the taproot gating and the `LSPStore.ts:216` comment do not get lost.

With the above this is ACK from me on the code side. Your reported matrix (Android and iOS across LDK Node, Embedded LND, LND REST, plus the LNC run for the branding fix) covers the affected flag combinations; a quick re-check of the `LSP` row on LNC with a custom REST host after the change would close it out.

shubhamkmr04:
@kaloudis Fixed in aae0ca56e — `lspDisplay` now falls through native → REST → custom message, matching `fetchGetInfo` and the rest of the screen.

Re-checked on LNC with a custom REST host: the `LSP` row in the service-info accordion now shows the REST URL instead of the Olympus pubkey, and reverts to "Olympus by ZEUS" once the host is back at its default. Also re-ran LDK Node and Embedded LND for no regression, on Android and iOS.

The two nits are up as #4705 (draft, cut from this branch — it will show just its own two commits once this merges).

kaloudis (approved):
Round 4 review of `aae0ca56e`. ACK.

The last item is resolved. `views/LSPS1/index.tsx:449-455` now falls through native, then REST, then custom message, so the `LSP` row in the service-info accordion (line 623) names the endpoint LSPS1 actually talks to. That matches `fetchGetInfo` (119-130), the retry button (1281-1293), the order-creation branch (1336-1341), `LSPStore.lsps1CheckOrderStatus`, and `LSPStore.isOlympus` (`stores/LSPStore.ts:219-239`). On LND or LNC with a custom LSPS1 REST host and the pubkey left at its default, the row reads the REST URL instead of Olympus's pubkey, and your reported LNC re-check confirms it reverts to "Olympus by ZEUS" once the host is back at its default.

Re-confirmed `views/LSPS7/index.tsx:313-317` needs no change: both the native and custom-message LSPS7 paths key off the same pubkey setting, so the pubkey fallback is the correct one there.

Verified on the PR head: `tsc`, `eslint` and `prettier` clean, `stores/SettingsStore.test.ts` 8/8, full `jest` 1790/1791 (the one failure is the known `PhotoUtils` path artifact of my review worktree, not this branch). All four CI checks green. Your matrix (Android and iOS across LDK Node, Embedded LND and LND REST, plus the LNC runs for the two branding fixes) covers the affected flag combinations.

Thanks for #4705 for the taproot gating and the `LSPStore.ts:215-217` comment. I'll review it once this lands.

### Out of scope here, but the same class, if you want a third issue

`views/OpenChannel.tsx:197-208` and `views/Channels/ChannelsPane.tsx:147-154` both label the channel partner `'Olympus by ZEUS'` unconditionally while sourcing the pubkey and host from `lspConfig.lsps1Pubkey` / `lspConfig.lsps1Host`, which are exactly the fields the LSPS1 settings screen lets you customize. The dropdown at `views/OpenChannel.tsx:678-699` does the same on selection. So a user who points LSPS1 at a third-party peer gets that peer offered as "Olympus by ZEUS" on the open-channel screen, and because the pubkey and host inputs only render under the `Custom` branch (line 711), there is nothing on screen showing where the channel would actually go.

Unlike the `lspDisplay` case, this one is untouched by this PR and is not newly reachable because of it: those screens never consulted `isOlympus`, so the label was already wrong on master for the same user. Worth its own issue rather than growing this branch.

kaloudis:
Filed the out-of-scope item as #4710 and assigned it to you, @shubhamkmr04. Not a blocker for this PR, and no rush on it.

One correction to what I wrote above: only `views/OpenChannel.tsx` has the user-visible defect. `views/Channels/ChannelsPane.tsx` sets the same `channelDestination` and `node_pubkey_string` state at `97-100` and `141-155`, but neither field is read anywhere in that file and the state is not spread into a child, so it is dead rather than wrong. The issue covers deleting it alongside the fix.

Issue #2390, opened 2024-09-01
Only last 500 invoices displayed when using LND


yanascz:
### Describe the bug

It seems that Zeus fetches only the last 500 invoices, no matter what date filters are used.

### Reproduce

1. Go to the _Activity_ screen.
2. Tap the filter icon and disable _On-chain payments_ and _Sent_.
3. Go back and scroll down to the last invoice.
4. Note the date of the last invoice.
5. Tap the filter icon again and set _End Date_ to the date from previous step.
6. Go back and notice that no older invoices are visible.

### ZEUS version

v0.9.0

### Node interface

LND (REST)

### Network

Clearnet

### Device

iPhone 15 Pro

### Device operating system

17.6.1

Wajid-0ne:
Hello @shubhamkmr04  I'd like to help with this. I've traced the flow and the root cause is clear, but the fix involves a deliberate performance tradeoff, so I'd like for maintainer to advice before proceeding.

> **Root cause**: 

ActivityStore.getActivity() fetches invoices, payments, and on-chain txs each capped at 500, most-recent-first, then merges and filters entirely client-side (including the date filter). So an older end-date can't surface anything beyond the most recent 500 per source — those records are never fetched. The Activity FlatList has onEndReachedThreshold set but no onEndReached handler, so there's no load-more today. The 500 cap looks intentional for performance (_and seems related to #1328 "slow to load with many transactions_"), which is why I don't want to just remove it blindly. Three approaches:

- A Raise/remove the cap (e.g. num_max_invoices=0): smallest change, fully fixes filtering, but loads entire history into memory → likely perf/memory regression on large wallets (the thing the cap prevents).

- B True pagination: wire onEndReached + index_offset, paginating and merging all three sources with loading states. Correct long-term, but a sizable architectural change that overlaps with #1328.

- C Filter-aware fetch (my suggestion): keep the 500 cap for the default view, but when a date filter is applied, fetch that query with a higher/unbounded limit so filters can reach older records — i.e. only pay the cost when the user actually filters.

> **Clarifications:**

Which direction do you prefer (A / B / C / other)?
Should this be solved together with #1328, or kept separate?
If C, is "unbounded fetch only while a date filter is active" an acceptable tradeoff, or do you want a hard upper bound?

_i'll be happy to implement once you advise on the approach._

kaloudis:
[...]

PR #4666, opened 2026-09-16
backends: filter LND activity by date

Wajid-0ne:
# Description

Relates to issue: **#2390**

Passes activity date bounds to supported LND backends while preserving the existing fallback for older nodes. Adds regression tests for date conversion and LND requests.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


Wajid-0ne:
hello @kaloudis made an update could you please give this a review when you available? thanks

kaloudis (commented):
Reviewed 1e01cac. The round-1 findings are all properly addressed, and I verified the trickier claims against the surrounding code:

- The upper-bound-only invoice fetch is correct: client-side date filtering uses `getDate`, which for paid invoices is the settle date (`models/Invoice.ts`), and settle >= creation, so `creation_date_end` at end-of-day cannot exclude an invoice settled in range. Payments filter by creation date on both sides, so both bounds are safe there.
- The boundary math matches `ActivityFilterUtils` exactly (local midnight start, inclusive end-of-day), both in device-local time against Unix timestamps.
- `VersionUtils.isSupportedVersion` returns false without throwing on an empty or missing version, and the nodeInfo prefetch covers the cold-start case with a graceful fallback to the unscoped fetch.
- Widening or clearing the date filter is safe because Activity refetches via `getActivityAndFilter` on every focus, and `ActivityExport` calls it directly.
- LndHub overrides both methods (and is not LND-based), external `getPayments`/`getInvoices` callers pass subsets of the new param types, and both the embedded `proto/lightning.proto` and the vendored lnc-core protos include the new fields.

A few remaining items:

1. `activityPayments` / `activityInvoices` in `stores/ActivityStore.ts` are never cleared. They persist across wallet switches and keep holding the previous node's Payment/Invoice objects until the next `getActivity`. Exposure today is narrow (only the LNC event listeners rebuild activity without going through `getActivity`, and only while the Activity screen is mounted), but this is one refactor away from a cross-wallet leak in the activity list. Please clear both at the top of `getActivity`, or add them to a reset invoked on disconnect.

2. With an `endDate` filter active, `updateInvoices` now refreshes only the private scoped list, so other consumers of `invoicesStore.invoices` no longer get the canonical refresh they used to get on LNC invoice events. If that is intentional (it is the point of the isolation), fine; just confirming it is deliberate.

3. Nit: `getLndInvoiceCreationDateRange(endDate)` is exactly `getLndCreationDateRange(undefined, endDate)`; consider delegating instead of duplicating the end-of-day math.

4. Nit: in `LND.getInvoices`, the `limit ? ...` check after `limit = params.limit ?? 500` means an explicit `limit: 0` silently drops the param and falls back to lnd's default of 100. No caller does that today.

Approach ACK. Before a tACK this needs device testing evidence:

- LND REST (0.16+), LNC, and embedded LND with a date filter spanning more than 500 items (the original #2390 repro)
- fallback behavior on a pre-0.16 node, or with the node version unavailable
- smoke check that CLN / NWC / LndHub activity is unchanged
- a date filter with an invoice created near local midnight, to validate the day-boundary math end to end

kaloudis (commented):
Reviewed `87464ca`. All four round-2 items are addressed:

1. `activityPayments` / `activityInvoices` are now cleared at the top of `getActivity`.
2. `updateInvoices` restores the canonical `invoicesStore.invoices` refresh alongside the scoped fetch.
3. `getLndInvoiceCreationDateRange` delegates to `getLndCreationDateRange(undefined, endDate)`.
4. `num_max_invoices` is always emitted, so an explicit `limit: 0` no longer silently drops the param.

Local checks on this head are clean: `tsc`, `eslint`, `prettier`, and the full Jest suite (the 6 suites touched here contribute 57 passing tests).

### Re-verified from first principles

- The boundary math is a provable superset of the client-side filter. `ActivityFilterUtils` uses local-midnight `>=` for the start and an exclusive `< endDate + 1 day` for the end; `getLndCreationDateRange` produces exactly local midnight and `nextDay - 1s`. Since `Payment.getDate` is the creation date and `Invoice.getDate` is the settle date (or `expires_at || created_at` when unpaid), and both are always at or after creation, no in-range item can fall outside a creation-date bound. Nothing can be silently dropped in either direction.
- The new fields exist on both native paths: `proto/lightning.proto:4225,4464` for embedded, and lnc-core's `lightning.d.ts:3249,3432` for LNC. Worth noting that `tsc` cannot validate these, because they are introduced via object spread and excess-property checking does not apply to spreads, so this needed manual confirmation. The snake_case request keys are correct: lnc's Go layer uses jsonpb, which accepts both field-name forms, and that is the same mechanism the existing `include_incomplete` and `num_max_invoices` keys rely on.
- Blast radius is contained. Only `LND`, `EmbeddedLND`, and `LightningNodeConnect` return `isLNDBased() === true`. LndHub returns false and also overrides both methods; CLN, NWC, and LDK are untouched. Request shapes for every existing caller (`SendingLightning`, `RebalancingChannels`, bare `getInvoices()`) are identical to before.

### New findings

**1. The `getNodeInfo()` prefetch has global side effects. (should fix)**

`stores/ActivityStore.ts:382-390`. On failure, `NodeInfoStore.getNodeInfo` runs `handleGetNodeInfoError()`, which sets `error = true` and `nodeInfo = {}`. Two consequences:

- `NodeInfoStore.error` drives the error banner on the wallet screen (`views/Wallet/Wallet.tsx:1484`, `views/Wallet/BalancePane.tsx:328`), so a failed prefetch triggered from the Activity screen flips a different screen into an error state.
- Wiping `nodeInfo` drops `nodeId`, which `getLSPOrders()` uses to select this node's orders, so LSPS1 and LSPS7 entries disappear from the very activity list being built.

It also supersedes any in-flight `getNodeInfo` through the `currentRequest` latch, which is live if the user reaches Activity before `Wallet.fetchData` settles.

The prefetch buys very little: by the time Activity is reachable, `fetchData` has populated `version`, and the unsupported path is exactly the pre-PR behavior and self-heals on the next focus refetch. I would drop the prefetch entirely. If you would rather keep it, please at least keep it from flipping global error state.

**2. Snapshot aliasing in the unfiltered branches. (should fix)**

`stores/ActivityStore.ts:398`, `:413`, `:442` assign `this.paymentsStore.payments` and `this.invoicesStore.invoices` by reference. MobX 5 replaces the ObservableArray instance on assignment, so those snapshots detach and `getSortedActivity` stops seeing later canonical refreshes. Leaving them `undefined` in the unfiltered branches makes the existing `??` fallback at `:307,:309` meaningful again, and it preserves the round-2 clearing behavior. Exposure today is bounded by the focus refetch, so this is about keeping the data flow honest rather than a user-visible bug.

**3. The canonical-refresh policy is now half-applied. (consistency)**

`updateInvoices` refreshes canonical plus scoped, but `getActivity` still skips the canonical refresh for both invoices and payments when a date range is active. `NostrWalletConnectStore` reads `invoicesStore.invoices` (`:2031`, `:2148`, `:3164`) and `paymentsStore.payments` (`:3156`). It does perform its own refreshes, so this looks survivable, but the two paths should follow one policy rather than two.

**4. `limit: 0` now behaves differently between the two methods. (nit)**

`getInvoices` always emits the limit, while `getPayments` keeps the `maxPayments ? ... : ''` check, and an absent `max_payments` means unlimited in lnd rather than 500. No caller passes 0 today, and the new test pins only the invoice half. Either drop both truthiness checks or keep both.

**5. Error-handling asymmetry in `getActivity`. (nit, pre-existing)**

The scoped invoice fetch swallows failures to `[]`, while the scoped payment fetch propagates and rejects `getActivityAndFilter`, which `Activity.componentDidMount` does not catch, leaving `state.loading` stuck true. `getPayments` already rethrew before this PR, so this is a follow-up rather than something to fix here.

**6. `backends/LND.test.ts` mutates the shared `nodeInfoStore` mock. (nit)**

The last case sets `nodeInfo = {}` and there is no `beforeEach` reset, which makes the suite order-dependent. A reset in `beforeEach` would fix it.

**7. One test-coverage gap worth closing.**

Nothing asserts the negative direction, that `isLNDBased() === false` leaves `getInvoices` and `getPayments` unscoped. That is the regression-risk path for CLN, NWC, and LDK, and it is a one-line mock flip in `stores/ActivityStore.test.ts`.

### Scope note

A start-date-only filter still returns only the newest 500 items, and the invoice half gets no server-side help at all without an end date. What this fixes is the #2390 symptom where an older range shows nothing, which is the right thing to fix here. Real pagination stays with #2671. Worth stating in the description so expectations are set.

### Before a tACK

Still owed, unchanged from round 2:

- Device evidence on LND (REST, 0.16+), LNC, and embedded LND with a date filter spanning more than 500 items, which is the original #2390 repro
- Fallback behavior on a pre-0.16 node, or with the node version unavailable
- Smoke check that CLN, NWC, and LndHub activity is unchanged
- A date filter with an invoice created near local midnight, to validate the day-boundary math end to end

Also please squash the two `fix:` commits into the feature commit before merge.

The approach continues to look right, and the test coverage you have added here is well above what I usually see for a change this size.

PR #4667, opened 2026-09-16
fix(modalbox): clamp fontScale-scaled sheet height to the container

kaloudis:
# Description

A user reported that they can select a UTXO in the coin-control picker but cannot confirm the selection — the sheet "won't scroll down any further" and the **Set** button is nowhere to be found.

image

Root cause is in `ModalBox`, not in `UTXOPicker`. `renderContent` multiplies a sheet's requested height by the device `fontScale` (added in 6addf53e9 to give larger text more room) but never clamps the result. Since `calculateModalPosition` clamps a `position="bottom"` sheet's offset at `0`, a sheet taller than its container pins to the **top** and spills its excess off the **bottom** of the screen — taking the footer with it.

`UTXOPicker` asks for `SCREEN_HEIGHT * 0.9`, so overflow begins at `fontScale > 1.11` — one step above default iOS Dynamic Type, or "Large" on Android. Past that point the Set/Cancel footer is off-screen, and since the sheet's only scrollable region is the UTXO `FlatList` in the middle, there is no way to reach it. The reporter's screenshot measures out at roughly `fontScale` 1.5–1.8, which makes the sheet about 1.5–1.7× the screen height.

`CurrencySelectorModal` uses the same `SCREEN_HEIGHT * 0.9` pattern and has the same defect — less visible there, since it has no footer to lose.

**This is a regression.** Before 89e113802 (v13.0.2) the picker requested `height: '95%'`, and `'95%' * FONT_SCALE` evaluates to `NaN`, which Yoga discards — so the scaling silently never applied. Switching to a numeric height activated it. Affected releases: v13.0.2 through v13.2.1.

The fix clamps the scaled height to `containerHeight` (seeded from `SCREEN_HEIGHT`, refreshed in `onContainerLayout`), so a sheet can never outgrow its container. This covers every `ModalBox` consumer that passes a numeric height, not just the UTXO picker.

### Follow-up not included here

At extreme accessibility font scales (~2.0+), `UTXOPicker`'s header (title, description, jumbo amount, account filter) is `flexShrink: 0` and the list body has `minHeight: 120`, so the header could still push the footer out even with a correctly sized sheet. My layout arithmetic says it holds to roughly `fontScale` 1.8 on a 914dp screen, but the margin is thin. The durable fix is moving the header into the `FlatList`'s `ListHeaderComponent` so the footer stays pinned — that changes the sheet's UX, so I left it out of this fix.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

No unit test added: the repo has no component-render tests (all 54 suites cover utils/models/stores), and exercising this path needs a mounted `ModalBox` inside a `SettingsStore` provider. **This change has not been verified on a device** — it is a static analysis of the layout math, so it needs a manual check at an elevated system font size before merge.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


kaloudis:
Follow-up after the reporter tried lowering their system font size and hit the same wall.

I pixel-measured the first screenshot to check my own diagnosis rather than assume it. The sheet's background is `modalBackground` `#31363F`, which decodes as rgb(49,54,62) — an exact match — and it runs from y=64 down to y=681, where it gives way to the body container's `background` `#1E2022`. Critically, **there is no band of `#31363F` below the list**: the sheet's own 16dp bottom padding is nowhere on screen, so the sheet's bottom edge is off-screen and the footer with it.

The decisive measurement is the sheet's **top** edge, which sits at y≈64 in the first screenshot and y≈62 in the second — **identical in both, despite the text being roughly 2.5× smaller in the second**. `ModalBox` renders inside an RN `<Modal>` without `statusBarTranslucent`, so the modal container starts just below the status bar — about 62px on this device. A sheet that *fits* would be positioned at `containerHeight - height`, i.e. visibly lower, and that offset would change with font size. Both screenshots being pinned at the container's top means `calculateModalPosition` returned a negative value and hit its `if (position < 0) position = 0` clamp **both times**. The sheet is taller than the screen in both.

So why didn't lowering the font size help? Because `FONT_SCALE` was captured once at module load:

```js
const { height: SCREEN_HEIGHT, width: SCREEN_WIDTH, fontScale: FONT_SCALE } = Dimensions.get('window');
```

Changing the system text size doesn't reload the JS bundle. On Android `MainActivity`'s `android:configChanges` is `keyboard|keyboardHidden|orientation|screenSize` — no `fontScale` — so the Activity is destroyed and recreated, but React Native holds the JS context at the Application level and reuses it, leaving module-level constants untouched. On iOS nothing restarts at all. The result is that `<Text>` re-renders at the new, smaller scale while the sheet keeps sizing itself from the old, larger one. That is exactly what the second screenshot shows: small text, sheet still overflowing, footer still unreachable. Only a full force-quit clears it.

The second commit reads `fontScale` from `Dimensions` inside `renderContent` instead, matching the existing pattern at `components/WalletHeader.tsx:476`. With the clamp, an over-tall sheet now self-corrects on the next render rather than persisting until the app is killed.

Note that `SCREEN_HEIGHT` is stale in the same way in the consumers (`UTXOPicker`, `CurrencySelectorModal`). I left those alone because the clamp bounds the damage: a stale-large `SCREEN_HEIGHT` now yields a full-height sheet that still lays its footer out correctly, rather than one that overflows.

Still unverified on a device — this remains static analysis plus screenshot forensics. The check that would confirm it: set the system font size well above default, open the UTXO picker, and confirm the **Usar UTXOs** / **Cancelar** buttons are reachable without force-quitting.

ajaysehwal (approved):
tACK

PR #4671, opened 2026-09-17
fix(navigation): prevent swipe-back to stale onboarding screen

ajaysehwal:
# Description

The boot-time transition from the loading `Wallet` screen to `IntroSplash`, and the equivalent transition when a user's last wallet is deleted from `WalletConfiguration`, used `navigation.navigate('IntroSplash')`. `navigate` pushes a new screen onto the stack when the target isn't already current — it doesn't replace the current one — so the calling screen (`Wallet` mid-loading, or a `WalletConfiguration` screen for a wallet that no longer exists) stayed in the stack underneath `IntroSplash`.

https://github.com/user-attachments/assets/fa102d9b-39b7-4400-9102-5f7932f5d468

## Root cause

`navigate` vs `replace`/`reset` for a screen that has nothing to go back to. iOS's interactive swipe-back gesture reveals the stale screen underneath. This is not iOS-only: `App.tsx`'s `handleBackPress` explicitly declines to intercept the hardware back button for `IntroSplash` (it's listed in `SCREENS_WITH_CUSTOM_BACK_HANDLER`), so Android's hardware back button has the identical exposure.

## Fix

Both call sites reset the navigation stack down to a single `IntroSplash` root: `navigation.reset({ index: 0, routes: [{ name: 'IntroSplash' }] })`. Nothing stale is left for swipe-back or hardware back to reveal, on either platform, regardless of how deep the stack was beforehand.

An earlier version of this fix used `navigation.replace('IntroSplash')` instead. Review caught that `replace` only swaps the *current* route, not the whole stack — it's sufficient for `Wallet.tsx` (the app's default initial route, so nothing precedes it on a normal boot), but `WalletConfiguration` is reached several pushes deep (`Menu`/`Wallets` → `WalletConfiguration`, all siblings in one flat top-level stack navigator in `App.tsx`), so `replace` there would still have left those earlier routes — now pointing at a deleted wallet — reachable underneath. `reset` removes that assumption entirely for both call sites rather than relying on stack position.

### Follow-up fix: advanced-setup first-wallet path

A second review pass caught a regression the `reset()` fix introduced: removing the stale `Wallet` root when there are no wallets configured broke `navigation.popTo('Wallet')` in `WalletConfiguration.tsx`'s `saveNodeConfiguration` (the calls that run after saving a wallet). `popTo` only collapses the stack down to a target route **if that route already exists on the stack** — verified directly against the installed `@react-navigation/routers` `StackRouter.js`:

```js
// POP_TO, when the target route isn't found in the stack:
const routes = state.routes.slice(0, currentIndex).concat(route);
```

When `Wallet` isn't present (exactly the case after the boot-time `reset()` above, on the "no wallets yet" path: `IntroSplash` → Advanced setup → `Menu` → `Wallets` → add wallet → save), `popTo('Wallet')` degrades to a plain pop-and-push, leaving `IntroSplash`/`Menu`/`Wallets` stranded underneath the newly active `Wallet` screen — the exact same stale-screen bug this PR exists to fix, now on the first-wallet creation path instead of the boot path.

Both `popTo('Wallet')` call sites in `saveNodeConfiguration` now use the same `reset({ index: 0, routes: [{ name: 'Wallet' }] })` shape. This is safe unconditionally: when `Wallet` is already the stack root (the normal case, since an existing install always boots there), `reset` produces the identical `[Wallet]` result `popTo` already did; when it's absent, `reset` is the correct behavior instead of the buggy one. A third `popTo('Wallet')` call site (`setWalletConfigurationAsActive`) was checked and left unchanged — it's only reachable when managing an already-existing wallet, which requires at least one wallet to have existed at boot, so the stale-root scenario never applies there.

## Files changed
- `views/Wallet/Wallet.tsx` — boot-time loading → onboarding transition
- `views/Settings/WalletConfiguration.tsx` — last-wallet-deleted → onboarding transition, and both `saveNodeConfiguration` post-save navigations (advanced-setup first-wallet regression fix)

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [ ] N/A

- [x] Android
- [x] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


views/Settings/WalletConfiguration.tsx
```diff
@@ -923,7 +923,16 @@ export default class WalletConfiguration extends React.Component<
...
                 // longer recoverable, and WAL/SHM sidecars).
                 await clearCDKDatabase();
-                navigation.navigate('IntroSplash');
+                // reset, not navigate/replace: this screen is normally
+                // reached several pushes deep (Menu/Wallets ->
+                // WalletConfiguration), and once its last wallet is
+                // deleted every one of those earlier routes is stale too
+                // - a plain replace only swaps this top entry and leaves
+                // them reachable underneath via swipe-back/hardware back
+                navigation.reset({
```
kaloudis, line 945: **SHOULD-FIX: this reset breaks `popTo('Wallet')` in `saveNodeConfiguration` for the first-wallet advanced-setup path**

The reset is correct here, but the stale `Wallet` root it removes was what made `navigation.popTo('Wallet')` (lines 739 and 745 in this file) collapse the stack after saving a first wallet. I checked the installed `@react-navigation/routers` `StackRouter`: when the target route is absent, `POP_TO` pops only the current route and pushes the target (`routes.slice(0, currentIndex).concat(route)`), it does not clear anything beneath.

Repro after this PR: boot with no wallets (or delete the last one) -> reset leaves `[IntroSplash]` -> Advanced set up -> Menu -> Wallets -> add wallet -> save. `index === activeNodeIndex` (0 === 0) so we hit `popTo('Wallet')`, and the stack becomes `[IntroSplash, Menu, Wallets, Wallet]` instead of `[Wallet]`. Consequences:

- iOS swipe-back from the main wallet screen (default `gestureEnabled`, no override for `Wallet`) walks back through Wallets, Menu, IntroSplash: the same bug class this PR fixes, inverted.
- Android hardware back on `Wallet` now pops to `Wallets` instead of exiting, because `handleBackPress` in `App.tsx` pops whenever `routes.length > 1`.

The quick Get Started flow is unaffected because `RecommendedSettings`/`WalletSettings` already `reset` to `[Wallet]` on success.

Suggested fix: make the two `popTo('Wallet')` sites in `saveNodeConfiguration` reset to `[{ name: 'Wallet' }]` as well. In every flow that reaches them, `Wallet` is either the stack root (reset is equivalent, and a reconnect is already being forced via `setConnectingStatus(true)`) or absent (reset is the correct behavior). If the remount side effects are a concern, gate on `navigation.getState().routes[0].name !== 'Wallet'`.

kaloudis (commented):
Reviewed the diff and traced both call sites, plus every other `IntroSplash`/`Wallet` navigation site, against master.

The diagnosis and both fixes are correct: `navigate` pushes and leaves the caller underneath, `IntroSplash` is in `SCREENS_WITH_CUSTOM_BACK_HANDLER` so Android hardware back has the same exposure as iOS swipe-back, and `reset` over `replace` is right for `WalletConfiguration` since it sits several pushes deep in the flat root stack. The reset shape matches the existing idiom in `RecommendedSettings.tsx` / `WalletSettings.tsx`, and `IntroSplash` reads no route params, so nothing is lost. Also verified the post-reset code in `Wallet.tsx` is safe: the `finally` setState and the deferred `processPendingShareIntent` now run after unmount, but setState after unmount is a no-op on React 18 and `pendingShareIntent` is never set on the no-wallets branch.

One should-fix, see the inline comment: the stale `Wallet` root this PR removes was silently load-bearing for `popTo('Wallet')` in `saveNodeConfiguration`, and removing it regresses the advanced-setup first-wallet path.

Nits, no action needed in this PR:

- `Lockscreen.tsx:253` still uses `popTo('IntroSplash')`. Same fallback semantics would leave `Wallet` underneath, but that branch (no passphrase, no pin, no nodes, no delete params) looks unreachable since `loginRequired()` is false without credentials. Fine to leave.
- On the last-wallet delete path, `justDeletedWallet: active` is written and now never cleared until the next `Wallets` visit; previously the stale `Wallets` screen could clear it on a swipe-back refocus. Pre-existing wart, fine as a follow-up.

Since the should-fix touches the same flow, please retest the advanced-setup path on device once addressed: IntroSplash -> Advanced set up -> Menu -> Wallets -> add wallet -> save -> then try swipe-back / hardware back from the wallet screen.

kaloudis:
One site got missed: `saveLdkNodeWalletConfiguration` also ends with `if (nodes.length === 1) navigation.popTo('Wallet')` (WalletConfiguration.tsx ~1243). That's the same first-wallet case, reachable via Advanced set up -> add wallet -> LDK Node, so it needs the same `reset` treatment as the two `saveWalletConfiguration` sites. `setWalletConfigurationAsActive` is fine to leave as noted.

views/Settings/WalletConfiguration.tsx
ajaysehwal: @kaloudis, confirmed against the installed `@react-navigation/routers` source — `POP_TO`’s “route not found” branch is exactly `state.routes.slice(0, currentIndex).concat(route)`, so your reasoning is correct: it doesn’t collapse anything below the current screen.

Fixed: both `popTo('Wallet')` calls in `saveNodeConfiguration` now use `reset({ index: 0, routes: [{ name: 'Wallet' }] })` unconditionally. That’s equivalent when `Wallet` is already the root and gives the correct result when it isn’t present.

ajaysehwal:
> One site got missed: `saveLdkNodeWalletConfiguration` also ends with `if (nodes.length === 1) navigation.popTo('Wallet')` (WalletConfiguration.tsx ~1243). That's the same first-wallet case, reachable via Advanced set up -> add wallet -> LDK Node, so it needs the same `reset` treatment as the two `saveWalletConfiguration` sites. `setWalletConfigurationAsActive` is fine to leave as noted.

fixed in [d2e24da](https://github.com/ZeusLN/zeus/pull/4671/commits/d2e24daf6071492333c5473be5d7f0ceea8f8a87)

kaloudis (approved):

PR #4674, opened 2026-09-18
locales: fix 17 broken localeString references, add en.json coverage test

kaloudis:
# Description

Relates to issue: N/A

17 locale keys referenced via `localeString()` did not exist in `locales/en.json`. The lookup chain in `utils/LocaleUtils.ts` falls back to `English[key]` with no final default, so each of these rendered as `undefined` at runtime: a blank header on the amount keypad, blank error messages in the Cashu, watchtower, and NWC flows, blank labels on the LSPS1/LSPS7 order screens, and more.

About half were typo'd references to strings that already exist under another name; those call sites were repointed with no new strings needed:

- `views.PendingHTLCs.noPendingHTLC` -> `noPendingHTLCs`
- `views.SetWalletPicture.choosePicture` -> `views.SetNodePicture.choosePicture`
- `views.Settings.SignMessage.selectAddress` -> mode-based `selectAddressSigning` / `selectAddressVerification`
- `views.Sweep.outputIndexNotFound` -> `views.Wif.outputIndexNotFound` (same `{{vout}}`/`{{txid}}` substitutions)
- `views.Tools.cashu.deleteData.subtitle` -> `deleteData.confirmMessage`
- `views.LSPS1.totalOrderValue` -> the previously orphaned `views.LSPS1.orderTotal` (LSPS1 and LSPS7 order response views)
- NostrRelays: removed a `Text` bound to a nonexistent key; it rendered nothing and duplicated the "Add relay" label directly below it

The remaining 10 strings were added to `en.json` only (per the Transifex rule): `general.amount`, `general.unknownError`, `general.clipboardError`, `components.QRSpeedMeterButton.speed`, `views.Invoice.invoiceAmount`, `stores.NostrWalletConnectStore.error.noPreimageReceived`, and four `stores.CashuStore` error strings (`errorPayingInvoice`, `errorDeletingData`, `invalidToken`, `sweepError`).

To lock this class of bug down, this PR adds `check-locales.test.ts` (modeled on `check-styles.test.ts`, runs in the normal Test CI job). It extracts every literal `localeString()` key from the codebase and fails with a key -> file listing if any is missing from `en.json`. Dynamic concatenation prefixes (e.g. `localeString('views.Channel.Total.' + kind)`) are skipped. Before the fixes it failed listing exactly the 17 keys above.

This pull request is categorized as a:

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [x] Quality assurance
[...]

PR #4680, opened 2026-09-18
tests: scan all literal keys in localeString calls

kaloudis:
# Description

Relates to issue: N/A

Follow-up to #4674, addressing @shubhamkmr04's post-merge review finding there: the locale coverage test only matched a string literal sitting immediately after `localeString(`, so it never saw shapes like `localeString(cond ? 'a' : 'b')`. 45 distinct keys across 82 call sites were unchecked; all are valid today, but a typo in any of them would have rendered as `undefined` without failing the test.

This adopts the extraction approach proposed in that review: a paren-balanced scan pulls every string literal out of each `localeString(...)` call span, and filtering on en.json's top-level namespaces keeps substitution values from being mistaken for keys. Interpolated template keys (backticks with `${...}`) are skipped alongside the existing dynamic-concatenation-prefix skip. One tightening over the proposal: the literal regex requires the closing quote to match the opening one, so an apostrophe inside a double-quoted string cannot produce a garbage match.

Verified by mutation: typo'ing the ternary key `views.Settings.SetPin.newPin` (views/Settings/SetPin.tsx) and the direct key `views.PaymentRequest.title` each fail the test with the correct key -> file listing; the clean tree passes, i.e. zero false positives across all 45 newly covered keys.

Round 1 (52f5e4d73) adopted both review fixes: the key argument (the leading literal of the call span) is now checked unconditionally, so first-segment typos in it stay covered as they were before this PR, and the paren-depth scan skips string literals, so a paren inside a string can no longer truncate or over-run a call span. The earlier claim here that both scan limits failed open was wrong: a `(` inside a string ran the span past the call and could report unrelated literals as missing keys.

Remaining accepted limits: a substitution value that is exactly a bare namespace word would only be flagged if it is not itself a key, and a first-segment typo inside a ternary branch (a non-leading position) is filtered out by the namespace check, a class no version of this test has covered. Neither occurs in the tree today.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

- [ ] Android
- [ ] iOS

N/A: CI-only test change, no runtime code touched.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


check-locales.test.ts
```diff
@@ -35,17 +35,50 @@ describe('locale keys', () => {
...
+            ) {
+                let j = i + call.length;
+                for (let depth = 1; j < source.length && depth > 0; j++) {
+                    if (source[j] === '(') depth++;
+                    else if (source[j] === ')') depth--;
+                }
+                for (const match of source
+                    .slice(i + call.length, j - 1)
+                    .matchAll(literalRegExp)) {
+                    if (namespaces.has(match[2].split('.')[0]))
```
shubhamkmr04, line 68: The namespace filter is load-bearing for keeping substitution values out, but it also drops any key whose first segment isn't already a namespace in en.json — and that class *was* covered before this PR.

Mutation on `views/AmountKeypad.tsx:233`, `localeString('general.amount')` -> `localeString('genral.amount')`:

- master (merged #4674): fails, reports `genral.amount`
- this branch: passes

So this trades one blind spot for a narrower one. The same applies to a key under a genuinely new top-level namespace added before its en.json entry.

Checking the key argument unconditionally restores it, since that position is exactly what the old regex matched:

```ts
const span = source.slice(i + call.length, j - 1);
// the key argument itself is checked unconditionally, so a
// typo in its first segment still fails
const lead = span.trimStart().match(/^(['"`])([^'"`\n]+)\1/);
if (lead) keys.push(lead[2]);
for (const match of span.matchAll(literalRegExp)) {
    if (namespaces.has(match[2].split('.')[0])) keys.push(match[2]);
}
```

Verified: clean tree passes, and the `genral.amount` and ternary `views.Settings.SetPin.newPinTYPO` mutations both fail.

Nice catch on requiring the closing quote to match the opener, by the way — that's more than cosmetic. With the unmatched form, an apostrophe earlier in a call span desynchronises the scan and can swallow the real keys after it.

check-locales.test.ts
```diff
@@ -35,17 +35,50 @@ describe('locale keys', () => {
...
+        const call = 'localeString(';
+        const keysIn = (source: string): string[] => {
+            const keys: string[] = [];
+            for (
+                let i = source.indexOf(call);
+                i !== -1;
+                i = source.indexOf(call, i + call.length)
+            ) {
+                let j = i + call.length;
+                for (let depth = 1; j < source.length && depth > 0; j++) {
```
shubhamkmr04, line 56: Small correction to the description's "both fail open (under-scan, never a false failure)": that holds for a `)` inside a string, but a `(` inside one runs the other way — depth never returns to 0 at the real closing paren, so the span over-runs into following code and literals there get reported as missing keys.

Constructed on `views/AmountKeypad.tsx`:

```ts
text: localeString('general.amount', { note: 'Zeus (beta' }),
translateKey: 'views.Totally.NotAKey',
```

fails with `views.Totally.NotAKey`, which is not inside any `localeString()` call.

Zero call spans in the tree hit either direction today (I diffed this walk against a string-aware one across every `localeString(` site — no disagreements), so it's not urgent. If you want it closed, skipping string literals here does it:

```ts
for (let depth = 1; j < source.length && depth > 0; j++) {
    const c = source[j];
    if (c === "'" || c === '"' || c === '`') {
        for (j++; j < source.length && source[j] !== c; j++)
            if (source[j] === '\\') j++;
    } else if (c === '(') depth++;
    else if (c === ')') depth--;
}
```

Clean tree still passes and the construction above no longer fails.

check-locales.test.ts
kaloudis: Good catch, this was a real regression against the merged test. Reproduced independently: `genral.amount` in views/AmountKeypad.tsx fails on master and passed on this branch. Adopted your fix verbatim in 52f5e4d73: the leading literal of the call span is checked unconditionally, so the position the old regex covered is covered again. Verified the full battery: `genral.amount` fails, the ternary `views.Settings.SetPin.newPinTYPO` fails, clean tree passes.

One residual worth naming: a first-segment typo inside a ternary branch (non-leading position) is still invisible, since it only flows through the namespace filter. That class was never covered by any version of the test, so I've listed it in the description's remaining limits rather than chasing it here.

check-locales.test.ts
kaloudis: You're right, and the description was wrong: the `(` direction fails closed, not open. Reproduced your construction (`'Zeus (beta'` plus a following non-key literal) and it reported `views.Totally.NotAKey` as missing. Took the string-aware depth scan in 52f5e4d73 even though no call span hits it today: it closes both directions, and I traced the escape handling through `'\\'` and `'\''` shapes to make sure the skipper can't desync on them. The construction no longer false-fails and the clean tree still passes. Corrected the fail-open claim in the PR description as well.

shubhamkmr04 (approved):
ACK 52f5e4d73

ajaysehwal (approved):
ACK

PR #4685, opened 2026-09-20
tests: cover StealthModeUtils, which had none

kunalKumar-13:
## What

`utils/StealthModeUtils.ts` had no test file. This adds 13 tests for it.

Stealth mode swaps the launcher alias so ZEUS presents as a calculator, VPN, QR scanner or notepad. Every method in the module is guarded twice — once by `isSupported()`, which requires both Android and a present native module, and once by a `try/catch` that returns `false` — and both guards exist for the same reason: **a caller that cannot hide has to be told so.** Returning `true` while the launcher icon still reads ZEUS would be worse than returning `false`.

That is the property the tests pin down.

## Covered

- **Platform gate** — true on Android with the native module; false on iOS even when the module is present; false on Android when it is missing.
- **Unsupported platform** — all four methods resolve `false` and the native module is not called at all.
- **Pass-through** — each of the five `StealthApp` values (`zeus`, `calculator`, `vpn`, `qrscanner`, `notepad`) reaches `enableStealthMode` unchanged.
- **`isStealthModeActive` reports what the native module reports**, rather than a fixed value — so a genuine `false` from the native side is distinguishable from the unsupported case, which is also `false`.
- **Rejecting native module** — every method resolves `false` rather than producing an unhandled rejection.

The module reads `Platform.OS` and captures `NativeModules.StealthMode` at module scope and exports a singleton, so each scenario loads it fresh inside `jest.isolateModules` with its own `react-native` stub. `console.warn`/`console.error` are spied and silenced so the guard paths do not fill the test output.

## Verification

```
PASS utils/StealthModeUtils.test.ts
Tests:       13 passed, 13 total
```

`prettier --check`, `tsc --noEmit` and `eslint` are all clean on the file.

I also checked the tests actually have teeth rather than just passing: changing the `enableStealthMode` catch to `return true` instead of `return false` fails **exactly two** tests — `every method fails closed rather than throwing` and `a failed enable does not leave the caller believing it is hidden` — and none of the other eleven. The source file is unchanged in this PR; that was a local mutation, reverted.

No production code is touched, so there is no UI behaviour to screenshot.

## Disclosure

This was written with AI assistance. Given CONTRIBUTING's note about AI-generated submissions that have not actually been tested, I have tried to make the testing the substance of the PR rather than a claim: the run output above is real, and the mutation check is there to show the assertions fail when the behaviour they describe is broken. Happy to adjust anything or drop it if you would rather this module were covered by hand.


kaloudis (changes requested):
Approach ACK, one ask before merge.

Verified on 669c97ca: 13/13 pass, `prettier --check`, `eslint`, `tsc --noEmit` clean. I re-ran your mutation check and added seven more; your claim holds (enable-catch-returns-true fails exactly two, no others), and these also die: dropping the `Platform.OS` check (2 fail), dropping the `StealthMode != null` check (1), and disable/fix/isActive catch returning true (1 each).

One mutant survives: make `isStealthModeActive` still call the native module but return a hardcoded `false`, and all 13 tests pass. That is the property the PR says it pins. The only test for it mocks the native answer as `false`, so a fixed `false` reads the same. The `true` direction is the load-bearing one: `StealthModeWrapper.checkStealthStatus` shows the decoy only when `nativeActive` is true, and `App.tsx:402` only calls `disableStealthMode` when it is true. So a `false`-when-actually-active answer drops the user into the real Zeus UI while the launcher still says Calculator, which is the exact failure this module exists to prevent.

Covering both directions kills it:

```ts
it.each([true, false])(
    'isStealthModeActive reports %s back from the native module',
    async (active) => {
        const native = makeNative();
        native.isStealthModeActive.mockResolvedValue(active);
        const utils = load('android', native);

        // The false case is distinct from the unsupported case, which is
        // also false: here the native module was consulted and answered.
        await expect(utils.isStealthModeActive()).resolves.toBe(active);
        expect(native.isStealthModeActive).toHaveBeenCalledTimes(1);
    }
);
```

Nit, optional: `let utils: any` in `load()` gives up type checking at every call site. `let utils: typeof import('./StealthModeUtils').default` keeps it.

The rest is well built. `jest.isolateModules` with a per-scenario `react-native` stub is the right call for a module that destructures `NativeModules` at import time, and it does not leak here since the only top-level import is `import type`. No production code touched, so no device testing owed.

kunalKumar-13:
Good catch, and thank you for actually re-running the mutation check and extending it — that is a more useful review than I had any right to expect.

You are right that the mutant survives and right about why it matters. My test only ever mocked the native answer as `false`, so a hardcoded `false` reads identically. And the `true` direction is the one with consequences: if `isStealthModeActive` lies in that direction, `checkStealthStatus` never shows the decoy and `App.tsx:402` never calls `disableStealthMode`, so the user lands in the real Zeus UI while the launcher icon still reads Calculator. That is precisely the failure the module exists to prevent, and my test was blind to it.

Taken your parametrised version, with a comment recording why the `true` case is the load-bearing one so it does not get simplified away later.

Verified your mutant is dead: making `isStealthModeActive` call the native module and return a hardcoded `false` now fails **exactly one** test — `isStealthModeActive reports true back from the native module` — and leaves the other 13 passing.

```
Tests: 14 passed, 14 total
```

Also took the nit. `load()`'s local is now

```ts
type StealthUtils = typeof import('./StealthModeUtils')['default'];
...
let utils!: StealthUtils;
```

so the call sites keep type checking. `prettier --check`, `eslint` and `tsc --noEmit` are clean.

kaloudis (commented):
Round 2 verified on 717f065d0. The ask is satisfied.

14/14 pass, `prettier --check`, `eslint` and `tsc --noEmit` clean. The surviving mutant is dead: making `isStealthModeActive` call the native module and then return a hardcoded `false` now fails exactly one test, `isStealthModeActive reports true back from the native module`, which matches your count. Hardcoded `true` dies as well.

I re-ran the round-1 battery plus twelve new mutants against the new suite, nineteen in all, and none survive:

- `isActive`: call native, return hardcoded `false`; and hardcoded `true`
- `isActive`: drop the `await`, so a native rejection escapes the catch
- `isActive`: never consult native
- `isSupported`: drop the `Platform.OS` check; drop the `StealthMode != null` check
- `enable` / `disable` / `isActive` / `fix`: catch returns `true`
- `enable`: ignore the `app` argument and always pass `'calculator'`
- `enable` / `disable` / `fix`: skip the native call and report success
- `enable` / `disable` / `isActive` / `fix`: unsupported guard returns `true`
- `disable`: drop the unsupported guard entirely

The nit turned out to be load-bearing rather than cosmetic. With the `StealthUtils` alias in place, tsc now rejects `utils.noSuchMethod()` (TS2339) and `enableStealthMode('nope')` (TS2345) at the `load()` call sites; under `any` both compiled silently. Good call keeping the comment about why the `true` case is the one that matters, that is exactly the thing that gets simplified away in a year.

Two mutants still survive and I am not asking for either, because both are equivalent under the code as written: wrapping the native answer in `Boolean()`, and `Platform.OS !== 'ios'` in place of `=== 'android'`, which cannot differ on a two-platform app.

One housekeeping item before merge: please squash the two commits into one. Zeus lands PRs as merge commits, so `tests: cover both directions of isStealthModeActive` would otherwise ship as a separate commit next to the original rather than folding into it. Nothing else wanted; I will flip my review once the squash is up.

kunalKumar-13:
Squashed to one commit, `edcef1946b3`. The tree is byte-identical to `717f065d0`, the revision you verified in round 2 — `git diff 717f065d0 edcef1946b3` is empty — so nothing you checked has moved; only the two commits became one.

kaloudis (approved):
Round 3: squash verified, approving.

`edcef1946b3` is one commit on `dd12ab334`, and the tree is byte-identical to `717f065d0`, the revision I verified in round 2: both resolve to tree `7fbb79dfcecef372804699d515ad9b5f71326812`, and `git diff 717f065d0 edcef1946b3` is empty. So nothing I checked in round 2 has moved, and the nineteen-mutant battery from that round still applies unchanged.

CI green on the squashed head: lint, prettier, tsc, and test all pass, with `PASS utils/StealthModeUtils.test.ts` inside 1811/1811. The commit message folded the round-2 rationale in rather than dropping it, which is the right outcome for a squash.

Nice piece of work. Test-only, so no device testing owed.

PR #4689, opened 2026-09-21
chore(deps): bump react-native-reanimated to 4.7.0

### LINKED ISSUE #4221: Remove reanimated packaging patch once react-native.config.js ships in the npm tarball

kaloudis:
# Description

Relates to issue: **[#4222](https://github.com/ZeusLN/zeus/issues/4222)**, closes **[#4221](https://github.com/ZeusLN/zeus/issues/4221)**

Bumps `react-native-reanimated` 4.6.0 -> 4.7.0, plus `react-native-worklets` 0.12.1 -> 0.13.0, which 4.7.0 requires (peer `0.13.x`). The `react-native` peer range is `0.86 - 0.88`, so this is compatible with the current 0.86.0 pin and with the 0.87.1 bump in #4566.

Supersedes #4549, which re-anchored the SET patches pre-emptively against a 4.7.0 nightly. 4.7.0 stable landed a later refactor than that nightly, so the re-anchoring is redone here against the actual release.

## What 4.7.0 fixes for us

**Packaging (#4221).** The tarball now ships `react-native.config.js` (upstream [#10375](https://github.com/software-mansion/react-native-reanimated/pull/10375)), so Android autolinking registers `REASharedTransitionBoundaryComponentDescriptor` on its own. `patches/patch-reanimated-boundary-android.mjs` is deleted.

**Cancelled-close wedge (iOS).** Upstream [#9945](https://github.com/software-mansion/react-native-reanimated/issues/9945) is fixed by [#10371](https://github.com/software-mansion/react-native-reanimated/pull/10371), which removed `closingScreenTag_` and `synchronized_` outright: an active gesture is now a `ProgressTransition`, and an uncommitted pop an `UncommittedScreenPop` bound to its own source screen, so a cancelled swipe on one screen can no longer block the pop of another. Verified in the 4.7.0 tarball that neither symbol exists. That edit is deleted from `patches/patch-reanimated-set-fixes.mjs`.

## What still needs patching

The Android pop race ([#9944](https://github.com/software-mansion/react-native-reanimated/issues/9944)) is still open and still unfixed in 4.7.0: the proxy remains a bare `findActiveBoundary(root)` with no fallback when the revealed screen's boundary has not received `isActive=true` yet. The pop-fallback edit stays.

4.7.0 renamed `LayoutAnimationsProxy_Experimental.cpp` to `LayoutAnimationsProxy.cpp` (split with `LayoutAnimationsProxyCommon.cpp` / `_Legacy`) and reshaped the surrounding code (`topScreen[surfaceId]` -> `topScreen_`), so the patch is re-anchored against the new file.

## Fail loudly

An anchor that silently stops matching reintroduces a SET failure with no error, no log, and nothing a test catches. The old patch warned and continued, and on 4.7.0 it would not even have reached the warning: the hardcoded `_Experimental.cpp` path no longer exists, so its `fs.existsSync` check returned early. The patch now fails the install in both cases, missing source file and unmatched anchor, and `patches/index.mjs` exits non-zero on a throw instead of printing "All patches applied successfully". `ZEUS_ALLOW_UNPATCHED_REANIMATED=1` installs anyway, with transitions broken.

## Notes

None of the other 4.7.0 breaking changes apply. The app uses no `useAnimatedRef`, `AnimatedRefOnUI`, `AnimatedRefOnJS`, or `USE_SYNCHRONIZABLE_FOR_MUTABLES`. The `ENABLE_SHARED_ELEMENT_TRANSITIONS` static flag still exists and the `reanimated` block in `package.json` is unchanged. `react-native-reanimated-carousel@4.0.3` peers on `>=3.0.0`.

`ios/Podfile.lock` carries only the RNReanimated and RNWorklets version and checksum lines. A full `pod install` here also picks up unrelated drift already on master (`react-native-get-random-values` 1.9.0 -> 2.0.0, a hermes-engine checksum, the CocoaPods version marker); that belongs in a separate lock refresh, not this bump.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [x] Other: dependency bump

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

Verified so far:

- `yarn install` applies the re-anchored patch to 4.7.0 on a clean tree, is idempotent on re-run, and fails the install with exit 1 both when the proxy source is missing and when the anchor does not match. The `ZEUS_ALLOW_UNPATCHED_REANIMATED=1` override warns and continues in both cases.
- `:react-native-reanimated:externalNativeBuildDebug` compiles the patched `LayoutAnimationsProxy.cpp` for arm64-v8a (NDK 28) with no errors.
- `yarn verify` passes: 1788 tests, tsc, prettier, lint.

Device testing of the shared element transitions themselves is still owed on both platforms. Contacts -> ContactDetails, push and pop, twice, plus the #9945 repro sequence that upstream documented: push, cancel a back swipe, push again, then `goBack()` from JS.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales


kaloudis:
Device testing checklist for this bump, for whoever picks it up (myself included).

**Before testing.** This is a native dependency change, so a full rebuild is required; a Metro restart is not enough. A stale native build against the old worklets will throw a JS/native version mismatch at startup, which is a setup failure and not a finding. On Android also clear the autolinking cache (`rm -rf android/build/generated/autolinking`, or `yarn android:clean`): it invalidates on hashes of the root `react-native.config.js`, `package.json` and `yarn.lock` only, so the entry we used to write inside `node_modules` can linger and mask whether 4.7.0's own config file is doing the work. Worth confirming the remaining patch landed before you start: `grep -c "ZEUS PATCH" node_modules/react-native-reanimated/Common/cpp/reanimated/LayoutAnimations/LayoutAnimationsProxy.cpp` should print 1.

**1. Android shared element transitions still run at all.** This is the direct risk of deleting `patch-reanimated-boundary-android.mjs`: if autolinking does not register `REASharedTransitionBoundaryComponentDescriptor`, the boundary gets flattened out of the shadow tree and transitions silently never run, with no error and no log. Settings -> Contacts -> tap a contact. The avatar and name should animate into place on ContactDetails, not cut. If they cut, the packaging fix is not reaching us and the patch has to come back.

**2. Return transitions on pop (Android).** Same screen pair, then back, both the header button and the system back gesture. Do it at least four times, entering a different contact each time. This is the failure the remaining patch exists for (upstream #9944, still open): the bug is intermittent, so a single clean pop proves nothing. Push is unaffected either way, only the return leg races.

**3. The #9945 sequence that 4.7.0 is supposed to fix.** Primarily iOS, worth repeating on Android. Upstream's repro, with the step our original report was missing:

1. open a contact, so a transition runs
2. start a back swipe on ContactDetails and release it early, so the close is cancelled
3. open a different contact, then leave with the in-app back button rather than a gesture
4. keep navigating

Every subsequent transition must keep animating. Before 4.7.0 this killed shared element transitions app-wide until restart, so wander the app for a minute afterwards and re-enter Contacts to confirm they are still alive. Repeat the cycle three or four times.

**4. Other boundary screens.** Nostr contact import (`NostrContacts`) and the channel detail screen (`views/Channels/Channel.tsx`) also render a boundary. Looking for the avatar vanishing mid-transition or flickering, which is the shape the flattening failure takes when it is partial.

**5. Wider reanimated surface.** 4.7.0 makes the new layout animations engine the default for everything, not just SET, and worklets goes 0.12 -> 0.13, so the gesture-driven components deserve a pass:

- swipeable rows on the main wallet screen (lightning, on-chain and ecash balances) and the swipeable order items: swipe open, swipe closed, then scroll the list and swipe again. Android had a stuck-rendering regression here before (f0e4f1c8f), so that is the one to watch for
- accordions in Receive, PaymentRequest, OpenChannel, OnChain Addresses and Wallet Configuration: expand and collapse
- the QR scanner overlay and the MultiQR carousel (BBQr/animated QRs), including rotating through frames

**6. Smoke.** Boot to wallet, no redbox, and no new warnings in the Metro or Xcode console mentioning reanimated or worklets.

Please note OS version and device or emulator for each platform when reporting back. The patch itself is already covered by automation: it applies to a clean 4.7.0 tree, is idempotent, fails the install both when the proxy source is missing and when the anchor does not match, and the patched C++ compiles for arm64-v8a on NDK 28. What none of that can tell us is whether the transitions actually animate, which is the whole point of the checks above.

ajaysehwal (approved):
tACK

PR #4694, opened 2026-09-21
fix(recovery): guard clearLegacyLocations behind __DEV__

kunalKumar-13:
# Description

Relates to issue: **N/A** — found while reading `utils/KeychainRecoveryUtils.ts`, which had no test file.

`copyToLegacyLocations` refuses to run outside a dev build, and the comment says why it is guarded in the util rather than only at the UI:

```ts
// Development seeding helper only. Guarded internally (not just at
// the UI) because it is the last remaining writer of unprefixed
// keychain entries: in a release build it could resurrect exactly
// the data the keychain cleanup purge deletes.
if (!__DEV__) { ... }
```

`clearLegacyLocations` is documented `DEV ONLY:` in the same style, is the more destructive of the two, and had no such guard. It deletes the unprefixed local keychain entry, the unprefixed iCloud entry and the `EncryptedStorage` copy of the wallet configurations — the exact entries `scanForRecoverableData` exists to find. For someone who has opened the recovery screen but not yet restored, those are the only remaining copy of their node configs.

**To be clear about severity: this is not reachable in a release build today.** Its only caller, `handleDevClearLegacy`, sits inside the `{__DEV__ && (` block at `views/Tools/NodeConfigExportImport.tsx:1608`. The argument for the guard is that the same was true of `copyToLegacyLocations` and it was guarded anyway, and that the two should not disagree about whether the UI gate is sufficient.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance
- [ ] Other

## What the tests cover

`utils/KeychainRecoveryUtils.ts` is 669 lines of recovery-critical code with no test file. 24 tests, covering the invariants that are easy to break without noticing:

- **The prefixed-cloud copy is suppressed when current storage holds the key.** After a completed desync migration the local copy is the fresher one, so offering the synchronizable copy beside it invites restoring stale data over live wallet configs. This is the whole safety property of that branch in `scanKey`.
- **The iCloud partitions are not read at all off iOS** — neither `getRawItem(..., true)` nor `getInternetCredentials(..., { cloudSync: true })`.
- **One throwing source does not abort the scan** — a locked keychain still lets the `EncryptedStorage` copy be found.
- **`otherDataFound` lists orphaned copies only**, never something current storage can already read.
- **`hasCurrentSettings` is not satisfied by the legacy key alone.**
- **`restoreSettings`** rejects malformed JSON, `null` and non-objects *without writing*, and reports failure both when the write is rejected and when the value cannot be read back afterwards.
- **`clearLegacyLocations` puts current settings back if they vanish** during the clear, and reports failure when that restore does not take.
- **`parseSettingsPreview`** falls back nickname → host → `Node N`, and returns `null` rather than throwing on garbage.

One note on the mocking: the module imports a dozen store modules purely for their key constants, and those pull native modules in behind them (`stores/SettingsStore` → `react-native-blob-util` → `new NativeEventEmitter()`), so the file cannot be required under Jest without stubbing them. The stubs supply only the constants.

## Checklist

```
PASS utils/KeychainRecoveryUtils.test.ts
Tests:       24 passed, 24 total
```

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Since you asked for mutants last time, here they are up front. Nine, none survive:

| mutation | tests failed |
|---|---:|
| drop the prefixed-cloud suppression in `scanKey` | 1 |
| `readFromPrefixedCloud` ignores `Platform.OS` | 2 |
| `readFromUnprefixedCloud` ignores `Platform.OS` | 2 |
| `otherDataFound` keeps `current` entries | 1 |
| `restoreSettings` skips the post-write verification | 1 |
| `restoreSettings` accepts any parsed JSON | 1 |
| `hasCurrentSettings` ignores the key | 1 |
| `clearLegacyLocations` skips the restore | 2 |
| remove the new `__DEV__` guard | 1 |

The last row is the one that matters for the fix: before the guard, that test failed with `resetInternetCredentials` and `EncryptedStorage.removeItem` both actually called with `__DEV__` false.

- [ ] Android
- [ ] iOS

**Neither** — I do not have a device or simulator for this, so the change is verified only by the unit tests above. The production change is a four-line early return in a function whose sole caller is already `__DEV__`-gated, so there should be no observable behaviour to exercise in a release build; if you would like it confirmed on-device before merging, that is a fair ask and I would rather you did it than take my word.


kunalKumar-13:
Correction: my `tsc` checkbox above was wrong. CI caught four `TS2345` errors in the
test file that my local run reported clean, and I should not have ticked it on the
strength of a run whose exit code I had not actually read.

The cause: `EncryptedStorage`'s methods are overloaded with promise and callback forms,
so `jest.Mocked<typeof EncryptedStorage>` collapses `mockResolvedValue`'s parameter to
`never`. Typing the stub as the three `jest.Mock`s it actually is fixes all four.

Pushed as `ea1ed0ce231`. Tests still 24/24, prettier and eslint clean, and `tsc --noEmit`
now exits 0 locally — but CI is the authority here, not me, so I will wait for it rather
than tick the box again.

kaloudis:
Squash the fixup commit please.

kunalKumar-13:
Squashed to one commit, `79e7ddd66d4`, tree byte-identical to `ea1ed0ce231`. I folded a note about the `EncryptedStorage` stub typing into the message, since the fixup it absorbed is no longer there to explain itself.

kaloudis (changes requested):
Concept ACK on the guard, and the reasoning in the description is right: the two helpers should not disagree about whether the UI gate is sufficient, and the clear side is the one with more at stake.

I verified the severity claim rather than taking it. `clearLegacyLocations` has exactly one caller, `handleDevClearLegacy`, and it sits inside the `{__DEV__ && (` block in `views/Tools/NodeConfigExportImport.tsx`, so this is unreachable in a release build today. Nothing observable changes, and I agree device testing is not owed for this one. CI is green; I reproduced 24/24, `tsc` clean, in a worktree at `79e7ddd`.

Thanks for putting the mutants up front. I ran all nine and they die exactly as your table says. I then ran fifteen more, and six survive. One of them undercuts a test in this PR, so I would like those addressed before merge.

## 1. The `copyToLegacyLocations` guard test is vacuous

Delete the pre-existing guard entirely and all 24 tests still pass:

```
M10 remove copyToLegacyLocations __DEV__ guard :: Tests: 24 passed, 24 total
```

`beforeEach` leaves `Storage.getItem` resolving `false`, so with the guard gone the body reaches `readFromCurrentStorage` first, finds nothing, and returns `{ success: false, locations: [], error: 'No current settings found to copy' }` without touching the keychain. Both of that test's assertions hold for a reason that has nothing to do with the guard.

This matters here more than it normally would, because the symmetry argument is the entire case for the PR. The new `clearLegacyLocations` guard test does have teeth (`M09` dies, as you said), so right now the two guards are asymmetric in the test file in the same way they were asymmetric in the source.

One line fixes it: seed current settings so the unguarded body would actually write.

## 2. Nothing asserts that either helper does its job

Four more survivors, all in the dev-build path:

```
M16 clearLegacyLocations: drop the unprefixed-local reset  :: 24 passed
M17 clearLegacyLocations: drop the iCloud reset            :: 24 passed
M18 clearLegacyLocations: drop the EncryptedStorage remove :: 24 passed
M20 copyToLegacyLocations: no-op all three writes          :: 24 passed
```

The guard tests are phrased as "did not call X". A negative assertion only carries weight next to a positive one showing X is called when it should be, otherwise it survives the function being gutted. `clearLegacyLocations`'s own `Platform.OS === 'ios'` branch is uncovered too; the iOS-only block covers the scanner, not this.

The "puts current settings back" test already runs with `__DEV__` true, so the clear side needs no new test, just assertions.

## 3. Nit: the nickname/host order is not pinned

```
M12 parseSettingsPreview: n.host || n.nickname :: 24 passed
```

No fixture node has both fields, so the fallback order the description claims to cover is not actually observable. Giving the first node a host fixes it.

## Patch

I wrote and verified all of the above. With it applied the suite is 25/25, `prettier` and `tsc` clean, and all six survivors die:

```
M10 :: 1 failed, 24 passed
M12 :: 1 failed, 24 passed
M16 :: 1 failed, 24 passed
M17 :: 1 failed, 24 passed
M18 :: 1 failed, 24 passed
M20 :: 1 failed, 24 passed
```

```diff
diff --git a/utils/KeychainRecoveryUtils.test.ts b/utils/KeychainRecoveryUtils.test.ts
index 5562ede80..511ada04f 100644
--- a/utils/KeychainRecoveryUtils.test.ts
+++ b/utils/KeychainRecoveryUtils.test.ts
@@ -84,7 +84,7 @@ const storage = Storage as jest.Mocked<typeof Storage>;
 const rawItem = getRawItem as jest.MockedFunction<typeof getRawItem>;
 
 const SETTINGS = JSON.stringify({
-    nodes: [{ nickname: 'alice' }, { host: '10.0.0.2' }, {}],
+    nodes: [{ nickname: 'alice', host: '10.0.0.1' }, { host: '10.0.0.2' }, {}],
     selectedNode: 1
 });
 
@@ -348,6 +348,8 @@ describe('KeychainRecoveryUtils', () => {
         };
 
         it('copyToLegacyLocations refuses to run outside a dev build', async () => {
+            storage.getItem.mockResolvedValue(SETTINGS as any);
+
             const result = await withDev(false, () =>
                 utils.copyToLegacyLocations()
             );
@@ -357,6 +359,37 @@ describe('KeychainRecoveryUtils', () => {
             expect(encrypted.setItem).not.toHaveBeenCalled();
         });
 
+        it('copyToLegacyLocations writes every legacy location in a dev build', async () => {
+            storage.getItem.mockResolvedValue(SETTINGS as any);
+            keychain.setInternetCredentials.mockResolvedValue(undefined as any);
+            encrypted.setItem.mockResolvedValue(undefined as any);
+
+            const result = await withDev(true, () =>
+                utils.copyToLegacyLocations()
+            );
+
+            expect(result.locations).toEqual([
+                'unprefixed-local',
+                'unprefixed-cloud',
+                'encrypted-storage'
+            ]);
+            expect(keychain.setInternetCredentials).toHaveBeenCalledWith(
+                STORAGE_KEY,
+                STORAGE_KEY,
+                SETTINGS
+            );
+            expect(keychain.setInternetCredentials).toHaveBeenCalledWith(
+                STORAGE_KEY,
+                STORAGE_KEY,
+                SETTINGS,
+                { cloudSync: true }
+            );
+            expect(encrypted.setItem).toHaveBeenCalledWith(
+                STORAGE_KEY,
+                SETTINGS
+            );
+        });
+
         it('clearLegacyLocations refuses to run outside a dev build', async () => {
             const result = await withDev(false, () =>
                 utils.clearLegacyLocations()
@@ -383,6 +416,19 @@ describe('KeychainRecoveryUtils', () => {
             );
 
             expect(storage.setItem).toHaveBeenCalledWith(STORAGE_KEY, SETTINGS);
+            expect(keychain.resetInternetCredentials).toHaveBeenCalledWith({
+                server: STORAGE_KEY
+            });
+            expect(keychain.resetInternetCredentials).toHaveBeenCalledWith({
+                server: STORAGE_KEY,
+                cloudSync: true
+            });
+            expect(encrypted.removeItem).toHaveBeenCalledWith(STORAGE_KEY);
+            expect(result.cleared).toEqual([
+                'unprefixed-local',
+                'unprefixed-cloud',
+                'encrypted-storage'
+            ]);
             expect(result.success).toBe(true);
         });
 
```

## One observation, not a change request

`restoreSettings` and `restoreDataKey` have no callers. The recovery UI restores through `saveNodeConfigs` in `handleRecoveryRestore`, and `handleRecoveryPreview` re-parses `result.data` itself rather than going through `parseSettingsPreview`. So six of the 24 tests pin down behaviour the app never executes. That is not something you introduced and I would not change it in this PR, but it is worth knowing that the coverage there buys less than it looks like it does, and it is something we should resolve on our side one way or the other.

kunalKumar-13:
You're right, and the `copyToLegacyLocations` one is a fair hit. I verified it rather than
taking it: removing that guard entirely leaves the suite at **24 passed, 24 total**.

The mechanism is exactly as you describe. `beforeEach` leaves `Storage.getItem` resolving
`false`, so the unguarded body reaches `readFromCurrentStorage`, finds nothing, and
returns `{ success: false, locations: [] }` at the "No current settings found to copy"
branch without ever reaching the keychain. Both my assertions held for a reason that had
nothing to do with the guard being there.

That is worse than an ordinary gap, for the reason you give: symmetry between the two
helpers is the whole argument for this PR, and the test file reproduced the asymmetry it
was meant to close. I had also written in the description that the mutants "all die",
which was true of the nine I ran and not of the property that actually mattered.

Applied your patch as written. All six survivors now die:

| mutant | result |
|---|---|
| `M10` remove the `copyToLegacyLocations` guard | 1 failed, 24 passed |
| `M12` `parseSettingsPreview`: `n.host \|\| n.nickname` | 1 failed, 24 passed |
| `M16` `clearLegacyLocations`: drop the unprefixed-local reset | 1 failed, 24 passed |
| `M17` `clearLegacyLocations`: drop the iCloud reset | 1 failed, 24 passed |
| `M18` `clearLegacyLocations`: drop the EncryptedStorage remove | 1 failed, 24 passed |
| `M20` `copyToLegacyLocations`: no-op the writes | 1 failed, 24 passed |

25/25 green, `prettier`, `eslint` and `tsc --noEmit` all clean. Production source is
untouched by this round — `git diff` against `79e7ddd` shows only the test file.

The general point lands beyond this file: a "did not call X" assertion is worth nothing
without a positive one proving X is called when it should be, because both survive the
function being gutted. I had that backwards in three places here. I'll write the positive
case first from now on.

kaloudis (approved):
Round 2. ACK, approving.

Re-ran it all rather than taking the table. `e4b0609`: 25/25 green, `tsc --noEmit`, `eslint` and `prettier --check` clean, and all four CI checks pass. The only delta from `79e7ddd` is the test file, +56/-1, production source untouched, as you said.

### The six survivors die

Sixteen mutants of my own rather than a replay of yours:

| mutant | result |
|---|---|
| remove the `copyToLegacyLocations` `__DEV__` guard | dies |
| remove the new `clearLegacyLocations` `__DEV__` guard | dies |
| `clearLegacyLocations`: local keychain reset never issued | dies |
| `clearLegacyLocations`: iCloud reset never issued | dies |
| `clearLegacyLocations`: `EncryptedStorage.removeItem` never issued | dies |
| `clearLegacyLocations`: `cleared[]` never records `unprefixed-local` | dies |
| `clearLegacyLocations`: skips the lost-settings restore | dies (2) |
| `copyToLegacyLocations`: local keychain write never issued | dies |
| `copyToLegacyLocations`: iCloud write never issued | dies |
| `copyToLegacyLocations`: `EncryptedStorage.setItem` never issued | dies |
| `parseSettingsPreview`: `n.host \|\| n.nickname` | dies |

The one that mattered is the first row: seeding `SETTINGS` before the negative call means the guard is now the only thing between that call and the keychain, so the test finally tests the guard and not the `No current settings found to copy` early return.

### Production change

Re-verified on this head. `__DEV__` is true in a dev build, so the guard is a no-op there; in a release build the function is unreachable, since `handleDevClearLegacy` and `handleDevCopyToLegacy` are bound only inside the `{__DEV__ && (` block at `views/Tools/NodeConfigExportImport.tsx:1608`, and nothing else in the tree calls either helper. There is no reachable configuration whose behaviour changes, so no device testing is owed and I am merging on CI plus the unit tests. Your offer to have it confirmed on-device was the right instinct, but there is nothing to observe.

### Three survivors left, all optional

None of these hold up the merge. Noting them so they are on the record, not as a third round:

1. `copyToLegacyLocations writes every legacy location in a dev build` seeds with a blanket `storage.getItem.mockResolvedValue(SETTINGS)`, so making the helper read `LEGACY_STORAGE_KEY` instead of `STORAGE_KEY` still passes 25/25. The rest of the file pins the key (`mockImplementation(async (key) => key === STORAGE_KEY ? SETTINGS : false)`); matching that here kills it.
2. `success: cleared.length > 0` to `success: true` survives in both helpers. There is no case where every write or delete throws, which is the only way that expression is false.
3. `Platform.OS === 'ios'` to `true` survives in the cloud branch of both dev helpers. You assert the off-iOS property both ways for the scan side; the write and delete sides only ever run at `ios`.

### Follow-up, not yours to fix here

`restoreSettings` and `restoreDataKey` have no callers anywhere in the tree. The recovery screen restores through `saveNodeConfigs` in `handleRecoveryRestore` (`NodeConfigExportImport.tsx:1017`). Six of your 25 tests therefore cover dead code. That is a pre-existing problem in the file, not something this PR introduced, and the tests are worth keeping either way. I will open an issue to decide between wiring them up and deleting them.

Last thing: the closing paragraph of your comment is the correct reading. A `not.toHaveBeenCalled()` assertion is a claim about a guard only if something else in the suite proves the guarded work happens when it should. Writing the positive case first is the right habit.

kaloudis:
Filed the follow-up: #4711.

It grew a little past the dead-code note in my review. The part worth building is `restoreDataKey` plus the `otherDataFound` half of the scan result: the scanner walks 20 non-settings keys across all five sources, works out which copies are orphaned, and then nothing reads it, so recoverable contacts and notes are found and discarded. `restoreSettings` I think is superseded rather than unwired, since the live path appends only the nodes the user selected instead of overwriting the whole blob, and the safer semantic is the one we want to keep.

Nothing there is yours to fix in this PR. Your six `restoreSettings` tests are noted in the issue either way: if we delete the function they go with it, and if we keep it the coverage is already sitting there.

PR #4695, opened 2026-09-22
tests(bolt11): cover every amount multiplier, including sub-satoshi

kunalKumar-13:
# Description

`Bolt11Utils.test.ts` asserted amounts for the **nano** multiplier and for no-amount
invoices. The other three multipliers BOLT11 defines were never exercised, and neither
was the case that matters most: an invoice asking for an amount that is **not a whole
number of satoshis**.

Relates to issue: **N/A** — found while reading the decoder.

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance
- [ ] Other

## What is covered

All from BOLT11's own [reference vectors](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#examples):

| vector | amount | note |
|---|---|---|
| `2500u` | 250,000 sat | vector was already in the file for signature recovery; its amount was never asserted |
| `20m` | 2,000,000 sat | milli was untested |
| `9678785340p` | 967,878,534 **msat** | 967878.534 sat — not a whole satoshi |

The pico one is the one worth having. The three amount fields disagree on purpose:

```ts
satoshis      === null          // refuses to round
num_satoshis  === '967878'      // floors, for display
num_msat      === '967878534'   // exact
```

`satoshis` is deliberately null rather than rounded so a caller cannot mistake a
truncated figure for an exact one. A wallet that reads the wrong one of those three
charges the wrong amount, and nothing was pinning which is which.

## The spec rule, and why it needed a built invoice

BOLT11 says that with multiplier `p` the least-significant digit of the amount MUST be
`0`, because a pico-bitcoin is a tenth of a millisatoshi and anything finer cannot be
paid. The spec publishes no vector for an amount it forbids, so that case is built with
`bech32.encode`:

```ts
const invoiceWithPrefix = (hrp: string) =>
    bech32.encode(hrp, new Array(110).fill(0), Infinity);
```

`lnbc1p` must give a null amount; `lnbc10p` must give `1` msat. **Both directions are
asserted deliberately** — a null on its own would be equally well explained by pico
amounts not being supported at all, so the test would have passed against a decoder that
simply ignored `p`.

The signature words are zeros. That is harmless here because recovery is lazy in this
decoder, so reading the amount fields never touches `destination` — which the existing
"lazy signature recovery" block in this same file already establishes.

## Checklist
- [x] I've run `yarn run tsc` — exits 0, no output
- [x] I've run `yarn run lint` — eslint exits 0 on the file
- [x] I've run `yarn run prettier` — clean
- [x] I've run `yarn run test`

```
PASS utils/Bolt11Utils.test.ts
Tests:       24 passed, 24 total
```

## Testing

- [ ] No, I'm a fool
- [x] Yes
- [ ] N/A

Mutants, since you ask for them. Five, none survive:

| mutation | tests failed |
|---|---:|
| `hrpToSat` rounds instead of refusing a non-whole satoshi amount | 1 |
| milli and micro divisors swapped | 2 |
| nano divisor wrong by a factor of ten | 2 |
| `num_satoshis` rounds instead of flooring | 1 |
| the pico last-digit rule dropped | 1 |

The last one is worth noting: it **survived** my first pass. I had the three spec vectors
in and every one of them is a *valid* amount, so removing the rule changed nothing. That
is what sent me to build the `1p` / `10p` pair rather than leave the gap.

I have tested this PR on the following platforms:

- [ ] Android
- [ ] iOS

Neither — test-only, no production code changes, so there is nothing on-device to
exercise.


kunalKumar-13:
The bot's finding on the `1p` test is worth taking seriously, and I think it is pointing
at the source rather than at the test. Verifying it turned up something a bit wider, so
raising it here rather than quietly changing the test.

**The test is deliberately characterising, not endorsing.** `decode` does not throw on an
invalid amount today — it catches and nulls:

```ts
if (amountValue) {
    try { satoshis = this.hrpToSat(...); } catch (e) { satoshis = null; }
    try { millisatoshis = this.hrpToMillisat(...); } catch (e) { millisatoshis = null; }
}
```

So asserting a throw would fail against current `master`. What I pinned is what the code
does.

**But the bot is right that the resulting object is indistinguishable from an amountless
invoice**, and it is not only the pico rule. Three different classes of invalid amount
collapse to the same state:

```console
lnbc1p                          -> satoshis=null msat=null num_sat='0'   # pico rule
lnbc22000000                    -> satoshis=null msat=null num_sat='0'   # 22,000,000 BTC
lnbc2100000000000000000000p     -> satoshis=null msat=null num_sat='0'   # overflow
lnbc10m                         -> satoshis=1000000 ...                  # valid, for contrast
```

The middle one is the one I would point at: an invoice demanding **more bitcoin than will
ever exist** decodes as amountless. And a genuine no-amount invoice produces exactly the
same three fields, so nothing downstream can tell them apart.

Downstream that resolves to a real zero. `ChoosePaymentMethod.decodeSatAmount` feeds the
decoded object to `new Invoice(...)`, and `Invoice.getRequestAmount` ends at:

```ts
return Number(this.num_satoshis || 0);
```

so the user is offered the "enter an amount" path for an invoice that should have been
rejected. The payment would fail at the node afterwards, so this is a correctness and UX
problem rather than a way to lose funds — but "malformed" and "you choose the amount" are
very different things to show someone.

**I have deliberately not changed `decode` in this PR.** Making it throw is a behaviour
change across every call site — `CashuStore`, `LightningAddressStore`, `SwapDetails`,
`ChoosePaymentMethod` — and several of those wrap the call in `try/catch` already, so the
blast radius needs your judgement, not mine. Three options as I see them:

1. Leave it, and treat the null-amount collapse as known.
2. Distinguish the states without throwing — e.g. an `amountInvalid` flag on the decoded
   object, so callers can refuse without every existing `try/catch` changing meaning.
3. Throw, and fix the call sites.

Happy to open an issue with the above so it does not live in a test PR's comments, or to
implement whichever you prefer as a separate change. This PR stays test-only either way.

One other thing the bot flagged that I can close out: it could not verify the `bech32`
import because the manifest was not in its context. It is a declared direct dependency,
`"bech32": "2.0.0"` in `dependencies`, and `Bolt11Utils.ts` itself already imports the
same named export the test uses.

kunalKumar-13:
Added a test for the supply-cap case from the comment above, since it is the clearest instance of the collapse and I would rather it be visible in the file than only in a thread. `lnbc22000000` — 22 million BTC — now has `satoshis` null and `num_satoshis` `'0'` pinned, with a comment saying it characterises current behaviour and is meant to fail loudly if `decode` is later changed to reject.

That also kills a sixth mutant: dropping the `MAX_MILLISATS` bound entirely used to pass. 25/25, prettier/eslint/tsc clean, still test-only, still one commit.

PR #4696, opened 2026-09-22
fix(cashu): don't report unpaid melts as success

kaloudis:
# Description

CDK reports a melt the mint has not settled by resolving with a non-Paid lifecycle state rather than throwing. `FinalizedMelt.state` is a `QuoteState` (`Unpaid | Paid | Pending | Issued`) and `preimage` is optional, so a mint that accepts the payment asynchronously (NUT-05), or is still settling the Lightning leg, comes back `Pending` with no preimage. Both bridges pass that straight through (`CashuDevKitModule.kt` `encodeMelted`, and the Swift equivalent).

The single-mint pay path never looked at it. `payLnInvoiceFromEcash` set `paymentSuccess = true` immediately after `CashuDevKit.melt()` resolved, so the user saw **payment successful** and the payment was recorded in history with an empty preimage, for an invoice nobody had paid. No hostile mint is required: a slow Lightning payment is enough. Someone acting on that confirmation (telling a merchant the invoice is paid) has been deceived. The multimint path in the same file already refused anything but `Paid`, which is what made the omission visible.

This guards the three paths that did not check:

| Path | Was |
|---|---|
| `payLnInvoiceFromEcash` single-mint | set `paymentSuccess` on any resolved melt |
| `meltCDK` | returned the result to the `claimToken` self-custody sweep, which reported `{success: true}` regardless of state |
| `sweepMint` | reported a successful sweep for sats still sitting at the mint |

The guard goes **inside `meltCDK`** rather than at its call site, so it covers the claimToken path without touching the self-custody sweep block being reworked in #4488. That PR inherits the guard when it rebases.

The rule lives in one place, `isMeltPaid` / `normalizeMeltState` in `utils/CashuUtils.ts`, where it is unit testable. The multimint path switches to the helper. That changes one thing there: the old inline logic treated an **absent** state as `PAID`, and an absent state is now not paid. Both bridges always emit a state (`CDKMelted.state` is required, `encodeMelted` always writes it), so this does not reject melts any real build produces, and a result with no state at all is exactly the ambiguity that must not be reported as success. It is no longer dropped either: see reconciliation below, where an unknown state is treated as in flight and the melt quote is re-checked to find out.

## Refusing to report it was only half of it

Reviewer feedback on the first version (thanks @ajaysehwal) was that the guard leaves a reconciliation gap, and that is right. A melt that resolves `Pending` is real: the mint has the ecash and may pay minutes later. CDK keeps the proofs reserved against the quote, so nothing is lost, but nothing would ever look at that quote again. The payment would either complete with no record, the balance simply dropping, or sit unresolved forever. The auto-sweep path is the easiest to miss, since it swallows the error.

So every melt is now recorded before it is executed, which also covers a crash or kill mid-melt, and cleared as soon as the mint settles it. What survives is only what is genuinely in flight. `checkPendingItems` re-checks each one with `CashuDevKit.checkMeltQuote`, which both bridges already implement and nothing called:

| Mint says | Then |
|---|---|
| Paid | recorded as the payment it turned out to be, with the preimage from the quote |
| Unpaid (or another settled-unpaid state) | dropped, along with the in-flight entry |
| Pending, or anything unrecognized | left for the next sweep |
| unreachable | kept: the mint being down says nothing about the payment |
| unreachable and past the quote expiry | dropped, since it can no longer settle |

The pay path also records the in-flight payment itself, with no preimage and no fee, so an invoice that may still be paid is visible in activity rather than vanishing, and the user is told it is in flight instead of that it failed. `CashuPayment` already had `isInTransit`, and the activity list already renders it. Sweeps and self-custody claims pay an invoice from the user's own node, which records the incoming payment itself, so those reconcile the balance without inventing an ecash payment entry.

This is the shape both comparable wallets settled on. Neither reports a non-paid melt as success: cashu.me throws `Invoice not paid.` on anything but `PAID` and Minibits refuses too. Both then keep the quote and re-check it, which is the part this adds. Cashu.me writes a pending history entry *before* the melt and polls outgoing quotes from a worker; Minibits keeps the transaction `PENDING` and resolves it over a mint websocket with a poller behind it. Zeus needs less of that machinery than either, because CDK owns the proof state.

**Multimint is not covered.** `meltPartial` creates its MPP quote inside the native module, so there is no quote id on the JS side to record before the melt. Returning it is a bridge change, and a follow-up.

## Two smaller fixes caught in review

- **Balances went stale on a failed payment.** The single-mint pay path only synced on success, so a melt the mint did not settle left the pre-attempt balance on screen even though the proofs it moved are reserved against the quote. `meltCDK` and `sweepMint` already sync before their guard; the pay path now does too, best effort so a failed sync cannot replace the payment error.
- **A claimed token was reported as a failed claim.** The self-custody claim receives the token *first* and only then sweeps it to the user's node, so a melt that does not pay leaves the ecash claimed, the proofs spent, and the token unclaimable again. It now returns success carrying a warning that the sweep did not complete, which is the channel the offline-queue claim already uses; the self-custody button was reading only `success` and `errorMessage`, so it would have dropped the warning.

Related: #4493 (draft) makes recorded transactions Pending-aware on the history side but never touches the pay path, so it is complementary, not a fix for this. If it lands a shared normalizer, these helpers should collapse into it.

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [X] Yes
- [ ] N/A

`utils/CashuUtils.test.ts` covers the helpers, including that an unknown or absent state reads as in flight rather than failed. `stores/CashuStore.test.ts` covers the finding's site and everything above it: a `Pending` melt surfaces an error, records nothing as paid, and leaves an in-flight payment and a tracked melt behind, while a `Paid` melt still succeeds with its preimage and leaves nothing tracked; a melt the mint settled unpaid leaves nothing at all; a melt whose execution never returned stays tracked. Reconciliation is covered for each outcome, including an unreachable mint, an expired quote, a sweep (no payment entry), and a mint the user has since removed. The two pay-path tests were verified to fail against the unguarded and untracked code. `utils/DataClearUtils.test.ts` already enforces that every `-cashu-` key the store writes is wiped, and caught the new one.

- [ ] Android
- [ ] iOS

Device testing still owed: an ecash payment against a mint that returns a non-instant melt, confirming the in-flight payment appears in activity and then settles or disappears once the mint makes up its mind, and a self-custody claim whose sweep does not complete. Not yet run, flagging rather than ticking.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Backend-independent: the Cashu pay path runs against the mint via CDK, not the node backend. The sweep and claimToken paths do create an invoice on the active backend.

### Locales

Two new strings: `stores.CashuStore.paymentInFlight`, for a payment the mint has not settled, and `stores.CashuStore.claimedButNotSwept`, for a token claimed whose sweep did not complete. Everything else reuses `stores.CashuStore.errorPayingInvoice`, as the multimint path does.

### Other:


stores/CashuStore.ts
```diff
@@ -1052,6 +1054,15 @@ export default class CashuStore {
         const result = await CashuDevKit.melt(mintUrl, quote.id);
         await this.syncCDKBalances();
 
+        // A resolved melt is not a paid one. Callers treat a returned result
```
ajaysehwal, line 1073: @kaloudis, I think there’s a reconciliation gap in the new guards: `meltCDK`, `payLnInvoiceFromEcash`, and `sweepMint` don’t persist a melt that resolves as `Pending` / `Unpaid` / `Issued` before throwing.

A melt can legitimately be pending while the mint is still settling the Lightning leg. In that case, Zeus surfaces the error but doesn’t retain any record to reconcile later. If the mint completes the payment afterward, Zeus has no way to discover and record that outcome, unlike the pending-payment tracking we now have for Lightning sends.

The auto-sweep path is especially easy to miss because it catches the error and intentionally fails silently, even though the mint balance may eventually decrease.

I think we need some pending-melt persistence/reconciliation here to avoid losing track of a payment that may still settle.

PR #4697, opened 2026-09-22
fix(security): purge node-config staging plaintext on wipe

kaloudis:
# Description

The node-config export and import flows have to stage plaintext in app-private cache, because the native crypto API is file-based and must be handed a file. The export writes the full node list, **seed phrases included**, to `zeus-nodeconfig-plain.tmp` before encrypting it; the import decrypts to `zeus-nodeconfig-import-plain.tmp` before parsing. Both unlink their own temporaries in a `finally`, so a remnant only survives a run that never reached it: a crash, an OS kill, or a force-stop in that window.

Nothing then removed it. `clearAllData()` purges the analogous rescue-key staging (added in #4373) but had **no step** for these files, and no sweep existed at launch, so the plaintext survived a full wipe, **including the duress wipe**, whose entire purpose is destroying seed material under coercion. `views/Lockscreen.tsx` routes both the duress and repeated-failed-login wipes through `clearAllData()`. The wipe logged "All data cleared successfully" while every seed phrase stayed readable to anyone with later local or forensic access to the device (root/jailbreak, backup extraction, forensics).

The export hardening series secured every other stage of this flow (AES-256-GCM in #4378, the save dialog, the password floor, iOS file-sharing in #4383), which is what makes the wipe gap stand out as the residual.

Fix:

- purge the staging files in `clearAllData()`, as its **first** step, before the settings read and the node teardown
- sweep at launch too, so a remnant does not sit in cache waiting for the next export to clear it
- also remove the encrypted envelope a finished export stages under its user-facing name (`*.zeus-wallet-config-backup`)

The paths and the sweep live in a new `utils/NodeConfigStagingUtils.ts` rather than in `NodeConfigUtils.ts`, which imports `SettingsStore`: the launch sweep runs from `SettingsStore`, so importing `NodeConfigUtils` there would close a cycle and pull the document-picker native module into every test that touches `SettingsStore` (this is not hypothetical, it broke `stores/SettingsStore.test.ts` on the first attempt). #4630 splits `ChannelExportStagingUtils` out for the same reason. Both flows now read their staging paths from that module, so a renamed temporary cannot silently fall out of the sweep.

Both call sites run the sweep **first**, and in both cases that is the point rather than a detail:

- In `clearAllData()` it used to sit behind the per-node keychain clear, the CDK database and the node shutdown. Those take seconds on a device (iOS pays a seed-derived node id per embedded wallet; each node shutdown retries up to three times at 500ms), and every one of those seconds was a window in which a force-stop left `zeus-nodeconfig-plain.tmp` on disk. That is the residual this change exists to close, and the duress wipe is exactly where a hostile interruption should be expected. Thanks @shubhamkmr04 for catching it. A test pins the ordering, since the cost of it drifting back is invisible.
- At launch it runs before the keychain migrations rather than after them, so a migration that rejects cannot skip it and leave the plaintext for that boot and every boot after while the failure persists.

The launch sweep is deliberately **not** MOD_KEY gated, unlike the one-off legacy purges beside it: a remnant can appear after any killed export, so it has to run every boot rather than once per install. It is latched once per process instead, because `getSettings()` also runs when Android backgrounds for the save dialog, and sweeping then would delete an active export before `saveDocuments` copies it. `clearAllData` calls the utility directly and always purges. The sweep itself cannot reject (each unlink and the cache listing swallow their own failures), so putting it ahead of the migrations cannot break settings loading.

Storage note, since this touches `stores/SettingsStore.ts` and `utils/MigrationUtils.ts`: no settings-blob shape change, no keychain access, no new persisted key, no migration. The only addition to the boot sequence is a file unlink, in the same idiom as the existing `purgeRescueKeyFiles()` call, ahead of it rather than beside it.

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [X] Yes
- [ ] N/A

`utils/NodeConfigStagingUtils.test.ts` covers the sweep against an in-memory FS: every staged export/import path removed, staged envelopes swept, unrelated cache files untouched, and the known paths still removed when the cache listing fails or a single unlink throws. `utils/DataClearUtils.test.ts` and `stores/SettingsStore.test.ts` cover the two call sites, including that the wipe sweeps before the slow node teardown and that a failing migration cannot skip the launch sweep; the `clearAllData` ones fail against the unpurged wipe (verified by disabling the call) and the ordering one fails against the old position (verified by moving the call back).

- [ ] Android
- [ ] iOS

Device testing still owed, and it is the part worth doing carefully here: kill the app mid-export (and mid-import), confirm the tmp is present, then confirm a wipe and a relaunch each remove it, on both platforms. Not yet run, flagging rather than ticking.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Backend-independent: this is local file hygiene in the export/import and wipe paths and never talks to a node.

### Locales

No user-facing strings: the sweep is silent.

### Other:


utils/DataClearUtils.ts
```diff
@@ -734,6 +735,13 @@ export async function clearAllData(): Promise<void> {
     await purgeLegacyRescueKeyFiles();
     await unlinkRescueKeyStagingFile();
 
+    // 2e. Delete anything the node-config export/import flows left staged in
+    // cache. A kill between staging the plaintext node list and unlinking it
+    // leaves every seed phrase sitting there in the clear, and nothing below
+    // touches app cache. A duress wipe that spares seed material is not a
+    // wipe.
+    await purgeNodeConfigStagingFiles();
```
shubhamkmr04: This lands after the per-node xprv re-derivation (step 2) and the node shutdown retries (step 2c, up to 3x500ms per node) — seconds in which a force-stop still leaves `zeus-nodeconfig-plain.tmp` on disk, which is the residual this fix is closing.

`purgeNodeConfigStagingFiles()` needs no settings, node state or keychain, just a few `exists`/`unlink` calls, so it could run right after `Storage.blockWrites()` above, ahead of everything slow.

shubhamkmr04 (commented):
One sequencing note below; the rest looks good.

PR #4699, opened 2026-09-22
fix(pos): restore the header on the POS connection error screen

kaloudis:
# Description

When POS mode is active and the app cannot reach the node, the top of the error message is rendered underneath the status bar and is cut off.

`views/POS/Layout.tsx` early-returns `renderErrorScreen()` before it renders `WalletHeader`:

```js
const error = NodeInfoStore?.error || SettingsStore?.error;

if (error) {
    return this.renderErrorScreen();
}
```

`WalletHeader` is the only thing on that screen that supplies the top safe-area inset. The root `SafeAreaView` in `App.tsx` is `edges={['left', 'right', 'bottom']}`, deliberately excluding `top`, and the wallet tab navigator sets `safeAreaInsets: { top: 0 }` with the comment "WalletHeader handles it". `components/Header.tsx` applies `marginTop: insets.top`. Drop the header and nothing pads the top.

The error text started at `paddingTop: 20` plus `marginTop: 20`, so 40pt from y=0. On notch and Dynamic Island iPhones `insets.top` is roughly 47 to 62pt, so the first line of a 20pt font sits under the status bar. On older iPhones and most Android devices the status bar is around 24dp and it looks fine, which is why it is device dependent.

This has been wrong since before the Layout extraction in 564d1c7c5; the pre-refactor `StandalonePosPane` had the same early return. The non-POS path already does it correctly: `BalancePane` renders `WalletHeader` inside its error pane.

## Changes

1. Render `WalletHeader` in the error branch, matching both the non-error branch of `Layout` and the `BalancePane` error pane.

   The `title` prop is load-bearing and is commented as such in the code. With `title` set, `centerComponent` renders a plain `<Body>`. Without it, it falls through to the `displayNickname` branch, whose nickname `Text` has `onPress={() => navigation.navigate('Wallets')}`, an ungated direct navigate to the wallet switcher that bypasses the POS lockscreen gate, along with status badges that navigate to `NodeInfo` and `Networking`.

2. Move the message, button and version string into a `ScrollView`, so a long error (cert and TLS failures, lnd error dumps) no longer overflows past the button and off the bottom of the screen.

3. Drop a `marginBottom: -40` negative margin that is no longer needed.

## Staff access controls

The error screen already exposed the menu through `protectedNavigation(navigation, 'Menu')`, and `MenuBadge` in the header uses the identical call, so the gated path is unchanged. `protectedNavigation` routes through `Lockscreen` whenever `posStatus === 'active'` and a PIN or passphrase is set, and `Lockscreen` only auto-proceeds without auth when there is no `pendingNavigation`, which `protectedNavigation` always supplies.

With `title` passed, the resulting control surface on the error screen is identical to the one staff already have on the normal POS screen: `MenuBadge` and `TempleButton` gated, `ActivityButton` open by design.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [ ] Yes
- [x] N/A

**Not yet tested on device.** This is a layout fix whose whole point is the safe-area inset, so it needs a run on a notched or Dynamic Island iPhone with POS active and a node that fails to connect, plus an Android check for no regression. Marking the platform boxes unchecked until that is done.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:

No locale strings added; the error screen reuses `views.Wallet.MainPane.error` and `views.Wallet.MainPane.goToSettings`.


views/POS/Layout.tsx
```diff
@@ -36,48 +42,62 @@ interface LayoutProps {
...
+                        icon={{ name: 'settings', size: 25, color: '#fff' }}
+                        title={localeString(
+                            'views.Wallet.MainPane.goToSettings'
+                        )}
+                        buttonStyle={{ backgroundColor: 'gray' }}
+                        containerStyle={{ alignItems: 'center' }}
+                        onPress={() => protectedNavigation(navigation, 'Menu')}
+                        adaptiveWidth
+                    />
+                    <Text
```
ajaysehwal, line 90: @kaloudis, I think the app version looks better in center?

Image

kaloudis: How did you generate an error message like this?

ajaysehwal: Oh, this is just a random error message I added to the code to test the scroll behavior. But the main thing is that the app version is currently aligned to the left, and I think it looks better when centered.

PR #4706, opened 2026-09-22
fix(pos): store per-item order tax in cents, add PosStore tests

shubhamkmr04:
# Description

`PosStore.recalculateCurrentOrder` writes `total_tax_money.amount` in two different units depending on which branch it takes:

- the global-rate branch writes **cents** — `totalFiat.div(100) × pct`, where `totalFiat` is already accumulated in cents
- the per-item-rate branch writes **dollars** — it normalises each line item down to major units first (`.div(100)` on the sats path, raw `.amount` on the fiat path) before applying the rate

Every reader assumes cents: `Order.getTaxMoney` divides by 100, as do the three call sites in `views/Order.tsx`. So a $10 order at 8% stores `0.8` instead of `80`.

**This is not user-visible today.** In the per-item case nothing reads the field — `getTaxMoney` is only consumed by `getOrdersHistorical`, which is Square-only; two of the `views/Order.tsx` reads are behind `posEnabled === Square`; and the third is behind `!hasIndividualTaxRates`, the exact negation of the branch that writes the bad value. Displayed and charged tax both come from `calculateTaxSats`, recomputed from `line_items` in sats. Verified on device: standalone POS, $10 product at 8% with global tax 0, shows $0.80 and invoices correctly via both checkout and Quick Pay.

So this is a correctness cleanup on a stored value, not a behaviour change. Reviewers should expect **no observable difference** in the app.

## Changes

- per-item branch now computes in cents like the global branch. The sats path drops its `.div(100)`, since `calcFiatAmountFromSats` already returns cents; the fiat path gains `.times(100)`. Locals renamed `…Fiat` → `…Cents` — that ambiguity is what produced the bug.
- the outer `if` had no `else`, so `total_tax_money.amount` was never reset to
  zero once no rate applied. Added. This is reachable today: set a global tax
  rate, add items, then clear `taxPercentage` in POS settings and add another
  item — `recalculateCurrentOrder` now takes the `else` and zeroes the stale
  value. The `resets the tax when the global rate is cleared` test covers it.
- first unit tests for `PosStore`: 13 cases covering both tax branches, mixed
  per-item rates, global-rate fallback, sats-priced items, baskets mixing
  sats-priced and fiat-priced items, quantity scaling, the reset, and
  `processCheckout` quick-pay totals.

Note on existing data: `saveStandaloneOrder` persists the whole order, so rows
already in `zeus-pos-standalone` keep the old major-unit value while new ones
are written in cents. No migration is included, since nothing reads the field
for standalone orders today — but a future reader would need to account for the
mixed units, or backfill first.

Without the fix, 8 of the 13 fail. The 3 `processCheckout` tests pass either way, which is the point — quick pay never reads this field.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- [ ] Android
- [X] iOS

On-device
- [ ] LDK Node
- [X] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


kaloudis (commented):
tACK `68a2b2a`, with a few notes below (none blocking).

I checked this out and verified the unit analysis independently rather than taking the description at face value:

- `calcFiatAmountFromSats` returns cents (`PosStore.ts:291`, `.multipliedBy(100)`), `totalFiat` accumulates in cents (`:337`, `:343`), and the global branch's `div(100).multipliedBy(pct)` does land on cents. So the per-item branch really was the odd one out.
- `base_price_money.amount` is major units in standalone (`StandalonePosPane.tsx:196-226` pushes `product.price` raw), which is what makes the new `.times(100)` correct.
- The "not user-visible" claim holds. The only readers are `getOrdersHistorical` (Square-only, hits the Square API at `PosStore.ts:546`), `views/Order.tsx:169` and `:607` (both gated on `posEnabled === Square`), and `views/Order.tsx:197` (gated on `!hasIndividualTaxRatesForReceipt`, the exact negation of the branch being fixed). I also checked that the `hasIndividualTaxRates` predicates in `PosStore` and `PosUtils.calculateTaxSats` select the same branch for every input, so the stored value and the charged value never disagree about which rule applies.

### Verification

| Check | Result |
| --- | --- |
| `yarn tsc` / `yarn prettier` / `yarn lint` | pass |
| `yarn test` (full suite) | 1823/1824; the one failure is the known `PhotoUtils` "groks out `preset://` values" path assertion, which fails in any non-default checkout directory. Not a regression. |
| New tests at PR head | 13/13 pass |
| New tests with `stores/PosStore.ts` reverted to master | 8 fail, 5 pass, exactly as described |

The 8/5 split reproduces precisely, so the tests genuinely pin the fix rather than just documenting current behaviour.

### Notes

**1. The description undersells the `else`.** It says the reset is "not reachable through the current UI (the POS has no item-removal path)". There is a second route that does not need item removal: set a global tax rate, add items, then clear `taxPercentage` in POS settings, come back and add another item. `recalculateCurrentOrder` takes the new `else` and zeroes the stale value. That is exactly what the `resets the tax when the global rate is cleared` test exercises. Worth correcting in the body, since it makes the `else` a live fix rather than defensive padding.

**2. Orders already on disk keep the old value.** `saveStandaloneOrder` pushes the whole `currentOrder` (including `total_tax_money`) into `zeus-pos-standalone`, so orders written before this lands are in dollars and ones written after are in cents, with no migration. Nothing reads the field for standalone today so nothing breaks, but a future reader would see mixed units across historical rows. One sentence in the PR body would close it out.

**3. Adjacent latent bug, follow-up rather than blocker.** `PosStore.ts:419-427`: on the update path, `saveStandaloneOrder` copies `line_items`, `updated_at` and `total_money` but never `total_tax_money`. It is dead today (`createCurrentOrder` always mints a fresh uuid and `currentOrder` is never repopulated from `openOrders`), but it is the other half of "the stored tax value is unreliable", and it would resurrect the same class of bug the moment an order gets re-saved.

### Nits

- `PosStore.ts:389`: the bare `.times(100)` is the exact ambiguity this PR exists to remove. A short comment along the lines of "line item prices are major units, `total_*` are cents" would save the next reader re-deriving it, especially since Square line items are in cents (see the `.div(100)` at `views/Order.tsx:225-227`).
- The new `else` resets `amount` but not `currency`. Harmless at zero, just asymmetric with the `if`.
- `PosStore.test.ts`: the module-scope `fiatStore` stub and the one inside `jest.mock('./Stores', ...)` are different objects sharing a name. Renaming the injected one to `fiatStoreStub` would read better. Its `formatAmountForDisplay` is never reached through `PosStore`, only via the mocked `Stores` in `models/Order`, so it is dead on the injected copy.
- `agrees between the global and per-item branches` restates the two assertions directly above it.
- No per-item test mixes a sats-priced and a fiat-priced item in one order, which is the case where the two normalisation paths would most plausibly diverge.

### Conventions

Mocking style matches the existing store tests (`UTXOsStore.test.ts`, `InvoicesStore.test.ts`): top-of-file `jest.mock` calls ahead of imports, `__esModule: true` defaults. Colocated `stores/PosStore.test.ts` matches the repo layout. No dependency, locale or settings changes.

### Out of scope, spotted while reading

`PosStore.ts:649-652` tallies `Number(order.getTotalMoney) + Number(order.getTaxMoney)` for the Square recon screen, but Square's `total_money.amount` already includes tax (which is why `views/Order.tsx:169` subtracts it to get the subtotal). That recon total looks double-counted. Separate PR if it is real.

Branch is behind master (based on `951539ae3`) but merges clean, and no other open PR touches this code.

shubhamkmr04:
Addressed in `1b48f0f8`, rebased onto `ad10ff62`. 

**Notes**

1. Corrected the body. The `else` is reachable without item removal: set a global rate, add items, clear `taxPercentage` in POS settings, add another item. That is what `resets the tax when the global rate is cleared` exercises, so it is a live fix rather than padding.
2. Added a note that rows already in `zeus-pos-standalone` keep major units while new ones are written in cents, with no migration, since nothing reads the field for standalone today.
3. Agreed, leaving it alone here.

**Nits**

- Added the units comment at the `.times(100)`.
- Renamed the injected stub to `fiatStoreStub` and dropped its `formatAmountForDisplay`, which was only ever reached through the mocked `Stores` in `models/Order`.
- Dropped `agrees between the global and per-item branches`.
- Added `normalises sats-priced and fiat-priced items alike`, covering a basket with one of each.
- Left the `currency` asymmetry. It only matters at zero, and setting it in the `else` would mean repeating the `fiatRates` lookup from the `if`.

**On the recon tally**  it is real. `views/Order.tsx:169` subtracts tax from `total_money.amount` to derive the Square subtotal, which confirms the total already includes it, so `Number(getTotalMoney) + Number(getTaxMoney)` at `PosStore.ts:649-652` double counts. Happy to take that and the `saveStandaloneOrder` gap separately if you want them.

Issue #4687, opened 2026-09-21
getEcashRowColors throws on the default theme: buttonGradient is an array, blendHexColors expects a string


kunalKumar-13:
### Summary

`getEcashRowColors` throws on the **default theme**, so the balance-based warning tint on the ecash row never appears. The failure is swallowed by a `try/catch`, so the row still renders — just always in the plain gradient, with a `console.error` on every call.

### Chain

`DEFAULT_THEME` is `kyriaki` (`stores/SettingsStore.ts:1271`), and Kyriaki is the only theme that defines `buttonGradient` (`utils/ThemeUtils.ts:26`):

```ts
buttonGradient: ['#FF9000', '#FFA900'],
```

It is an **array**. `components/LayerBalances/index.tsx:112` takes it:

```ts
const buttonBg = themeColor('buttonGradient') || themeColor('secondary');
```

The `||` does not fall through, because an array is truthy. The guard below checks only truthiness, not shape:

```ts
if (!buttonBg || !warningColor || !errorColor) { ... }
```

and then `blendHexColors` calls a string method on it (`utils/ThemeUtils.ts:498`):

```ts
hex1 = hex1.replace('#', '');
```

```
TypeError: hex1.replace is not a function
```

caught at `index.tsx:130`, logged, `return false`.

Every other theme falls back to `Dark['buttonGradient']`, which is undefined, so `buttonBg` becomes `themeColor('secondary')` — a hex string — and the path works. Only the default theme has the array.

### Effect

`index.tsx:182` reads:

```ts
isEcash && ecashRowColors
    ? ecashRowColors
    : themeColor('buttonGradient')
    ...
```

With `ecashRowColors === false` it falls back to the flat gradient. So nothing looks broken — the row simply never tints toward `warning`/`error` as the ecash balance grows past 10,000 sats. On the default theme that nudge has never been visible.

### Reproducing without a device

Running `blendHexColors` as it stands in the file:

```
array (Kyriaki buttonGradient)  -> TypeError: hex1.replace is not a function
valid hex pair                  -> #803637
3-char shorthand   ('#abc')     -> #c62cNaN
named colour       ('white')    -> #NaNNaN2d
```

The last two are worth noting separately: `blendHexColors` has no input validation, so a short hex or a named colour produces a syntactically invalid colour string rather than failing. Themes do contain named colours — Kyriaki has `text: 'white'` — so that is reachable if the function is ever pointed at one.

[...]

PR #4709, opened 2026-09-22
ThemeUtils: align upgrade tint ramp with getUpgradeIntensity

### LINKED ISSUE #4686: getUpgradeBackgroundColor and getUpgradeIntensity use different ramps for the same balance
### Summary

`getUpgradeIntensity` and `getUpgradeBackgroundColor` in `utils/ThemeUtils.ts` describe the same 10k–100k upgrade ramp but compute it two different ways. `Menu.tsx` drives one view from both, so the background tint and the glow disagree — most visibly at exactly the balance where the card first appears.

### The two ramps

```ts
// utils/ThemeUtils.ts:523
export function getUpgradeIntensity(balance: number): number {
    if (balance < 10_000) return 0;
    return Math.min((balance - 10_000) / 90_000, 1);
}

// utils/ThemeUtils.ts:532
if (balance >= 10_000) {
    const ratio = Math.min(balance / 100_000, 1);   // <- not getUpgradeIntensity
    return blendHexColors(baseColor, errorColor, ratio * 0.6);
}
```

One subtracts the 10k floor before normalising; the other does not.

| balance | `getUpgradeIntensity` | blend fraction actually used | what it would be via `getUpgradeIntensity` |
|---|---|---|---|
| 10,000 | **0.000** | **0.060** | 0.000 |
| 20,000 | 0.111 | 0.120 | 0.067 |
| 50,000 | 0.444 | 0.300 | 0.267 |
| 90,000 | 0.889 | 0.540 | 0.533 |
| 100,000 | 1.000 | 0.600 | 0.600 |

They agree at the top of the range and diverge everywhere below it.

### Why it shows

`views/Menu.tsx:292-297` takes both from the same balance:

```tsx
const bgColor = getUpgradeBackgroundColor(themeColor('secondary'), balance) || themeColor('secondary');
const t = getUpgradeIntensity(balance);
```

and `t` drives `shadowOpacity: 0.4 + t * 0.4`, `shadowRadius: 4 + t * 8`, `elevation: 4 + t * 8` and the border offsets.

The card is gated on `CashuStore.totalBalanceSats >= 10_000`, so **the first balance at which it renders is exactly 10,000** — and there `t = 0` puts the glow fully at rest while the background is already 6% blended toward the error colour. The two cues are meant to express the same urgency and they contradict each other at the threshold.

### Suggested fix

Have the colour use the ramp that already exists:

```ts
const ratio = getUpgradeIntensity(balance);
return blendHexColors(baseColor, errorColor, ratio * 0.6);
```

That makes 10,000 a clean "just appeared, nothing yet" state and keeps 100,000 unchanged at 0.6, so only the middle of the range shifts, slightly downward.

### Caveat

If the offset tint at 10k is deliberate — so the card does not appear completely flat — then the mismatch is intentional and this is only a documentation problem, since `getUpgradeIntensity`'s docstring says "10k → 0, 100k → 1" for a band the colour function walks differently.

I have not opened a PR because it changes what the screen looks like and I have no device to capture before/after on, which CONTRIBUTING asks for. Happy to send the one-line change if you would rather have it, or to leave it if the current curve is intended.

*Disclosure: found and written with AI assistance. The table above comes from running both functions as they stand in the file.*

kaloudis:
# Description

Relates to issue: #4686

`getUpgradeBackgroundColor` normalised the ecash balance as `balance / 100_000`, while `getUpgradeIntensity` (which drives the Menu upgrade card's glow) ramps from the 10k threshold: `(balance - 10_000) / 90_000`. Both came from 19ab242f0: the tint helper was extracted as-is from the older upgrade-modal code, and the intensity helper was written new for the Menu card. So the two cues for the same urgency followed different curves. The gap was biggest at 10k, where the card first appears: the glow sat at its baseline but the background was already blended 6% toward `error`.

This change makes the tint use `getUpgradeIntensity(balance) * 0.6`, so both follow one curve. 10k is now untinted and 100k is unchanged at a 60% blend.

Resulting colours on the default dark theme (`secondary` blended toward `error`):

| Balance | Before | After |
|---|---|---|
| 10k | `#37353b` | `#31363f` (no tint) |
| 25k | `#413436` | `#3b3439` |
| 50k | `#50312c` | `#4d322e` |
| 100k+ | `#6f2c19` | `#6f2c19` |

**Note:** the upgrade InfoModal (`CashuStore.getUpgradeModalBackgroundColor`) uses the same helper, so the first modal shown at 10k is now untinted too.

Screenshots still needed: before and after of the Menu upgrade card and the upgrade modal at about 10k, 25k and 100k ecash. I haven't captured these on a device yet.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

Adds `utils/ThemeUtils.test.ts` with 8 cases for `getUpgradeIntensity` and `getUpgradeBackgroundColor`. The two ramp-alignment cases fail on the previous implementation.

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

This is a pure colour calculation and doesn't depend on the backend. The card only shows with Cashu enabled, at least 10k ecash sats, and no channels.

### Locales

### Other:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kunalKumar-13:
Took a look as asked. Concept ACK and the reasoning in the description holds up — I traced both helpers back to 19ab242f0 and the two curves really did diverge from a common origin rather than by design.

I ran it rather than read it, and then went after the tests the way you went after mine.

## Ran

Checked out `fa9dda3`, `npx jest utils/ThemeUtils.test.ts`: **8 passed, 8 total**, exit 0.

## Mutants

Six, against the two functions this PR touches:

| mutant | result |
|---|---|
| revert the fix — `getUpgradeIntensity(balance)` → `Math.min(balance / 100_000, 1)` | **KILLED** (2 failed) |
| drop the scaling — `ratio * 0.6` → `ratio` | **KILLED** (2 failed) |
| widen the guard — `balance >= 10_000` → `> 10_000` | **KILLED** (2 failed) |
| wrong theme key — `themeColor('error')` → `themeColor('warning')` | **KILLED** (2 failed) |
| drop the cap — `Math.min((balance - 10_000) / 90_000, 1)` → unclamped | **KILLED** (2 failed) |
| `if (balance < 10_000)` → `<= 10_000` in `getUpgradeIntensity` | SURVIVED |

The survivor is equivalent, not a gap: at exactly 10,000, `(10_000 - 10_000) / 90_000` is 0, so both branches return the same thing. There is no test that could kill it and none should try.

The one I most expected to survive was the third — `follows getUpgradeIntensity across the ramp` computes its expectation by calling `getUpgradeIntensity`, which is the shape that usually makes a test vacuous. It isn't here, because `applies no tint at exactly 10k` and `reaches a 60% blend at 100k` pin two literal values (`'#31363f'`, `'#6f2c19'`) either side of it. The self-referential test checks the *shape* of the curve and the literal ones anchor it. That combination is why reverting the production change fails two tests instead of none.

## The 10k behaviour change, at both call sites

Below 10k the function returns `undefined` as before, but at exactly 10,000 it now returns a colour — the base, 0% blended — where it used to return a 6% tint. I checked that this cannot surprise either caller:

- `views/Menu.tsx:293` does `getUpgradeBackgroundColor(themeColor('secondary'), balance) || themeColor('secondary')`, so base-colour and `undefined` land on the same value. The returned hex is lowercased by `blendHexColors` where `themeColor` gives `'#31363F'`; irrelevant for RN styling, just noting I looked.
- `stores/CashuStore.ts:3076` and `:3133` pass it straight to `toggleInfoModal({ backgroundColor })`, where a 0% blend of `modalBackground` is the modal background.

So the only visible change is the intended one, and the card and its glow now agree at the moment the card first appears — which was the point.

## Two small things, neither blocking

**`blendHexColors` has no assertion of its own here.** It is imported and used to compute expectations, so it is exercised but never checked — if it were wrong, the ramp test would compare wrong against wrong and pass. The literal-value tests do cover it indirectly, so this is a note rather than a request.

**This file is the right home for #4687.** Your mock —

```ts
jest.mock('../stores/Stores', () => ({
    settingsStore: { settings: { display: { theme: 'dark' } } }
}));
```

— is all `ThemeUtils` needs, since it imports only `settingsStore`. That is the thing that was blocking a test file here, and now it exists. I have just posted a runnable repro on #4687 using the same mock with `theme: 'kyriaki'`; it belongs in this file rather than a new one. Happy to add it on top of this once it lands, so the two do not conflict.

Nice change. The description's before/after table is the part I would keep — it makes the 10k discontinuity obvious in a way the diff does not.

Issue #4711, opened 2026-09-22
Keychain recovery: orphaned non-settings data is scanned but unreachable; restoreSettings is superseded


kaloudis:
Found while reviewing #4694, which adds the first test file for `utils/KeychainRecoveryUtils.ts`. Three related loose ends, none of them introduced by that PR.

### 1. Orphaned non-settings data is scanned and then discarded

`scanForRecoverableData` walks the 20 keys in `OTHER_KEYS` (contacts, notes, POS categories and products, lightning address hashes and activation flag, unit preferences, hidden accounts, currency codes, activity filters, backup status, LSP orders, swaps, reverse swaps, swap rescue data) across all five sources, filters out anything current storage can already read, and returns the remainder as `otherDataFound`.

Nothing reads it. `views/Tools/NodeConfigExportImport.tsx` uses `scanResult.settingsFound` only (line 1134); `otherDataFound` appears nowhere in the tree. `restoreDataKey` (`KeychainRecoveryUtils.ts:351`), the function that would write one of those entries back, has no callers either.

So on a device where the desync migration stranded data, the scan establishes that the user's contacts and notes are sitting in the legacy keychain, logs a count, and drops the result on the floor with no way to act on it. This is the half worth building rather than deleting: the scan work is already done and the write path already exists.

### 2. `restoreSettings` looks superseded, not merely unwired

`restoreSettings` (`KeychainRecoveryUtils.ts:287`) also has no callers, but its semantics are the reason. It overwrites `STORAGE_KEY` wholesale with the recovered blob. The live path does something safer: `handleRecoveryPreview` shows the recovered nodes, the user ticks the ones they want, and `handleRecoveryRestore` appends just those through `saveNodeConfigs` (`NodeConfigExportImport.tsx:1017`), the same call the import flow uses. That cannot clobber wallets that are currently working; a whole-blob overwrite can.

If we agree the append-selected behaviour is the one we want, `restoreSettings` should be deleted rather than wired up.

### 3. Both verifications are truthiness-only

Whichever way the above goes, note that the post-write check in `restoreSettings` and `restoreDataKey` is:

```ts
const verifyData = await Storage.getItem(key);
if (!verifyData) return { success: false, error: 'Verification failed after write' };
```

It never compares `verifyData` against what was written, so a stale value left behind by a failed write passes verification and the caller is told the data is safe. If either function survives, compare against the payload.

### Note for whoever picks this up

[...]

PR #4712, opened 2026-09-22
fix(recovery): verify a restore by comparing the read-back, not its truthiness

kunalKumar-13:
# Description

Part 3 of #4711. Both restore paths write, read the key back, and then check only that
something came back:

```ts
const verifyData = await Storage.getItem(STORAGE_KEY);
if (!verifyData) return { success: false, error: 'Verification failed after write' };
```

That cannot tell a key that was never written from one still holding an older copy. **A
stale value is truthy**, so the caller is told the restore succeeded while the data on
disk is still the data being replaced — which, for wallet configurations, is the one
outcome this screen exists to prevent.

There is a second, quieter part: `Storage.getItem` resolves `false` for a missing key
rather than throwing, so today the missing case is caught only because `false` happens to
be falsy. Comparing against the payload covers the missing key and the stale one, and
stops depending on that coincidence.

Relates to issue: **#4711**

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [x] Quality assurance
- [ ] Other

## Changes Made

`verifyData !== recoveryResult.data` in place of `!verifyData`, at both sites in
`utils/KeychainRecoveryUtils.ts`.

Tests: the stale read-back on both paths, plus **the first tests for `restoreDataKey`**,
which had none — write, rejected write, stale read-back, absent key. That function is the
one #4711 wants wired to the UI, so it seemed worth covering before anyone builds on it.

## Checklist
- [x] `yarn run tsc` — exits 0, no output
- [x] `yarn run lint` — eslint exits 0
- [x] `yarn run prettier` — clean
- [x] `yarn run test`

```
PASS utils/KeychainRecoveryUtils.test.ts
Tests:       30 passed, 30 total
```

## Testing

- [x] Yes

Mutants, one that matters:

| mutation | tests failed |
|---|---:|
| revert both sites to `!verifyData` | **2** — exactly the two stale-value tests |

Nothing else in the suite moves, which is the point: the existing tests passed against the
old check because none of them ever put a *plausible* wrong value in front of it.

I have tested this PR on the following platforms:

- [ ] Android
- [ ] iOS

Neither — no device or simulator here. The production change is a comparison operator on a
path reached only from the recovery screen, so if you would rather see it exercised there
before merging, that is a fair ask.

## Additional Notes

Deliberately scoped to part 3. Whether `restoreSettings` is kept or deleted, and whether
`otherDataFound` gets wired up, are the open questions on #4711 — this change is correct
either way, so it seemed better not to tangle it with them.

If `restoreSettings` does get deleted, its half of this reverts with it and the
`restoreDataKey` half stands alone.


utils/KeychainRecoveryUtils.ts
```diff
@@ -370,9 +371,14 @@ class KeychainRecoveryUtils {
                 };
             }
 
-            // Verify the write
+            // Verify the write by reading the value back and comparing it
+            // against what was sent. Checking only that something is there
+            // cannot tell an unwritten key from one still holding an older
+            // copy -- and Storage.getItem resolves `false` for a missing key
+            // rather than throwing, so both pass a truthiness check as long
+            // as the stale value is non-empty.
```
kaloudis, line 379: This comment contradicts itself, and the half that is wrong is the half a future reader would act on.

`Storage.getItem` resolves `false` for a missing key, and `false` is falsy, so an unwritten key *fails* a truthiness check. The two cases do not "both pass"; the old check reported failure for the missing key and success for the stale one. What the old check could not distinguish is the value just written from some other value that happens to be sitting there, which is the actual bug, and it caught the missing case only incidentally.

The PR description gets this right ("the missing case is caught only because `false` happens to be falsy"); the comment lost it somewhere. Suggested rewrite:

```suggestion
            // Verify the write by reading the value back and comparing it
            // against what was sent. A truthiness check cannot tell the value
            // just written from an older copy the key is still holding, so it
            // reports success while the data on disk is the data being
            // replaced. It catches the unwritten case only incidentally,
            // because Storage.getItem resolves `false` for a missing key
            // rather than throwing.
```

utils/KeychainRecoveryUtils.test.ts
```diff
@@ -291,6 +291,80 @@ describe('KeychainRecoveryUtils', () => {
...
+        const recovered = (data: string): RecoveryResult => ({
+            key: 'contacts',
+            source: 'encrypted-storage',
+            data,
+            description: 'Contacts'
+        });
+
+        it('writes the payload under the recovered key', async () => {
+            storage.setItem.mockResolvedValue(true as any);
+            storage.getItem.mockResolvedValue('["alice"]' as any);
```
kaloudis, line 323: Blanket `mockResolvedValue` here leaves the read-back *key* unpinned, which is the one mutant that survived my sweep: point the verification at the wrong key and nothing fails.

```
datakey: verify reads STORAGE_KEY instead of recoveryResult.key   SURVIVED
settings: verify reads LEGACY_STORAGE_KEY instead of STORAGE_KEY  SURVIVED
```

That is worth closing in this PR specifically, because a verification aimed at the wrong key is the same failure mode you are fixing: it reads back something real, compares it to the payload, and the comparison is meaningless. See the review body for the two-hunk patch that kills both (I ran it: 30/30 still pass).

kaloudis (commented):
Concept ACK, and the production change is right as written. Two asks before I merge, neither of which touches the operator you changed.

## What I verified

Branch `5607c25` is level with master (`bbb779b`), all four CI checks green. Locally, in a worktree off the PR head: `tsc` clean, `eslint` clean, `prettier` clean, `npx jest utils/KeychainRecoveryUtils.test.ts` 30/30.

I confirmed the round-trip assumption the comparison rests on, since a strict `!==` is only safe if the storage layer is byte-transparent. `Storage.setItem` passes a string value through untouched (`typeof value === 'string' ? value : JSON.stringify(value)`, `storage/index.ts:111`) and `getItem` returns `response.password` verbatim, same key prefix, same `cloudSync: false` partition. `RecoveryResult.data` is typed `string` and every reader in the scan produces one, so there is no shape on either side that could make an honest write fail the new check. The empty-string case routes through `removeItem` and fails verification under both the old and the new check, and `scanKey` never emits an empty `data` anyway.

Your mutant table reproduces:

| mutant | result |
|---|---|
| `restoreSettings` back to `!verifyData` | KILLED, exactly the new stale test |
| `restoreDataKey` back to `!verifyData` | KILLED, exactly the new stale test |
| either verify never fails | KILLED (2, 2) |
| either comparison inverted | KILLED (3, 3) |
| `restoreDataKey` writes under `STORAGE_KEY` | KILLED |
| either write-failure check disabled | KILLED |

One survivor, which is ask 2 below.

## Ask 1: the comment in `restoreDataKey` states the opposite of the behaviour

Left inline. Short version: a missing key reads back `false`, which fails a truthiness check, so the missing and stale cases do not "both pass". Your PR description has it right and the comment does not.

## Ask 2: pin the read-back key in both success tests

The only mutant to survive, in both functions:

```
datakey: verify reads STORAGE_KEY instead of recoveryResult.key   SURVIVED
settings: verify reads LEGACY_STORAGE_KEY instead of STORAGE_KEY  SURVIVED
```

Blanket `storage.getItem.mockResolvedValue(...)` answers for every key, so a verification pointed at the wrong key still reads back the payload and still passes. That is worth closing here rather than later: a read-back aimed at the wrong key is the same class of defect as a read-back that only checks truthiness. It looks like verification and confirms nothing.

Keying the two success tests kills both. I ran this and the suite stays 30/30:

```ts
        it('writes the recovered payload verbatim and counts the nodes', async () => {
            storage.setItem.mockResolvedValue(true as any);
            // Keyed, not blanket: the read-back has to hit the key that was
            // written, or the verification is reading someone else's value.
            storage.getItem.mockImplementation(async (key: string) =>
                key === STORAGE_KEY ? SETTINGS : (false as any)
            );
```

```ts
        it('writes the payload under the recovered key', async () => {
            storage.setItem.mockResolvedValue(true as any);
            storage.getItem.mockImplementation(async (key: string) =>
                key === 'contacts' ? '["alice"]' : (false as any)
            );
```

## On device testing

Not owed here, and I will say so on the record rather than leave you guessing. `restoreSettings` and `restoreDataKey` have no callers anywhere in the tree (confirmed by grep; `NodeConfigExportImport` uses `scanForRecoverableData`, `parseSettingsPreview`, `getSourceDisplayName`, and the two dev helpers only). Nothing a user can do today reaches either line you changed, so there is no device behaviour to test. When #4711 part 1 wires `restoreDataKey` to the UI, that PR carries the device testing, and it will carry it against a verification that means something.

## Follow-up worth filing, not for this PR

`MigrationUtils.migrateKey` has the identical truthiness-only verify at `utils/MigrationUtils.ts:151`, and unlike the recovery path it runs in production, on iOS, at boot, and then deletes the legacy keychain entry at line 164 on the strength of it. The early return at line 125 means a pre-existing stale value cannot be what it reads back, so the exposure is narrower than here: a truncated or partial write, or a concurrent writer landing between the write and the read. But that is precisely the case where the only remaining copy is about to be deleted. The same two-line change applies, `verifyData !== credentials.password`, and `utils/MigrationUtils.test.ts` already exists to hold the test. Yours if you want it; say the word and I will open the issue.

One nuance I considered and am happy to accept: after this change, a concurrent writer landing between the write and the read-back turns a real success into "Verification failed after write". For a recovery screen, failing loudly on an ambiguous outcome is the right side to err on.

Deliberate scoping to part 3 was the right call, and the note about which half survives if `restoreSettings` gets deleted is exactly the context that makes this easy to merge. Push those two and I will approve.

kunalKumar-13:
Both pushed, exactly as written. Thank you for running the sweep and handing me the patch rather than the finding — that saved a round trip.

## Ask 1

You're right and I can see where I lost it. The description says the missing case is caught "only because `false` happens to be falsy", and then the comment says both cases pass, which is the opposite. Taken your rewrite verbatim.

`restoreSettings` only says "See restoreDataKey", so fixing the one fixes both.

## Ask 2

Keyed both stubs. Re-ran the two survivors against the new tests:

| mutant | before | after |
|---|---|---|
| `settings`: verify reads `LEGACY_STORAGE_KEY` instead of `STORAGE_KEY` | SURVIVED | **KILLED** — 1 failed, 29 passed |
| `datakey`: verify reads `STORAGE_KEY` instead of `recoveryResult.key` | SURVIVED | **KILLED** — 1 failed, 29 passed |

Baseline back to 30/30 after reverting each. `tsc --noEmit` exit 0, `eslint` exit 0, `prettier --check` exit 0 — read the exit codes this time rather than the scrollback.

The point lands: a verification aimed at the wrong key reads back something real, compares it to the payload, and means nothing — same shape as comparing against truthiness. It should have been keyed from the start.

## On the `MigrationUtils` follow-up

Yes, and I'll take it — but one correction to the premise first, because it changes how
urgent this is and I'd rather say so than quietly agree.

You have the lines right: the vacuous verify is `utils/MigrationUtils.ts:151-152`, the
early return is `:125`, and the delete is `:164`. But `deleteFromOldKeychain` doesn't
delete anything:

```ts
    private async deleteFromOldKeychain(key: string) {
        // Safety measure: skip deletion to prevent any potential data loss
        console.log(`[Migration] Skipping delete for ${key} (safety measure)`);
        return;

        // Original delete code - commented out for safety:
        // try {
        //     await Keychain.resetInternetCredentials({
```

It logs and returns, with the `resetInternetCredentials` calls commented out beneath the
early return. `:164` is its only caller anywhere in the tree. So today the legacy entry
survives a failed verify, and "the only remaining copy is about to be deleted" is not
currently true.

That makes it a smaller bug than your framing, and I don't think it changes the answer.
The check is still vacuous, `migrateKey` still runs at boot on iOS rather than behind a
dev-only screen, and the delete is sitting right there commented out with an explanation
of why — which is the shape of thing someone re-enables once migration looks settled. The
right order is to fix the verify while it's cheap, so that whoever turns the delete back on
inherits a check that means something rather than one that has always passed.

Worth me saying plainly: I only found this because you gave me a line number and I went to
read it. If I've misread the dead code, tell me.

Open the issue and assign it to me. `utils/MigrationUtils.test.ts` already exists, so it's
the same shape of PR as this one — change, test, mutant table. Its own PR, not a widening
of this one.

## On the concurrent-writer nuance

Agreed, and worth having on the record: after this change a concurrent writer landing between write and read-back turns a real success into `Verification failed after write`. On a recovery screen, a false failure sends the user to try again with the data still on disk, and a false success tells them their data is back when it isn't. Erring toward the first is right.

## On device testing

Noted, and thank you for saying it on the record — I'd have offered to find a device and wasted both our time. Agreed that when #4711 part 1 wires `restoreDataKey` to the UI, that PR carries the device testing, and it'll be testing against a verification that means something.

PR #4713, opened 2026-09-22
fix(security): authenticate share intents on the POS cold start

kaloudis:
# Description

With POS enabled, every cold start enters POS mode without a PIN. That is the documented POS boot. The same startup path also reads the Android share intent, and it handed the payload to `ShareIntentProcessing` without the `requiresAuth` / `requiresWalletSelection` flags that the sibling `LinkingUtils` path sets. That screen gated only on a **truthy** `requiresAuth`, so a missing flag read as "no gate needed": it decoded the QR and navigated wherever the QR pointed.

Any local app can send an `ACTION_SEND image/*` to the exported `MainActivity`, no permission required. So an attacker's QR could put the victim on `Send` with an invoice already loaded, on `WalletConfiguration` pre-filled from an lndconnect URI, or on `LnurlAuth`, with the PIN never asked for. One tap from there is the whole attack. There is no auto-payment, but leaving POS is otherwise only possible through the Lockscreen, so this is a real escalation out of the mode rather than a shortcut to something already reachable.

The chain, for review:

| Step | Where |
|---|---|
| `posStatus` is a non-persisted observable defaulting to `'unselected'` | `stores/SettingsStore.ts` |
| every POS cold start therefore takes the branch that sets `unlocked: true` with no auth | `views/Wallet/Wallet.tsx` |
| `processSharedQRImageFast()` returned `{ base64Image }` and nothing else | `utils/ShareIntentProcessor.ts` |
| the payload is stashed and later pushed to the processing screen | `views/Wallet/Wallet.tsx` `processPendingShareIntent` |
| the screen gates on truthy flags, then decodes and `navigation.replace(route, params)` | `views/ShareIntentProcessing.tsx` |

## The fix

**The flags are attached where the payload enters the app**, in `processSharedQRImageFast`, rather than by each caller. An intent read from the OS has passed no authentication, and a caller cannot forget what it never had to remember. `Wallet.tsx` needs no change at all as a result, which is worth something in a file with 30-plus open PRs against it, and `LinkingUtils` loses the duplicated flag setting it was doing by hand.

**`ShareIntentProcessing` fails closed.** Only an explicit `requiresAuth: false`, from a caller that knows the user is through or that no login is configured, skips the Lockscreen. This is the half that matters: it is what stops the next call site from reintroducing the same hole.

**The flags become a protocol, so each gate clears its own once satisfied.** The Lockscreen marks the payload authenticated on its way out, wallet selection marks it selected, and the processing screen preserves the gate it is not currently handling while it delegates the other. Without that, a payload bounces back to the screen it just cleared, forever.

**The Lockscreen's POS waiver no longer applies to a share intent.** This one is easy to miss and the fix is incomplete without it: the Lockscreen proceeds without authentication when POS is active, so the payload would have reached it and been waved straight through, one screen later than before. Nothing sends an already-authenticated payload to the Lockscreen, so a share intent that arrives there is one that still has to authenticate. The waiver still covers its actual purpose, an active POS terminal resuming without a PIN prompt.

Not fixed here: the startup path never calls `clearSharedIntent()`, which `LinkingUtils` does on both of its paths, so the same intent looks re-processable on a later launch. It is authenticated either way now.

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [X] Yes
- [ ] N/A

`views/ShareIntentProcessing.test.ts` is new and covers the gate at the finding's site: a payload with no auth flag goes to the Lockscreen and the QR is never decoded, a pending wallet selection survives the trip through the Lockscreen, and an explicitly satisfied payload still processes and navigates. `utils/ShareIntentProcessor.test.ts` is new and covers the flags being attached at the source and each continuation clearing its own. Verified that the first test fails against the unfixed screen.

- [ ] Android
- [ ] iOS

Device testing still owed, and it is the part worth doing carefully here: with POS enabled and a PIN set, share a QR image to ZEUS from another app on a cold start and confirm the Lockscreen appears and the destination is only reached after the PIN; then the same with no PIN configured, and with select-wallet-on-startup on, to confirm neither path loops between the Lockscreen, wallet selection and the processing screen. Not yet run, flagging rather than ticking.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

Backend-independent: this is navigation authorization, and it runs before any backend call.

### Locales

No new strings: the Lockscreen step already has one.

### Other:


utils/ShareIntentProcessor.ts
```diff
@@ -62,11 +63,26 @@ export const processSharedQRImageFast =
...
+            // asked for. Screens that continue this payload after a
+            // successful unlock or wallet selection clear the corresponding
+            // flag themselves.
             return {
                 success: true,
                 route: 'ShareIntentProcessing',
-                params: { base64Image }
+                params: {
+                    base64Image,
+                    requiresAuth: settingsStore.loginRequired(),
```
ajaysehwal, line 82: @kaloudis, really nice fix overall. Moving the flags to where the payload enters the app and making the screen fail closed are the right calls. I think one path is still open, though.

`requiresAuth` comes from `loginRequired()`, which is based on `!loggedIn`, but the Lockscreen’s POS waiver calls `setLoginStatus(true)` without a PIN. After a POS cold start, one background/resume can go through `Wallet.handleAppStateChange` → `Lockscreen` → waiver, leaving `loggedIn = true` while POS is still active.

From there, a share intent picked up by `handleAndroidIntents` (or iOS `handleOpenURL`) gets `requiresAuth: false` and can go straight to `Send` / `WalletConfiguration` without authentication.

Could we treat an active POS session as unauthenticated here, e.g. `loginMethodConfigured() && (!loggedIn || posStatus === 'active')`, and ideally re-check `posStatus` inside `ShareIntentProcessing` too? `handleDeepLink` has the same `loginRequired()` check and may need the same treatment.

A test for “logged in but POS active → `requiresAuth: true`” would be useful to lock this down.

views/Lockscreen.tsx
ajaysehwal, line 1: @kaloudis, views/Lockscreen.tsx:601 small POS UX issue:  with this change, a shared image sends the terminal to the PIN screen, but there’s no Back button for `shareIntentData`, and Android back exits the app. So staff without the PIN can’t dismiss the share intent and return to POS without restarting the app.

Could the Lockscreen show Back when `shareIntentData` is present, clear the intent via `MobileTools.clearSharedIntent()`, and return to `Wallet` without authenticating?

utils/ShareIntentProcessor.ts
```diff
@@ -62,11 +63,26 @@ export const processSharedQRImageFast =
...
+                    requiresWalletSelection:
+                        settingsStore.settings?.selectNodeOnStartup
```
ajaysehwal, line 84: @kaloudis, one behavior change to flag: `requiresWalletSelection` is now always derived from `selectNodeOnStartup`, so warm shares handled through `Wallet.tsx` (e.g. iOS `handleOpenURL`) can show the wallet picker every time.

Before this PR, only the cold-start/`initialStart` path requested wallet selection. Is the new behavior intentional? If not, it may be better to require selection only when no wallet has been selected yet in the current session.

views/Lockscreen.tsx
```diff
@@ -343,7 +359,9 @@ export default class Lockscreen extends React.Component<
                 // Only handle wallet selection when NOT modifying security
                 this.resetAuthenticationAttempts();
 
-                const shareIntentData = route.params?.shareIntentData;
+                const shareIntentData = authenticatedShareIntent(
```
shubhamkmr04: This branch never calls `setPosStatus('inactive')`, unlike the ones at 213 and 381. It used to be unreachable with POS active, but a share intent lands here now, so after the PIN (with select wallet on startup on) POS stays active while `loggedIn` is true.

views/ShareIntentProcessing.tsx
```diff
@@ -64,12 +64,22 @@ class ShareIntentProcessing extends React.Component<
...
 
-        if (requiresAuth) {
+        // Fail closed. This screen decodes whatever QR it is handed and
+        // navigates wherever that QR points, including Send with an invoice
+        // loaded and WalletConfiguration pre-filled from a link, so a caller
+        // that arrives without saying whether the payload was authenticated
+        // must be treated as not having authenticated it. Only an explicit
+        // false, from a caller that knows the user has unlocked or that no
+        // login is configured, skips the Lockscreen.
+        if (requiresAuth !== false) {
```
shubhamkmr04: `requiresAuth` is decided when the intent is read, but the payload can sit in `pendingShareIntent` until connecting finishes. With lock-on-background on: share while it's connecting, background the app, resume to the Lockscreen, then `processPendingShareIntent` does `popToTop()` (dropping the Lockscreen) and pushes the old `requiresAuth: false`. Maybe worth checking `loginRequired()` live here?

utils/ShareIntentProcessor.ts
```diff
@@ -77,6 +93,24 @@ export const processSharedQRImageFast =
         }
     };
 
+/**
+ * Marks a share-intent payload as having cleared the app lock, so the
+ * processing screen does not send the user straight back to the Lockscreen
+ * they just came from. Only for callers where the user has actually
+ * authenticated.
+ */
+export const authenticatedShareIntent = (shareIntentData?: any) =>
```
shubhamkmr04: nit: a shared payload type instead of `any` here (and for `pendingShareIntent` in Wallet, which doesn't include the flags) would make a dropped `requiresAuth` a type error.

utils/ShareIntentProcessor.ts
kaloudis: Confirmed, and it holds under default settings (`loginBackground` is off, so the waiver's `loggedIn = true` sticks). Fixed in b56fee249: new `SettingsStore.externalInputAuthRequired()`, which is `loginMethodConfigured() && (!loggedIn || posStatus === 'active')`. It replaces `loginRequired()` in `processSharedQRImageFast` and `handleDeepLink`, and `ShareIntentProcessing` re-checks it live. Deep links received while POS is active are now held, as on the POS cold start. Added the "logged in but POS active" test in `stores/SettingsStore.test.ts`.

views/Lockscreen.tsx
kaloudis: Agreed. In b56fee249 the Lockscreen shows Back when a share intent arrives while POS is active, and Android back does the same. `App.handleBackPress` no longer exits the app on that screen and lets the Lockscreen handler run. Back drops the payload and pops to `Wallet` without authenticating. The intent is now cleared in `processSharedQRImageFast` as soon as the image is read, so the same image no longer comes back on the next resume.

utils/ShareIntentProcessor.ts
kaloudis: Not intentional. The `LinkingUtils` paths already used bare `selectNodeOnStartup`, but the `Wallet.tsx` paths passed no flag, so warm shares there now hit the picker. Changed to `selectNodeOnStartup && initialStart` in b56fee249. `initialStart` goes false once a wallet is selected or connected, which matches "no wallet selected yet this session". Test added.

views/Lockscreen.tsx
kaloudis: Right, that branch was unreachable with POS active before this PR. Fixed in b56fee249: it now calls `setPosStatus('inactive')` like its siblings. This was also needed for the live POS check in `ShareIntentProcessing`, which would otherwise have sent the payload back to the Lockscreen in a loop.

views/ShareIntentProcessing.tsx
kaloudis: Good catch. b56fee249 re-checks live: the gate is now `requiresAuth !== false || settingsStore.externalInputAuthRequired()`, so a payload that waited in `pendingShareIntent` goes back to the Lockscreen if auth is required now. It can't loop: the PIN path sets `loggedIn` and leaves POS before continuing. Test added.

utils/ShareIntentProcessor.ts
kaloudis: Done in b56fee249: `ShareIntentPayload` is exported from `ShareIntentProcessor` and used for the continuation helpers, the `Lockscreen` and `ShareIntentProcessing` params, and `pendingShareIntent` and `getSettingsAndNavigate` in `Wallet`.

PR #4714, opened 2026-09-22
fix(security): apply the LNURL host policy to every fetch, not just pay

kaloudis:
# Description

`isLnurlCallbackAllowed` exists to stop the wallet fetching a callback in cleartext or aiming one at loopback, RFC1918, link-local or CGNAT space. Its own docstring names the class: SSRF-by-callback, with the cloud-metadata endpoint as the example. It was added with the LNURL-pay binding work in #4374 and hardened in the IP-literal canonicalization that followed.

Until #4576 it had **one** caller. Three callback fetches still go out unchecked, each from the phone's network position, each reachable from a scanned or pasted LNURL:

| Site | What the attacker-chosen host receives |
|---|---|
| `stores/InvoicesStore.ts` | `k1` and the invoice (lnurl-withdraw) |
| `stores/CashuStore.ts` | `k1` and the invoice (ecash withdraw route) |
| `views/LnurlChannel.tsx` | the node pubkey |

All three now check before building the URL, and tell the user why they stopped. LNURL-auth is deliberately untouched: `verifyLnurlAuthCallback` from #4576 already runs this check on its way to the stricter domain binding, at both of its call sites.

## The step before the callbacks was open too

Fixing only the callbacks would have left the more direct version of the same thing, and it needs no attacker server at all:

- `js-lnurl` decodes a bech32 lnurl to an **arbitrary** URL and fetches whatever comes out, so `lnurl1...` encoding `http://192.168.1.1/` is a LAN request with no service, no response parsing and no callback involved.
- `getParams` short-circuits a `withdrawRequest` whose `k1` and `callback` are already in the query string and returns them **without any fetch**, so a single QR can carry the entire payload, callback included.

`isLnurlEndpointAllowed` applies the same policy to the decoded endpoint, which covers every tag at once rather than tag by tag. The existing comment in `handleAnything` already described that fetch as "a live HTTP GET of an attacker-controllable target"; it was only being deferred for clipboard probes, not policed.

Not every resolve goes through `handleAnything`. The ChoosePaymentMethod rows (`LightningSwipeableRow`, `EcashSwipeableRow`) called js-lnurl's `getParams` themselves, and a BIP21 URI such as `bitcoin:<addr>?lightning=lnurl1...` reaches them unchecked. So the check lives in `getLnurlParams` (`utils/LnurlResolveUtils.ts`), a wrapper around `getParams`, and every resolve goes through it. It is kept out of `LnurlPayUtils` because it needs `localeString`, and importing `LocaleUtils` there pulls the store graph into every importer's tests.

## Lightning addresses

Zeus resolves LUD-16 addresses itself, without js-lnurl, and the address grammar accepts IP-literal domains. `user@192.168.1.1` was fetched as `https://192.168.1.1/.well-known/lnurlp/user` by both `handleAnything` and `LightningSwipeableRow.handleLightningAddress`. Both now check the well-known URL with `isLightningAddressEndpointAllowed` before fetching.

A value the decoder cannot decode is deliberately left alone: js-lnurl decodes before it fetches and reports an undecodable value as an error without touching the network, so there is no request for the check to prevent, and rejecting it would only replace a clear error with a misleading security one.

Scope is unchanged from the existing guard, and stated where the check lives: IP literals, not DNS rebinding. A hostname that resolves to an internal address still gets through.

## Note for reviewers

`utils/handleAnything.test.ts` stubbed `decodelnurl` to `() => null`. With the new check that stub would have meant nothing to inspect, so the suite now uses a per-test decoder that is undecodable by default, which is what most of those inputs actually are.

- [ ] New feature
- [X] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

## Testing

- [ ] No, I’m a fool
- [X] Yes
- [ ] N/A

`utils/LnurlPayUtils.test.ts` covers the endpoint check against real bech32-encoded targets: public https and onion allowed, loopback, RFC1918, link-local and cleartext refused, undecodable left to js-lnurl. `utils/LnurlResolveUtils.test.ts` covers `getLnurlParams` refusing LAN and loopback endpoints without a fetch, including a withdraw link js-lnurl would short-circuit, and pins the js-lnurl 0.6.0 behavior the undecodable pass-through relies on (no request for a value it cannot decode). `utils/handleAnything.test.ts` covers the resolve step refusing those targets and never calling `getlnurlParams`, and IP-literal lightning addresses being refused without a fetch. `stores/InvoicesStore.test.ts` covers the withdraw callback: the four unsafe shapes are not fetched, a public https callback still is, with `k1` and the invoice intact. Verified that the withdraw tests fail against the unguarded path.

Rebased onto master after #4576 landed, which is what removed `LnurlAuth` from the list above.

- [ ] Android
- [ ] iOS

Device testing still owed: a normal lnurl-withdraw against a real service, and an lnurl-channel request, to confirm the guard is not in the way of the legitimate flows, which is the risk with a change like this. Not yet run, flagging rather than ticking.

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

The withdraw paths create an invoice on the active backend before the callback is fetched, so any backend exercises them; the check itself is backend-independent.

### Locales

Three strings: `utils.lnurl.unsafeCallback`, `utils.lnurl.unsafeEndpoint` and `utils.lnurl.unsafeLightningAddress`. The existing `views.LnurlPay.LnurlPay.unsafeCallback` is payment-specific ("Payment cancelled"), which does not fit a withdraw, an auth or a channel request.

### Other:


utils/LnurlPayUtils.ts
```diff
@@ -305,3 +306,27 @@ export const verifyLnurlAuthCallback = (
...
+export const isLnurlEndpointAllowed = (lnurl: string): LnurlCheckResult => {
+    let decoded: string;
+    try {
+        decoded = decodelnurl(lnurl);
+    } catch (err: any) {
+        // Nothing to police: js-lnurl decodes before it fetches and reports
+        // an undecodable value as an error without touching the network, so
+        // no request can come of this one. Rejecting it here would only
+        // replace that error with a misleading security one.
+        return { ok: true };
```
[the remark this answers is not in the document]
kaloudis: The mechanism doesn't hold for the pinned js-lnurl (0.6.0). `getParams` decodes with the same `decodelnurl` this check calls, on the same input, and when that throws it returns `{ status: 'ERROR', reason: "invalid lnurl '...'" }` without making a request. There is no broader decoder in this version and no lightning-address support. `getParams('user@192.168.1.1')` returns that error and never reaches `cross-fetch`. So the undecodable pass-through stays: rejecting those values would only replace a clear error with a misleading security one.

The example input was still a real SSRF, through a different path. Zeus resolves LUD-16 addresses itself, and `isValidLightningAddress` accepts IP-literal domains, so `user@192.168.1.1` was fetched as `https://192.168.1.1/.well-known/lnurlp/user` by both `handleAnything` and `LightningSwipeableRow.handleLightningAddress`. While checking that, I found a second gap: the ChoosePaymentMethod rows called `getParams` directly, so `bitcoin:<addr>?lightning=lnurl1...` got around the endpoint check.

Fixed in e9d761c3b:
- `utils/LnurlResolveUtils.ts` adds `getLnurlParams`, which runs the endpoint check before `getParams`. Every resolve goes through it.
- Both lightning-address fetches check the well-known URL first (`isLightningAddressEndpointAllowed`).
- `utils/LnurlResolveUtils.test.ts` includes a test that pins the js-lnurl behavior described above (no request for an undecodable value), so a future js-lnurl version that resolves lightning addresses will fail it.

utils/handleAnything.ts
```diff
@@ -1003,6 +1025,7 @@ const handleAnything = async (
         // to when the user actually acts on the value.
         if (isClipboardValue) return true;
         const raw: string = findlnurl(value) || lnurl || value || '';
+        assertLnurlEndpointAllowed(raw);
```
ajaysehwal, line 1028: nit: `assertLnurlEndpointAllowed` duplicates the validation already performed inside `getLnurlParams`, so the endpoint is checked twice on every resolve.

If `getLnurlParams` throws a named `UnsafeLnurlError`, `handleAnything` could rethrow that directly, and the extra assertion could be removed.

Also optional: the same OK-only `Alert.alert` block now appears in five places, so a small helper could tidy that up.

utils/LnurlResolveUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+ * (the ChoosePaymentMethod rows, fed by a BIP21 `lightning=` parameter) is
+ * policed the same way. Rejects with a localized, user-facing message.
+ *
+ * Kept apart from LnurlPayUtils so that module stays free of localization.
+ */
+export const getLnurlParams = async (lnurl: string) => {
+    if (!isLnurlEndpointAllowed(lnurl).ok) {
+        throw new Error(localeString('utils.lnurl.unsafeEndpoint'));
+    }
+    return getParams(lnurl);
```
ajaysehwal, line 18: One gap worth flagging: getParams (via cross-fetch) and ReactNativeBlobUtil.fetch both follow 3xx redirects, so a public HTTPS endpoint/callback could redirect to something like http://192.168.x.x/... and bypass the IP-literal check.

This isn’t a regression since the existing pay callback has the same behavior, but could we mention redirects alongside DNS rebinding in the scope/PR description? A follow-up could disable redirects for these fetches (e.g. followRedirect: false for callbacks) or re-validate each redirect target.

stores/InvoicesStore.ts
```diff
@@ -490,6 +491,29 @@ export default class InvoicesStore {
                 }
 
                 if (lnurl) {
+                    if (!isLnurlCallbackAllowed(lnurl.callback).ok) {
```
ajaysehwal, line 494: @kaloudis, small UX issue: the withdraw callback is checked only after `createInvoice` (and any LSP JIT wrapping), so if validation fails, the user is left with an invoice that the withdraw service will never pay.

Since `getLnurlParams` already has the resolved `callback`, could we validate it there for the withdraw/channel/pay flows and reject before creating the invoice? The store-level checks can stay as defense in depth for other entry points.

utils/handleAnything.ts
kaloudis: Agreed, done in d5df78565. `getLnurlParams` throws errors named `UnsafeLnurlError` (checked by name through `isUnsafeLnurlError`, since `instanceof` on transpiled Error subclasses isn't reliable), and both lnurl catch blocks in `handleAnything` rethrow them, so `assertLnurlEndpointAllowed` is gone. I left the Alert blocks as they are: the same inline OK-only alert appears about 40 times across the codebase, so a helper belongs in its own cleanup PR.

utils/LnurlResolveUtils.ts
kaloudis: Confirmed. It also works for `http://` targets, since iOS sets `NSAllowsArbitraryLoads` and Android permits cleartext. Added to the scope paragraph in the description, and filed #4728 for the follow-up. The callbacks and lightning address lookups can use `followRedirect: false` and check each hop. The resolve can't: js-lnurl calls `cross-fetch` itself and RN fetch has no way to stop following redirects, so `getLnurlParams` would have to fetch and parse the response itself.

stores/InvoicesStore.ts
kaloudis: Done in d5df78565. `getLnurlParams` now checks `params.callback` after `getParams`, which covers withdraw, pay and channel before anything is created. The store-level checks stay for params that arrive another way. This needed the named error from your other comment, since `handleAnything`'s catches would otherwise have replaced the unsafe-callback message with the invalid-params one.

PR #4563, opened 2026-09-03
locales: fix contradictory incorrect-payment-details error copy

kaloudis:
# Description

Relates to issue: **#2001** (follow-up to #4085)

Fixes the self-contradictory copy in `error.failureReasonIncorrectPaymentDetails`.

`FAILURE_REASON_INCORRECT_PAYMENT_DETAILS` is only ever surfaced on payments in the terminal `FAILED` state: LND sets `failure_reason` only once every HTLC attempt has been resolved, and `handlePayment` in `TransactionsStore` excludes `IN_FLIGHT` from the error path. By the time the user sees this message the payment is definitively failed and any held HTLC has been failed back. Telling users to "Check Activity for the final payment state" reintroduced exactly the uncertainty the first sentence resolved; that phrasing belongs to the genuinely non-terminal messages (`inTransit`, `paymentTimedOut`) it was copied from.

Also drops "your funds are not sent", which is implied by "Payment failed". The released-funds reassurance is kept (reworded) for hold invoice payers who watched their spendable balance drop while the HTLC was held, which was the original complaint in #2001. The message stays generic because the wire error is the same `incorrect_or_unknown_payment_details` code for canceled hold invoices, unknown hashes, and already-settled or expired invoices, so Zeus cannot distinguish those causes at this layer.

Before:

> Payment failed because the invoice was rejected, canceled, or is no longer valid. For a canceled hold invoice, the held payment is released and your funds are not sent. Check Activity for the final payment state.

After:

> Payment failed: the recipient rejected it, or the invoice was canceled or is no longer valid. Any held funds have been returned to your balance.

The keysend-specific suffix ("The receiving node might not accept keysend payments.") is unchanged and still appended through the locale-key path.

This pull request is categorized as a:

- [ ] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [x] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist
- [x] I’ve run `yarn run tsc` and made sure my code compiles correctly
- [x] I’ve run `yarn run lint` and made sure my code didn’t contain any problematic patterns
- [x] I’ve run `yarn run prettier` and made sure my code is formatted correctly
- [x] I’ve run `yarn run test` and made sure all of the tests pass

## Testing

If you modified or added a utility file, did you add new unit tests?

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

[...]

Issue #2001, opened 2024-02-25
vague error when HOLD invoice canceled


AndySchroder:
### Describe the bug

Vague error when HOLD invoice canceled.

There should be a clearer error message and a smoother UX for hold invoice cancellation. A HOLD invoice cancellation may be a good thing because you are getting your deposit back, so a nasty error message can be misleading in a lot of circumstances.

Would be nice to fix this as a part of https://github.com/ZeusLN/zeus/issues/1865

### Reproduce

1. Display HOLD invoice
2. Scan and pay hold invoice
3. Zeus says "Sending Transaction" with yellow squiggle animation.
4. Wait 10 seconds
5. Cancel HOLD invoice
6. Get `Error Payment details incorrect (unknown hash, invalid amt or invalid final cltv delta)`

### ZEUS version

v0.8.1

### Node interface

LND (REST)

### Network

Clearnet

### Device

android

### Device operating system

android

kaloudis:
Should this be something that is handled in LND instead?

AndySchroder:
Is that error coming directly from LND?

I don't know that LND is having any problems, it just seems like the Zeus UI isn't that smooth because it doesn't understand what LND is doing.

AndySchroder:
Are you suggesting that LND could provide a more specific error and that would give you more direction on how to more cleanly handle this in the UI? Can LND differentiate between different failure types? I guess I'm wondering what the other reasons could cause this error? Do we know that there was a good route to the destination before getting this error?

AndySchroder:
Would like to add a bounty of 200,000 sat for fixing this issue.

kaloudis:
ZEUS v13.2.1-rc1 which includes a fix for this, is now available for testing: https://github.com/ZeusLN/zeus/releases/tag/v13.2.1-rc1

It will be submitted to app stores shortly. Please reopen if you hit any problems.

Thank you for reporting.

kaloudis:
@AndySchroder ZEUS v13.2.1, which includes a fix, is now available on app stores https://github.com/ZeusLN/zeus/releases/tag/v13.2.1

Please reopen the issue and report back if it recurs. Thanks again for reporting.

AndySchroder:
Now getting

<img width="1080" height="2238" alt="Image" src="https://github.com/user-attachments/assets/2d0aea7b-0b79-41d4-95f3-de10bb048347" />

[...]

PR #4720, opened 2026-09-23
Payments: neutral result and hold duration when the recipient cancels

kaloudis:
| Before (master), held 134 s then canceled | After, held 134 s then canceled | After, rejected within 1 s |
| --- | --- | --- |
|  |  |  |

Rendered on the iOS 26.5 simulator from an lnd REST failure fed through `TransactionsStore.handlePayment`, not from a live node. Details in [this comment](https://github.com/ZeusLN/zeus/pull/4720#issuecomment-5802970872).

# Description

Relates to issue: **#2001**

Related: #4492 and #4495. For LND REST and embedded LND, `payLightningInvoice` swaps the result for the `paymentTimedOut` shape after `timeout_seconds + 1` (61 s by default), so a hold longer than that never reaches `handlePayment` on the Sending screen. #4492 tracks the payment to its final state and #4495 adds the targeted lookups. Once those land, a long hold that is later canceled comes through `handlePayment` and gets the copy added here.

Follows up on the feedback in #2001 after #4085 and #4563 (the "Check Activity for the final payment state" sentence in the reporter's screenshot is the v13.2.1 copy; #4563 removed it in v13.2.2).

**Commit 1: neutral result for payments the recipient rejected.** When the recipient's node rejects or cancels a payment (LND `FAILURE_REASON_INCORRECT_PAYMENT_DETAILS`, LDK Node `recipientRejected`), the Sending screen showed a red X and an "Error" heading and offered Try Again. For a canceled hold invoice that is often the expected outcome (a bond or deposit coming back), and retrying the same invoice fails the same way.

- `TransactionsStore.paymentRejectedByRecipient` is set in `handlePayment` and `handlePaymentError` from a new `isRecipientRejection` in `utils/ErrorUtils.ts` (shares the matcher with `errorToUserFriendly`).
- `PaymentErrorView` shows a neutral return icon (`PaymentReturnedIcon.svg`) and a "Payment not completed" heading for these failures.
- `SendingLightning` hides Try Again and Troubleshooting for them. Go to Wallet stays.

**Commit 2: report how long the recipient held a canceled HTLC.** The wire failure (`incorrect_or_unknown_payment_details`) is the same for a canceled hold invoice and for an unknown, expired, or already paid invoice. The HTLC attempts in `lnrpc.Payment` can tell them apart: an attempt failed by the final hop (`failure_source_index == route.hops.length`) with `INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS`, where `resolve_time_ns` is well after `attempt_time_ns`, was held by the recipient before being failed back.

- `utils/RecipientRejectionUtils.ts`: `getRecipientRejection` reads both the REST/LNC shape (enum names, int64 strings) and the embedded LND protobuf shape (enum numbers, Long timestamps). A hold is reported at 5 s or more (`MIN_REPORTED_HOLD_SECONDS`) and is the longest one across MPP shards. `formatHeldDuration` rounds down to the largest whole unit.
- When a hold is found, the message becomes: "The recipient held this payment for about 2 minutes, then canceled it. The funds have been returned to your balance."
- The check does not depend on `failure_reason`. lnd's payment lifecycle calls `FailPayment(FailureReasonTimeout)` when `timeout_seconds` expires with HTLCs still in flight. I have not confirmed whether the recipient's later cancel replaces that reason, so a final-hop rejection overrides a TIMEOUT reason either way.
- This is measured, so no guess about cause is needed. The pending time is measured by the sender's node and includes path latency, hence "about".
- Covers LND REST, embedded LND, and LNC. CLN and LDK Node do not expose per-attempt timings and keep the generic copy from commit 1.

Not in scope: `CashuSendingLightning` (melt failures), and the Activity payment detail view, which still shows the raw failure.

- [ ] New feature
- [x] Bug fix
- [ ] Code refactor
- [ ] Configuration change
- [ ] Locales update
- [ ] Quality assurance
- [ ] Other

## Checklist

Full jest: 1878 pass; the one failure is `utils/PhotoUtils.test.ts`, which fails on a path check in any worktree checkout and is unrelated.

## Testing

- [ ] No, I’m a fool
- [x] Yes
- [ ] N/A

- `utils/ErrorUtils.test.ts`: `isRecipientRejection` for LND reason strings, the v1 "Payment details incorrect" text, LDK `recipientRejected`, and non-matches.
- `utils/RecipientRejectionUtils.test.ts` (11): REST and embedded shapes, prompt vs held rejection, missing timestamps, MPP max, intermediate-hop and other-code failures ignored, duration formatting.
- `stores/TransactionsStore.recipientRejection.test.ts` (8): flag set and cleared across LND REST, embedded LND and LDK Node; held copy; TIMEOUT reason overridden by a final-hop rejection. Kept in its own file to avoid colliding with the `TransactionsStore.test.ts` that #4492 adds.

Device tests still owed. Plan on Polar (lnd): `lncli addholdinvoice`, pay from ZEUS, `lncli cancelinvoice` at about 10 s (expect held copy, neutral screen, no Try Again) and at about 90 s (expect the timed-out screen until #4492, and record whether lnd reports `failure_reason` TIMEOUT or INCORRECT_PAYMENT_DETAILS); pay an already paid invoice (expect the generic rejection copy with no duration).

- [ ] Android
- [ ] iOS

On-device
- [ ] LDK Node
- [ ] Embedded LND

Remote
- [ ] LND (REST)
- [ ] Nostr Wallet Connect
- [ ] LndHub

### Locales

### Other:


kaloudis:
Before and after on the Sending screen (iOS 26.5 simulator, LND REST wallet). Both builds are the same native app loading JS from Metro, one from `upstream/master` (85a6f624b) and one from this branch. The failure was fed through the real `TransactionsStore.handlePayment` as an lnd REST payment: `status: FAILED`, `failure_reason: FAILURE_REASON_INCORRECT_PAYMENT_DETAILS`, one HTLC failed by the final hop (`failure_source_index: 2` on a two-hop route) with `INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS`. These are not from a live node; the Polar device test is still owed.

| Before (master), held 134 s then canceled | After, held 134 s then canceled | After, rejected within 1 s |
| --- | --- | --- |
|  |  |  |

On master, both cases look like the first screenshot.

stores/TransactionsStore.ts
```diff
@@ -840,6 +856,31 @@ export default class TransactionsStore {
                           result.failure_reason,
                           isKeysend ? ['Keysend'] : undefined
                       )) || errorToUserFriendly(result.payment_error);
+
+            // The per-HTLC failure shows the recipient's node rejected the
+            // payment and how long it held it first. It does not depend on
+            // failure_reason, which lnd may set to FAILURE_REASON_TIMEOUT
+            // when a hold outlasts timeout_seconds.
+            const htlcRejection = getRecipientRejection(result);
```
ajaysehwal, line 864: @kaloudis, `getRecipientRejection` and handling both the REST/LNC and protobuf shapes looks good.

One thing I noticed while tracing the flow: `LND.payLightningInvoice` and `sendPaymentV2Sync` race the call against `forcedTimeout(timeout_seconds + 1)`, so any HTLC held longer than ~61s returns as `paymentTimedOut` before LND’s stream ends. LNC gives up at around 120s.

That means the `FAILURE_REASON_TIMEOUT` override and the minutes/hours/days duration formatting aren’t reachable with the default timeouts, and longer-held invoices — the main use case here — still show “Payment timed out” with “Try Again”.

Could we either narrow the PR description/tests to the short-hold case, or also surface this on the payment details/Activity view where the final HTLC data is available through `listpaym

views/SendingLightning.tsx
```diff
@@ -375,6 +376,11 @@ export default class SendingLightning extends React.Component<
 
         const success = this.successfullySent(TransactionsStore);
         const inTransit = this.inTransit(TransactionsStore);
+        // retrying an invoice the recipient rejected or canceled fails the
+        // same way, so only offer Try Again and Troubleshooting for other
+        // failures
+        const showRetryActions =
```
ajaysehwal, line 382: small UX concern with `showRetryActions`: `INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS` can also mean the final CLTV is too soon, which can happen when the sender’s node is a few blocks behind the chain. For embedded/Neutrino wallets, retrying after the node catches up can succeed.

Hiding both “Try Again” and “Troubleshooting” removes the recovery path in that case, especially since Troubleshooting is only available for local wallets.

Could we hide the actions only when we’re sure it was a held-payment rejection (`heldSeconds !== null`) or LDK `recipientRejected`? At minimum, it might be worth keeping Troubleshooting available for local wallets.

stores/TransactionsStore.ts
```diff
@@ -825,6 +834,13 @@ export default class TransactionsStore {
             (result.payment_error && result.payment_error !== '')
         ) {
             this.error = true;
+            const failureReason =
```
ajaysehwal, line 837: Minor DRY nit: `failureReason` is computed and then the same `lnrpc.PaymentFailureReason[result.failure_reason]` lookup is repeated immediately below. Could we reuse `failureReason` in the `payment_error` expression?

Also, when `heldSeconds === null`, could we keep the `payment_error` that was already computed instead of creating a synthetic `Error('FAILURE_REASON_INCORRECT_PAYMENT_DETAILS')` just to look up the locale key?

Small cleanup, not blocking.

utils/RecipientRejectionUtils.ts
```diff
@@ -0,0 +1,94 @@
+import { lnrpc } from '../proto/lightning';
+import { localeString } from './LocaleUtils';
+
+// Below this, a rejection at the final hop is an ordinary round trip. At or
+// above it, the recipient's node held the HTLC before failing it back, which
+// is what a canceled hold invoice looks like to the payer.
+export const MIN_REPORTED_HOLD_SECONDS = 5;
```
ajaysehwal, line 7: @kaloudis, one concern with the `heldSeconds` threshold: the duration is measured from our side, so it includes the full multi-hop round trip, not just how long the recipient held the HTLC.

A slow route, peer, or reconnect could make an ordinary expired/unknown-invoice failure take 5s+ and get reported as “the recipient held this payment for about 5 seconds, then canceled it,” which may be misleading.

Would a higher threshold (e.g. 15–30s) or softer wording like “didn’t complete after about X” be safer?

utils/ErrorUtils.ts
```diff
@@ -218,10 +218,31 @@ const errorToUserFriendly = (error: Error, errorContext?: string[]) => {
...
+// (unknown or canceled invoice, including a canceled hold invoice). Paying
+// the same invoice again fails the same way.
+const RECIPIENT_REJECTION_LOCALE_KEYS = [
+    'error.failureReasonIncorrectPaymentDetails',
+    'error.ldk.recipientRejected'
+];
+
+// accepts an Error or a bare failure reason string, like errorToUserFriendly
+const isRecipientRejection = (error: any) => {
+    if (typeof error !== 'string' && !(error instanceof Error)) return false;
```
shubhamkmr04: `errorToUserFriendly` also handles plain objects with a `.message` (e.g. the `reject(data.error)` paths in LND.ts), but this guard returns false for them. So we'd show the incorrect-details copy under the red "Error" heading with Try Again. Maybe check `typeof error?.message === 'string'` instead of `instanceof Error`?

utils/ErrorUtils.ts
```diff
@@ -218,10 +218,31 @@ const errorToUserFriendly = (error: Error, errorContext?: string[]) => {
...
         ? userFriendlyErrors[matchingPattern]
         : null;
 
+    return { errorMsg, localeKey };
+};
+
+// Failures where the recipient's node rejected or canceled the payment
+// (unknown or canceled invoice, including a canceled hold invoice). Paying
+// the same invoice again fails the same way.
+const RECIPIENT_REJECTION_LOCALE_KEYS = [
```
shubhamkmr04: I don't think CLN hits either of these. xpay failures come through `handlePaymentError` with CLN's own wording ("Destination said it doesn't know invoice…" iirc), so CLN still gets the red Error + Try Again. The description says CLN keeps the commit 1 copy. Is it meant to be covered here?

locales/en.json
```diff
@@ -970,6 +970,12 @@
     "views.Sending.goToSwap": "Go to swap",
     "views.SendingLightning.lowFeeLimitMessage": "This payment may have failed due to a low fee limit. Try again with a higher fee limit",
     "views.SendingLightning.tryAgain": "Try Again",
+    "views.SendingLightning.paymentNotCompleted": "Payment not completed",
+    "views.SendingLightning.recipientCanceledHeldPayment": "The recipient held this payment for about {{duration}}, then canceled it. The funds have been returned to your balance.",
+    "views.SendingLightning.heldDuration.seconds": "{{count}} seconds",
```
shubhamkmr04: nit: we already have `time.seconds/minutes/hours/days` and `localizedExpiryDuration` in ExpiryUtils. Could we reuse those instead of adding four new strings for translators?

stores/TransactionsStore.ts
kaloudis: Confirmed. `LND.payLightningInvoice` and `sendPaymentV2Sync` return the `paymentTimedOut` shape at `timeout_seconds + 1`, and LNC gives up at `timeout_seconds + 60`, so today the TIMEOUT override and anything past about 120 s cannot be reached from the Sending screen. The description had this under "Related" but the 134 s screenshot overstated it, so the caption now says that case depends on #4492. The override and the longer units stay: #4492 tracks the payment to its final state and hands it to `handlePayment`, which is when they apply. Showing the pending time on the payment detail view from the `listpayments` HTLCs is a good idea; I would do it as a follow-up rather than grow this PR.

views/SendingLightning.tsx
kaloudis: Good catch, confirmed. lnd maps both an unknown invoice and a too-soon final expiry to `reasonIncorrectDetails` (`routing/result_interpretation.go`), so a lagging embedded/Neutrino node would have lost its retry path. Fixed in 431857d74 with a direct check rather than the `heldSeconds` proxy: for a prompt final-hop rejection, `getRecipientRejection` compares `route.hops[last].expiry` with `failure.height + min_final_cltv_expiry` from the decoded payment request, which is the check the recipient runs. lnd pads the final expiry by 3 blocks, so a synced sender does not trip it. CLN xpay uses the failure blockheight the same way. When it trips, the payment is not treated as a recipient rejection, Try Again stays, and the message says the node is behind the chain. Troubleshooting is now shown for local wallets on every failure; only Try Again is hidden for rejections.

stores/TransactionsStore.ts
kaloudis: Both done in 431857d74. `failureReason` is reused in the `payment_error` expression. The synthetic error was there to replace a TIMEOUT message when the HTLC shows a final-hop rejection, so it now runs only when `failureReason` is not already a recipient rejection; otherwise the computed `payment_error` is kept.

utils/RecipientRejectionUtils.ts
kaloudis: Agreed that the measured time includes the route round trip. In 431857d74 the threshold is 15 s and the copy says what is measured: "The payment was pending for about {{duration}}, then the recipient canceled it." That keeps the 15 to 61 s holds reachable today while leaving room for slow routes and reconnects.

utils/ErrorUtils.ts
kaloudis: The object paths you mention do not reach payments today (`wsReq` has no callers, and the `reject(data.error)` at `LND.ts:506` is the PSBT channel-open stream), but the check costs nothing, so 431857d74 accepts any value with a string `message`. It still returns false for objects without one, which would otherwise throw in the matcher.

utils/ErrorUtils.ts
kaloudis: Right, the description was wrong on this. xpay fails with "Destination said it doesn't know invoice: %s" (`plugins/xpay/xpay.c`) and legacy pay with "failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)". Neither matched. 431857d74 maps both to the recipient-rejection copy, so CLN gets the neutral screen. xpay already waits and retries when the failure blockheight is ahead of its own before it gives up, so the lagging-sender case does not reach this message.

locales/en.json
kaloudis: I kept the separate strings. `localizedExpiryDuration` returns capitalized units ("2 Minutes"), which read wrong mid-sentence, and it joins count and unit in code. With `{{count}} minutes` as one string, translators control word order and the unit inflection.

